PR #1, opened 2019-05-19
Adding basic cloudevents pubsub binding.

n3wscott:


adrcunha:
Presubmit tests are failing because of https://github.com/GoogleCloudPlatform/cloud-run-events/issues/7

docs/pubsub-cloudevents.md
```diff
@@ -0,0 +1,374 @@
...
+
+### 1.3. Content Modes
+
+This specification defines three content modes for transferring events:
+_binary_, _structured_ and _batched_. Every compliant implementation SHOULD
+support the _structured_ and _binary_ modes.
+
+In the _binary_ content mode, the value of the event `data` attribute is placed
+into the Pub/Sub payload data as-is, with the `datacontenttype`
+attribute value declaring its media type; all other event attributes are mapped
```
kamalaboulhosn: Are there limits to the number of attributes that can appear in an event? Pub/Sub does have limitations on the number and size of attributes (https://cloud.google.com/pubsub/quotas#resource_limits).

n3wscott: Really good to know! The cloudevents spec has some min-max supported size ideas and they are 64kB.

It would be good to link to this table from this bindings spec.

vaikas-google:
What's the plan here @n3wscott

n3wscott:
@vaikas-google I have implemented this spec in the golang sdk. I am looking for more reviews from the pubsub owners.

hrasadi:
I think it worth mentioning what content mode(s) we are going to use for GCP Sources.

nachocano:
Old PR. Closing it for now. We still have it around for future reference.

jroper:
Is this work being continued anywhere? We're looking at publishing/consuming events over Google Pubsub in [Cloudstate](https://cloudstate.io), and would very much like a consistent means for doing so.

docs/pubsub-cloudevents.md
```diff
@@ -0,0 +1,374 @@
...
+
+Except for attributes
+[explicitly handled in this specification](#2-use-of-cloudevents-attributes),
+the naming convention for the Pub/Sub attribute mapping of well-known CloudEvents
+attributes is that each attribute name MUST be prefixed with "ce-".
+
+Examples:
+
+    * `time` maps to `ce-time`
+    * `id` maps to `ce-id`
```
jroper, line 164: Something to consider, how do the `time` and `id` attributes specified here relate to the `message_id` and `publish_time` fields on the Google [PubsubMessage](https://cloud.google.com/pubsub/docs/reference/rpc/google.pubsub.v1#pubsubmessage)? My thought is that they should be ignored, with the caveat that if something is receiving non cloud event messages through Google Pubsub, and they want to express these messages as cloud events, then those fields may be used to seed those values.

jroper: Ah didn't see that. Thanks!

PR #11, opened 2019-06-10
Adding Channel CRD.

n3wscott:
Only the Channel CRD, no controller.


mattmoor:
Fix 0% coverage?

config/300-channel.yaml
```diff
@@ -0,0 +1,130 @@
...
+                  status:
+                    type: string
+                  type:
+                    type: string
+                required:
+                  - type
+                  - status
+                type: object
+              type: array
+          type: object
```
vaikas-google: nit: missing newline

config/300-channel.yaml
```diff
@@ -0,0 +1,130 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+ name: channels.pubsub.cloud.run
```
vaikas-google: Just jotting this here. Can you also modify the 201-addressable-resolvers-clusterrole.yaml here (or add a new one) for additional verbs, I don't think as it stands today we'll be able to create channels because it only lists: get, list and watch. But perhaps I missed something along the way? Something like this:
https://github.com/knative/eventing/blob/master/config/channels/in-memory-channel/200-channelable-manipulator-clusterrole.yaml

pkg/apis/pubsub/v1alpha1/channel_types.go
```diff
@@ -93,8 +93,8 @@ type ChannelStatus struct {
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
-// PubSubChannelList is a collection of in-memory channels.
-type PubSubChannelList struct {
+// ChannelList is a collection of in-memory channels.
```
vaikas-google: fix comment? not in-memory channels?

config/300-channel.yaml
n3wscott: thanks, forgot this. Added.

PR #26, opened 2019-06-18
Adding Channel controller.

n3wscott:
Using  #24, stage the channel controller.


dicarlo2:
@n3wscott do you have an eta on when this will be merged? We're eager to use knative eventing with GCP PubSub and would prefer to use the new Channel CRD paradigm rather than the deprecated provisioners.

n3wscott:
@dicarlo2 I am trying to get this to work now, I think within a week. Attempting to make it happen before the 0.8 release of Knative.

pkg/apis/events/v1alpha1/channel_lifecycle.go
```diff
@@ -0,0 +1,69 @@
...
+func (cs *ChannelStatus) SetAddress(url *apis.URL) {
+	if cs.Address == nil {
+		cs.Address = &v1beta1.Addressable{}
+	}
+	if url != nil {
+		cs.Address.URL = url
+		channelCondSet.Manage(cs).MarkTrue(ChannelConditionAddressable)
+	} else {
+		cs.Address.URL = nil
+		channelCondSet.Manage(cs).MarkFalse(ChannelConditionAddressable, "emptyUrl", "url is empty")
```
spencer-p: Should there be a check for `url != nil && url.String() == ""`?

n3wscott: there is a fair amount of trust at this point of the app and if the controller wants to set a "" url, I guess that is its choice, I don't return an error here so there is no way to signal back up the call that this was odd.

config/300-channel.yaml
```diff
@@ -56,15 +56,11 @@ spec:
             secret:
               type: object
               # TODO: document we default.
-              description: "Credential to use to manage Cloud Pub/Sub. The value of the secret entry must be a service account key in the JSON format (see https://cloud.google.com/iam/docs/creating-managing-service-account-keys)."
+              description: "Credential to use to manage Cloud Pub/Sub. The value of the secret entry must be a service account key in the JSON format (see https://cloud.google.com/iam/docs/creating-managing-service-account-keys). Defaults to secret.name of 'google-cloud-key' and secret.key of 'key.json'."
             project:
               type: string
               # TODO: document we default.
```
vaikas-google: might as well document now?

pkg/apis/events/v1alpha1/channel_types.go
```diff
@@ -60,25 +59,21 @@ type ChannelSpec struct {
 	// Topic and Subscriptions. The value of the secret entry must be a
 	// service account key in the JSON format
 	// (see https://cloud.google.com/iam/docs/creating-managing-service-account-keys).
+	// +optional
```
vaikas-google: we should probably document the default behaviour here?

pkg/apis/events/v1alpha1/channel_types.go
```diff
@@ -60,25 +59,21 @@ type ChannelSpec struct {
...
 	Secret *corev1.SecretKeySelector `json:"secret,omitempty"`
 
 	// Project is the ID of the Google Cloud Project that the Pub/Sub
 	// Topic and Subscriptions will be created in.
```
vaikas-google: And here, esp. the bit about metadata?

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+// Reconciler implements controller.Reconciler for Channel resources.
+type Reconciler struct {
+	*reconciler.Base
+
+	// listers index properties about resources
+	channelLister      listers.ChannelLister
+	topicLister        pubsublisters.TopicLister
+	subscriptionLister pubsublisters.PullSubscriptionLister
+
+	tracker tracker.Interface // TODO: use tracker for sink.
```
vaikas-google: Consider turning this into an issue. Something about status of sinks might not get reflected, there might be issues with this on error conditions.

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+			Owner:      channel,
+			Name:       genName,
+			Project:    channel.Spec.Project,
+			Topic:      channel.Status.TopicID,
+			Secret:     channel.Spec.Secret,
+			Labels:     resources.GetPullSubscriptionLabels(controllerAgentName, channel.Name, genName),
+			Subscriber: s,
+		})
+		ps, err := c.RunClientSet.PubsubV1alpha1().PullSubscriptions(channel.Namespace).Create(ps)
+		if err != nil {
```
vaikas-google: Might want to check the error case of "already exists" if something created it already (name collision) and check the ownerref and if it's not us, propagate.

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+			Name:       genName,
+			Project:    channel.Spec.Project,
+			Topic:      channel.Status.TopicID,
+			Secret:     channel.Spec.Secret,
+			Labels:     resources.GetPullSubscriptionLabels(controllerAgentName, channel.Name, genName),
+			Subscriber: s,
+		})
+
+		existingPs, found := pullsubs[genName] //c.RunClientSet.PubsubV1alpha1().PullSubscriptions(channel.Namespace).Get(genName, metav1.GetOptions{})
+		//if apierrs.IsNotFound(err) {
```
vaikas-google: remove?

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+				if !found || got.ObservedGeneration != want.Generation {
+					subUpdates = append(subUpdates, want)
+				}
+			}
+			// Remove want from exists.
+			delete(exists, want.UID)
+		}
+	}
+
+	// Remaining exists will be deleted.
```
vaikas-google: This is a pretty long function, took me a while to get through and figure out under which cases what happens. Might want to include more comments here as well as what the overall method does. Also might consider breaking add/update/delete loops into their own methods?

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+	exists := make(map[types.UID]eventingduck.SubscriberStatus)
+	for _, s := range channel.Status.Subscribers {
+		exists[s.UID] = s
+	}
+
+	if channel.Spec.Subscribable != nil {
+		for _, want := range channel.Spec.Subscribable.Subscribers {
+			if got, ok := exists[want.UID]; !ok {
+				// If it does not exist, then update it.
+				subCreates = append(subCreates, want)
```
vaikas-google: update or create? Comment says update.

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+		channel.Status.Subscribers = append(channel.Status.Subscribers, eventingduck.SubscriberStatus{
+			UID:                s.UID,
+			ObservedGeneration: s.Generation,
+			// TODO: do I need the other fields?
+		})
+		return nil // Signal a re-reconcile.
+	}
+	for _, s := range subUpdates {
+		genName := resources.GenerateSubscriptionName(s.UID)
+		c.Logger.Infof("Channel %q will update subscription %s", channel.Name, genName)
```
vaikas-google: This is where I found some confusion, coupled with the lines below where we actually create it. They seem very similar to just create. Or put another way, why do you need subCreates and don't just throw everything into subUpdates. I think this is where the comments would help. It was difficult to visually diff what really was different about these two loops for the create case.

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+				channel.Status.MarkTopicReady()
+			} else if c.IsUnknown() {
+				channel.Status.MarkTopicOperating(c.Reason, c.Message)
+			} else if c.IsFalse() {
+				channel.Status.MarkNoTopic(c.Reason, c.Message)
+			}
+		}
+	}
+	// 2. Sync all subscriptions.
+	//   a. create all subscriptions that are in spec and not in status.
```
vaikas-google: Maybe also add the case about update. Also, consider moving these comments (also maybe expand a little bit) to the actual syncSubscribers. By the time I was reading that function, I didn't remember reading this comment, and tracking to the callsite to see the behaviour of the function is odd :)

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+	return nil
+}
+
+func (r *Reconciler) createTopic(ctx context.Context, channel *v1alpha1.Channel) (*pubsubv1alpha1.Topic, error) {
+	topic, err := r.getTopic(ctx, channel)
+	if err != nil && !apierrors.IsNotFound(err) {
+		logging.FromContext(ctx).Error("Unable to get an Topic", zap.Error(err))
+		return nil, err
+	}
+	if topic != nil {
```
vaikas-google: Should we check that we actually own this topic?

pkg/reconciler/channel/channel.go
```diff
@@ -0,0 +1,412 @@
...
+	}))
+	if err != nil {
+		logging.FromContext(ctx).Info("Topic created.", zap.Error(err), zap.Any("topic", topic))
+		r.Recorder.Eventf(channel, corev1.EventTypeNormal, "TopicCreated", "Created Topic %q", topic.GetName())
+	}
+	return topic, err
+}
+
+func (r *Reconciler) getTopic(ctx context.Context, channel *v1alpha1.Channel) (*pubsubv1alpha1.Topic, error) {
+	tl, err := r.RunClientSet.PubsubV1alpha1().Topics(channel.Namespace).List(metav1.ListOptions{
```
vaikas-google: Since you don't use generateName, you can just do a get instead of listing things?

pkg/reconciler/channel/channel.go
vaikas-google: ... time passes... you actually do check the owner, but if it's not owned by us, you return a notfound, which will try to recreate, which will fail.

vaikas-google: Serving does a get, verifies the owner and fails the resource as 'XYZNotOwned'

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -29,7 +29,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"knative.dev/pkg/kmeta"
 
-	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/apis/pubsub/v1alpha1"
+	v1alpha12 "github.com/GoogleCloudPlatform/cloud-run-events/pkg/apis/events/v1alpha1"
```
vaikas-google: typo? v1alpha12?

cmd/pubsub/invoker/main.go
```diff
@@ -0,0 +1,72 @@
...
+			logger.Fatal("failed to find project id. ", zap.Error(err))
+		}
+		env.Project = project
+	}
+
+	logger.Info("using project.", zap.String("project", env.Project))
+
+	startable := &invoker.Invoker{
+		ProjectID: env.Project,
+		// TODO: this is not done.
```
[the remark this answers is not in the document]
n3wscott: WHOOPS! This is old dead code. Thanks, deleted.

config/201-clusterrole.yaml
```diff
@@ -44,6 +44,21 @@ rules:
     - update
     - patch
 
+- apiGroups:
+    - events.cloud.run
```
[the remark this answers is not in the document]
n3wscott: that seems fine to me.

pkg/apis/events/v1alpha1/channel_types.go
n3wscott: it is documented in the openapi spec, but this bit of code is not tied to the webhook code and it could change.

pkg/apis/events/v1alpha1/channel_types.go
n3wscott: I want to do this as a followup because it is a concept that touches all the resources. I will add an issue to track.

config/300-channel.yaml
n3wscott: I filed an issue as a follow because it applies to all resouces

pkg/reconciler/channel/channel.go
n3wscott: turned into an issue.

pkg/reconciler/channel/channel.go
n3wscott: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/113

pkg/reconciler/channel/channel.go
n3wscott: I throw not Owned now. Thanks again.

pkg/reconciler/channel/channel.go
n3wscott: I now get, not list. Thanks.

pkg/reconciler/channel/channel.go
n3wscott: When this function gets reworked I will do this too.

pkg/reconciler/channel/channel.go
n3wscott: #113 will do this.

pkg/reconciler/channel/channel.go
n3wscott: Comment fixed

pkg/reconciler/channel/channel.go
n3wscott: I will try this out in the e2e tests and see what it does. It will error at the moment and that error will be logged but not as nice as you want.

PR #28, opened 2019-06-18
[WIP] Add Topic API and Controller.

n3wscott:


pkg/apis/pubsub/v1alpha1/topic_lifecycle.go
```diff
@@ -0,0 +1,104 @@
...
+		ts.Address.URL = url
+		topicCondSet.Manage(ts).MarkTrue(TopicConditionAddressable)
+	} else {
+		ts.Address.Hostname = ""
+		ts.Address.URL = nil
+		topicCondSet.Manage(ts).MarkFalse(TopicConditionAddressable, "emptyHostname", "hostname is the empty string")
+	}
+}
+
+func (ts *TopicStatus) PropagateDeploymentAvailability(d *appsv1.Deployment) {
```
mattmoor: I would use a ksvc per Topic, it makes just about every aspect of this trivial.

cmd/pubsub/subscription/main.go
```diff
@@ -97,6 +97,13 @@ func main() {
 	}
 
 	switch env.Action {
+	case "exists":
```
mattmoor: Split this into a separate change?

PR #32, opened 2019-06-19
Adding exists to pubsubs base.

n3wscott:
/assign @mattmoor 

breaks apart https://github.com/GoogleCloudPlatform/cloud-run-events/pull/28

### DIFF
--- a/config/300-pullsubscription.yaml
+++ b/config/300-pullsubscription.yaml
@@ -28,7 +28,6 @@ spec:
     - all
     - knative
     - pubsub
-    - eventing
     kind: PullSubscription
     plural: pullsubscriptions
   scope: Namespaced
--- a/config/500-controller.yaml
+++ b/config/500-controller.yaml
@@ -37,6 +37,10 @@ spec:
         env:
         - name: PUBSUB_RA_IMAGE
           value: github.com/GoogleCloudPlatform/cloud-run-events/cmd/pubsub/receive_adapter
+        - name: PUBSUB_PUBLISHER_IMAGE
+          value: github.com/GoogleCloudPlatform/cloud-run-events/cmd/pubsub/publisher
+        - name: PUBSUB_TOPIC_IMAGE
+          value: github.com/GoogleCloudPlatform/cloud-run-events/cmd/pubsub/topic
         - name: PUBSUB_SUB_IMAGE
           value: github.com/GoogleCloudPlatform/cloud-run-events/cmd/pubsub/subscription
         - name: SYSTEM_NAMESPACE
--- a/pkg/pubsub/adapter/adapter.go
+++ b/pkg/pubsub/adapter/adapter.go
@@ -17,11 +17,8 @@ limitations under the License.
 package adapter
 
 import (
-	"errors"
 	"fmt"
 
-	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/apis/pubsub/v1alpha1"
-	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/kncloudevents"
 	cloudevents "github.com/cloudevents/sdk-go"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport"
 	cepubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub"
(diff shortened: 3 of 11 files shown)


config/300-pullsubscription.yaml
```diff
@@ -28,7 +28,6 @@ spec:
     - all
     - knative
     - pubsub
-    - eventing
```
grantr: Why remove this?

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -61,15 +61,23 @@ const (
 	OpsJobOngoing            OpsJobStatus = "JOB_ONGOING"
 )
 
-func (c *PubSubBase) EnsureSubscription(ctx context.Context, owner kmeta.OwnerRefable, project, topic, subscription string) (OpsJobStatus, error) {
+func (c *PubSubBase) EnsureSubscriptionExists(ctx context.Context, owner kmeta.OwnerRefable, project, topic, subscription string) (OpsJobStatus, error) {
```
grantr: Where is this function used, or is that in future PRs?

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -61,15 +61,23 @@ const (
...
 func (c *PubSubBase) EnsureSubscriptionDeleted(ctx context.Context, owner kmeta.OwnerRefable, project, topic, subscription string) (OpsJobStatus, error) {
 	return c.ensureSubscriptionJob(ctx, operations.ActionDelete, owner, project, topic, subscription)
 }
 
-func (c *PubSubBase) EnsureTopic(ctx context.Context, owner kmeta.OwnerRefable, project, topic string) (OpsJobStatus, error) {
+func (c *PubSubBase) EnsureTopicExists(ctx context.Context, owner kmeta.OwnerRefable, project, topic string) (OpsJobStatus, error) {
+	return c.ensureTopicJob(ctx, operations.ActionExists, owner, project, topic)
+}
+
+func (c *PubSubBase) EnsureTopicCreated(ctx context.Context, owner kmeta.OwnerRefable, project, topic string) (OpsJobStatus, error) {
```
grantr: Where are there EnsureTopic* functions used, or is that in future PRs?

pkg/pubsub/adapter/adapter.go
```diff
@@ -86,13 +85,6 @@ func (a *Adapter) Start(ctx context.Context) error {
 	return a.inbound.StartReceiver(ctx, a.receive)
 }
 
-func (a *Adapter) DeletePubSubSubscription(ctx context.Context) error {
```
grantr: Did this move somewhere else, or was it dead code?

config/500-controller.yaml
```diff
@@ -37,6 +37,10 @@ spec:
         env:
         - name: PUBSUB_RA_IMAGE
           value: github.com/GoogleCloudPlatform/cloud-run-events/cmd/pubsub/receive_adapter
+        - name: PUBSUB_PUBLISHER_IMAGE
```
grantr: Where are these new images used, or is that in future PRs?

config/300-pullsubscription.yaml
n3wscott: the pullsubscription is not part of eventing, it is part of pubsub was the reasoning. The channel will be part of eventing, and moved to events.cloud.run group.

pkg/reconciler/pubsub/reconciler.go
n3wscott: I do not use it at the moment, I added it for symmetry to topic as this is a base lib.

pkg/reconciler/pubsub/reconciler.go
n3wscott: Future PR, because I split a larger PR up into parts. This is one part that uses Topic, but the impl of topic is later.

pkg/pubsub/adapter/adapter.go
n3wscott: this was removed from adapter, it was implemented and then replaced with PubSubBase, and this PR is an improvement to PubSubBase v1

config/500-controller.yaml
n3wscott: future PR. they are in the stack...

PR #34, opened 2019-06-19
Adding Topic API.

n3wscott:
breaks apart #28

/assign @mattmoor


config/300-topic.yaml
```diff
@@ -0,0 +1,101 @@
...
+# limitations under the License.
+apiVersion: apiextensions.k8s.io/v1beta1
+kind: CustomResourceDefinition
+metadata:
+ name: topics.pubsub.cloud.run
+ labels:
+   events.cloud.run/release: devel
+   events.cloud.run/crd-install: "true"
+spec:
+  group: pubsub.cloud.run
```
akashrv: Rather than exposing topic and pull_subscriptions as public interface, we should expose the importer as the public resource (which will leverage pull_subscription). And keep pull_subscription and topic as internal objects.

pkg/apis/pubsub/v1alpha1/topic_validation.go
```diff
@@ -0,0 +1,73 @@
...
+)
+
+func (t *Topic) Validate(ctx context.Context) *apis.FieldError {
+	return t.Spec.Validate(ctx).ViaField("spec")
+}
+
+func (ts *TopicSpec) Validate(ctx context.Context) *apis.FieldError {
+	var errs *apis.FieldError
+
+	if ts.Topic == "" {
```
akashrv: make Topic required in openapi schema too.

config/300-topic.yaml
```diff
@@ -0,0 +1,101 @@
...
+              description: "Credential to use to manage Cloud Pub/Sub. The value of the secret entry must be a service account key in the JSON format (see https://cloud.google.com/iam/docs/creating-managing-service-account-keys)."
+            project:
+              type: string
+              # TODO: document we default.
+              description: "ID of the Google Cloud Project to own the Pub/Sub credentials. E.g. 'my-project-1234' rather than its display name, 'My Project' or its number '1234567890'."
+            topic:
+              type: string
+              description: "ID of the Cloud Pub/Sub Topic to create. It must be in the form of the unique identifier within the project, not the entire name. E.g. it must be 'laconia', not 'projects/my-gcp-project/topics/laconia'."
+            propagationPolicy:
+              type: string
```
akashrv: Doesn't OpenAPI schema support enums - for policy?

akashrv:
/hold 
we need to discuss which objects we would make public v/s internal.

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,149 @@
...
+	// +optional
+	Status TopicStatus `json:"status,omitempty"`
+}
+
+// Check that Topic can be validated, can be defaulted, and has immutable fields.
+var _ runtime.Object = (*Topic)(nil)
+var _ webhook.GenericCRD = (*Topic)(nil)
+
+// Check that PullSubscription implements the Conditions duck type.
+var _ = duck.VerifyType(&Topic{}, &duckv1beta1.Conditions{})
```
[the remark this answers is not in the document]
n3wscott: those are part of webhook.GenericCRD

pkg/apis/pubsub/v1alpha1/topic_lifecycle.go
```diff
@@ -0,0 +1,104 @@
...
+		ts.Address = &v1alpha1.Addressable{}
+	}
+	if url != nil {
+		ts.Address.Hostname = url.Host
+		ts.Address.URL = url
+		topicCondSet.Manage(ts).MarkTrue(TopicConditionAddressable)
+	} else {
+		ts.Address.Hostname = ""
+		ts.Address.URL = nil
+		topicCondSet.Manage(ts).MarkFalse(TopicConditionAddressable, "emptyHostname", "hostname is the empty string")
```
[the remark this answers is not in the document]
n3wscott: when we drop the alpha support I will.

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,149 @@
...
+	// * ObservedGeneration - the 'Generation' of the Service that was last processed by the controller.
+	// * Conditions - the latest available observations of a resource's current state.
+	duckv1beta1.Status `json:",inline"`
+
+	// Topic is Addressable. It currently exposes the endpoint as a
+	// fully-qualified DNS name which will distribute traffic over the
+	// provided targets from inside the cluster.
+	//
+	// It generally has the form {Topic}.{namespace}.svc.{cluster domain name}
+	duckv1alpha1.AddressStatus `json:",inline"`
```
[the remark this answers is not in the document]
n3wscott: I am mirroring what is in eventing. I will migrate when we migrate eventing.

config/300-topic.yaml
n3wscott: they are part of pubsub.cloud.run, you can choose to view that as internal.

config/300-topic.yaml
n3wscott: hm, we have never used that feature. Followup.

n3wscott: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/38

n3wscott:
/hold cancel
will follow up about names: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/39

config/300-topic.yaml
grantr: The `Channel` type is also part of pubsub.cloud.run. Will that move to a different api group?

pkg/apis/pubsub/v1alpha1/topic_defaults.go
```diff
@@ -0,0 +1,29 @@
...
+
+package v1alpha1
+
+import "context"
+
+func (t *Topic) SetDefaults(ctx context.Context) {
+	t.Spec.SetDefaults(ctx)
+}
+
+func (ts *TopicSpec) SetDefaults(ctx context.Context) {
```
grantr: The comments in the OpenAPI schema above note that several fields are defaulted, but I don't see their defaults set here. Are they set elsewhere?

pkg/apis/pubsub/v1alpha1/topic_lifecycle.go
```diff
@@ -0,0 +1,104 @@
...
+				ts.MarkDeployed()
+			case "False":
+				ts.MarkNotDeployed(cond.Reason, cond.Message)
+			default:
+				ts.MarkDeploying(cond.Reason, cond.Message)
+			}
+			return
+		}
+	}
+	ts.MarkDeploying("DeploymentStatus", "Failed to inspect Deployment status.")
```
grantr: This message should be clearer about what actually happened (the deployment didn't have the expected condition).

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+}
+
+// Check that Topic can be validated, can be defaulted, and has immutable fields.
+var _ runtime.Object = (*Topic)(nil)
+var _ webhook.GenericCRD = (*Topic)(nil)
+
+// Check that Topic implements the Conditions duck type.
+var _ = duck.VerifyType(&Topic{}, &duckv1beta1.Conditions{})
+
+// TopicSpec defines an parameters for creating or publishing to a Cloud
```
grantr: typo nit: -an

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+// Pub/Sub Topic depending the PropagationPolicy.
```
grantr: typo nit: depending **on**

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+type TopicSpec struct {
+	// Secret is the credential to be used to create and publish into the
+	// Cloud Pub/Sub Topic. The value of the secret entry must be a service
+	// account key in the JSON format
```
grantr: Missing period. What should the key for this value be?

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+// Check that Topic can be validated, can be defaulted, and has immutable fields.
+var _ runtime.Object = (*Topic)(nil)
+var _ webhook.GenericCRD = (*Topic)(nil)
+
+// Check that Topic implements the Conditions duck type.
+var _ = duck.VerifyType(&Topic{}, &duckv1beta1.Conditions{})
+
+// TopicSpec defines an parameters for creating or publishing to a Cloud
+// Pub/Sub Topic depending the PropagationPolicy.
+type TopicSpec struct {
```
grantr: Are any of these values defaulted?

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+	// Topic will be created in or used from.
+	Project string `json:"project,omitempty"`
+
+	// Topic is the ID of the Topic to create/use in Google Cloud Pub/Sub.
+	Topic string `json:"topic,omitempty"`
+
+	//PropagationPolicy defines how Topic controls the Cloud Pub/Sub topic for lifecycle changes.
+	PropagationPolicy PropagationPolicyType `json:"propagationPolicy,omitempty"`
+
+	// ServiceAccoutName is the name of the ServiceAccount that will be used to
```
grantr: typo: `ServiceAccountName`

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+	// run the Topic Deployment.
+	ServiceAccountName string `json:"serviceAccountName,omitempty"`
+}
+
+// PropagationPolicyType defines enum type for TopicPolicy
+type PropagationPolicyType string
+
+const (
+	// TopicPolicyCreateDelete defines the Cloud Pub/Sub topic management
+	// policy for creating topic if not present, and delete topic when the
```
grantr: creating and delete don't match; they should be creating and deleting

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+	// Topic resource is deleted.
+	TopicPolicyCreateDelete PropagationPolicyType = "CreateDelete"
+
+	// TopicPolicyCreateRestrictDelete defines the Cloud Pub/Sub topic
+	// management policy for creating topic if not present, and do not delete
+	// topic when the Topic resource is deleted.
+	TopicPolicyCreateRestrictDelete PropagationPolicyType = "CreateRestrictDelete"
```
grantr: Restrict is a bit confusing for me because it implies there is a range of possible restrictions that could be applied. It's harder to determine what the behavior will be. I suggest renaming to something like `CreateNoDelete` or `CreateNeverDelete`.

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+
+	// TopicPolicyRestrictCreateRestrictDelete defines the Cloud Pub/Sub topic
+	// management policy for only using existing topics, and do not delete
```
grantr: These tenses don't match and the pluralization of topics doesn't match the other comments. Should be `only using an existing topic, and not deleting a topic`

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+type PropagationPolicyType string
+
+const (
+	// TopicPolicyCreateDelete defines the Cloud Pub/Sub topic management
+	// policy for creating topic if not present, and delete topic when the
+	// Topic resource is deleted.
+	TopicPolicyCreateDelete PropagationPolicyType = "CreateDelete"
+
+	// TopicPolicyCreateRestrictDelete defines the Cloud Pub/Sub topic
+	// management policy for creating topic if not present, and do not delete
```
grantr: These tenses don't match. Should be `and not deleting topic`

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+	TopicConditionPublisherReady,
+)
+
+const (
+	// TopicConditionReady has status True when all subconditions below have
+	// been set to True.
+	TopicConditionReady = apis.ConditionReady
+
+	// TopicConditionAddressable has status true when this Topic meets the
+	// Addressable contract and has a non-empty hostname.
```
grantr: `hostname or URL`? Shouldn't we target the v1beta1 Addressable even if we support v1alpha1?

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+	TopicConditionAddressable apis.ConditionType = "Addressable"
+
+	// TopicConditionTopicExists has status True when the Topic has had a
+	// Pub/Sub topic created for it.
+	TopicConditionTopicExists apis.ConditionType = "TopicExists"
+
+	// TopicConditionPublisherReady has status True when the Topic has had
+	// its publisher deployment created.
```
grantr: Created and ready, or just created?

pkg/apis/pubsub/v1alpha1/topic_types.go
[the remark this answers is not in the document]
grantr: Seems like if we target v1beta1 right away, we can avoid the work of deprecating v1alpha1 in this code and creating a new release for that. What's the downside of using v1beta1? Will we be incompatible with older deployments?

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+	duckv1beta1.Status `json:",inline"`
+
+	// Topic is Addressable. It currently exposes the endpoint as a
+	// fully-qualified DNS name which will distribute traffic over the
+	// provided targets from inside the cluster.
+	//
+	// It generally has the form {Topic}.{namespace}.svc.{cluster domain name}
+	duckv1alpha1.AddressStatus `json:",inline"`
+
+	// ProjectID is the resolved project ID in use by the Topic.
```
grantr: Is this different from the projectID in the spec?

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,160 @@
...
+	// +optional
+	ProjectID string `json:"projectId,omitempty"`
+
+	// TopicID is the created topic ID used by the Topic.
```
grantr: Is this different from the topicID in the spec?

pkg/apis/pubsub/v1alpha1/topic_validation_test.go
```diff
@@ -0,0 +1,121 @@
...
+			},
+			Key: "secret-key",
+		},
+		Project:            "my-eventing-project",
+		Topic:              "pubsub-topic",
+		ServiceAccountName: "service-account-name",
+	}
+)
+
+func TestTopicValidation(t *testing.T) {
```
grantr: Can you add a test using an invalid propagationPolicy value?

pkg/apis/pubsub/v1alpha1/topic_types.go
n3wscott: project id in spec is optional.

pkg/apis/pubsub/v1alpha1/topic_types.go
n3wscott: this one is the same, I bet I can delete this... I added it just if I wanted to support generate topic. punt on that.

pkg/apis/pubsub/v1alpha1/topic_validation_test.go
n3wscott: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/40

pkg/apis/pubsub/v1alpha1/topic_defaults.go
grantr: Why set them in the controller? Is the default impossible/difficult to set in the webhook?

pkg/apis/pubsub/v1alpha1/topic_lifecycle.go
grantr: Sure, and that should be explained in the message. :)

pkg/apis/pubsub/v1alpha1/topic_types.go
grantr: That doc doesn't tell me what the content of the Secret should be. Should there be a single key with a JSON string value? If so what is the key? Should the JSON fields be elevated into the Secret as individual kv pairs?

It would be fine to reference some other doc that explains how the Secret should look, but the GCP doc referenced doesn't say anything about how to format a k8s Secret.

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,161 @@
...
+var _ = duck.VerifyType(&Topic{}, &duckv1beta1.Conditions{})
+
+// TopicSpec defines parameters for creating or publishing to a Cloud Pub/Sub
+// Topic depending on the PropagationPolicy.
+type TopicSpec struct {
+	// Secret is the credential to be used to create and publish into the
+	// Cloud Pub/Sub Topic. The value of the secret entry must be a service
+	// account key in the JSON format
+	// (see https://cloud.google.com/iam/docs/creating-managing-service-account-keys).
+	Secret *corev1.SecretKeySelector `json:"secret,omitempty"`
```
grantr: Does this need the `+optional` tag?

pkg/apis/pubsub/v1alpha1/topic_types.go
```diff
@@ -0,0 +1,161 @@
...
+
+	// Project is the ID of the Google Cloud Project that the Pub/Sub
+	// Topic will be created in or used from.
+	Project string `json:"project,omitempty"`
```
grantr: Does this need the +optional tag?

pkg/apis/pubsub/v1alpha1/topic_types.go
grantr: I'm fine with this sticking around as long as it's always filled in

akashrv:
> /hold cancel
> will follow up about names: #39

Just for the record, Scott and I had an offline discussion and decided to unblock development with the current apigroup and names, and follow-up on it in a separate issue based on the PRD and UX that Donna is working on.

pkg/apis/pubsub/v1alpha1/topic_types.go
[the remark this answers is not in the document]
akashrv: Eventing already checks the existence of beta field - url, and then if it is not there then falls back to hostname. Given that there is no need to add this now and then deprecate later, which will be unnecessary work. You can directly got with beta, unless you know of other reasons.

n3wscott:
/hold cancel

I will fix the optional issues in the controller PR to follow.

PR #35, opened 2019-06-19
Adding Topic Controller

### LINKED ISSUE #50: Pubsub Publisher should be a ksvc.
Publisher should be a ksvc.

_Originally posted by @n3wscott in https://github.com/GoogleCloudPlatform/cloud-run-events/pull/35/review_comment/create_

n3wscott:
fixes #50


n3wscott:
dep on https://github.com/GoogleCloudPlatform/cloud-run-events/pull/34

pkg/reconciler/topic/resources/publisher.go
```diff
@@ -0,0 +1,139 @@
...
+		LocalObjectReference: corev1.LocalObjectReference{
+			Name: "google-cloud-key",
+		},
+		Key: "key.json",
+	}
+}
+
+// MakePublisher generates (but does not insert into K8s) the Invoker Deployment for
+// Channels.
+func MakePublisher(args *PublisherArgs) *v1.Deployment {
```
n3wscott: Publisher should be a ksvc.

n3wscott: Will fix: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/50

n3wscott: fixed.

pkg/apis/pubsub/v1alpha1/topic_lifecycle.go
```diff
@@ -55,21 +53,19 @@ func (ts *TopicStatus) SetAddress(url *apis.URL) {
...
+			ts.MarkDeployed()
+		case "False":
+			ts.MarkNotDeployed(ready.Reason, ready.Message)
+		default:
+			ts.MarkDeploying(ready.Reason, ready.Message)
 		}
+		return
 	}
-	ts.MarkDeploying("DeploymentStatus", "Failed to inspect Deployment status.")
+	ts.MarkDeploying("PublisherStatus", "Failed to Publisher has no Ready type status.")
```
jonjohnsonjr: bad english

pkg/apis/pubsub/v1alpha1/topic_lifecycle.go
```diff
@@ -55,21 +53,19 @@ func (ts *TopicStatus) SetAddress(url *apis.URL) {
...
-			case "False":
-				ts.MarkNotDeployed(cond.Reason, cond.Message)
-			default:
-				ts.MarkDeploying(cond.Reason, cond.Message)
-			}
-			return
+func (ts *TopicStatus) PropagatePublisherStatus(ready *apis.Condition) {
+	if ready != nil {
+		switch ready.Status {
+		case "True":
```
jonjohnsonjr: use ready.Status.IsTrue, etc.

pkg/reconciler/testing/topic.go
```diff
@@ -0,0 +1,151 @@
...
+		s.UID = uid
+	}
+}
+
+// WithInitTopicConditions initializes the Topics's conditions.
+func WithInitTopicConditions(s *v1alpha1.Topic) {
+	s.Status.InitializeConditions()
+}
+
+func WithTopicTopic(topicID string) TopicOption {
```
grantr: :smile: Should this be `WithTopicTopic` or `WithTopicTopicID`?

pkg/reconciler/topic/controller.go
```diff
@@ -0,0 +1,99 @@
...
+}
+
+// NewController initializes the controller and is called by the generated code
+// Registers event handlers to enqueue events
+func NewController(
+	ctx context.Context,
+	cmw configmap.Watcher,
+) *controller.Impl {
+
+	//deploymentInformer := deploymentinformer.Get(ctx)
```
grantr: Dead code, delete?

pkg/reconciler/topic/controller.go
```diff
@@ -0,0 +1,99 @@
...
+	"knative.dev/pkg/logging"
+
+	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/apis/pubsub/v1alpha1"
+	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/reconciler"
+	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/reconciler/pubsub"
+
+	serviceinformer "github.com/knative/serving/pkg/client/injection/informers/serving/v1beta1/service"
+	jobinformer "knative.dev/pkg/injection/informers/kubeinformers/batchv1/job"
+
+	topicinformers "github.com/GoogleCloudPlatform/cloud-run-events/pkg/client/injection/informers/pubsub/v1alpha1/topic"
```
grantr: nit: why is this plural and the other informers aren't?

pkg/reconciler/topic/resources/labels.go
```diff
@@ -0,0 +1,32 @@
...
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+func GetLabelSelector(controller, source string) labels.Selector {
+	return labels.SelectorFromSet(GetLabels(controller, source))
+}
+
+func GetLabels(controller, topic string) map[string]string {
+	return map[string]string{
+		"cloud-run-pubsub-topic":      controller,
```
grantr: Should the label prefix here be `pubsub.cloud.run/` to be consistent with the rest of knative?

pkg/reconciler/topic/resources/publisher.go
```diff
@@ -0,0 +1,114 @@
...
+)
+
+// DefaultSecretSelector is the default secret selector used to load the creds
+// for the publisher to auth with Google Cloud.
+func DefaultSecretSelector() *corev1.SecretKeySelector {
+	return &corev1.SecretKeySelector{
+		LocalObjectReference: corev1.LocalObjectReference{
+			Name: "google-cloud-key",
+		},
+		Key: "key.json",
```
grantr: This `key.json` is what I wanted to be documented earlier in the Topic type's Secret ref field. How does the user know to use this key otherwise?

pkg/reconciler/topic/resources/publisher.go
```diff
@@ -0,0 +1,114 @@
...
+				},
+			},
+		}},
+	}
+
+	return &servingv1beta1.Service{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:       args.Topic.Namespace,
+			GenerateName:    fmt.Sprintf("pubsub-publisher-%s-", args.Topic.Name),
+			Labels:          args.Labels, // TODO: not sure we should use labels like this.
```
grantr: What's the problem with using labels like this?

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,420 @@
...
+
+	} else if _, updated, fErr := c.updateFinalizers(ctx, Topic); fErr != nil {
+		logger.Warnw("Failed to update Topic finalizers", zap.Error(fErr))
+		c.Recorder.Eventf(Topic, corev1.EventTypeWarning, "UpdateFailed",
+			"Failed to update finalizers for Topic %q: %v", Topic.Name, fErr)
+		return fErr
+	} else if updated {
+		// There was a difference and updateFinalizers said it updated and did not return an error.
+		c.Recorder.Eventf(Topic, corev1.EventTypeNormal, "Updated", "Updated Topic %q finalizers", Topic.GetName())
+	}
```
grantr: This if-else sequence is not clear to me. I think it would be clearer as two nested ifs:

```go
if !equality.Semantic.DeepEqual(original.Finalizers, Topic.Finalizers) {
  _, updated, fErr := c.updateFinalizers(ctx, Topic)
  if ferr != nil {
    // handle error
    // if this were a helper function, you could return here and avoid the else
  } else if updated {
    // handle success
  }
}
```

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,420 @@
...
+
+	if equality.Semantic.DeepEqual(original.Status, Topic.Status) {
```
grantr: Same as above, this is clearer as an outer if

```go
if !equality.Semantic.DeepEqual(original.Status, Topic.Status) {
  if _, uErr := c.updateStatus(ctx, Topic); uErr !== nil {
    // handle error and return
  }
  if reconcileErr == nil {
    // notify update happened
  }
}
```

Gopkg.toml
```diff
@@ -26,6 +26,11 @@ required = [
   # HEAD as of 2919-06-26
   revision = "677d3d17e84b7337f006220d6bb465af4f416f81"
 
+[[override]]
+  name = "github.com/knative/serving"
+  # HEAD as of 2919-06-26
+  revision = "50cdf8762deae58ed89763adef26a88e7efecac7"
```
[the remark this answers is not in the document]
n3wscott: no I am using `knative.dev/pkg`

cmd/pubsub/publisher/main.go
```diff
@@ -0,0 +1,77 @@
...
+
+import (
+	"flag"
+	"log"
+
+	"cloud.google.com/go/compute/metadata"
+	"github.com/kelseyhightower/envconfig"
+	"go.uber.org/zap"
+	"go.uber.org/zap/zapcore"
+	"golang.org/x/net/context"
```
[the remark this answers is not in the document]
n3wscott: ah weird... damn you IDE.

Gopkg.toml
```diff
@@ -26,6 +26,11 @@ required = [
   # HEAD as of 2919-06-26
   revision = "677d3d17e84b7337f006220d6bb465af4f416f81"
 
+[[override]]
```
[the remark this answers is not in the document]
n3wscott: Constraints:

- Can be declared by any project's manifest, yours or a dependency
- Apply only to direct dependencies of the project declaring the constraint
- Must not conflict with the constraint entries declared in any other project's manifest

Overrides:

- Are only utilized from the current/your project's manifest
- Apply globally, to direct and transitive dependencies
- Supersede constraints declared in all manifests, yours or a dependency's

I think I want overrides.

cmd/pubsub/publisher/main.go
```diff
@@ -0,0 +1,77 @@
...
+
+	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/pubsub/publisher"
+)
+
+type envConfig struct {
+	// Environment variable containing project id.
+	Project string `envconfig:"PROJECT_ID"`
```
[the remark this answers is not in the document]
n3wscott: There are many samples that use `PROJECT_ID` and I was trying to match that.

cmd/pubsub/publisher/main.go
```diff
@@ -0,0 +1,77 @@
...
+
+	// Topic is the environment variable containing the PubSub Topic being
+	// subscribed to's name. In the form that is unique within the project.
+	// E.g. 'laconia', not 'projects/my-gcp-project/topics/laconia'.
+	Topic string `envconfig:"PUBSUB_TOPIC_ID" required:"true"`
```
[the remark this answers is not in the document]
n3wscott: There are many samples that use `TOPIC_ID` and I was trying to match that.

pkg/pubsub/publisher/publisher.go
```diff
@@ -0,0 +1,91 @@
...
+	"golang.org/x/net/context"
+
+	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/kncloudevents"
+)
+
+// Publisher implements the Pub/Sub adapter to deliver Pub/Sub messages from a
+// pre-existing topic/subscription to a Sink.
+type Publisher struct {
+	// ProjectID is the pre-existing eventing project id to use.
+	ProjectID string
```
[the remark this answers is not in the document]
n3wscott: matching the style of the cloudevents lib.

pkg/reconciler/testing/topic.go
n3wscott: topic == topicID for better or worse.

pkg/reconciler/topic/resources/publisher.go
n3wscott: ack https://github.com/GoogleCloudPlatform/cloud-run-events/issues/64

pkg/reconciler/topic/resources/publisher.go
n3wscott: It depends on the upstream caller to use the correct label creation method.

pkg/reconciler/topic/topic.go
n3wscott: I am keeping formatting with sample-controller:

https://github.com/knative/sample-controller/blob/master/pkg/reconciler/addressableservice/addressableservice.go#L92-L106

cmd/pubsub/publisher/main.go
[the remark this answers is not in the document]
grantr: Choose one or the other (either is fine) and make all variables match. So if we want to match samples, this should be `ProjectID` everywhere.

pkg/reconciler/topic/topic.go
n3wscott: https://github.com/knative/sample-controller/blob/master/pkg/reconciler/addressableservice/addressableservice.go#L92-L106

cmd/pubsub/publisher/main.go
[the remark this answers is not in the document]
grantr: Choose one or the other (either is fine) and make all variables match. So if we want to match samples, this should be `TopicID` everywhere.

pkg/pubsub/publisher/publisher.go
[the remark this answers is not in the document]
grantr: Let's make all the variables `ProjectID` and `TopicID` then.

pkg/reconciler/testing/topic.go
grantr: Worse :) Make this `WithTopicID` so it doesn't stutter.

pkg/reconciler/topic/resources/publisher.go
grantr: Can you add that to the comment so future maintainers will know what you were thinking about?

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,420 @@
...
+		switch state {
+		case pubsub.OpsJobGetFailed:
+			logger.Error("Failed to get topic ops job.",
+				zap.Any("propagationPolicy", topic.Spec.PropagationPolicy),
+				zap.Any("state", state),
+				zap.Error(err))
+			return err
+
+		case pubsub.OpsJobCreated:
+			// If we created a job to make a subscription, then add the finalizer and update the status.
```
grantr: Seems like this comment should be `// If we created a job to make a topic`

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,420 @@
...
```
grantr: This comment seems inaccurate... doesn't this job just verify the topic exists? Or does it create a subscription too?

pkg/reconciler/testing/topic.go
n3wscott: `WithTopicID` will conflict with `WithChannelTopicID` and `WithPullSubscriptionTopic`

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,420 @@
...
+		case v1alpha1.TopicPolicyCreateDelete:
+			// Ensure the Topic is deleted.
+			state, err := c.EnsureTopicDeleted(ctx, topic, topic.Spec.Project, topic.Status.TopicID)
+			switch state {
+			case pubsub.OpsJobGetFailed:
+				logger.Error("Failed to get Topic ops job.", zap.Any("state", state), zap.Error(err))
+				return err
+
+			case pubsub.OpsJobCreated:
+				// If we created a job to make a subscription, then add the finalizer and update the status.
```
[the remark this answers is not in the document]
n3wscott: the entire reason for finalizers is to allow the controller to see and act on deleted objects before removal.

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,420 @@
...
+					"Created Job to delete topic %q.",
+					topic.Status.TopicID)
+				return nil
+
+			case pubsub.OpsJobCompleteSuccessful:
+				topic.Status.MarkNoTopic("Deleted", "Successfully deleted topic %q.", topic.Status.TopicID)
+				topic.Status.TopicID = ""
+				removeFinalizer(topic)
+
+			case pubsub.OpsJobCreateFailed, pubsub.OpsJobCompleteFailed:
```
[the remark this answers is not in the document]
n3wscott: I don't really agree.

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,420 @@
...
+
+		return nil
+	}
+
+	// Set the topic being used.
+	topic.Status.TopicID = topic.Spec.Topic
+
+	switch topic.Spec.PropagationPolicy {
+	case v1alpha1.TopicPolicyCreateDelete, v1alpha1.TopicPolicyCreateNoDelete:
+		state, err := c.EnsureTopicCreated(ctx, topic, topic.Spec.Project, topic.Status.TopicID)
```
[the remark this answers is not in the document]
n3wscott: in the controller for topic:

```golang
	jobInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
		FilterFunc: controller.Filter(v1alpha1.SchemeGroupVersion.WithKind("Topic")),
		Handler:    controller.HandleAll(impl.EnqueueControllerOf),
	})
```

pkg/reconciler/topic/topic.go
n3wscott: oops. copy/pasta

pkg/reconciler/topic/topic.go
n3wscott: yeah, updated, thanks

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+		return nil, false, err
+	}
+
+	// Don't modify the informers copy.
+	existing := source.DeepCopy()
+
+	var finalizers []string
+
+	// If there's nothing to update, just return.
+	exisitingFinalizers := sets.NewString(existing.Finalizers...)
```
grantr: Typo.

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+		finalizers = append(existing.Finalizers, finalizerName)
+	} else {
+		if !exisitingFinalizers.Has(finalizerName) {
+			// Nothing to do.
+			return desired, false, nil
+		}
+		// Remove the finalizer.
+		exisitingFinalizers.Delete(finalizerName)
+		finalizers = exisitingFinalizers.List()
+	}
```
grantr: This code seems unnecessary. `updateFinalizers` is only called if we changed the finalizers during reconcile. IIUC the only thing this is protecting against is if something else changes the object while we reconciled, but we can let the patch fail and try again, right?

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+			"resourceVersion": existing.ResourceVersion,
+		},
+	}
+
+	patch, err := json.Marshal(mergePatch)
+	if err != nil {
+		return desired, false, err
+	}
+
+	update, err := c.RunClientSet.PubsubV1alpha1().Topics(existing.Namespace).Patch(existing.Name, types.MergePatchType, patch)
```
grantr: Why use a patch instead of an update here?

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+	return update, true, err
+}
+
+func addFinalizer(s *v1alpha1.Topic) {
```
grantr: This should be a method on the type.

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+	finalizers := sets.NewString(s.Finalizers...)
+	finalizers.Insert(finalizerName)
+	s.Finalizers = finalizers.List()
+}
+
+func removeFinalizer(s *v1alpha1.Topic) {
```
grantr: This should be a method on the type.

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+		Labels: resources.GetLabels(controllerAgentName, topic.Name),
+	})
+
+	if existing == nil {
+		svc, err := r.ServingClientSet.ServingV1beta1().Services(topic.Namespace).Create(desired)
+		logging.FromContext(ctx).Desugar().Info("Publisher created.", zap.Error(err), zap.Any("publisher", svc))
+		return svc, err
+	}
+
+	if diff := cmp.Diff(desired.Spec, existing.Spec); diff != "" {
```
grantr: Use Semantic.DeepEqual here, not cmp.Diff. cmp.Diff is meant for tests. https://github.com/google/go-cmp

pkg/reconciler/topic/topic_test.go
```diff
@@ -0,0 +1,560 @@
...
+	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/reconciler/pubsub"
+	"github.com/GoogleCloudPlatform/cloud-run-events/pkg/reconciler/topic/resources"
+
+	. "knative.dev/pkg/reconciler/testing"
+
+	. "github.com/GoogleCloudPlatform/cloud-run-events/pkg/reconciler/testing"
+)
+
+const (
+	topicName = "hubbub"
```
grantr: pubsubhubbub 😃

pkg/reconciler/topic/topic.go
n3wscott: scope creep. followup.

pkg/reconciler/topic/topic_test.go
evankanderson: This is now WebSub:

https://github.com/pubsubhubbub/
and is a W3C recommendation:
https://www.w3.org/TR/websub/

pkg/reconciler/topic/topic.go
n3wscott: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/68

pkg/reconciler/topic/topic.go
n3wscott: we are only trying to change the finalizers

pkg/reconciler/topic/topic.go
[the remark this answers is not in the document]
grantr: I'm saying the comment is wrong - it says this block is adding a finalizer but it's not.

pkg/reconciler/topic/topic.go
grantr: Update semantics are the same - either way, ResourceVersion will cause the op to fail if it changes in the meantime.

pkg/reconciler/topic/topic_test.go
grantr: Didn't know it had become a W3C rec, nice! I'd say we should implement it in knative if anybody actually used it.

Gopkg.toml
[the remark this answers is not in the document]
grantr: I still don't understand why this needs to be an override. From https://golang.github.io/dep/docs/Gopkg.toml.html#override:

> Overrides are primarily intended as a way of eliminating disagreements between multiple irreconcilable [[constraint]] declarations on a single dependency. However, they will also be your primary recourse if you need to constrain a transitive dependency's version?

> Overrides should be used cautiously and temporarily, when possible.

Are you eliminating disagreements between constraints, or constraining a transitive dependency's version? If so, explain which in a comment above the override.e

Gopkg.toml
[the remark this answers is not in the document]
grantr: IIUC, you can't use version 0.7.0 because you have an existing dependency on `knative.dev/pkg` and serving 0.7 uses `github.com/knative/pkg`. I suggest using `github.com/knative/pkg` to avoid this and depending on serving 0.7, but if you don't want to do that, add a comment explaining why this is a revision constraint so future maintainers know when it can be updated to a version constraint.

cmd/pubsub/publisher/main.go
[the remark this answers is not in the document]
n3wscott: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/72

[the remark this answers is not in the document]
n3wscott: I don't want to do this. It makes sense for it to be just project in the spec and PROJECT_ID in the env vars. Defer with #72.

cmd/pubsub/publisher/main.go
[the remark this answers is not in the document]
n3wscott: defer. #72

pkg/pubsub/publisher/publisher.go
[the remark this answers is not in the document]
n3wscott: defer #72

n3wscott:
Let's follow back up on Project vs ProjectID. I prefer just Project; but there are samples that use either depending on author.

Merge this so we can make channels.

Gopkg.toml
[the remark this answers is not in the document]
n3wscott: I plan to not integrate 0.7 with this code. It will dep on master, aka 0.8 and above.

Gopkg.toml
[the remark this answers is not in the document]
n3wscott: tbh it is just easier to use overrides to preemptively avoid issues.

Gopkg.toml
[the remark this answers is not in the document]
grantr: Is your plan for it to ALWAYS depend on master, or will it depend on 0.8 when 0.8 is released?

Gopkg.toml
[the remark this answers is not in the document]
grantr: I don't agree, but it's not a huge deal as long as nothing ever depends on this repo.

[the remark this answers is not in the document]
grantr: Still need a comment saying "This is a preemptive override"

Gopkg.toml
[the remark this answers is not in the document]
grantr: Plan is to always depend on serving and eventing master.

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+		if topic.Spec.PropagationPolicy == v1alpha1.TopicPolicyCreateDelete {
+			// Ensure the Topic is deleted.
+			state, err := c.EnsureTopicDeleted(ctx, topic, topic.Spec.Project, topic.Status.TopicID)
+			switch state {
+			case pubsub.OpsJobGetFailed:
+				logger.Error("Failed to get Topic ops job.", zap.Any("state", state), zap.Error(err))
+				return err
+
+			case pubsub.OpsJobCreated:
+				// If we created a job to make a topic, then add the finalizer and update the status.
```
grantr: Remove `add the finalizer` because that's already happened

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+	return nil
+}
+
+func (c *Reconciler) updateStatus(ctx context.Context, desired *v1alpha1.Topic) (*v1alpha1.Topic, error) {
+	topic, err := c.topicLister.Topics(desired.Namespace).Get(desired.Name)
+	if err != nil {
+		return nil, err
+	}
+	// If there's nothing to update, just return.
+	if reflect.DeepEqual(topic.Status, desired.Status) {
```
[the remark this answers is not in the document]
grantr: #87

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+	}
+	becomesReady := desired.Status.IsReady() && !topic.Status.IsReady()
+	// Don't modify the informers copy.
+	existing := topic.DeepCopy()
+	existing.Status = desired.Status
+
+	ch, err := c.RunClientSet.PubsubV1alpha1().Topics(desired.Namespace).UpdateStatus(existing)
+	if err == nil && becomesReady {
+		duration := time.Since(ch.ObjectMeta.CreationTimestamp.Time)
+		c.Logger.Infof("Topic %q became ready after %v", topic.Name, duration)
```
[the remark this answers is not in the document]
grantr: We could use `now - LastTransitionTime`, but maybe not if LastTransitionTime is updated when the message changes.

According to @n3wscott we actually want to timestamp the point when `observedGeneration != spec.generation` and use that as the subtrahend.

pkg/reconciler/topic/topic.go
```diff
@@ -0,0 +1,418 @@
...
+
+		if err := c.StatsReporter.ReportReady("Topic", topic.Namespace, topic.Name, duration); err != nil {
+			logging.FromContext(ctx).Infof("failed to record ready for Topic, %v", err)
+		}
+	}
+
+	return ch, err
+}
+
+func (c *Reconciler) updateFinalizers(ctx context.Context, desired *v1alpha1.Topic) (*v1alpha1.Topic, bool, error) {
```
grantr: Add a comment saying "This method is generic for future compatibility with an SDK or some such"

pkg/reconciler/topic/topic.go
grantr: @n3wscott says: This is patch to defend against bad controller practice of changing the spec in the reconciler.

pkg/reconciler/topic/topic.go
grantr: @n3wscott says: This is here so this method can be generic (and replaced by a generic method) independent of whatever the reconciler is doing with finalizers.

PR #54, opened 2019-06-24
Adding a wrapper interface to pubsub to allow us to test the pubsub client.

n3wscott:
Also adding a dumb pubsub client.


pkg/pubsub/client.go
```diff
@@ -0,0 +1,45 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return &PubSubClient{
+		client: client,
+	}, nil
+}
+
+// PubSubClient wraps pubsub.Client.
+type PubSubClient struct {
```
mattmoor: Does this even need to be public?

n3wscott: nope!

PR #65, opened 2019-06-28
Allow for sink and transformer to be updated.

n3wscott:


pkg/apis/pubsub/v1alpha1/pull_subscription_validation.go
```diff
@@ -80,12 +81,9 @@ func (current *PullSubscription) CheckImmutableFields(ctx context.Context, og ap
...
 	}
 
-	// TODO: revisit this.
-
-	// All of the fields are immutable because the controller doesn't understand when it would need
-	// to delete and create a new Receive Adapter with updated arguments. We could relax it slightly
-	// to allow a nil Sink -> non-nil Sink, but I don't think it is needed yet.
-	if diff := cmp.Diff(original.Spec, current.Spec); diff != "" {
+	// Modification of Sink and Transform allowed. Everything else is immutable.
+	if diff := cmp.Diff(original.Spec, current.Spec,
```
grantr: I think it's a better experience for users to see a list of the field names rather than the very noisy diff output. Optional.

PR #75, opened 2019-07-01
Use desugared loggers in the pubsub main.go files

Harwayne:
Use desugared loggers in the pubsub main.go files, fixing (and preventing future) bugs.

### DIFF
--- a/cmd/pubsub/receive_adapter/main.go
+++ b/cmd/pubsub/receive_adapter/main.go
@@ -51,10 +51,11 @@ type envConfig struct {
 func main() {
 	flag.Parse()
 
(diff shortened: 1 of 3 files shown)


Harwayne:
> It's odd that `WithLogger` is taking a Sugared logger, but we're avoiding it elsewhere. Something on your list, I presume?

It is a difference of opinion between pkg's logging, which always uses SugaredLoggers and eventing's logging, which always uses desugared loggers. I have a strong preference for desugared, but I don't think many agree.

grantr:
> I have a strong preference for desugared, but I don't think many agree.

I do!

n3wscott:
> I have a strong preference for desugared, but I don't think many agree.

I don't.

PR #78, opened 2019-07-01
Config clean ups.

Harwayne:
Config clean ups.

### DIFF
--- a/config/201-addressable-resolvers-clusterrole.yaml
+++ b/config/201-addressable-resolvers-clusterrole.yaml
@@ -12,56 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Use this aggregated ClusterRole when you need readonly access to "Addressables".
-
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: addressable-resolver
-aggregationRule:
-  clusterRoleSelectors:
-  - matchLabels:
-      duck.knative.dev/addressable: "true"
-rules: []
-
----
-
-kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: service-addressable-resolver
-  labels:
-      duck.knative.dev/addressable: "true"
-# Do not use this role directly. These rules will be added to the "addressable-resolver" role.
-rules:
-- apiGroups:
-  - ""
-  resources:
-  - services
-  verbs:
-  - get
-  - list
-  - watch
-
----
-
-kind: ClusterRole
-apiVersion: rbac.authorization.k8s.io/v1
-metadata:
-  name: kservice-addressable-resolver
-  labels:
-    duck.knative.dev/addressable: "true"
-# Do not use this role directly. These rules will be added to the "addressable-resolver" role.
-rules:
-  - apiGroups:
-      - serving.knative.dev
-    resources:
-      - services
-      - routes
-    verbs:
-      - get
-      - list
-      - watch
+# Use the 'addressable-resolver' aggregated ClusterRole when you need readonly access to
+# "Addressables". It is defined in knative/eventing:
+# https://github.com/knative/eventing/blob/master/config/200-addressable-resolvers-clusterrole.yaml
 
 ---
 
--- a/config/201-channelable-manipulator-clusterrole.yaml
+++ b/config/201-channelable-manipulator-clusterrole.yaml
@@ -12,12 +12,16 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Use this aggregated ClusterRole when you nead and update permissions on "Channelables".
+# Use the 'channelable-manipulator' aggregated ClusterRole when you need read and update permissions
+# on "Channelables". Is is defined in knative/eventing:
+# https://github.com/knative/eventing/blob/master/config/200-channelable-manipulator-clusterrole.yaml
+
+---
 
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: pubsub-channel-addressable-resolver
+  name: pubsub-channel-channelable-manipulator
   labels:
     duck.knative.dev/channelable: "true"
 # Do not use this role directly. These rules will be added to the "channelable-manipulator" role.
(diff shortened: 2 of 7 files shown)


config/201-addressable-resolvers-clusterrole.yaml
```diff
@@ -12,56 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Use this aggregated ClusterRole when you need readonly access to "Addressables".
-
-apiVersion: rbac.authorization.k8s.io/v1
-kind: ClusterRole
-metadata:
-  name: addressable-resolver
```
Harwayne: I removed all of these because they duplicate what is already in https://github.com/knative/eventing/blob/master/config/200-addressable-resolvers-clusterrole.yaml. Is there a reason we would also want them defined here?

config/400-webhook-service.yaml
```diff
@@ -21,8 +21,8 @@ metadata:
   name: webhook
   namespace: cloud-run-events
 spec:
+  selector:
```
Harwayne: Switched the order to match `400-controller-service.yaml`.

config/500-controller.yaml
```diff
@@ -28,6 +28,7 @@ spec:
     metadata:
       labels:
         control-plane: cloud-run-events-controller
+        events.cloud.run/release: devel
```
Harwayne: To match `500-webhook.yaml`.

config/500-webhook.yaml
```diff
@@ -24,12 +24,12 @@ spec:
   selector:
     matchLabels:
       app: webhook
```
Harwayne: Is this label needed or wanted?

config/201-addressable-resolvers-clusterrole.yaml
n3wscott: I need these here. This repo does not dep on eventing to be installed for the pubsub parts.

config/201-channelable-manipulator-clusterrole.yaml
```diff
@@ -12,12 +12,16 @@
...
+# on "Channelables". Is is defined in knative/eventing:
+# https://github.com/knative/eventing/blob/master/config/200-channelable-manipulator-clusterrole.yaml
+
+---
 
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: pubsub-channel-addressable-resolver
+  name: pubsub-channel-channelable-manipulator
```
n3wscott: no no, this is just addressable-resolver.

config/500-webhook.yaml
n3wscott: yes: https://github.com/knative/serving/blob/master/config/webhook.yaml#L22-L36

config/500-webhook.yaml
```diff
@@ -24,12 +24,12 @@ spec:
...
-      role: webhook
+      role: cloud-run-events-webhook
```
n3wscott: why this change? just leave it `webhook`

config/500-webhook.yaml
```diff
@@ -24,12 +24,12 @@ spec:
...
   template:
     metadata:
       labels:
         app: webhook
-        role: webhook
+        role: cloud-run-events-webhook
```
n3wscott: same: leave as `webhook`

config/400-webhook-service.yaml
```diff
@@ -21,8 +21,8 @@ metadata:
...
+    role: cloud-run-events-webhook
```
n3wscott: `webhook`

n3wscott:
I don't really think most of these changes should happen just because I am attempting to let cloud run pubsub to be useful without eventing installed and the structure is suppose to match serving as much as possible.

config/201-addressable-resolvers-clusterrole.yaml
Harwayne: Can we move these out of `config/`, so that it is an optional install instruction, rather than by default overwriting eventing's ClusterRoles? Something like `config/without-eventing/201-addressable-resolvers-clusterrole.yaml`.

config/201-channelable-manipulator-clusterrole.yaml
Harwayne: Why does it include create, update, and patch?

config/400-webhook-service.yaml
Harwayne: 400-controller.yaml uses the [control-plane: cloud-run-events-controller](https://github.com/GoogleCloudPlatform/cloud-run-events/blob/8da14eb2bc52d58c256a884e11205645dcc91e66/config/400-controller-service.yaml#L22), I would like standardize on one style. My preference is to prefix with `cloud-run-events-`, but could be convinced to omit it.

Although, now that I look again, I should change the name of the label to match the controller too.

config/500-webhook.yaml
Harwayne: Same reasoning as https://github.com/GoogleCloudPlatform/cloud-run-events/pull/78/files#r299373680, to match the controller's `control-plane: cloud-run-events-controller`.

config/500-webhook.yaml
Harwayne: The pair of labels goes all the way back to the [initial commit](https://github.com/knative/serving/blob/7f5da0b7438f562ea4a45c7021a674e2f90a5ddf/webhook.yaml#L25-L26), but that doesn't answer why it is there. Does anything look at the `app` label and do something special with it?

config/201-addressable-resolvers-clusterrole.yaml
n3wscott: move sounds ok, how about `config/addressable`

config/500-webhook.yaml
n3wscott: I think that is how the rp and the pods know about eachother?

config/201-channelable-manipulator-clusterrole.yaml
n3wscott: ah, whoops, yeah my bad. you are right. thanks

config/400-webhook-service.yaml
n3wscott: I want just `webhook`.

vaikas-google:
@n3wscott @Harwayne what's the plan here? Should we close this, or get it rebased?

Harwayne:
/close

At this point it is hopelessly out of date, and if we want some of these changes, they should be made in a new PR.

PR #79, opened 2019-07-01
Docs updates.

Harwayne:
Docs updates.

### DIFF
--- a/docs/install/README.md
+++ b/docs/install/README.md
@@ -1,6 +1,6 @@
 ## Installing Cloud Run Events
 
-To install Cloud Run Events from master,
+Installing Cloud Run Events from master.
 
 ### Using [ko](http://github.com/google/ko)
 
@@ -10,13 +10,16 @@ ko apply -f ./config
 
 Applying the `config` dir will:
 
-- create a new namespace, `cloud-run-events`.
-- install new CRDs,
+- Create a new namespace, `cloud-run-events`.
+- Install new CRDs:
+  - `channels.pubsub.cloud.run`
   - `pullsubscriptions.pubsub.cloud.run`
-- install a controller to operate on the new resources,
+  - `topics.pubsub.cloud.run`
+- Install a controller Deployment to operate on the new resources,
   `cloud-run-events/cloud-run-events-controller`.
-- install a new service account, `cloud-run-events/cloud-run-events-controller`.
-- provide RBAC for that service account.
+- Install a webhook Deployment to validate and mutate the new resources, `cloud-run-events/cloud-run-events-webhook`.
+- Install a new ServiceAccount, `cloud-run-events/cloud-run-events-controller`.
+- Provide RBAC for that ServiceAccount.
 
 ## Uninstalling Cloud Run Events
 
--- a/docs/pullsubscription/README.md
+++ b/docs/pullsubscription/README.md
@@ -12,10 +12,14 @@ Pub/Sub topic, such as
 1. Create a
    [Google Cloud project](https://cloud.google.com/resource-manager/docs/creating-managing-projects)
    and install the `gcloud` CLI and run `gcloud auth login`. This sample will
-   use a mix of `gcloud` and `kubectl` commands. The rest of the sample assumes
-   that you've set the `$PROJECT_ID` environment variable to your Google Cloud
-   project id, and also set your project ID as default using
-   `gcloud config set project $PROJECT_ID`.
+   use a mix of `gcloud` and `kubectl` commands.
+
+1. Save the Google Cloud Project's ID to an environment variable and ensure `gcloud` is using that
+   project.
+   ```shell
+   export PROJECT_ID="my-gcp-project" # Replace 'my-gcp-project' with your project's ID.
+   gcloud config set project $PROJECT_ID
+   ```
 
 1. [Install Cloud Run Events](../install).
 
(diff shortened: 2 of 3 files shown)


docs/install/README.md
```diff
@@ -10,13 +10,16 @@ ko apply -f ./config
...
+- Install new CRDs:
+  - `channels.pubsub.cloud.run`
   - `pullsubscriptions.pubsub.cloud.run`
-- install a controller to operate on the new resources,
+  - `topics.pubsub.cloud.run`
+- Install a controller Deployment to operate on the new resources,
   `cloud-run-events/cloud-run-events-controller`.
-- install a new service account, `cloud-run-events/cloud-run-events-controller`.
-- provide RBAC for that service account.
+- Install a webhook Deployment to validate and mutate the new resources, `cloud-run-events/cloud-run-events-webhook`.
```
Harwayne: Note that this is actually incorrect, the Deployments are named `webhook` and `controller`. I think we should change the names of the Deployments to match these docs. Thougths?

docs/install/README.md
```diff
@@ -10,13 +10,16 @@ ko apply -f ./config
 
 Applying the `config` dir will:
 
-- create a new namespace, `cloud-run-events`.
-- install new CRDs,
+- Create a new namespace, `cloud-run-events`.
+- Install new CRDs:
+  - `channels.pubsub.cloud.run`
```
n3wscott: not yet for channels. It will not be here.

docs/install/README.md
n3wscott: I want them to be called `webhook` and `controller`. Can you update the doc?

docs/pullsubscription/README.md
```diff
@@ -68,44 +72,47 @@ Pub/Sub topic, such as
...
+           ```shell
+           kubectl apply --filename pullsubscription.yaml
+           ```
+      1. Running outside Google Kubenetes Engine or want to use a Pub/Sub Topic from a different
+         project than your GKE cluster:
 
-   If you're in the samples directory, you can replace `MY_PROJECT` and
-   `TOPIC_NAME` and then apply in one command:
+           If you're in the samples directory, you can replace `MY_PROJECT`,
+           `TOPIC_NAME`, uncomment `#secret:`, and then apply in a single command:
```
n3wscott: secret has a default value and is not required it you follow this guide.

docs/pullsubscription/README.md
```diff
@@ -191,7 +198,7 @@ If you are replaced `TOPIC_NAME` manually, then make sure you delete the
 resulting YAML:
```
n3wscott: This sentence originally does not make sense. maybe 

`then make sure you delete the  resulting YAML`

becomes

`then you can use the updated files to delete using:`

docs/pullsubscription/README.md
```diff
@@ -12,10 +12,14 @@ Pub/Sub topic, such as
 1. Create a
    [Google Cloud project](https://cloud.google.com/resource-manager/docs/creating-managing-projects)
    and install the `gcloud` CLI and run `gcloud auth login`. This sample will
-   use a mix of `gcloud` and `kubectl` commands. The rest of the sample assumes
```
n3wscott: this language came from the docs writers

docs/pullsubscription/README.md
```diff
@@ -25,7 +29,7 @@ Pub/Sub topic, such as
    gcloud services enable pubsub.googleapis.com
    ```
 
-1. Create a
+1. [TODO - This is only required outside of GKE] Create a
```
n3wscott: This is required for GKE too

docs/pullsubscription/README.md
```diff
@@ -68,44 +72,47 @@ Pub/Sub topic, such as
    gcloud pubsub topics create $TOPIC_NAME
    ```
 
-1. Update `TOPIC_NAME` in the [`pullsubscription.yaml`](pullsubscription.yaml)
-   and apply it.
+1. Replace the placeholders with real values.
```
n3wscott: again, this language was taken from the docs writers.

docs/pullsubscription/README.md
```diff
@@ -68,44 +72,47 @@ Pub/Sub topic, such as
...
 
-   If you're in the samples directory, you can replace `TOPIC_NAME` and apply in
-   one command:
+      1. Placeholders:
+          - `TOPIC_NAME` - Must be replaced.
+          - `MY_PROJECT` - Replaced only if you are running outside of Google Kubernetes Engine.
```
n3wscott: Or if you are targeting a topic from another project.

docs/pullsubscription/README.md
```diff
@@ -68,44 +72,47 @@ Pub/Sub topic, such as
...
+            Note that if you are replacing this, then you should also uncomment the `#secret:` line.
```
n3wscott: This is not true if the secret is from the other project.

docs/pullsubscription/README.md
```diff
@@ -68,44 +72,47 @@ Pub/Sub topic, such as
    gcloud pubsub topics create $TOPIC_NAME
    ```
 
-1. Update `TOPIC_NAME` in the [`pullsubscription.yaml`](pullsubscription.yaml)
-   and apply it.
+1. Replace the placeholders with real values.
 
-   If you're in the samples directory, you can replace `TOPIC_NAME` and apply in
-   one command:
+      1. Placeholders:
```
n3wscott: I don't understand why this is now a substep, there are only a few steps, seems to make sense to just have steps 1-4.

n3wscott:
/close 

out of date says Adam.

PR #80, opened 2019-07-02
Revert "Use desugared loggers in the pubsub main.go files"

n3wscott:
Serving is fine with sugared loggers.

Reverts GoogleCloudPlatform/cloud-run-events#75

### DIFF
--- a/cmd/pubsub/receive_adapter/main.go
+++ b/cmd/pubsub/receive_adapter/main.go
@@ -51,11 +51,10 @@ type envConfig struct {
 func main() {
 	flag.Parse()
 
-	sl, _ := logging.NewLogger("", "INFO") // TODO: use logging config map.
-	logger := sl.Desugar()
+	logger, _ := logging.NewLogger("", "INFO") // TODO: use logging config map.
 	defer logger.Sync()
 
-	ctx := logging.WithLogger(signals.NewContext(), logger.Sugar())
+	ctx := logging.WithLogger(signals.NewContext(), logger)
 
 	startable := adapter.Adapter{}
 	if err := envconfig.Process("", &startable); err != nil {
--- a/cmd/pubsub/subscription/main.go
+++ b/cmd/pubsub/subscription/main.go
@@ -19,10 +19,11 @@ package main
 import (
 	"flag"
 
+	"knative.dev/pkg/signals"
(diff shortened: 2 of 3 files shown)


Harwayne:
/hold

> Serving is fine with sugared loggers.

Serving has the same sugared logging issues.

When sugared, the zap Fields (zap.Error, zap.Any, etc.) are logged poorly. In all three of the files you are reverting, there are log lines that assume desugared. 
- receive_adapter: lines 62, 68, 73, and 77.
- subscription: line 41
- topic: line 43

As an example, see how the sugared and desugared loggers log an error and a struct:

```golang
func main() {
	dl, _ := zap.NewProduction()
	sl := dl.Sugar()

	err := fmt.Errorf("golang error")
	s := envConfig{
		Namespace:  "foo",
		Mode:       "bar",
		ApiVersion: []string{"baz"},
	}
	dl.Info("Desugared", zap.Error(err), zap.Any("s", s))
	sl.Info("Sugared", zap.Error(err), zap.Any("s", s))
}
```

```shell
{"level":"info","ts":1562085479.4341152,"caller":"apiserver_receive_adapter/main.go:65","msg":"Desugared","error":"golang error","s":{"Namespace":"foo","Mode":"bar","SinkURI":"","ApiVersion":["baz"],"Kind":null,"Controller":null}}
{"level":"info","ts":1562085479.434289,"caller":"apiserver_receive_adapter/main.go:66","msg":"Sugared{error 25 0  golang error} {s 22 0  {foo bar  [baz] [] []}}"}
```

Is it preferred that the error and struct are simply dumped as structs, rather than printed? Or that everything is concatenated into a single string, rather than given individual fields?

n3wscott:
Define poorly?

n3wscott:
Now at a computer and I am able to see the logged lines you have provided. Interesting. OK it seems like desugared logs make better logs. What is sugar for??

Harwayne:
> Now at a computer and I am able to see the logged lines you have provided. Interesting. OK it seems like desugared logs make better logs. What is sugar for??

Sugar is useful if you want to do format strings, with Infof, Errorf, etc. It's also easier in that you don't need to wrap everything in a zap field, if you are OK with it being dumped.

PR #90, opened 2019-07-08
Fix a minor typo and update a comment for finalizers.

n3wscott:
fixes: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/89


grantr:
Also need 

> Add a comment saying "This method is generic for future compatibility with an SDK or some such" https://github.com/GoogleCloudPlatform/cloud-run-events/pull/35/files#diff-1d699b9cd7b3d4b5ad8802c07777a902R312 https://github.com/GoogleCloudPlatform/cloud-run-events/pull/35/files#r301248760

PR #92, opened 2019-07-08
Adding dep on master of serving and eventing.

n3wscott:
Fixes: https://github.com/GoogleCloudPlatform/cloud-run-events/issues/86

Also adding a script to update them.

### DIFF
--- a/Gopkg.toml
+++ b/Gopkg.toml
@@ -21,15 +21,15 @@ required = [
   name = "knative.dev/pkg"
   branch = "master"
 
+# This is a preemptive override.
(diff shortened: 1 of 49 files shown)


Gopkg.toml
```diff
@@ -21,15 +21,15 @@ required = [
   name = "knative.dev/pkg"
   branch = "master"
 
+# This is a preemptive override.
```
vaikas-google: What does this comment mean? Why is knative.dev/pkg not commented with this? I just don't understand what preemptive override mean.

PR #96, opened 2019-07-09
Remove service account from PullSubscription

n3wscott:
The service account is not used for DataPlane PullSubscriptions. Removing.


cmd/webhook/main.go
```diff
@@ -111,12 +111,19 @@ func SharedMain(handlers map[schema.GroupVersionKind]webhook.GenericCRD, factori
 		SecretName:     "webhook-certs",
 		WebhookName:    fmt.Sprintf("webhook.%s.events.cloud.run", system.Namespace()),
 	}
+
+	stats, err := webhook.NewStatsReporter()
```
n3wscott: This was broken by https://github.com/knative/pkg/issues/517

grantr: Should be a separate PR

PR #108, opened 2019-07-25
Enable Sink and Transformer to be URIs in the spec.

n3wscott:
This is work to support Channel.


config/300-pullsubscription.yaml
```diff
@@ -80,13 +80,13 @@ spec:
               type: string
               description: "ID of the Cloud Pub/Sub Topic to Subscribe to. It must be in the form of the unique identifier within the project, not the entire name. E.g. it must be 'laconia', not 'projects/my-gcp-project/topics/laconia'."
             ackDeadline:
-              type: string
+              type: integer
```
vaikas-google: this seems wrong? see the comments about units in the description?

config/300-pullsubscription.yaml
```diff
@@ -80,13 +80,13 @@ spec:
...
               description:  "The default maximum time after a subscriber receives a message before the subscriber should acknowledge the message. Defaults to `30s`. Valid time units are `s`, `m`, `h`."
             retainAckedMessages:
               type: boolean
               description: "Whether to retain acknowledged messages. If true, acknowledged messages will not be expunged until they fall out of the RetentionDuration window."
             retentionDuration:
-              type: string
+              type: integer
```
vaikas-google: ditto

pkg/apis/pubsub/v1alpha1/pull_subscription_types.go
```diff
@@ -94,22 +94,27 @@ type PullSubscriptionSpec struct {
 	// +optional
 	RetentionDuration *time.Duration `json:"retentionDuration,omitempty"`
 
-	// Sink is a reference to an object that will resolve to a domain name to
-	// use as the sink.
+	// Sink is a reference to an object that will resolve to a domain name or a
+	// URI directly to use as the sink.
 	// +optional
-	Sink *corev1.ObjectReference `json:"sink,omitempty"`
+	Sink *Destination `json:"sink,omitempty"`
```
vaikas-google: Why is this optional? Just curious.

vaikas-google: Looking at the validation code, doesn't seem so optional? :)

pkg/apis/pubsub/v1alpha1/pull_subscription_validation.go
```diff
@@ -74,6 +74,23 @@ func (current *PullSubscriptionSpec) Validate(ctx context.Context) *apis.FieldEr
 	return errs
 }
 
+func validateDestination(dest *Destination) *apis.FieldError {
+	if dest == nil {
+		return apis.ErrMissingField(apis.CurrentField)
```
vaikas-google: this seems to make sink not optional.

pkg/reconciler/pullsubscription/pullsubscription.go
```diff
@@ -191,21 +193,28 @@ func (c *Reconciler) reconcile(ctx context.Context, source *v1alpha1.PullSubscri
 		return nil
 	}
 
-	sinkURI, err := duck.GetSinkURI(ctx, c.DynamicClientSet, source.Spec.Sink, source.Namespace)
-	if err != nil {
-		source.Status.MarkNoSink("NotFound", "")
-		return err
+	// Sink is required. // TODO: this is a problem for channels.
```
vaikas-google: why is this a problem for channels? Is there something that needs to be done for channels? perhaps create an issue for it?

pkg/apis/pubsub/v1alpha1/pull_subscription_types.go
n3wscott: nice catch! it should not be optional.

config/300-pullsubscription.yaml
n3wscott: you are correct. It was half wired wrong... fixing.

pkg/reconciler/pullsubscription/pullsubscription.go
n3wscott: I was wrong, I have a plan. Todo removed.

pkg/apis/pubsub/v1alpha1/pull_subscription_types.go
vaikas-google: If it should not be optional, why not:
Sink Destination `json:"sink"`

pkg/apis/pubsub/v1alpha1/pull_subscription_types.go
```diff
@@ -89,27 +88,51 @@ type PullSubscriptionSpec struct {
...
+	// Sink is a reference to an object that will resolve to a domain name or a
+	// URI directly to use as the sink.
+	Sink *Destination `json:"sink,omitempty"`
 
 	// Transformer is a reference to an object that will resolve to a domain
-	// name to use as the transformer.
+	// name or a URI directly to use as the transformer or a URI directly.
 	// +optional
-	Transformer *corev1.ObjectReference `json:"transformer,omitempty"`
+	Transformer *Destination `json:"transformer,omitempty"`
```
vaikas-google: this makes sense to be a pointer.

n3wscott: Now sink is a pointer.

vaikas-google: it was supposed to be paired with the comment about Sink not being a pointer, but this makes sense being a pointer. The comments were supposed to be linked but whatevs. I was trying to make sure you don't change them both to a non-pointer :)

PR #124, opened 2019-08-01
Update pubsub to use secrets and fix serving version to v1alpha1.

n3wscott:


vaikas-google:
can we bring up the test coverage for the components on a followup PR...
/lgtm
/approve

PR #135, opened 2019-08-02
Fix service account generation.

n3wscott:


n3wscott:
/close

merged this into add-e2e

PR #150, opened 2019-08-12
adding storage converter files

vaikas-google:
/hold
Just taking a temp, seem fine? I'll add UT if this looks like a plausible direction for extending the converter logic.

### DIFF
--- a/pkg/pubsub/adapter/adapter.go
+++ b/pkg/pubsub/adapter/adapter.go
@@ -27,7 +27,6 @@ import (
 	"go.uber.org/zap"
 	"knative.dev/pkg/logging"
 
-	"github.com/google/knative-gcp/pkg/apis/pubsub/v1alpha1"
 	"github.com/google/knative-gcp/pkg/kncloudevents"
 	"github.com/google/knative-gcp/pkg/pubsub/adapter/converters"
 )
@@ -55,7 +54,7 @@ type Adapter struct {
 
 	// SendMode describes how the adapter sends events.
 	// One of [binary, structured, push]. Default: binary
-	SendMode ModeType `envconfig:"SEND_MODE" default:"binary" required:"true"`
+	SendMode converters.ModeType `envconfig:"SEND_MODE" default:"binary" required:"true"`
 
 	// inbound is the cloudevents client to use to receive events.
 	inbound cloudevents.Client
@@ -65,36 +64,14 @@ type Adapter struct {
 
 	// transformer is the cloudevents client to transform received events before sending.
 	transformer cloudevents.Client
-
-	// converters is the map for handling Source specific event
-	// conversions. For example, a GCS event will need to be
-	// converted differently from the PubSub. The key into
-	// this map will be "knative-gcp" CloudEvent attribute.
-	// If there's no such attribute, we assume it's a native
-	// PubSub message and a default one will be used.
-	converters map[string]cloudevents.ConvertFn
 }
-
-// ModeType is the type for mode enum.
-type ModeType string
-
-const (
-	// Binary mode is binary encoding.
-	Binary ModeType = "binary"
-	// Structured mode is structured encoding.
-	Structured ModeType = "structured"
-	// Push mode emulates Pub/Sub push encoding.
-	Push ModeType = "push"
-	// DefaultSendMode is the default choice.
-	DefaultSendMode = Binary
-)
 
 // Start starts the adapter. Note: Only call once, not thread safe.
 func (a *Adapter) Start(ctx context.Context) error {
 	var err error
 
 	if a.SendMode == "" {
-		a.SendMode = DefaultSendMode
+		a.SendMode = converters.DefaultSendMode
 	}
 
 	// Receive Events on Pub/Sub.
@@ -109,11 +86,6 @@ func (a *Adapter) Start(ctx context.Context) error {
 		if a.outbound, err = a.newHTTPClient(a.Sink); err != nil {
 			return fmt.Errorf("failed to create outbound cloudevent client: %s", err.Error())
 		}
-	}
-
-	// Add the converters...
-	a.converters = map[string]cloudevents.ConvertFn{
-		"google-storage": converters.ConvertStorage,
 	}
 
 	// Make the transformer client in case the TransformerURI has been set.
@@ -149,7 +121,7 @@ func (a *Adapter) receive(ctx context.Context, event cloudevents.Event, resp *cl
 	}
 
 	// If send mode is Push, convert to Pub/Sub Push payload style.
-	if a.SendMode == Push {
+	if a.SendMode == converters.Push {
 		event = ConvertToPush(ctx, event)
 	}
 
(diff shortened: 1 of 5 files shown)


pkg/pubsub/adapter/converters/storage.go
```diff
@@ -0,0 +1,60 @@
...
+import (
+	"context"
+	"fmt"
+
+	cloudevents "github.com/cloudevents/sdk-go"
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport"
+	cepubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub"
+	"knative.dev/pkg/logging"
+)
+
```
n3wscott: you could make converters a package global and then add an init method in this file like:

```go
func init() {
	converters["google-storage"] = converters.ConvertStorage
}
```

it kinda keeps new converters only changing their one file? You can reject if this idea is too "cute".

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -0,0 +1,60 @@
...
+	logger := logging.FromContext(ctx)
+	logger.Debug("Converting event from transport.")
+	if msg, ok := m.(*cepubsub.Message); ok {
+		tx := cepubsub.TransportContextFrom(ctx)
+		// Make a new event and convert the message payload.
+		event := cloudevents.NewEvent()
+		event.SetID(tx.ID)
+		event.SetTime(tx.PublishTime)
+		if msg.Attributes != nil {
+			if val, ok := msg.Attributes["notificationConfig"]; ok {
```
n3wscott: you also want to delete `notificationConfig` from the map, or else you will get it dup'ed in the extensions from line 52

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -0,0 +1,60 @@
...
+	"knative.dev/pkg/logging"
+)
+
+func ConvertStorage(ctx context.Context, m transport.Message, err error) (*cloudevents.Event, error) {
+	logger := logging.FromContext(ctx)
+	logger.Debug("Converting event from transport.")
+	if msg, ok := m.(*cepubsub.Message); ok {
+		tx := cepubsub.TransportContextFrom(ctx)
+		// Make a new event and convert the message payload.
+		event := cloudevents.NewEvent()
```
n3wscott: Does using subject make sense for this event? any properties make an easy transition into source and subject?

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -0,0 +1,60 @@
...
+		if msg.Attributes != nil {
+			if val, ok := msg.Attributes["notificationConfig"]; ok {
+				event.SetSource(val)
+			} else {
+				return nil, fmt.Errorf("received event did not have notificationConfig")
+			}
+		}
+		event.SetType("google.storage")
+		event.SetDataContentType(*cloudevents.StringOfApplicationJSON())
+		event.Data = msg.Data
```
n3wscott: What is msg.Data? it is a json string? If so then what you are doing here is fine.

it it is a json struct, you can use `event.SetData(msg.Data)` and you do not have to set `DataEncoded`.

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -0,0 +1,60 @@
...
+	"context"
+	"fmt"
+
+	cloudevents "github.com/cloudevents/sdk-go"
+	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport"
+	cepubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub"
+	"knative.dev/pkg/logging"
+)
+
+func ConvertStorage(ctx context.Context, m transport.Message, err error) (*cloudevents.Event, error) {
```
n3wscott: I would change this signature to 

```go
func ConvertStorage(ctx context.Context, msg *cepubsub.Message, err error) (*cloudevents.Event, error) {
```

because you already have access to the `cepubsub.Message`: `if msg, ok := m.(*cepubsub.Message); ok {`

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -0,0 +1,60 @@
...
+	logger := logging.FromContext(ctx)
+	logger.Debug("Converting event from transport.")
+	if msg, ok := m.(*cepubsub.Message); ok {
```
n3wscott: And then you do not need this cast if the sig changes.

pkg/pubsub/adapter/adapter.go
```diff
@@ -151,7 +165,17 @@ func (a *Adapter) receive(ctx context.Context, event cloudevents.Event, resp *cl
...
+				logger.Infof("Converting Knative-GCP event: %q", val)
+				if c, ok := a.converters[val]; ok {
+					return c(ctx, m, err)
+				}
+			}
+		}
+
 		tx := cepubsub.TransportContextFrom(ctx)
 		// Make a new event and convert the message payload.
 		event := cloudevents.NewEvent()
```
n3wscott: One question I had when writing this was is it important to pass along the topic and/or the subscription that delivered this? I could see a user encoding some logic in the name of the topic in some cases. But it does not make sense for like, a channel... What do you think?

pkg/pubsub/adapter/adapter.go
```diff
@@ -151,26 +137,9 @@ func (a *Adapter) receive(ctx context.Context, event cloudevents.Event, resp *cl
...
-		// Attributes are extensions.
-		if msg.Attributes != nil && len(msg.Attributes) > 0 {
-			for k, v := range msg.Attributes {
-				event.SetExtension(k, v)
-			}
-		}
 
-		return &event, nil
+	if msg, ok := m.(*cepubsub.Message); ok {
+		converters.Convert(ctx, msg, a.SendMode)
```
n3wscott: you need `return converters.Convert(ctx, msg, a.SendMode)`

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -0,0 +1,53 @@
...
+		return nil, fmt.Errorf("nill pubsub message")
+	}
+	tx := cepubsub.TransportContextFrom(ctx)
+	// Make a new event and convert the message payload.
+	event := cloudevents.NewEvent()
+	event.SetID(tx.ID)
+	event.SetTime(tx.PublishTime)
+	if msg.Attributes != nil {
+		if val, ok := msg.Attributes["notificationConfig"]; ok {
+			event.SetSource(val)
```
n3wscott: still need a delete `notificationConfig` from `msg.Attributes`

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -0,0 +1,53 @@
...
+		} else {
+			return nil, fmt.Errorf("received event did not have notificationConfig")
+		}
+	}
+	event.SetType("google.storage")
+	event.SetDataContentType(*cloudevents.StringOfApplicationJSON())
+	event.SetData(msg.Data)
```
n3wscott: you also might want to add a:

```
event.SetSchemaURL(fmt.Sprintf("//events.cloud.run/storage/schema.json?mode=%s", sendMode))
```

or whatever, I am not using real urls yet. but they do mean something

n3wscott:
/hold cancel
/lgtm
/approve

THANKS

PR #153, opened 2019-08-13
Gcs controller prototype

vaikas-google:
works end to end, but needs tests


pkg/apis/events/v1alpha1/storage_defaults.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"context"
+)
+
+func (g *Storage) SetDefaults(ctx context.Context) {
+	g.Spec.SetDefaults(ctx)
+}
+
+func (gs *StorageSpec) SetDefaults(ctx context.Context) {
+	// TODO? What defaults?
```
n3wscott: Default GCS secrets.

PR #167, opened 2019-08-15
Updates to PubSub CloudEvent

nachocano:
- Promoting `schema` to be the schemaURL of the CloudEvent, in case it came in the PubSub attributes map.
- Setting a `knativecemode` to the mode configured: push, structured, and binary.


schemas/pubsub/schema.json
```diff
@@ -0,0 +1,105 @@
...
+              "pattern": "^(.*)$"
+            }
+          }
+        }
+      },
+      "additionalProperties": false
+    },
+    "binary": {
+      "$id": "#binary",
+      "oneOf": [
```
n3wscott: binary and structured are the same for data. But data can be many more things and encoded in many more ways. Are we talking about the http version of the data? or the data converted?

nachocano: Agreed to remove the schema, and promote the schema if provided in the event...
We'll set the mode as an extension attribute as well

PR #169, opened 2019-08-15
Storage follows the duck type for attribute overrides.

Harwayne:
Storage follows the duck type for attribute overrides.


pkg/reconciler/storage/storage.go
```diff
@@ -280,8 +280,10 @@ func (c *Reconciler) reconcileNotification(ctx context.Context, storage *v1alpha
 	}
 
 	customAttributes := make(map[string]string)
-	for k, v := range storage.Spec.CustomAttributes {
-		customAttributes[k] = v
+	if storage.Spec.CloudEventOverrides != nil {
```
n3wscott: the cloudEventOverrides should propagate down into the PullSubscription, not added to the customAttributes

n3wscott:
/hold

Leverage the PullSubscription CloudEventsOverrides

pkg/reconciler/storage/storage.go
Harwayne: I don't understand the tradeoffs, so I made the minimum change to match the desired interface.

Harwayne:
> /hold
> 
> Leverage the PullSubscription CloudEventsOverrides

I don't understand the tradeoffs, so I made the minimum change to match the desired interface.

/hold cancel

n3wscott:
/lgtm

I made an issue to follow up

PR #178, opened 2019-08-16
Make Channel implement Channelable

nachocano:
- Adding `SubscribableTypeStatus` to `ChannelStatus`.
- Marking the Subscriber as Ready. It should actually be read from the PullSubscription status... Maybe in a future PR, or in this one if I find the way.

/cc @n3wscott


pkg/reconciler/channel/channel.go
```diff
@@ -250,10 +252,11 @@ func (c *Reconciler) syncSubscribers(ctx context.Context, channel *v1alpha1.Chan
 		}
 		c.Recorder.Eventf(channel, corev1.EventTypeNormal, "CreatedSubscriber", "Created Subscriber %q", genName)
 
-		channel.Status.Subscribers = append(channel.Status.Subscribers, eventingduck.SubscriberStatus{
+		channel.Status.SubscribableStatus.Subscribers = append(channel.Status.SubscribableStatus.Subscribers, eventingduck.SubscriberStatus{
 			UID:                s.UID,
 			ObservedGeneration: s.Generation,
-			// TODO: do I need the other fields?
+			// TODO read the condition from PullSubscription, and add a message if it's not ready.
```
n3wscott: read the condition from PullSubscription, and add a message if it's not ready.

nachocano: done!

PR #183, opened 2019-08-16
Storage's spec.project is optional

Harwayne:
Storage's spec.project is optional.

### DIFF
--- a/pkg/apis/events/v1alpha1/storage_types.go
+++ b/pkg/apis/events/v1alpha1/storage_types.go
@@ -65,6 +65,7 @@ type StorageSpec struct {
 
 	// Project is the ID of the Google Cloud Project that the Bucket being
 	// subscribed to exists in.
+	// +optional
 	Project string `json:"Project,omitempty"`
 
 	// Bucket to subscribe to.
--- a/pkg/reconciler/storage/storage.go
+++ b/pkg/reconciler/storage/storage.go
@@ -22,17 +22,9 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/google/uuid"
-	"go.uber.org/zap"
-	"k8s.io/apimachinery/pkg/api/errors"
-	"k8s.io/apimachinery/pkg/util/runtime"
-	"k8s.io/client-go/tools/cache"
-	"knative.dev/pkg/controller"
-	"knative.dev/pkg/logging"
-
+	"cloud.google.com/go/compute/metadata"
 	"cloud.google.com/go/pubsub"
 	storageClient "cloud.google.com/go/storage"
-
 	"github.com/google/knative-gcp/pkg/apis/events/v1alpha1"
 	pubsubsourcev1alpha1 "github.com/google/knative-gcp/pkg/apis/pubsub/v1alpha1"
 	pubsubsourceclientset "github.com/google/knative-gcp/pkg/client/clientset/versioned"
@@ -42,12 +34,19 @@ import (
 	"github.com/google/knative-gcp/pkg/duck"
 	"github.com/google/knative-gcp/pkg/reconciler"
 	"github.com/google/knative-gcp/pkg/reconciler/storage/resources"
+	"github.com/google/uuid"
+	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
 	gstatus "google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/api/equality"
+	"k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/apimachinery/pkg/util/runtime"
 	"k8s.io/apimachinery/pkg/util/sets"
+	"k8s.io/client-go/tools/cache"
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/logging"
 )
 
 const (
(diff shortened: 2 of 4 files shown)


pkg/apis/events/v1alpha1/storage_types.go
```diff
@@ -91,6 +92,10 @@ type StorageSpec struct {
 	// Sink is a reference to an object that will resolve to a domain name to use
 	// as the sink.
 	Sink corev1.ObjectReference `json:"sink"`
+
+	// defaulterError is used exclusively to pass an error from the webhook defaulter to the webhook
+	// validator.
+	defaulterError error
```
Harwayne: This is really gross, but I think it is the best way to do what I want. 

I could just do the defaulting in the reconciler, but then I need to be sure everything uses the local project variable, not spec.project. 

I could do the defaulting in the validator, but that feels wrong for some reason.

pkg/apis/events/v1alpha1/storage_defaults.go
```diff
@@ -34,4 +35,13 @@ func (gs *StorageSpec) SetDefaults(ctx context.Context) {
 	if equality.Semantic.DeepEqual(gs.GCSSecret, &corev1.SecretKeySelector{}) {
 		gs.GCSSecret = *v1alpha1.DefaultGoogleCloudSecretSelector()
 	}
+
+	if gs.Project == "" {
```
n3wscott: This will not work in the operations world, the webhook will not have the secret that is required to call the metadata server I think

pkg/apis/events/v1alpha1/storage_validation.go
```diff
@@ -29,9 +29,16 @@ func (current *Storage) Validate(ctx context.Context) *apis.FieldError {
 }
 
 func (current *StorageSpec) Validate(ctx context.Context) *apis.FieldError {
-	// TODO
 	var errs *apis.FieldError
 
+	if current.defaulterError != nil {
```
n3wscott: Does the validation webhook chain inside of the mutation webhook?

pkg/apis/events/v1alpha1/storage_types.go
Harwayne: Actually this method won't work at all, it seems we can't have unexported fields in a spec. Switched to defaulting in the webhook validator.

Harwayne:
/hold

This doesn't work yet.

pkg/apis/events/v1alpha1/storage_types.go
Harwayne: This didn't work either, the changes validation made were ignored.

Harwayne:
/hold cancel

Doing this in a similar way as PullSubscription, namely doing it in the reconciler. Manually tested as working.

pkg/apis/events/v1alpha1/storage_validation.go
Harwayne: I think it does. Regardless, this doesn't work so it was reverted out.

pkg/apis/events/v1alpha1/storage_defaults.go
Harwayne: Switched to just doing it in the reconciler.

pkg/reconciler/storage/storage.go
```diff
@@ -129,6 +128,17 @@ func (c *Reconciler) Reconcile(ctx context.Context, key string) error {
 }
 
 func (c *Reconciler) reconcile(ctx context.Context, csr *v1alpha1.Storage) error {
+	// spec.project is optional, so if it is not specified, find the correct value and use it,
+	// rather than spec.project for the entire reconciliation.
+	project := csr.Spec.Project
+	if project == "" {
+		var err error
+		project, err = metadata.ProjectID()
```
n3wscott: Sorry this will still not work, the secrets tied to the resource need to be used to ask the correct metadata server. This will find out the gke project, not the secrets project. Ville is working on a migration to the jobs and will be PRed soon

n3wscott:
/hold

Wait for @vaikas-google to do the refactor.

pkg/reconciler/storage/storage.go
vaikas-google: My understanding is indeed that this will find the project that the Cluster belongs to. The credential (from the secret) is in the json that it was created in. I think it's fine if we use this project, I could also see having to pull it off mounted secret. If the other parts use the cluster project, then we should just be consistent.

Harwayne: Isn't this how the Topic operation works?

https://github.com/google/knative-gcp/blob/e0b81bfb474052be268ac900533095bcca38df8d/pkg/operations/pubsub/topic.go#L95

But I agree that we should create one function that everything in this repo uses, so that we consistently choose projects (e.g. spec.project -> secret's project -> metadata server's project).

vaikas-google: Yes, I guess i worded poorly. What I meant to say that I can see that's how we do it today, what I was trying to say is that if we're now thinking that perhaps we should use the project the secret was created, then we should do it everywhere.
I think it's fine to do it this way now, it's consistent with other usages in this repo, so let's do it.

n3wscott: this happens inside the job using the provided secret, see the path:

`knative-gcp/pkg/operations/pubsub/topic.go`

It should not happen in the reconciler. You will get the wrong project id.

Harwayne: > this happens inside the job using the provided secret, see the path:
> 
> `knative-gcp/pkg/operations/pubsub/topic.go`
> 
> It should not happen in the reconciler. You will get the wrong project id.

I completely agree that it happens inside the Job, not the reconciler. But I don't see anything in that Job code that looks at the secret to get the project. Nor do I see anything in the docs for the metadata library that says it looks at anything but the instance's metadata server. Does the use of the secret somehow modify the values in the metadata server?

n3wscott: It is a pending feature. The goal is to keep spec.projectId blank and use the metadata server, and use the code Ville just added to get the resolved project id from the job and save it in status.projectId. I did not have a good way to get that info so it was left as a todo.

PR #187, opened 2019-08-19
Convert storage to operations model

vaikas-google:
- run storage notification / topic related operations in jobs in the importers namespace. Controller does not need to have the secret anymore.
 - move operations related code to pkg/operations/<sub type> for easier organization
 - use fixed name rather than generate name for storage related jobs
 - use jobs (well, their pods) termination messages to communicate information / error details to the controller
 -


pkg/operations/storage/names.go
```diff
@@ -0,0 +1,36 @@
...
+
+import (
+	"strings"
+
+	"knative.dev/pkg/kmeta"
+)
+
+// NotificationJobName creates the name of a notification ops job.
+func NotificationJobName(owner kmeta.OwnerRefable, action string) string {
+	return strings.ToLower(
```
n3wscott: You will want to use the kmeta.ChildName method here to avoid overly long names

n3wscott:
/lgtm
/hold

I think https://github.com/google/knative-gcp/pull/187#pullrequestreview-276597099 could be an issue created, tracked and fixed in a followup.

PR #201, opened 2019-08-21
Updates to storage schema

nachocano:
- Updating the storage schema to match to match https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations


nachocano:
/assign @n3wscott 
Scott if we can get this in for the pilot it'd be great! Thanks!

PR #203, opened 2019-08-21
make Storage conform to duckv1beta1.Source duck type

vaikas-google:
* Make Storage type conform to duckv1beta1.Source duck type
 * Don't resolve the sink since we don't use it, have PullSubscription do it and use it
 * Start adding unit tests, tighten the validation for Storage types.


pkg/apis/events/v1alpha1/storage_types.go
```diff
@@ -37,16 +41,28 @@ type Storage struct {
...
+	_ apis.Validatable   = (*Storage)(nil)
+	_ apis.Defaultable   = (*Storage)(nil)
+	_ runtime.Object     = (*Storage)(nil)
+	_ kmeta.OwnerRefable = (*Storage)(nil)
+	_ webhook.GenericCRD = (*Storage)(nil)
+)
+
 // Check that Storage implements the Conditions duck type.
 var _ = duck.VerifyType(&Storage{}, &duckv1beta1.Conditions{})
+var _ = duck.VerifyType(&Storage{}, &duckv1beta1.Source{})
```
n3wscott: I guess there is something to do with performance or whatever so serving has been making this test instead of doing the asserts like above: https://github.com/google/knative-gcp/blob/master/pkg/apis/pubsub/v1alpha1/implements_test.go#L26

it lets you test round trips in json form so that is useful.

n3wscott:
one nit about adding a test. Added an issue to track.

/lgtm
/approve

pkg/apis/events/v1alpha1/storage_types.go
vaikas-google: cool, done. Also added briefly a dummy one (duckv1beta1.Addressable) and validated that the test failed, obvs took it out now but it does indeed work.

PR #208, opened 2019-08-21
Update storage schema

nachocano:
- Using the schema from https://raw.githubusercontent.com/googleapis/google-api-go-client/master/storage/v1/storage-api.json.
- Need to figure out a way not to host it ourselves, and point to a public language agnostic link


n3wscott:
/ok-to-test

I will let ville look at the schema details

PR #214, opened 2019-08-23
Storage field rename

vaikas-google:
Changing couple of externally visible field names for consistency. 
** This is a breaking change**
You have to delete existing ones before upgrading.

Storage.Spec.GCSSecret => Secret
Storage.Spec.PullSubscriptionSecret => PubSubSecret

Rename some internal conditions for shorter names.

### DIFF
--- a/pkg/apis/events/v1alpha1/storage_types.go
+++ b/pkg/apis/events/v1alpha1/storage_types.go
@@ -53,19 +53,19 @@ type StorageSpec struct {
 	// This brings in CloudEventOverrides and Sink
 	duckv1beta1.SourceSpec
 
-	// GCSSecret is the credential to use to create the Notification on the GCS bucket.
+	// Secret is the credential used to manage Notifications on the GCS bucket.
 	// The value of the secret entry must be a service account key in the JSON format (see
 	// https://cloud.google.com/iam/docs/creating-managing-service-account-keys).
 	// +optional
-	GCSSecret corev1.SecretKeySelector `json:"gcsSecret,omitempty"`
+	Secret *corev1.SecretKeySelector `json:"Secret,omitempty"`
 
-	// PullSubscriptionSecret is the credential to use for the GCP PubSub Subscription.
+	// PubSubSecret is the credential to use for the GCP PubSub Subscription.
 	// It is used for the PullSubscription that is used to deliver events from GCS.
 	// The value of the secret entry must be a service account key in the JSON format (see
 	// https://cloud.google.com/iam/docs/creating-managing-service-account-keys).
-	// If omitted, uses GCSSecret from above.
+	// If omitted, uses Secret from above.
 	// +optional
-	PullSubscriptionSecret *corev1.SecretKeySelector `json:"pullSubscriptionSecret,omitempty"`
+	PubSubSecret *corev1.SecretKeySelector `json:"pullSubscriptionSecret,omitempty"`
 
 	// ServiceAccountName holds the name of the Kubernetes service account
 	// as which the underlying K8s resources should be run. If unspecified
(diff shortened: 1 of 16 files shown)


Harwayne:
Note that you will need to update the commands in the guide too.

pkg/apis/events/v1alpha1/storage_types.go
```diff
@@ -53,19 +53,19 @@ type StorageSpec struct {
 	// This brings in CloudEventOverrides and Sink
 	duckv1beta1.SourceSpec
 
-	// GCSSecret is the credential to use to create the Notification on the GCS bucket.
+	// Secret is the credential used to manage Notifications on the GCS bucket.
 	// The value of the secret entry must be a service account key in the JSON format (see
 	// https://cloud.google.com/iam/docs/creating-managing-service-account-keys).
 	// +optional
-	GCSSecret corev1.SecretKeySelector `json:"gcsSecret,omitempty"`
+	Secret *corev1.SecretKeySelector `json:"Secret,omitempty"`
```
akashrv: camel case?

pkg/apis/events/v1alpha1/storage_types.go
```diff
@@ -107,14 +107,15 @@ const (
 	// TopicReady has status True when the underlying GCP PubSub topic is ready
 	TopicReady apis.ConditionType = "TopicReady"
 
-	// GCSReady has status True when GCS has been configured properly to send Notification events
-	GCSReady apis.ConditionType = "GCSReady"
+	// StorageNotificationReady has status True when GCS has been configured properly to
+	// send Notification events
+	NotificationReady apis.ConditionType = "NotificationReady"
```
akashrv: StorageNotificationReady? That will make it more relevant for end user. II haven't gone over the code in detail but assuming that the notificationJob will also fail in case the bucket doesn't exist.

pkg/reconciler/storage/storage.go
```diff
@@ -231,11 +231,11 @@ func (c *Reconciler) reconcile(ctx context.Context, csr *v1alpha1.Storage) error
 	if err != nil {
 		// TODO: Update status with this...
 		c.Logger.Infof("Failed to reconcile Storage Notification: %s", err)
-		csr.Status.MarkGCSNotReady("GCSNotReady", "Failed to create Storage notification: %s", err)
+		csr.Status.MarkNotificationNotReady("GCSNotReady", "Failed to create Storage notification: %s", err)
```
akashrv: GCSNotReady?  Just wanted to make sure you left this on purpose.

pkg/operations/storage/notification.go
```diff
@@ -153,7 +153,7 @@ type NotificationOps struct {
 	// Topic is the environment variable containing the PubSub Topic being
 	// subscribed to's name. In the form that is unique within the project.
 	// E.g. 'laconia', not 'projects/my-gcp-project/topics/laconia'.
-	Topic string `envconfig:"PUBSUB_TOPIC_ID" required:"true"`
+	Topic string `envconfig:"PUBSUB_TOPIC_ID" required:"false"`
```
akashrv: why false? And if it is not required then don't you want to check during operations.ActionCreate case that it is given?

vaikas-google:
> Note that you will need to update the commands in the guide too.

@Harwayne I thought this would not touch that.

pkg/reconciler/storage/storage.go
vaikas-google: nope, fixed.

pkg/apis/events/v1alpha1/storage_types.go
vaikas-google: My reasoning is that this is already part of the Storage object, so having StorageNotification is redundant there.

pkg/operations/storage/notification.go
vaikas-google: not needed for the delete. The call will fail without it and the error messages are logged, so figured that was fine.
Anyways, added a check for it.

vaikas-google:
@Harwayne @akashrv Thanks for the review, PTAL.

Harwayne:
> > Note that you will need to update the commands in the guide too.
> 
> @Harwayne I thought this would not touch that.

My mistake, I thought we still specified the secret, but looking at the guide I see that we let the webhook do it for us.

PR #247, opened 2019-09-04
pubsub reconciler and pod_test unit tests

vaikas-google:


n3wscott:
I never knew I would get excited about unit tests, but I do.
/lgtm

Thanks!!!

PR #250, opened 2019-09-09
[WIP] PullSubscription and Storage importer metrics

nachocano:
- Adding `event_count` metrics to PullSubscription and Storage.
- Commenting out current observability pattern. I think we can do a proper refactor if we think that is the right way to go later on. 
- ConfigMaps are still base64-based. Need to change that.
- Changes due to updates to pkg and cloudevents SDK versions. 

- Tested PullSubscription and is working. The service account the receive adapter is running on needs to have `roles/monitoring.metricWriter` permissions to use the stackdriver monitoring APIs... 
Example to set it up:

```
gcloud projects add-iam-policy-binding $PROJECT_ID \
  --member=serviceAccount:cloudrunevents-pullsub@$PROJECT_ID.iam.gserviceaccount.com \
  --role roles/monitoring.metricWriter
```
@n3wscott seems to have a better idea...

Note: I think we will be tagging storage counts with pullsubscription value tags, as I'm setting up the receive_adapter right now.. That will need to change...


Gopkg.toml
```diff
@@ -21,7 +21,8 @@ required = [
 
 [[override]]
   name = "knative.dev/pkg"
-  branch = "master"
+  source = "github.com/nachocano/pkg"
```
n3wscott: /hold 
until Nacho can finish upstreaming that change into pkg.

cmd/pubsub/receive_adapter/main.go
```diff
@@ -99,19 +90,22 @@ func mainMetrics(logger *zap.SugaredLogger, opts *metrics.ExporterOptions) {
...
+	//  They seem to be private ones and more profiling related ones.
+	//  Commenting them for now, as we will use pkg stats_reporter.
+	// Register the views.
+	//if err := view.Register(
+	//	client.LatencyView,
+	//	transporthttp.LatencyView,
+	//	json.LatencyView,
+	//	xml.LatencyView,
+	//	datacodec.LatencyView,
+	//  adapter.LatencyView,
```
n3wscott: `adapter.LatencyView` is how we get dataplane counts

config/500-controller.yaml
```diff
@@ -60,7 +60,7 @@ spec:
         - name: CONFIG_OBSERVABILITY_NAME
           value: config-observability
         - name: METRICS_DOMAIN
-          value: cloud.run/events
+          value: knative.dev/eventing
```
n3wscott: This seems wrong

config/500-webhook.yaml
```diff
@@ -59,6 +59,8 @@ spec:
                   fieldPath: metadata.namespace
             - name: CONFIG_LOGGING_NAME
               value: config-logging
+            - name: METRICS_DOMAIN
+              value: knative.dev/eventing
```
n3wscott: how will we know this is not knative eventing?

pkg/pubsub/adapter/adapter.go
```diff
@@ -128,30 +144,29 @@ func (a *Adapter) Start(ctx context.Context) error {
...
-	return err
-}
-
-func (a *Adapter) obsReceive(ctx context.Context, event cloudevents.Event, resp *cloudevents.EventResponse) error {
 	logger := logging.FromContext(ctx).With(zap.Any("event.id", event.ID()), zap.Any("sink", a.Sink))
 
+	args := &metrics.ReportArgs{
+		Name:          a.Name,
+		Namespace:     a.Namespace,
+		EventType:     event.Type(),
```
n3wscott: I bet this will not be upstreamed into pkg, it assumed cloudevents...

pkg/pubsub/adapter/adapter.go
```diff
@@ -261,3 +245,15 @@ func (a *Adapter) obsNewHTTPClient(ctx context.Context, target string) (cloudeve
 	// Use the transport to make a new CloudEvents client.
 	return cloudevents.NewClient(t)
 }
+
+// resourceGroupFromEventType returns the resource group for the given eventType.
+func (a *Adapter) resourceGroupFromEventType(eventType string) string {
```
n3wscott: thiz seems like it should be in the converter function area, not here.

pkg/pubsub/adapter/converters/pubsub.go
```diff
@@ -19,20 +19,27 @@ package converters
...
+
+func convertPubsub(ctx context.Context, msg *pubsub.Message, sendMode ModeType) (*cloudevents.Event, error) {
 	tx := cepubsub.TransportContextFrom(ctx)
 	// Make a new event and convert the message payload.
 	event := cloudevents.NewEvent(cloudevents.VersionV03)
 	event.SetID(tx.ID)
 	event.SetTime(tx.PublishTime)
 	event.SetSource(v1alpha1.PubSubEventSource(tx.Project, tx.Topic))
 	event.SetDataContentType(*cloudevents.StringOfApplicationJSON())
-	event.SetType(v1alpha1.PubSubEventType)
```
n3wscott: why not use the versioned type?

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -23,17 +23,18 @@ import (
...
 )
 
 var (
 	// Mapping of GCS eventTypes to CloudEvent types.
 	storageEventTypes = map[string]string{
-		"OBJECT_FINALIZE":        "google.storage.object.finalize",
-		"OBJECT_ARCHIVE":         "google.storage.object.archive",
-		"OBJECT_DELETE":          "google.storage.object.delete",
-		"OBJECT_METADATA_UPDATE": "google.storage.object.metadataUpdate",
+		"OBJECT_FINALIZE":        Finalize,
```
n3wscott: `Finalize` and all other keys need to be storage namedspaced. This is at the converter package level so it will conflict in the future with the next component that has `Finalize`.

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -44,13 +45,20 @@ const (
...
+
+	StorageResourceGroup = "storages.events.cloud.run"
 )
 
 func storageSource(bucket string) string {
 	return fmt.Sprintf("%s/%s", storageSourcePrefix, bucket)
 }
 
-func convertStorage(ctx context.Context, msg *cepubsub.Message, sendMode ModeType) (*cloudevents.Event, error) {
+func convertStorage(ctx context.Context, msg *pubsub.Message, sendMode ModeType) (*cloudevents.Event, error) {
```
n3wscott: are you sure `cepubsub` to `pubsub` rename is what should happen here?

pkg/pubsub/adapter/push.go
```diff
@@ -6,7 +22,7 @@ import (
 	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go"
-	cepubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub"
+	cepubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub/context"
```
n3wscott: this seems wrong

pkg/pubsub/adapter/push_test.go
```diff
@@ -8,7 +24,7 @@ import (
 	"github.com/google/go-cmp/cmp"
 
 	cloudevents "github.com/cloudevents/sdk-go"
-	cepubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub"
+	cepubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub/context"
```
n3wscott: ?? why is this changing?

pkg/reconciler/pullsubscription/pullsubscription.go
```diff
@@ -56,7 +56,8 @@ const (
 	// ReconcilerName is the name of the reconciler
 	ReconcilerName = "PullSubscriptions"
 
-	component = "pullsubscriptions"
+	// TODO change to importer once StackDriver MonitorResources are approved.
+	component = "gcpimporter"
```
n3wscott: We woould like to avoid using `importer` in the code anywhere.

cmd/pubsub/receive_adapter/main.go
nachocano: yeah, changed it to use that stats_reporter we are using in other places... 
I put a TODO there to see if we do a refactor later on...

config/500-controller.yaml
nachocano: In order to register monitored resources with stackdriver, we need to specify the metrics domain. We are proposing three new resources: knative_broker, knative_trigger, knative_importer, which have metrics in certain domain. E.g.,  
`knative.dev/eventing/importer/event_count `... 
`knative.dev/eventing/importer/dispatch_latencies`, ...

I agree it would be nice to keep the cloud.run/events, but I think this is the generic way allowing importers from different parties to leverage SD monitored resources. Otherwise, they will need to go over a SD approval process to add new metrics to those resources for their own domains... Does it make sense?

config/500-webhook.yaml
nachocano: We don't. serving does the same, according to my understanding. Do you see a problem with this?

pkg/pubsub/adapter/adapter.go
nachocano: yeah, you are probably right... But you will be surprised the things that have been upstreamed to pkg... Metrics in particular is something that was pretty coupled with serving. It is now also coupled with eventing :), and needs a proper refactor... See this [issue](https://github.com/knative/pkg/issues/608)...

As I saw the [source_types](https://github.com/knative/pkg/blob/528ad1c1dd627059b95aef17ccf27f9ab0f46c10/apis/duck/v1/source_types.go#L59) duck type there, which has a CloudEventOverrides field, I also moved this one down to pkg in order to avoid duplicating code all over... 
If there are objections of moving this down, we can further discuss and see the best quickest approach

pkg/pubsub/adapter/adapter.go
nachocano: yeah, good point... let me move it there

pkg/pubsub/adapter/converters/pubsub.go
nachocano: this is the only place we are using it, so I moved it here... But if you want it in the versioned type, I can move it back...

pkg/pubsub/adapter/converters/storage.go
nachocano: will do, thanks!

pkg/pubsub/adapter/converters/storage.go
nachocano: is still the cepubsub, but with a different alias... is confusing, you are right...
things have moved around in the sdk, that is why we need to different imports...
let me know your preference of aliases.

pkg/pubsub/adapter/push.go
nachocano: hmm.. See [here](https://github.com/cloudevents/sdk-go/blob/5088975a189005f96a59d4e375b9e88ead0ebe5c/pkg/cloudevents/transport/pubsub/context/context.go#L83)... it seems a change from 21 days back

pkg/pubsub/adapter/push_test.go
nachocano: same as above...

pkg/reconciler/pullsubscription/pullsubscription.go
nachocano: let's sync on this offline... We are proposing a SD monitored resource called importer, so we may need to revisit that as well.

config/500-webhook.yaml
n3wscott: Serving does not have anything like this repo as extensions, so they have not had to think about or solve this.

pkg/pubsub/adapter/converters/storage.go
n3wscott: I prefer the alias.

pkg/pubsub/adapter/push.go
n3wscott: Then the import should change to `pubsubcontext`

pkg/pubsub/adapter/push_test.go
n3wscott: the import should change to `pubsubcontext`

pkg/pubsub/adapter/converters/storage.go
nachocano: left it as is, and changed the other one to cepubsubcontext

pkg/pubsub/adapter/push.go
nachocano: left is as `cepubsubcontext` to match the other ones

pkg/pubsub/adapter/push_test.go
nachocano: same as above, put it as `cepubsubcontext`. If you don't like it, I can change it to `pubsubcontext`

nachocano:
need to wait until https://github.com/knative/pkg/pull/648 and https://github.com/knative/pkg/pull/656 are merged

n3wscott:
this should fix your deps issues: https://github.com/google/knative-gcp/pull/254

Gopkg.lock
```diff
@@ -1137,15 +1139,15 @@
...
   name = "knative.dev/eventing"
   packages = ["pkg/apis/duck/v1alpha1"]
   pruneopts = "NUT"
-  revision = "9014140a6b7df139bc08c91c3a1ae0628d842665"
+  revision = "8679b4ef150dc92fef25345d62cd36ffea7b4b15"
 
 [[projects]]
-  branch = "master"
-  digest = "1:80ebfb8d1bf38f2d1e3bf52eaa074a62e04c6b506ea5f00dc39596425e00ee5a"
+  branch = "source_stats"
```
n3wscott: this is your merge issue.

Gopkg.toml
nachocano: /hold cancel
changed has been upstreamed

nachocano:
@n3wscott ready for another look...

nachocano:
@n3wscott any idea of what can be wrong with the e2e here?

nachocano:
e2e tests are failing due to a bug that this [PR](https://github.com/cloudevents/sdk-go/pull/185) will solve. Will update the CloudEvents SDK version then, and give it another try

PR #253, opened 2019-09-10
Pubsubable into an interface for easier use.

vaikas-google:
Now with the combination of using the PubSub duck type and implementing the PubSubable interface, you can call the pubsub_reconciler ReconcilePubSub with the original object rather than pull it apart yourself

### DIFF
--- a/pkg/apis/events/v1alpha1/storage_types.go
+++ b/pkg/apis/events/v1alpha1/storage_types.go
@@ -81,7 +81,7 @@ const (
 	// StorageConditionReady has status True when the Storage is ready to send events.
 	StorageConditionReady = apis.ConditionReady
 
-	// StorageNotificationReady has status True when GCS has been configured properly to
+	// NotificationReady has status True when GCS has been configured properly to
 	// send Notification events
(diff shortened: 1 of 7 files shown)


pkg/apis/events/v1alpha1/storage_types.go
```diff
@@ -106,6 +106,23 @@ func (storage *Storage) GetGroupVersionKind() schema.GroupVersionKind {
...
+}
+
+// PubSubStatus returns the PubSubStatus portion of the Status.
+func (s *Storage) PubSubStatus() *duckv1alpha1.PubSubStatus {
+	return &s.Status.PubSubStatus
+}
+
+// ConditionSet returns the apis.ConditionSet of the embedding object
+func (s *Storage) ConditionSet() *apis.ConditionSet {
+	return &StorageCondSet
```
n3wscott: You might want to change this to the thing returned after you call .Manage on the condition set

n3wscott: It depends on the impk of the mark methods

n3wscott:
Nice that this mirrors how we do this same thing in other places.

PR #273, opened 2019-09-16
Only fail if the project ID is missing when doing a Create operation for notification

grac3gao:
When trying to delete storage, it will fail for notification was unable to delete. Error: missing project_ID will be returned 
Sync with @vaikas-google, only fail if the project ID is missing when doing a Create operation for notification (not for Delete operation)

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,8 +10,19 @@ independently, but is best mixed with [Knative](https://knative.dev).
 
 To get started, [install Knative with GCP](./docs/install/README.md).
 
-Then use one of the implemented resources,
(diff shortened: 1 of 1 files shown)


vaikas-google:
Could you also please modify the unit test to make sure that without projectid things work?
https://github.com/google/knative-gcp/blob/master/pkg/operations/storage/notification_test.go#L302

And, then move this inside create only section in the L349 vicinity.
https://github.com/google/knative-gcp/blob/master/pkg/operations/storage/notification_test.go#L339

PR #278, opened 2019-09-17
force events to be v03 in channel.

n3wscott:


pkg/pubsub/publisher/publisher.go
```diff
@@ -62,6 +62,11 @@ func (a *Publisher) Start(ctx context.Context) error {
 }
 
 func (a *Publisher) receive(ctx context.Context, event cloudevents.Event, resp *cloudevents.EventResponse) error {
+	// Upgrade to supported transport version.
```
vaikas-google: Just to make sure, this is totally lossless? If I put a v2 there, then it gets upgraded to v3, but my function expects v2, then this is fine and nothing is lost?

n3wscott: pubsub does not support anything except for v03. The clients all should expect this

n3wscott: the diff between v02 and v03 is minimal

PR #280, opened 2019-09-17
[e2e] adding knock down test for channel.

n3wscott:


n3wscott:
this might be an issue with eventing being old.... hang tight

nachocano:
/close

as @chizhg just merged some channel e2e tests. This is old

PR #283, opened 2019-09-17
E2E test for the storage source/importer

### LINKED ISSUE #205: E2E test for the storage source/importer
We need an E2E for the storage source/importer. It will have to do the proper cluster setup (e.g., creating secrets, a bucket, etc.), then upload a file, and see that a notification is received on the sink.

There is a working example with PullSubscription, where we might be able to reuse some of the code. Need to sync with @n3wscott

grac3gao:
fix #205 

1. To run manually using go test and an existing cluster
2. run ./e2e-tests.sh

### DIFF
--- a/test/cmd/storageTarget/main.go
+++ b/test/cmd/storageTarget/main.go
@@ -1,0 +1,64 @@
+
+package main
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io/ioutil"
+	"log"
+	"os"
+	"strings"
+
+	cloudevents "github.com/cloudevents/sdk-go"
+	"github.com/kelseyhightower/envconfig"
+)
+
+func main() {
+	client, err := cloudevents.NewDefaultClient()
+	if err != nil {
+		panic(err)
+	}
+
+	r := Receiver{}
+	if err := envconfig.Process("", &r); err != nil {
+		panic(err)
+	}
+
+	fmt.Printf("Type prefix to match: %q.\n", r.Type)
+
+	if err := client.StartReceiver(context.Background(), r.Receive); err != nil {
+		log.Fatal(err)
+	}
+}
+
+type Receiver struct {
+	Type string `envconfig:"TYPE" required:"true"`
+}
+
+func (r *Receiver) Receive(event cloudevents.Event) {
+	eventType := event.Context.GetType()
+	var success bool
+	if strings.HasPrefix(eventType, r.Type) {
+		fmt.Printf("Target prefix matched, %q.\n", r.Type)
+		success = true
+	} else {
+		fmt.Printf("Target prefix did not match, %q != %q.\n", eventType, r.Type)
+		success = false
+	}
+	// Write the termination message.
+	if err := r.writeTerminationMessage(map[string]interface{}{
+		"success": success,
+	}); err != nil {
+		fmt.Printf("failed to write termination message, %s.\n", err)
+	}
+	os.Exit(0)
+}
+
+func (r *Receiver) writeTerminationMessage(result interface{}) error {
+	b, err := json.Marshal(result)
+	if err != nil {
+		return err
+	}
+	return ioutil.WriteFile("/dev/termination-log", b, 0644)
+}
--- a/test/cmd/storageTarget/main.go
+++ b/test/cmd/storageTarget/main.go
@@ -1,4 +1,3 @@
-
 package main
 
 import (
@@ -8,7 +7,9 @@ import (
 	"io/ioutil"
 	"log"
 	"os"
+	"strconv"
 	"strings"
+	"time"
 
 	cloudevents "github.com/cloudevents/sdk-go"
 	"github.com/kelseyhightower/envconfig"
@@ -25,7 +26,23 @@ func main() {
 		panic(err)
 	}
 
-	fmt.Printf("Type prefix to match: %q.\n", r.Type)
+	fmt.Printf("Subject to match: %q.\n", r.Subject)
+
+	// Create a timer
+	duration, _ := strconv.Atoi(r.Time)
+	timer := time.NewTimer(time.Second * time.Duration(duration))
+	defer timer.Stop()
+	go func() {
+		<-timer.C
+		// Write the termination message if time out
+		fmt.Printf("time out to wait for event with subject %q.\n", r.Subject)
+		if err := r.writeTerminationMessage(map[string]interface{}{
+			"success": false,
+		}); err != nil {
+			fmt.Printf("failed to write termination message, %s.\n", err)
+		}
+		os.Exit(0)
+	}()
 
 	if err := client.StartReceiver(context.Background(), r.Receive); err != nil {
 		log.Fatal(err)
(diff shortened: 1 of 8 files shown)


test/cmd/storageTarget/main.go
```diff
@@ -0,0 +1,64 @@
...
+
+type Receiver struct {
+	Type string `envconfig:"TYPE" required:"true"`
+}
+
+func (r *Receiver) Receive(event cloudevents.Event) {
+	eventType := event.Context.GetType()
+	var success bool
+	if strings.HasPrefix(eventType, r.Type) {
+		fmt.Printf("Target prefix matched, %q.\n", r.Type)
```
n3wscott: just matching type will not tell you which event you are getting

test/e2e/config/storage_test/target.yaml
```diff
@@ -0,0 +1,39 @@
...
+spec:
+  selector:
+    targetUID: {{ .targetUID }}
+  ports:
+    - name: http
+      port: 80
+      protocol: TCP
+      targetPort: 8080
+  sessionAffinity: None
+  type: ClusterIP
```
n3wscott: nit: need a new line here.

test/e2e/test_storage.go
```diff
@@ -0,0 +1,181 @@
...
+	"fmt"
+	"github.com/google/uuid"
+	"google.golang.org/api/iterator"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"knative.dev/pkg/test/helpers"
+	"os"
+	"strings"
+	"time"
+
+	// Uncomment the following line to load the gcp plugin (only required to authenticate against GKE clusters).
```
n3wscott: this comment should be attached to the `_ "k8s.io...` line.

test/e2e/test_storage.go
```diff
@@ -0,0 +1,181 @@
...
+		t.Fatalf("failed to create storage client, %s", err.Error())
+	}
+	it := client.Buckets(ctx, project)
+	bucketName := helpers.AppendRandomString(bucketPrefix)
+	for {
+		bucketAttrs, err := it.Next()
+		if err == iterator.Done {
+			// Create a new bucket if there is no qualified existing bucket
+			bucket := client.Bucket(bucketName)
+			if e := bucket.Create(ctx, project, &storage.BucketAttrs{}); e != nil {
```
n3wscott: you are creating a bucket for each test run but not deleting the bucket. 

I think you need to delete the bucket or use the same bucket for each.

test/e2e/test_storage.go
```diff
@@ -0,0 +1,181 @@
...
+		}
+		if strings.HasPrefix(bucketAttrs.Name, bucketPrefix) {
+			bucketName = bucketAttrs.Name
+			break
+		}
+	}
+	return bucketName
+}
+
+func getBucketHandler(t *testing.T, bucketName string) *storage.BucketHandle {
```
n3wscott: it seems it is not a handler, it is a handle. `getBucketHandler` -> `getBucketHandle`. Which is also pretty funny api name. 😆

test/e2e/test_storage.go
```diff
@@ -0,0 +1,181 @@
...
+		Resource: "jobs",
+	}
+
+	if err := client.WaitForResourceReady(client.Namespace, storageName, gvr); err != nil {
+		t.Error(err)
+	}
+
+	// Add a file in the bucket
+	bucketHandler := getBucketHandler(t, bucketName)
+	wc := bucketHandler.Object("testFile").NewWriter(ctx)
```
n3wscott: I think the test should be you generate a random file name, and pass this as the knock down key on the target. Otherwise you don't know the target it falling down because of _this_ file upload or something else.

test/e2e-common.sh
```diff
@@ -84,16 +86,40 @@ function pubsub_setup() {
...
+function teardown() {
+  # When not running on Prow we need to delete the service account created
   if (( ! IS_PROW )); then
     echo "Tear down ServiceAccount for Pub/Sub Admin"
     gcloud iam service-accounts keys delete -q ${PUBSUB_SERVICE_ACCOUNT_KEY} \
       --iam-account=${PUBSUB_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com
     gcloud projects remove-iam-policy-binding ${E2E_PROJECT_ID} \
       --member=serviceAccount:${PUBSUB_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
       --role roles/pubsub.editor
+    cho "Tear down ServiceAccount for Storage Admin"
```
n3wscott: echo

vaikas-google:
Looks like service account might be needing some attention?

--- PASS: TestSmokeChannel (20.36s)
    lifecycle.go:53: namespace is : "test-smoke-channel-hccimkyp"
2019/09/17 21:30:34 storage-e2e-test-utermioa-storage is not ready, NotificationNotReady: Failed to create Storage notification: operation failed: googleapi: Error 403: The service account 'service-758498956531@gs-project-accounts.iam.gserviceaccount.com' does not have permission to publish messages to to the Cloud Pub/Sub topic '//pubsub.googleapis.com/projects/knative-boskos-68/topics/storage-54292b44-d992-11e9-a940-42010a800073', or that topic does not exist., forbidden
{"level":"info","ts":1568755837.4579425,"logger":"fallback","caller":"operations/job.go:106","msg":"Looking for Pod with jobname: \"ps-e2e-test-ncmkpdwz-target\""}

grac3gao:
Made following modification in this commit, according to code review and advice of attention for service account:
1. for the bucket, generate a random name file into bucket (used to be a fixed name), and delete it after testing
2. for the target pod, use the random file name as the knock down key, set a duration time (currently set as 2 min). If the target pod doesn't receive an event with knock down key as its subject in this period, failed the test. (used to only check the prefix for the first event's eventType)
3. for the service account (storage admin) for integration test, it seems that it takes some time for the storage admin role to setup for the bucket. That is the reason why it will get error (does not have permission), even I have updated storage/admin role in e2e-common.sh for the service account. Sleep for 2 min to wait for all resources setup ready. This runs well locally in my project(runing ./e2e-tests.sh). However, according to e2e-common.sh, we use GOOGLE_APPLICATION_CREDENTIALS in PROW, and don't set up a new service account for admin, not sure it will also work in PROW. Use this commit to try out...

grac3gao:
Currently, skip the e2e storage test in PROW integration test (works locally with ./e2e-tests.sh, and manually with go test, but doesn't work in PROW because of a GOOGLE_APPLICATION_CREDENTIALS issue), issue #290

test/cmd/storageTarget/main.go
```diff
@@ -0,0 +1,80 @@
+package main
```
nachocano: let's rename the folder to be `storage_target` instead of `storageTarget`

test/cmd/storageTarget/main.go
```diff
@@ -0,0 +1,80 @@
...
+
+type Receiver struct {
+	Subject string `envconfig:"SUBJECT" required:"true"`
+	Time    string `envconfig:"TIME" required:"true"`
+}
+
+func (r *Receiver) Receive(event cloudevents.Event) {
+	eventSubject := event.Context.GetSubject()
+	fmt.Printf(event.Context.String())
+	if strings.Compare(eventSubject, r.Subject) == 0 {
```
nachocano: can we do `if eventSubject == r.Subject` instead?

test/cmd/storageTarget/main.go
```diff
@@ -0,0 +1,80 @@
...
+		fmt.Printf("subject matches, %q.\n", r.Subject)
+		// Write the termination message if the subject successfully matches
+		if err := r.writeTerminationMessage(map[string]interface{}{
+			"success": true,
+		}); err != nil {
+			fmt.Printf("failed to write termination message, %s.\n", err)
```
nachocano: shouldn't it be `err.Error()` ?

test/e2e/e2e_test.go
```diff
@@ -29,6 +29,7 @@ import (
 
 var packages = []string{
 	"github.com/google/knative-gcp/test/cmd/target",
+	"github.com/google/knative-gcp/test/cmd/storageTarget",
```
nachocano: will change to `storage_target`

test/e2e/test_storage.go
```diff
@@ -0,0 +1,191 @@
...
+			if e := bucket.Create(ctx, project, &storage.BucketAttrs{}); e != nil {
+				t.Fatalf("failed to create bucket, %s", e.Error())
+			}
+			break
+		}
+		if err != nil {
+			t.Fatalf("failed to list buckets, %s", err.Error())
+		}
+		// Break iteration if there has a bucket for e2e test
+		if strings.Compare(bucketAttrs.Name, bucketName) == 0 {
```
nachocano: same as above, can we compare with ==?

test/e2e/test_storage.go
```diff
@@ -0,0 +1,191 @@
...
+	"time"
+
+	"testing"
+
+	// Uncomment the following line to load the gcp plugin (only required to authenticate against GKE clusters).
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
+)
+
+// makeBucket will get an existing bucket or create a new one if it is not existed
+func makeBucket(t *testing.T) string {
```
nachocano: // makeBucket retrieves the bucket name for the test. If it does not exist, it will create it.

test/e2e/test_storage.go
```diff
@@ -0,0 +1,191 @@
...
+		}
+		// Break iteration if there has a bucket for e2e test
+		if strings.Compare(bucketAttrs.Name, bucketName) == 0 {
+			break
+		}
+	}
+	return bucketName
+}
+
+func getBucketHandle(t *testing.T, bucketName string) *storage.BucketHandle {
```
nachocano: pass the context as first argument, so that you don't keep on creating a new one...
`ctx context.Context`

test/e2e/test_storage.go
nachocano: pass the context as first argument, so that you don't keep on creating a new one...
`ctx context.Context`

test/e2e/test_storage.go
nachocano: also you can pass the projectId that you read only once from StorageWithTestImpl..
same for the makeBucket

nachocano:
/lgtm
Thanks so much for doing this @grac3gao!

Issue #303, opened 2019-09-25
Implement PubSub source


n3wscott:
We need to implement a `PubSub` source object that is similar to `PullSubscription` except it decouples the _how_ (pull subscription) from the _what_ (deliver events from a topic). 

Requirements:

[] Is in the `events.cloud.run` api group.
[] Implements the `Source` ducktype.

Likely there is not a ton of work here, we need to cherrypick the spec parts of PullSubscription spec (avoid mode), (promote the secret). This lets a user source events from Pub/Sub without being coupled to the fact that the implementation. For example, in kubernetes needs to be a PullSubscription, but Cloud Run, this could be a Push Subscription.

/assign @nachocano

PR #306, opened 2019-09-26
Pubsub Source

nachocano:
Fixes https://github.com/google/knative-gcp/issues/303

* Adding a PubSub source CRD. 
* Adding types, defaults, validation.
* Adding reconciler, which just creates a PullSubscription
* Currently exposing all stuff from PullSubscription but the mode.

* Pending (will sync with @n3wscott personally)
  * need to move PubSub base reconciler to some other folder, otherwise it is somewhat confusing... Can be done in a subsequent PR

/cc @n3wscott @vaikas-google


pkg/reconciler/pubsub/pubsub.go
```diff
@@ -0,0 +1,212 @@
...
+
+const (
+	finalizerName = controllerAgentName
+
+	resourceGroup = "pubsubs.events.cloud.run"
+)
+
+// Reconciler is the controller implementation for the PubSub source.
+type Reconciler struct {
+	*reconciler.PubSubBase
```
n3wscott: I think we can just have normal base.

pkg/reconciler/pubsub/pubsub.go
```diff
@@ -0,0 +1,212 @@
...
+
+	if topic == "" {
+		topic = ps.Spec.Topic
+	}
+
+	// See if the source has been deleted.
+	deletionTimestamp := ps.DeletionTimestamp
+
+	if deletionTimestamp != nil {
+		if err := c.PubSubBase.DeletePubSub(ctx, ps.Namespace, ps.Name); err != nil {
```
n3wscott: This rec should not worry about talking to punbsub, just delete the PullSubscription.

n3wscott: or let the owner refs cleanup

pkg/reconciler/pubsub/pubsub.go
```diff
@@ -0,0 +1,212 @@
...
+			c.Logger.Infof("Unable to delete pubsub resources : %s", err)
+			return fmt.Errorf("failed to delete pubsub resources: %s", err)
+		}
+		c.removeFinalizer(ps)
+		return nil
+	}
+
+	// Ensure that there's finalizer there, since we're about to attempt to
+	// change external state with the topic, so we need to clean it up.
+	err := c.ensureFinalizer(ps)
```
n3wscott: This resource likely does not need a finalizer

pkg/reconciler/pubsub/pubsub.go
```diff
@@ -0,0 +1,212 @@
...
+	if err != nil {
+		return err
+	}
+
+	t, psubs, err := c.PubSubBase.ReconcilePubSub(ctx, ps, topic, resourceGroup)
```
n3wscott: The main role of the PubSub will be to propagate the status of the PullSubscription to this resource

nachocano:
> The resource looks good, I think the reconciler should be much simpler:
> 
> 1. Just make and own a PullSubscription
> 2. Propagate the status of that PullSubscription into the status of the PubSub
> 3. No need to talk to the PubSub API, if the topic does not exist, then that is fine... the pullsubscription will let you know there is an issue.

Sounds good! Will do. Should be done by Monday at most... Sorry for the delay, been entertained with eventing-contrib

n3wscott:
>Sorry for the delay

No need to hurry. :D

pkg/reconciler/pubsub/pubsub.go
nachocano: Done

pkg/reconciler/pubsub/pubsub.go
nachocano: done! thanks!

pkg/reconciler/pubsub/pubsub.go
nachocano: removed, thanks!

pkg/reconciler/pubsub/pubsub.go
nachocano: removed, thanks!

nachocano:
/hold cancel

@n3wscott ready for another pass

PR #311, opened 2019-09-30
Move to k8s 1.15.3

vaikas-google:
* Move to 1.15.3
 * Use serving resources to v1 resources.


n3wscott:
/lgtm

Thanks for getting that done. It's huge.

PR #326, opened 2019-10-06
Add JobArgs interface and Shared Job Reconciler

ian-mi:
JobArgs encapsulate job arguments for an operation which should be
passed using env vars. Other common operation parameters are moved out
of job arg structs into an a separate OpCtx struct.

In addition to defining a Validate and Env conversion method, valid
JobArgs instances must define methods OperationGroup,
OperationSubgroup, Action, and LabelKey to help the job reconciler
name and label jobs.

Since the required job arguments often vary by action type, existing
job arg structs have been split into separate per-action arg structs
which embed common args. Behavior of these new structs is mostly the
same as before, except for scheduler args which previously used the
"notification" label key.

### Ideas for Improvements

-  Generate env vars using tags, dual to the envconfig tags. Could maybe even use the same struct for args and op env.
- Could also use tags for validation, maybe using govalidator.
- Move more job reconciliation code to job.Reconciler.
- Use injection to create the job reconciler.


vaikas-google:
Great refactor!
/approve
@n3wscott for lgtm

pkg/operations/scheduler/job.go
```diff
@@ -59,95 +56,116 @@ type JobActionResult struct {
...
+		},
+		corev1.EnvVar{
+			Name:  "DATA",
+			Value: s.Data,
+		})
+}
+
+func (s SchedulerJobCreateArgs) Validate() error {
+	if s.TopicID == "" {
+		return fmt.Errorf("missing TopicID")
```
n3wscott: These errors should use the json casing of the field, ie `topicId`, etc

n3wscott:
This is really awesome work!!!

/lgtm
/hold

the hold is if you want to fix the error messages here or remove it and do the rename in a followup.

Strong work!!

ian-mi:
Obsoleted by #444

Issue #310, opened 2019-09-30
Performance test for PubSub Channel


grantr:
We need a perf test, similar to the ones in knative/eventing and eventing-contrib, that measures latency and throughput of the PubSub Channel and records results in Mako.

grantr:
This should test with a Broker.

chizhg:
The benchmarks are now up and running:
https://mako.dev/benchmark?benchmark_key=6592390798245888&tseconds=604800
https://mako.dev/benchmark?benchmark_key=6246352530964480&tseconds=604800

/close

PR #312, opened 2019-09-30
Perf testing for PubSub channel

n3wscott:

PR #328, opened 2019-10-08
[WIP]added perf test for broker with pubsub channel

capri-xiyue:
Release note:
Still working on the perf tests with only pubsub channel


test/performance/tools/common.sh
```diff
@@ -0,0 +1,183 @@
...
+
+# Create serice account secret on the cluster.
+# $1 -> cluster_name, $2 -> cluster_zone
+function create_secret() {
+  echo "Create service account on cluster $1 in zone $2"
+  gcloud container clusters get-credentials $1 --zone=$2 --project=${PROJECT_NAME} || abort "Failed to get cluster creds"
+  kubectl create secret generic service-account --from-file=robot.json=${PERF_TEST_GOOGLE_APPLICATION_CREDENTIALS}
+}
+
+# Set up the user credentials for cluster operations.
```
capri-xiyue: I'm not sure whether I need to set up pubsub related service accounts and secrets in the function "setup_user" or not

test/performance/tools/common.sh
```diff
@@ -0,0 +1,183 @@
...
+# limitations under the License.
+
+# Setup env vars
+export PROJECT_NAME="knative-gcp-performance"
+export USER_NAME="mako-job@knative-gcp-performance.iam.gserviceaccount.com"
+export TEST_ROOT_PATH="$GOPATH/src/github.com/google/knative-gcp/test/performance"
+export KO_DOCKER_REPO="gcr.io/knative-gcp-performance"
+export TEST_CONFIG_VARIANT="continuous"
+export TEST_NAMESPACE="default"
+export PUBSUB_SERVICE_ACCOUNT="performance-pubsub-test"
```
capri-xiyue: I'm not sure whether it's ok that some variables names are the same in performance tests shell script and e2e tests shell script. For example, in https://github.com/google/knative-gcp/blob/master/test/e2e-common.sh, it also have variable like PUBSUB_SERVICE_ACCOUNT, PUBSUB_SERVICE_ACCOUNT_KEY and PUBSUB_SECRET_NAME

chizhg:
/hold
Let's test it before getting it merged.

config/201-channelable-manipulator-clusterrole.yaml
```diff
@@ -17,7 +17,7 @@
 kind: ClusterRole
 apiVersion: rbac.authorization.k8s.io/v1
 metadata:
-  name: events-channel-addressable-resolver
+  name: cloud-run-events-channel-manipulator
```
chizhg: This change seems irrelevant

capri-xiyue: Yep. I think that's just a trivial renaming which is irrelevant to this ticket. I just followed the PR of Scott https://github.com/google/knative-gcp/pull/312/files to rename it.

grantr:
Fixes #310

capri-xiyue:
Current pause the progress of this PR since the tasks related to perf test is in an unstable stage. In addition, have some blockers in knative gcp since one pod with prefix pubsub-s-cre-sub will end in error stage in most cases which prevent the perf test from running. Need to talk with @n3wscott to find out an elegant way to solve this problem.

chizhg:
We can close this PR in favor of https://github.com/google/knative-gcp/pull/395.
Not sure if the problem described in the PR has been solved though.

PR #334, opened 2019-10-10
Channel tracing

Harwayne:
Add tracing to the Channel and thus Topic and PullSubscription. 

There is an E2E test, but it does not pass due to knative/eventing#2046, so it is skipped.


Harwayne:
I think this is ready for review, PTAL.

Harwayne:
Ping.

n3wscott:
> Ping.

for me, most github messages go into a black hole so dark it is hard to see them...

Harwayne:
@n3wscott merged master and fixed the unit test, PTAL.

PR #335, opened 2019-10-10
Switch to desugared logger.

Harwayne:
Switch to desugared logger.

There were actual problems with the PubSub receive adapter. Switched webhook, as well, even though there weren't any issues.


cmd/pubsub/receive_adapter/main.go
```diff
@@ -53,14 +53,15 @@ func main() {
 		}
 	}
 
-	logger, _ := logging.NewLoggerFromConfig(loggingConfig, component)
+	sl, _ := logging.NewLoggerFromConfig(loggingConfig, component)
+	logger := sl.Desugar()
 	defer flush(logger)
-	ctx := logging.WithLogger(signals.NewContext(), logger)
+	ctx := logging.WithLogger(signals.NewContext(), logger.Sugar())
```
evankanderson: Why not pass in `sl` here?

cmd/webhook/main.go
```diff
@@ -58,9 +58,9 @@ func SharedMain(resourceHandlers map[schema.GroupVersionKind]webhook.GenericCRD)
 	if err != nil {
 		log.Fatal("Error parsing logging configuration:", err)
 	}
-	logger, atomicLevel := logging.NewLoggerFromConfig(config, component)
-	defer logger.Sync()
-	logger = logger.With(zap.String("cloud.google.com/events", component))
+	sl, atomicLevel := logging.NewLoggerFromConfig(config, component)
+	defer sl.Sync()
```
evankanderson: Receive Adapter calls `metrics.FlushExporter()` as well as `sl.Sync`. Do we want to do that, here?

n3wscott:
I think you should make this switch from pkg, not from each repo.

cmd/pubsub/receive_adapter/main.go
Harwayne: Done.

My reasoning was that I wanted to pretend that `sl` didn't exist. But it's not a good reason, so switched.

cmd/webhook/main.go
Harwayne: Done.

Harwayne:
Ping.

Harwayne:
@evankanderson All comments have been addressed, PTAL.

Issue #313, opened 2019-09-30
PropagateStatus when Topic Fails in PullSubscription


nachocano:
- We should propagate the status message when the topic does not exist in a pull subscription.
- This will also help debug the new `PubSub` source.

nachocano:
fyi @n3wscott

akashrv:
@nachocano What happens today? Could you explain actual and expected behavior from users perspective? Will help in triage.

nachocano:
the message that the topic does not exist is not propagated back up, thus is not that easy to debug... this is not an alpha blocker, it will be a nice to have.
@n3wscott mentioned that there is a message in the jobs that we can use to add that info...

nachocano:
For example, if you create a PubSub as follows, with a topic that does not exist, you won't get a clear error.

```
apiVersion: events.cloud.google.com/v1alpha1
kind: PubSub
metadata:
  name: wrong-topic
spec:
  topic: non-existing-topic
  sink:
    apiVersion: eventing.knative.dev/v1alpha1
    kind: Broker
    name: default
```

PR #341, opened 2019-10-14
Propagate pubsub subscription/topic failure reason from job pod

yolocs:
To fix https://github.com/google/knative-gcp/issues/313
---
1. Added "reason" to subscription/topic job pod termination message
2. Refactored some common code to retrieve termination reason


n3wscott:
/lgtm
/approve

Looks like it needs a minor rebase.

Issue #71, opened 2019-06-28
Write tests for pubsub adapter


grantr:
Stub tests need to be filled in: https://github.com/GoogleCloudPlatform/cloud-run-events/pull/66/files#diff-e4aeeba1df8e80967ec05b8535f5d297

akashrv:
@grantr please verify and update the issue

grantr:
Unit tests for the pubsub adapter are now in https://github.com/google/knative-gcp/blob/master/pkg/pubsub/adapter/adapter_test.go. They are all stub tests.

PR #345, opened 2019-10-15
Add unit tests for pubsub adapter

yolocs:
To fix: https://github.com/google/knative-gcp/issues/71

### DIFF
--- a/pkg/pubsub/adapter/adapter_test.go
+++ b/pkg/pubsub/adapter/adapter_test.go
@@ -108,8 +108,8 @@ func TestInboundConvert(t *testing.T) {
 			e.SetSource(v1alpha1.PubSubEventSource("proj", "topic"))
 			e.SetDataContentType(*cloudevents.StringOfApplicationJSON())
 			e.SetType(v1alpha1.PubSubPublish)
-			e.SetSchemaURL("http://example.com")
-			e.SetExtension("knativecemode", converters.DefaultSendMode)
+			e.SetDataSchema("http://example.com")
+			e.SetExtension("knativecemode", string(converters.DefaultSendMode))
 			e.Data = []byte("some data")
 			e.DataEncoded = true
 			e.SetExtension("key1", "value1")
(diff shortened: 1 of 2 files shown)


pkg/pubsub/adapter/adapter_test.go
```diff
@@ -17,97 +17,355 @@ limitations under the License.
...
+		Project:          "proj",
+		Topic:            "top",
+		Subscription:     "sub",
+		Sink:             "http://localhost:8081",
+		Transformer:      "http://localhost:8080",
+		ExtensionsBase64: "eyJrZXkxIjoidmFsdWUxIiwia2V5MiI6InZhbHVlMiJ9Cg==",
 	}
+	// This test only does sanity checks to see if all fields are
+	// initialized.
+	a.Start(context.Background())
```
n3wscott: is start blocking?

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -88,7 +88,8 @@ func convertStorage(ctx context.Context, msg *cepubsub.Message, sendMode ModeTyp
 		}
 	}
 	event.SetDataContentType(*cloudevents.StringOfApplicationJSON())
-	event.SetData(msg.Data)
+	event.Data = msg.Data
```
n3wscott: what was the issue here? why did you have to directly set?

pkg/pubsub/adapter/adapter_test.go
yolocs: No. Because it will actually return error as it can't really connect to pubsub.

pkg/pubsub/adapter/converters/storage.go
yolocs: I see in https://github.com/google/knative-gcp/blob/master/pkg/pubsub/adapter/converters/pubsub.go#L44-L45, it directly sets the `Data`. I'm not sure if it's intended to be different. For unit tests here, it only affects how I construct the wanted event.

pkg/pubsub/adapter/adapter_test.go
yolocs: Note added.

pkg/pubsub/adapter/converters/storage.go
n3wscott: ah, this in in convert. +1

PR #351, opened 2019-10-21
Update to latest CloudEvents sdk, adding v1.0-ce support.

n3wscott:


pkg/pubsub/adapter/adapter_test.go
```diff
@@ -46,6 +46,7 @@ func (r *mockStatsReporter) ReportEventCount(args *ReportArgs, responseCode int)
 }
 
 func TestStartAdapter(t *testing.T) {
+	t.Skipf("need to fix the error from call to newPubSubClient: %s", `pubsub: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.`)
```
Harwayne: Do you want to submit with this skipped?

n3wscott: yeah the test is broken

PR #363, opened 2019-10-24
Use the updated Destination

Harwayne:
Related to knative/eventing#1918

- Update PullSubscription to use SourceSpec. 
- Update the following to use the new Destination from knative.dev/pkg.
    - PubSub
    - PullSubscription
    - Scheduler
    - Storage


Harwayne:
Ready for review, PTAL.

pkg/apis/pubsub/v1alpha1/pull_subscription_types.go
```diff
@@ -59,6 +59,9 @@ var _ = duck.VerifyType(&PullSubscription{}, &duckv1.Conditions{})
 
 // PullSubscriptionSpec defines the desired state of the PullSubscription.
 type PullSubscriptionSpec struct {
+	// This brings in CloudEventOverrides and Sink.
+	duckv1.SourceSpec
```
n3wscott: I think this needs a json inline tag here

config/300-pubsub.yaml
```diff
@@ -65,6 +65,41 @@ spec:
...
+                        apiVersion:
+                          type: string
+                          minLength: 1
+                        kind:
+                          type: string
+                          minLength: 1
+                        name:
+                          type: string
+                          minLength: 1
+                - type: object
```
grantr: Should this one also have the `required` from above?

config/300-pubsub.yaml
```diff
@@ -65,6 +65,41 @@ spec:
...
               type: object
               description: "Reference to an object that will resolve to a domain name to use as the sink."
+              anyOf:
+                - type: object
+                  properties:
+                    uri:
+                      type: string
+                      minLength: 1
+                - type: object
+                  properties:
```
grantr: Should this list also include a uri field?

pkg/apis/pubsub/v1alpha1/pull_subscription_types.go
Harwayne: Done.

config/300-pubsub.yaml
Harwayne: Isn't that the first 'branch' of the anyOf?

config/300-pubsub.yaml
Harwayne: Probably, but this is the deprecated version. Would you like me to do so?

Harwayne:
/hold cancel

Fixed the issue (at least I think so...).

config/300-pubsub.yaml
grantr: I won't complain either way :smiley:

Issue #352, opened 2019-10-22
Channel live locks when PullSubscription already exists


Harwayne:
I created ~320 Triggers on a Broker that is backed by a PubSub Channel. The Channel has created PullSubscriptions for ~5 of the subscribers, and has gotten permanently live locked. 

```yaml
  subscribablestatus:
    subscribers:
    - observedGeneration: 1
      ready: "True"
      uid: 9f79a0ca-f515-11e9-b13c-42010aa80029
    - observedGeneration: 1
      ready: "True"
      uid: 7975668a-f517-11e9-b13c-42010aa80029
    - observedGeneration: 1
      ready: "True"
      uid: 7cdf8679-f517-11e9-b13c-42010aa80029
    - message: PullSubscription cre-sub-90da4450-f517-11e9-b13c-42010aa80029 is not
        ready
      observedGeneration: 1
      ready: "False"
      uid: 90da4450-f517-11e9-b13c-42010aa80029
```
```yaml
k get pullsubscription -oyaml cre-sub-90e327a7-f517-11e9-b13c-42010aa80029
apiVersion: pubsub.cloud.google.com/v1alpha1
kind: PullSubscription
metadata:
  annotations:
    metrics-resource-group: channels.messaging.cloud.google.com
    metrics-resource-name: default-kne-trigger
  creationTimestamp: "2019-10-22T22:01:57Z"
  finalizers:
  - cloud-run-events-pubsub-pullsubscription-controller
  generation: 1
  labels:
    events.cloud.google.com/channel: cloud-run-events-channel-controller
    events.cloud.google.com/channel-name: default-kne-trigger
    events.cloud.google.com/channel-subscriber: cre-sub-90e327a7-f517-11e9-b13c-42010aa80029
    events.cloud.google.com/controller-uid: c8b19f38-f514-11e9-b13c-42010aa80029
  name: cre-sub-90e327a7-f517-11e9-b13c-42010aa80029
  namespace: default
  ownerReferences:
  - apiVersion: messaging.cloud.google.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Channel
    name: default-kne-trigger
    uid: c8b19f38-f514-11e9-b13c-42010aa80029
  resourceVersion: "18345"
  selfLink: /apis/pubsub.cloud.google.com/v1alpha1/namespaces/default/pullsubscriptions/cre-sub-90e327a7-f517-11e9-b13c-42010aa80029
[...]

PR #370, opened 2019-10-25
Handle subscriber already exists error gracefully for pubsub channel

yolocs:
To fix: https://github.com/google/knative-gcp/issues/352


yolocs:
@Harwayne

Issue #375, opened 2019-10-28
The default permission we give SA cloud-run-events doesn't seem allow adding finalizer to secret


yolocs:
For topic and pullsubscription, we want to add finalizer the the secret to prevent it from being deleted before topic/pullsubscription. The change was introduced in: https://github.com/google/knative-gcp/pull/339.
But from the log, it seems the service account doesn't really have permission to add finalizer to the secret. This caused reconciliation failure.

```
jsonPayload: {
  caller: "topic/topic.go:378"   
  events.cloud.google.com/controller: "cloud-run-events-pubsub-topic-controller"   
  knative.dev/key: "pschan-1/cre-tstchan-0-chan"   
  knative.dev/traceid: "1492d27e-a552-4809-bc5e-1a6145dba8d6"   
  level: "error"   
  logger: "controller.cloud-run-events-pubsub-topic-controller"   
  msg: "Failed to update Topic Secret's finalizers%!(EXTRA zapcore.Field={error 25 0  secrets "google-cloud-key" is forbidden: User "system:serviceaccount:cloud-run-events:controller" cannot patch resource "secrets" in API group "" in the namespace "pschan-1"})"   
  stacktrace: "github.com/google/knative-gcp/pkg/reconciler/topic.(*Reconciler).updateSecretFinailizer
	/usr/local/google/home/cshou/golang/src/github.com/google/knative-gcp/pkg/reconciler/topic/topic.go:378
github.com/google/knative-gcp/pkg/reconciler/topic.(*Reconciler).reconcile
	/usr/local/google/home/cshou/golang/src/github.com/google/knative-gcp/pkg/reconciler/topic/topic.go:218
github.com/google/knative-gcp/pkg/reconciler/topic.(*Reconciler).Reconcile
	/usr/local/google/home/cshou/golang/src/github.com/google/knative-gcp/pkg/reconciler/topic/topic.go:105
github.com/google/knative-gcp/vendor/knative.dev/pkg/controller.(*Impl).processNextWorkItem
	/usr/local/google/home/cshou/golang/src/github.com/google/knative-gcp/vendor/knative.dev/pkg/controller/controller.go:342
github.com/google/knative-gcp/vendor/knative.dev/pkg/controller.(*Impl).Run.func2
	/usr/local/google/home/cshou/golang/src/github.com/google/knative-gcp/vendor/knative.dev/pkg/controller/controller.go:291"   
  ts: "2019-10-28T16:32:23.720Z"   
 }
```

PR #376, opened 2019-10-28
Allow cloud-run-events service account to patch secrets

yolocs:
Allow cloud-run-events service account to patch secrets so that it can add finalizer to secrets
To fix: https://github.com/google/knative-gcp/issues/375

### DIFF
--- a/config/201-clusterrole.yaml
+++ b/config/201-clusterrole.yaml
@@ -115,6 +115,15 @@ rules:
     - ""
   resources:
     - secrets # TODO: Does it really need global secrets access?
+  verbs:
+    - get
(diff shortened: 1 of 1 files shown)


config/201-clusterrole.yaml
```diff
@@ -115,6 +115,15 @@ rules:
     - ""
   resources:
     - secrets # TODO: Does it really need global secrets access?
+  verbs:
+    - get
+    - list
+    - watch
+    - patch # For adding finalizer to secrets
```
n3wscott: I think we only need access to `secrets/finalize`

PR #387, opened 2019-10-30
E2E tests of Knative Serving as eventing consumer

grac3gao:
- Broker using PubSub Channel
- Trigger pointing to Knative Service


test/cmd/receiver/main.go
```diff
@@ -0,0 +1,41 @@
+package main
```
nachocano: copyright?

test/cmd/receiver/main.go
```diff
@@ -0,0 +1,41 @@
...
+
+import (
+	"context"
+	cloudevents "github.com/cloudevents/sdk-go"
```
nachocano: importer formatter

test/cmd/sender/main.go
```diff
@@ -0,0 +1,62 @@
+package main
```
nachocano: same here, copyright

test/cmd/sender/main.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"context"
```
nachocano: formatter, and in all files..

test/cmd/receiver/main.go
```diff
@@ -0,0 +1,58 @@
...
+	}
+	if err := r.client.StartReceiver(context.Background(), r.Receive); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func (r *Receiver) Receive(ctx context.Context, event cloudevents.Event, resp *cloudevents.EventResponse) {
+	if event.ID() == "dummy" {
+		resp.Status = http.StatusAccepted
+		event = cloudevents.NewEvent()
```
nachocano: I would call all Set methods on this event, and then at the end set the `resp.Event = &event`

test/cmd/receiver/main.go
```diff
@@ -0,0 +1,58 @@
...
+		resp.Event = &event
+		resp.Event.SetID("target")
+		resp.Event.SetType("e2e-testing-resp")
+		resp.Event.SetSource("e2e-testing")
+		resp.Event.SetDataContentType("application/json")
```
nachocano: `application/json` I think you can use a constant from the CE SDK

test/cmd/receiver/main.go
nachocano: Also set the version to 0.3 maybe? Something like `cloudevents.NewEvent(cloudevents.VersionV03)`

test/cmd/receiver/main.go
```diff
@@ -0,0 +1,58 @@
...
+	r := &Receiver{
+		client: client,
+	}
+	if err := r.client.StartReceiver(context.Background(), r.Receive); err != nil {
+		log.Fatal(err)
+	}
+}
+
+func (r *Receiver) Receive(ctx context.Context, event cloudevents.Event, resp *cloudevents.EventResponse) {
+	if event.ID() == "dummy" {
```
nachocano: add a comment to understand what this function does. Why we are checking for dummy, and things like that... so that someone else that grabs it later on can easily understand.

test/cmd/sender/main.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+func main() {
+	brokerURL := os.Getenv(brokerURLEnvVar)
+
+	ceClient, err := kncloudevents.NewDefaultClient(brokerURL)
+	if err != nil {
+		fmt.Printf("Unable to create ceClient: %s ", err)
+	}
+	event := dummyCloudEvent()
```
nachocano: I would just inline the creation here instead of in a function, unless you are using it somewhere else.
Inline will be clearer IMO

test/e2e-common.sh
```diff
@@ -55,6 +55,7 @@ function knative_setup() {
 
 # Setup resources common to all eventing tests.
 function test_setup() {
+  kubectl apply -f $(dirname $0)/e2e/config/istio-patch/istio-knative-extras.yaml
```
nachocano: maybe create a function istio_setup, and add this there?

nachocano: @chizhg do you know if we are doing something like this in eventing?

test/e2e/e2e_test.go
```diff
@@ -155,6 +157,13 @@ func TestPubSubStackDriverMetrics(t *testing.T) {
 	PubSubWithTargetTestImpl(t, packageToImageConfig, true /*assertMetrics */)
 }
 
+// TestPubSubWithBroker tests we can knock a knative service from a broker with PubSub channel.
```
nachocano: // TestPubSubWithBroker tests we can knock a Knative Service from a Broker with PubSub Channel.

nachocano: actually maybe we can call it `TestBrokerWithPubSubChannel`... `TestPubSubWithBroker` might get confusing as it might mean a PubSub source to a Broker.

test/e2e/test_broker_pubsub.go
```diff
@@ -0,0 +1,185 @@
...
+              1     v	    2         | v        3           |
+(Sender) ---> Broker(PubSub) ---> dummyTrigger -------> Knative Service(Receiver)
+                    |
+                    |    6                   7
+                    |-------> respTrigger -------> Service(Target)
+
+Note: the number denotes the sequence of the event that flows in this test case.
+*/
+
+func PubSubWithBrokerTestImpl(t *testing.T, packages map[string]string) {
```
nachocano: nit: same here, small rename

test/e2e/test_broker_pubsub.go
```diff
@@ -0,0 +1,185 @@
...
+	if err := client.WaitForResourceReady(client.Namespace, dummyTriggerName, triggerGVR); err != nil {
+		t.Error(err)
+	}
+	if err := client.WaitForResourceReady(client.Namespace, respTriggerName, triggerGVR); err != nil {
+		t.Error(err)
+	}
+
+	ksvcGVR := schema.GroupVersionResource{
+		Group:    "serving.knative.dev",
+		Version:  "v1alpha1",
```
nachocano: can we use `v1`

nachocano: might need to change the yamls as well

test/e2e/test_broker_pubsub.go
```diff
@@ -0,0 +1,185 @@
...
+
+	// Get broker URL
+	metaAddressable := resources.NewMetaResource(brokerName, client.Namespace, common.BrokerTypeMeta)
+	u, err := base.GetAddressableURI(client.Dynamic, metaAddressable)
+	if err != nil {
+		t.Error(err.Error())
+	}
+	config["brokerURL"] = u.String()
+
+	time.Sleep(5 * time.Second)
```
nachocano: add a comment why the sleep

test/e2e/test_broker_pubsub.go
```diff
@@ -0,0 +1,185 @@
...
+		"senderName":       senderName,
+		"targetName":       targetName,
+		"clusterRoleName":  clusterRoleName,
+	}
+	for k, v := range packages {
+		config[k] = v
+	}
+
+	// Create resources
+	brokerInstaller := createresource(client, config, []string{"pubsub_broker", "istio"}, t)
```
nachocano: can we camelCase this method... `createResource`

test/e2e/test_broker_pubsub.go
```diff
@@ -0,0 +1,185 @@
...
+	ksvcGVR := schema.GroupVersionResource{
+		Group:    "serving.knative.dev",
+		Version:  "v1alpha1",
+		Resource: "services",
+	}
+	if err := client.WaitForResourceReady(client.Namespace, kserviceName, ksvcGVR); err != nil {
+		t.Error(err)
+	}
+
+	// Get broker URL
```
nachocano: nit: add . at the end of every comment

test/e2e/test_broker_pubsub.go
```diff
@@ -0,0 +1,185 @@
...
+	defer deleteResource(senderInstaller, t)
+
+	jobGVR := schema.GroupVersionResource{
+		Group:    "batch",
+		Version:  "v1",
+		Resource: "jobs",
+	}
+	// Check if dummy cloudevent is sent out
+	if done := jobDone(client, senderName, t, jobGVR); !done {
+		t.Error("dummy event didn't send to broker")
```
nachocano: nit: dummy event wasn't sent to the broker

test/e2e/test_broker_pubsub.go
```diff
@@ -0,0 +1,185 @@
...
+	}
+	return installer
+}
+
+func deleteResource(installer *Installer, t *testing.T) {
+	if err := installer.Do("delete"); err != nil {
+		t.Errorf("failed to delete, %s", err)
+	}
+	// Just chill for tick.
+	time.Sleep(15 * time.Second)
```
nachocano: hmm, can we add a better comment. Or maybe this sleep should be made in the caller?

test/e2e/test_broker_pubsub.go
```diff
@@ -0,0 +1,185 @@
...
+}
+
+func jobDone(client *Client, podName string, t *testing.T, jobGVR schema.GroupVersionResource) bool {
```
nachocano: If these are utility methods for all tests, I think we should try to move them to another go file. And use them from there everywhere. The fact that all are in the same package, gives them visibility, but I'd rather not include pubsub stuff in a storage test.
Does it make sense?

test/e2e-common.sh
chizhg: No we do not have things like this in eventing.
I talked with Grace and I'm not sure why knative-gcp will need this, it's a bit hacky IMO, e.g. how do you guarantee users will have this patch?

grac3gao:
done

grac3gao:
ready for review

nachocano:
/lgtm
/approve

thanks for doing this @grac3gao !!!

PR #389, opened 2019-10-31
[WIP] Update webhook to latest style

grantr:
Mostly copied from eventing's webhook.

Because the webhook config is now expressed as yaml, we can't insert the
system namespace into the webhook config name as easily as before. That
feature of the webhook has been removed. This means on upgrade there
will be an obsolete webhook configuration. We'll need to ask users to
delete that or delete it in code.

/cc @n3wscott


grantr:
This is necessary to upgrade to the latest pkg. There might be a lighter touch fix than refactoring everything. WIP on that.

grantr:
Also this won't be the latest style for much longer... :smile: https://github.com/knative/eventing/pull/2127

PR #396, opened 2019-11-04
Adding Span From TraceParent only for Channels

nachocano:
- Adding the span from trace parent only if it's a channel...
- For sources this is not needed, and we get plenty of info errors because of that.
- Also moved the log to Debug, but I'm OK to move it back to Info if @Harwayne prefers.

### DIFF
--- a/pkg/pubsub/adapter/adapter.go
+++ b/pkg/pubsub/adapter/adapter.go
@@ -27,7 +27,7 @@ import (
 
 	nethttp "net/http"
 
-	"github.com/cloudevents/sdk-go"
+	cloudevents "github.com/cloudevents/sdk-go"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport"
 	"github.com/cloudevents/sdk-go/pkg/cloudevents/transport/http"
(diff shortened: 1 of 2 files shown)


pkg/pubsub/adapter/adapter.go
```diff
@@ -167,10 +176,15 @@ func (a *Adapter) receive(ctx context.Context, event cloudevents.Event, resp *cl
...
-	if err != nil {
-		logger.Infow("Unable to attach tracing to context", zap.Error(err))
+	var err error
+	// If the resource group is the Channel one, then we attach the span from the traceparent attribute.
+	// Otherwise, it means it's a Source, thus it does not have traceparent.
+	if a.ResourceGroup == channelGVR.GroupResource().String() {
+		// a.Sink is likely not exactly what we want...
+		ctx, err = tracing.AddSpanFromTraceparentAttribute(ctx, a.Sink, event)
+		if err != nil {
+			logger.Debugw("Unable to attach tracing to context", zap.Error(err))
```
Harwayne: Please leave it as a warning, I want to be aware of it happening.

nachocano: warning it is then! thanks!

PR #403, opened 2019-11-09
Setting "application/octet-stream" as ContentType for PubSub

nachocano:
- As we don't know the content-type, we set it to be this generic one.


grantr:
We should create an issue in cloudevents/sdk-go to add this content-type as a constant, along with text/plain.

nachocano:
> We should create an issue in cloudevents/sdk-go to add this content-type as a constant, along with text/plain.

Let me do that and I'll link it here. Thanks!

nachocano:
Issue: https://github.com/cloudevents/sdk-go/issues/254

PR #410, opened 2019-11-13
Fix performance tests setup

chizhg:
Fix a couple of issues in the performance tests setup process:
1. Install the Istio addon on the cluster
2. Install Serving and patch Istio
3. Clean up the error pod after benchmark setup

Also change the cluster size since 2 is enough based on the observation.

/cc @nachocano 
/cc @grantr

### DIFF
--- a/test/performance/benchmarks/broker-pubsub/cluster.yaml
+++ b/test/performance/benchmarks/broker-pubsub/cluster.yaml
@@ -16,6 +16,6 @@
 
 GKECluster:
   location: "us-west1"
-  nodeCount: 3
+  nodeCount: 2
   nodeType: "n1-standard-4"
-  addons: "HorizontalPodAutoscaling,HttpLoadBalancing"
+  addons: "HorizontalPodAutoscaling,HttpLoadBalancing,Istio"
--- a/test/performance/benchmarks/channel-pubsub/cluster.yaml
+++ b/test/performance/benchmarks/channel-pubsub/cluster.yaml
@@ -16,6 +16,6 @@
 
 GKECluster:
   location: "us-west1"
-  nodeCount: 3
+  nodeCount: 2
   nodeType: "n1-standard-4"
-  addons: "HorizontalPodAutoscaling,HttpLoadBalancing"
+  addons: "HorizontalPodAutoscaling,HttpLoadBalancing,Istio"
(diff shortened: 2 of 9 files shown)


test/performance/performance-tests.sh
```diff
@@ -57,7 +45,29 @@ function update_benchmark() {
   git clone https://github.com/knative/eventing
   popd
   ko delete -f ${BENCHMARK_ROOT_PATH}/$1/${TEST_CONFIG_VARIANT} --ignore-not-found=true
+  sleep 60
   ko apply -f ${BENCHMARK_ROOT_PATH}/$1/${TEST_CONFIG_VARIANT} || abort "failed to apply benchmark $1"
+
+  echo "Sleeping 2 min to wait for all resources to setup"
+  sleep 120
+  # In the current implmentation, for some reason there can be error pods after the setup, but it does not necessarily
```
nachocano: nit: implementation

test/performance/benchmarks/broker-pubsub/cluster.yaml
```diff
@@ -16,6 +16,6 @@
 
 GKECluster:
   location: "us-west1"
-  nodeCount: 3
```
nachocano: Can we keep it the same as eventing? I think we use 3 there? 
Just to be able to do more of an apple to apple comparison, and have an idea of how we compare to the IMC...

test/performance/benchmarks/channel-pubsub/cluster.yaml
```diff
@@ -16,6 +16,6 @@
 
 GKECluster:
   location: "us-west1"
-  nodeCount: 3
+  nodeCount: 2
```
nachocano: same here

nachocano:
Thanks for doing this @chizhg !!!

test/performance/benchmarks/broker-pubsub/cluster.yaml
chizhg: I just changed the cluster size to 1 in eventing, see https://github.com/knative/eventing/pull/2174, since I found the benchmarks had very low resource utilization. The node count here is also derived from my observation when the benchmark were run.

I'm not sure if cluster size is one factor that we need to consider when we compare it with IMC, since obviously pubsub channel will need more resources (it requires Serving, Istio, and some other pubsub related resources). IMO as long as the cluster is big enough for running the benchmarks and getting metrics, it would be fine.

test/performance/performance-tests.sh
chizhg: Done.

test/performance/benchmarks/channel-pubsub/cluster.yaml
chizhg: Ditto.

test/performance/benchmarks/broker-pubsub/cluster.yaml
nachocano: OK, I can live with this! Thanks Chi

PR #416, opened 2019-11-20
Using new pkg webhook and v1.Destination

### LINKED ISSUE #414: Update to new pkg webhook code
Added in https://github.com/knative/pkg/pull/848. Need a PR similar to https://github.com/knative/eventing/pull/2147.

nachocano:
Fixes #414

**Proposed Changes**

- Updating pkg, serving, and eventing.
- Adopting v1.Destination, removing deprecation messages.
- Changing webhook stuff
- Few other cosmetics.

**Release Note**

```release-note
* Moved to v1.Destination. If any of your yamls is using the v1alpha1.Destination with deprecated fields, you should update them to match the new API.
* Remove the old webhook called `webhook.cloud-run-events.events.cloud.google.com`
```


nachocano:
/assing @yolocs 

trying one more time after being added to collaborators

nachocano:
this is ready for a look. Tested in a clean cluster and things seem to be working as expected...
@grantr 
@yolocs

nachocano:
updated documentation here: https://github.com/knative/docs/pull/1993

Gopkg.lock
```diff
@@ -178,6 +178,14 @@
   revision = "0ca988a254f991240804bf9821f3450d87ccbb1b"
   version = "v1.3.0"
 
+[[projects]]
+  branch = "master"
+  digest = "1:e2b86e41f3d669fc36b50d31d32d22c8ac656c75aa5ea89717ce7177e134ff2a"
+  name = "github.com/golang/glog"
```
grantr: I wonder why this dependency was added?

cmd/webhook/main.go
```diff
@@ -18,135 +18,122 @@ package main
...
-	if err != nil {
-		log.Fatal("Error loading logging configuration:", err)
-	}
-	config, err := logging.NewConfigFromMap(cm)
-	if err != nil {
-		log.Fatal("Error parsing logging configuration:", err)
+func NewDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+	// Decorate contexts with the current state of the config.
+	// store := defaultconfig.NewStore(logging.FromContext(ctx).Named("config-store"))
+	// store.WatchConfigs(cmw)
```
grantr: Why are these lines commented? If they're unused but here for documentation, they should have a comment explaining why they exist and when they should be uncommented.

cmd/webhook/main.go
```diff
@@ -18,135 +18,122 @@ package main
...
+	ctxFunc := func(ctx context.Context) context.Context {
+		// return v1.WithUpgradeViaDefaulting(store.ToContext(ctx))
```
grantr: Same here as the above commented lines.

cmd/webhook/main.go
```diff
@@ -18,135 +18,122 @@ package main
...
+		// Name of the configmap webhook.
+		"config.webhook.events.cloud.google.com",
+
+		// The path on which to serve the webhook.
+		"/config-validation",
+
+		// The configmaps to validate.
+		configmap.Constructors{
+			tracingconfig.ConfigName: tracingconfig.NewTracingConfigFromConfigMap,
+			// metrics.ConfigMapName():   metricsconfig.NewObservabilityConfigFromConfigMap,
```
grantr: Why is this commented?

cmd/webhook/main.go
```diff
@@ -18,135 +18,122 @@ package main
...
 
-func flush(logger *zap.Logger) {
-	_ = logger.Sync()
-	metrics.FlushExporter()
+func main() {
+	// Set up a signal context with our webhook options
+	ctx := webhook.WithOptions(signals.NewContext(), webhook.Options{
+		ServiceName: logconfig.WebhookName(),
+		Port:        8443,
+		SecretName:  "webhook-certs",
```
grantr: This needs a comment explaining that this value must match the name of the Secret created in config.

config/201-webhook-clusterrole.yaml
```diff
@@ -0,0 +1,111 @@
...
+    verbs:
+      - "get"
+      - "list"
+      - "create"
+      - "update"
+      - "delete"
+      - "patch"
+      - "watch"
+
+  # Our own resources and statuses we care about.
```
grantr: Why are these permissions needed to process webhook requests?

config/500-webhook.yaml
```diff
@@ -32,34 +32,21 @@ spec:
...
         - name: webhook
+          terminationMessagePolicy: FallbackToLogsOnError
           # This is the Go import path for the binary that is containerized
           # and substituted here.
           image: github.com/google/knative-gcp/cmd/webhook
-          ports:
-          - name: metrics
-            containerPort: 9090
-          imagePullPolicy: Always
-          resources:
```
grantr: Seems like we should set requests and limits on the webhook pods so they don't get default values.

pkg/apis/events/v1alpha1/pubsub_validation.go
```diff
@@ -77,11 +77,7 @@ func (current *PubSubSpec) Validate(ctx context.Context) *apis.FieldError {
 	return errs
 }
 
-func (current *PubSub) CheckImmutableFields(ctx context.Context, og apis.Immutable) *apis.FieldError {
-	original, ok := og.(*PubSub)
-	if !ok {
-		return &apis.FieldError{Message: "The provided original was not a PubSub"}
-	}
+func (current *PubSub) CheckImmutableFields(ctx context.Context, original *PubSub) *apis.FieldError {
```
grantr: Why did this signature change? Seems like it removed compatibility with the `apis.Immutable` interface (see https://github.com/google/knative-gcp/pull/416/files#diff-130541c895e1b10fab4757caecea7ec8L49).

pkg/apis/pubsub/v1alpha1/pull_subscription_validation.go
```diff
@@ -91,11 +91,7 @@ func (current *PullSubscriptionSpec) Validate(ctx context.Context) *apis.FieldEr
 	return errs
 }
 
-func (current *PullSubscription) CheckImmutableFields(ctx context.Context, og apis.Immutable) *apis.FieldError {
-	original, ok := og.(*PullSubscription)
-	if !ok {
-		return &apis.FieldError{Message: "The provided original was not a PullSubscription"}
-	}
+func (current *PullSubscription) CheckImmutableFields(ctx context.Context, original *PullSubscription) *apis.FieldError {
```
grantr: Same question as above.

pkg/apis/pubsub/v1alpha1/topic_validation.go
```diff
@@ -49,11 +49,7 @@ func (ts *TopicSpec) Validate(ctx context.Context) *apis.FieldError {
 	return errs
 }
 
-func (current *Topic) CheckImmutableFields(ctx context.Context, og apis.Immutable) *apis.FieldError {
-	original, ok := og.(*Topic)
-	if !ok {
-		return &apis.FieldError{Message: "The provided original was not a Topic"}
-	}
+func (current *Topic) CheckImmutableFields(ctx context.Context, original *Topic) *apis.FieldError {
```
grantr: Same question as above.

pkg/reconciler/decorator/decorator.go
```diff
@@ -219,8 +220,8 @@ func (r *Reconciler) createOrUpdateDecorator(ctx context.Context, decorator *v1a
 	return nil
 }
 
-func (c *Reconciler) resolveDestination(ctx context.Context, destination pkgv1alpha1.Destination, decorator *v1alpha1.Decorator) (string, error) {
-	dest := pkgv1alpha1.Destination{
+func (c *Reconciler) resolveDestination(ctx context.Context, destination duckv1.Destination, decorator *v1alpha1.Decorator) (string, error) {
+	dest := duckv1beta1.Destination{
```
grantr: Why does this use duckv1beta1.Destination instead of duckv1? Is it because uriResolver still expects duckv1beta1? Seems like that should be fixed in pkg.

Gopkg.lock
nachocano: it seems to be from pkg, [e2e_flags.go](https://github.com/knative/pkg/commit/d9e3e244a6c044379b71d90b7440daad5b7910f3)

cmd/webhook/main.go
nachocano: I'm on your same page here. Don't know much about this webhook change... Copy/paste from [eventing](https://github.com/knative/eventing/blob/8e38791d3e4cba16df132c0b42ba68f62a746eac/cmd/webhook/main.go#L70). Will remove them.

cmd/webhook/main.go
nachocano: same as above, removed!

cmd/webhook/main.go
nachocano: Done!

config/500-webhook.yaml
nachocano: Copied it from eventing. I think these were just numbers copied from serving, due to their e2e tests. See [here](https://github.com/knative/serving/blob/e91fd663c77348557dcfa8720f2a78c167f6d5b9/config/webhook.yaml#L50). 
I'll add them back.

pkg/apis/events/v1alpha1/pubsub_validation.go
nachocano: `apis.Immutable` is not a thing anymore. It was removed as part of the webhook PR. See [here](https://github.com/knative/pkg/commit/4836f680bbe5dc615e9c822f21c6685d97c6d553).

config/201-webhook-clusterrole.yaml
nachocano: copied from eventing :) 
we were previously using the controller cluster role for the webhook as well, so it had much more things...  I splitted into a separate cluster role. Let me remove these. Thanks!

pkg/apis/pubsub/v1alpha1/pull_subscription_validation.go
nachocano: yes, same thing. They removed Immutable.

pkg/reconciler/decorator/decorator.go
nachocano: Yes, that's right.
There should be some follow up change in pkg to receive the v1 duck.

pkg/apis/pubsub/v1alpha1/topic_validation.go
nachocano: I ack that is a weird change... apparently it was already done in eventing as well.

nachocano:
> Sorry @nachocano, I have lots of questions, probably because I'm not very familiar with the new webhook code. Thanks for tackling all this!

Thanks for the comments @grantr . I'm not very familiar with the new code either, it was mainly a copy/paste things from eventing. 

It should be ready for another look.

PR #437, opened 2019-12-04
CloudEvents v1 events

### LINKED ISSUE #436: Send CE v1.0
**Problem**
All GCP sources should send CloudEvents v1.0. They currently send 0.3

Developer

**Exit Criteria**
E2E tests sending CE v1.0

**Time Estimate (optional):** 
1d

nachocano:
Fixes #436

## Proposed Changes

- Sending CE v1 events.
- Updating serving, eventing and pkg
- Only promoting pubsub message attributes to extensions if they are valid v1 extensions (i.e., lower case alphanumeric values). This is rather an arbitrary decision, and we may need to revisit.

**Release Note**

```release-note
knative-gcp now sends CloudEvents v1.0 events. Note that CE v1 extensions should be lower case alphanumeric values. We only promote the pubsub message attributes to CE extensions if they are valid lower case alphanumeric values. This is a breaking change if you were relying on filtering on such attributes. For example, in the case of storage, we are not setting: notificationConfig, objectGeneration, nor payloadFormat.
```


pkg/pubsub/adapter/converters/pubsub.go
```diff
@@ -48,7 +49,8 @@ func convertPubsub(ctx context.Context, msg *cepubsub.Message, sendMode ModeType
 	// Attributes are extensions.
 	if msg.Attributes != nil && len(msg.Attributes) > 0 {
 		for k, v := range msg.Attributes {
-			event.SetExtension(k, v)
+			// V1 attributes MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9').
+			event.SetExtension(strings.ToLower(k), v)
```
grantr: Is there a unit test for this lowercasing?

pkg/pubsub/adapter/converters/storage.go
```diff
@@ -93,7 +94,8 @@ func convertStorage(ctx context.Context, msg *cepubsub.Message, sendMode ModeTyp
 	// Attributes are extensions.
 	if msg.Attributes != nil && len(msg.Attributes) > 0 {
 		for k, v := range msg.Attributes {
-			event.SetExtension(k, v)
+			// V1 attributes MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9').
+			event.SetExtension(strings.ToLower(k), v)
```
grantr: Is there a unit test for this lowercasing?

pkg/pubsub/adapter/converters/pubsub.go
nachocano: There are no UT for converters, which is even worse...
If we don't lowercase them, then the CE library panics... 
Without this E2E tests don't pass.
If you don't mind, I can create issues to add UTs for converters, and as part of the big refactor we are making, I'll add those...

pkg/pubsub/adapter/converters/storage.go
nachocano: same as above... If you want them now, I can go for it... Otherwise in the several follow ups we need to do in this repo

pkg/pubsub/adapter/converters/pubsub.go
nachocano: Let me take that back... In order to fix the mistakes from the past... let me actually add them now.. thanks for the comment!

grantr: Thanks!

pkg/pubsub/adapter/converters/pubsub.go
```diff
@@ -50,7 +50,10 @@ func convertPubsub(ctx context.Context, msg *cepubsub.Message, sendMode ModeType
 	if msg.Attributes != nil && len(msg.Attributes) > 0 {
 		for k, v := range msg.Attributes {
 			// V1 attributes MUST consist of lower-case letters ('a' to 'z') or digits ('0' to '9').
-			event.SetExtension(strings.ToLower(k), v)
+			// Only setting them if the attributes are valid alphanumeric, otherwise the sdk will panic.
+			if IsAlphaNumericLowercaseLetters(k) {
```
grantr: Is this an issue with the cloudevents sdk or the pubsub sdk?  Perhaps an error should be returned rather than silently dropping event metadata.

pkg/pubsub/adapter/converters/pubsub.go
nachocano: Added UTs, and changed the logic a bit.

pkg/pubsub/adapter/converters/storage.go
nachocano: done!

pkg/pubsub/adapter/converters/pubsub.go
nachocano: I think is not an issue in neither. In v1, extensions should be lower-case letters or digits. 
It was rather our opinionated implementation to promote all pubsub attributes to extensions. If someone upstream sends an attribute that cannot be an v1 extension, do you think we should fail? I was thinking more of not to, and add as much metadata as possible, if it's valid.

grantr: @nachocano and I talked about this. While this seems like it should be transport-specific, it's actually [a MUST in the base CloudEvents v1 specification](https://github.com/cloudevents/spec/blob/master/spec.md#attribute-naming-convention), so it's not even possible for a conformant transport to allow non-base36 characters even if the underlying technology allows it. This seems like an issue in the specification.

@nachocano please clarify in the comment that CloudEvents v1.0 requires this behavior, then I'll lgtm.

nachocano: Done, thanks @grantr !!!

nachocano:
> /lgtm
> 
> Thanks @nachocano! Please mention this (breaking) change in the release notes section.

Done! Thanks!

Issue #327, opened 2019-10-07
Topic can't be deleted in a Terminating Namespace


Harwayne:
Related to #281.

Deleting a namespace puts that namespace into TERMINATING, tries to delete everything inside it, and blocks any new resources from being created in that namespace. According to https://github.com/eBay/Kubernetes/blob/master/docs/design/namespaces.md#phases (this is years old, but we couldn't find anything more recent), the NS is immediately put into the Terminating phase when the deletion timestamp exists. And "Admission control blocks creation of new resources in that namespace". 

The current design of Topic requires that when a Topic is deleted, a Job is created (to delete the GCP Cloud PubSub Topic). So when I delete a namespace with a Topic in it, the Topic is marked for deletion, but can never make progress.

```shell
Name:         cre-reply-ch-chan
Namespace:    includes-incoming-trace-id-mhlk5
Labels:       events.cloud.run/channel=cloud-run-events-channel-controller
              events.cloud.run/channel-name=reply-ch
              events.cloud.run/controller-uid=17633f18-e955-11e9-b584-42010a800089
Annotations:  <none>
API Version:  pubsub.cloud.run/v1alpha1
Kind:         Topic
Metadata:
  Creation Timestamp:             2019-10-07T22:52:08Z
  Deletion Grace Period Seconds:  0
  Deletion Timestamp:             2019-10-07T22:56:54Z
  Finalizers:
    cloud-run-events-pubsub-topic-controller
  Generation:  2
  Owner References:
    API Version:           messaging.cloud.run/v1alpha1
    Block Owner Deletion:  true
    Controller:            true
    Kind:                  Channel
    Name:                  reply-ch
    UID:                   17633f18-e955-11e9-b584-42010a800089
  Resource Version:        28963734
  Self Link:               /apis/pubsub.cloud.run/v1alpha1/namespaces/includes-incoming-trace-id-mhlk5/topics/cre-reply-ch-chan                                                                           
  UID:                     1766bce7-e955-11e9-91f5-42010a8000ac
Spec:
  Propagation Policy:  CreateDelete
  Secret:
    Key:   key.json
    Name:  google-cloud-key
  Topic:   cre-chan-17633f18-e955-11e9-b584-42010a800089
Status:
  Address:
    Hostname:  cre-reply-ch-chan-publish.includes-incoming-trace-id-mhlk5.svc.cluster.local
[...]

Issue #442, opened 2019-12-10
Add retries for Setting Status


nachocano:
**Problem**
As per PR in the serving stack, there should be retries / fallbacks on the code paths that set the Status. See knative/serving#6162

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?
Developer

**Exit Criteria**
Once we wrap the Status of our reconcilers

**Time Estimate (optional):** 
2d

nlopezgi:
Created https://github.com/google/knative-gcp/pull/508. Have left one question in the PR about code reuse from knative/serving#6162, please advise.

nachocano:
taking a look

PR #444, opened 2019-12-11
Removing Job-based reconcilers

### LINKED ISSUE #112: Make sure that sources track their sinks
Consider turning this into an issue. Something about status of sinks might not get reflected, there might be issues with this on error conditions.

Consider turning this into an issue. Something about status of sinks might not get reflected, there might be issues with this on error conditions.

_Originally posted by @vaikas-google in https://github.com/_render_node/MDE3OlB1bGxSZXF1ZXN0UmV2aWV3MjY4MDUyNjQy/pull_request_reviews/more_threads_

### LINKED ISSUE #325: Storage CRD does not have ceOverrides property
The [Storage event spec](https://github.com/google/knative-gcp/blob/master/pkg/apis/events/v1alpha1/storage_types.go#L53) has duckv1alpha1.PubSubSpec which brings in multiple fields including CloudEventOverrides. However, this field is not included in the [storage source CRD](https://github.com/google/knative-gcp/blob/master/config/300-storage.yaml).

### LINKED ISSUE #354: Topic creation+deletion race condition
This is theorized, I have not seen it in practice.

The Topic reconciler creates a Job to create and delete Topics in GCP. I believe that we can cause a race condition which leaks a GCP Topic. 

1. Create a Topic.
    - This creates a Job to create the Topic in GCP. 
        - This takes longer than expected. E.g. the Pod is on a node which does not have the image cached and needs to fetch it.
2. Delete the Topic.
    - This creates a Job to delete the Topic in GCP. But it doesn't do anything with the Job to create the Topic, which relies on K8s GC for deletion.
        - This Job's Pod runs before the create Job's Pod. E.g. the Pod is on a node which has the image cached.
          - A short time later, the Create Job's Pod runs, creating the GCP Topic.
3. The Channel reconciler notices the delete job has finished and removes the finalizer from the Channel.

The GCP Topic has now been leaked.

### LINKED ISSUE #358: Bulk PullSubscription deletion can cause live lock style issues
This is most likely a Kubernetes issue, not one specific to this repo. But it may affect whether we use a Job based design or not.

**Setup**
I created a Broker backed by a PubSub Channel. I then created ~320 Triggers on that Broker. Sometime later, I deleted all the Triggers (`kubectl delete triggers --all`). 

**Observation**
The PullSubscriptions, corresponding to the Triggers, never disappeared. They all had deletion timestamps, but even after 20 minutes, not a single one had actually been removed.

**Theory**
The number of Pods being created vastly overwhelmed the cluster, to the point that no new Pods were getting scheduled. (Note that I did not test adding a new Pod to see what would happen, I should have). 

**Work around**
For my case, I increased the number of nodes from 3 to 6, which allowed the PullSubscriptions to slowly clean themselves up and get removed. 

It is also possible that GKE had scaled the masters up/out and the change in number of Nodes had no effect.

### LINKED ISSUE #359: Channel permanently failed without adequate error message
I found this due to #353.

**Setup**
I created a Broker backed by the PubSub Channel, but the `google-cloud-key` had an invalid credential (for a Service Account that no longer existed). After figuring out the error, I updated the `google-cloud-key` to a correct value.

**Expectation**
The Topic (and thus the Channel and Broker) would heal itself. This might take a while due to exponential backoff in retries, but should occur in a reasonable time frame. 

Or failing the self healing, it would at least expose a status indicating it would never try again.

**Observed**
The Topic never makes any forward progress. Its status is not very helpful:
```yaml
apiVersion: pubsub.cloud.google.com/v1alpha1
kind: Topic
metadata:
  creationTimestamp: "2019-10-22T22:57:28Z"
  finalizers:
  - cloud-run-events-pubsub-topic-controller
  generation: 1
  labels:
    events.cloud.google.com/channel: cloud-run-events-channel-controller
    events.cloud.google.com/channel-name: default-kne-trigger
    events.cloud.google.com/controller-uid: 5205790d-f51f-11e9-b13c-42010aa80029
  name: cre-default-kne-trigger-chan
  namespace: new
  ownerReferences:
  - apiVersion: messaging.cloud.google.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Channel
    name: default-kne-trigger
    uid: 5205790d-f51f-11e9-b13c-42010aa80029
  resourceVersion: "41126"
  selfLink: /apis/pubsub.cloud.google.com/v1alpha1/namespaces/new/topics/cre-default-kne-trigger-chan
  uid: 5207f285-f51f-11e9-b13c-42010aa80029
spec:
  propagationPolicy: CreateDelete
  secret:
    key: key.json
    name: google-cloud-key
  topic: cre-chan-5205790d-f51f-11e9-b13c-42010aa80029
status:
  conditions:
  - lastTransitionTime: "2019-10-22T22:57:28Z"
    status: Unknown
    type: Addressable
  - lastTransitionTime: "2019-10-22T22:59:29Z"
    message: Configuration "cre-default-kne-trigger-chan-publish" does not have any
      ready Revision.
    reason: RevisionMissing
    status: "False"
    type: PublisherReady
  - lastTransitionTime: "2019-10-22T22:59:29Z"
    message: Configuration "cre-default-kne-trigger-chan-publish" does not have any
      ready Revision.
    reason: RevisionMissing
    status: "False"
    type: Ready
  - lastTransitionTime: "2019-10-22T22:57:28Z"
    message: Created Job to create topic "cre-chan-5205790d-f51f-11e9-b13c-42010aa80029".
    reason: Creating
    status: Unknown
    type: TopicExists
  observedGeneration: 1
  topicId: cre-chan-5205790d-f51f-11e9-b13c-42010aa80029

```

The 'create' Job had a FAILED status.

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  creationTimestamp: "2019-10-22T22:57:28Z"
  labels:
    events.cloud.google.com/topic: cre-default-kne-trigger-chan-Topic-createops
  name: pubsub-t-cre-default-kne-trigger-chan-topic-create
  namespace: new
  ownerReferences:
  - apiVersion: pubsub.cloud.google.com/v1alpha1
    blockOwnerDeletion: true
    controller: true
    kind: Topic
    name: cre-default-kne-trigger-chan
    uid: 5207f285-f51f-11e9-b13c-42010aa80029
  resourceVersion: "44861"
  selfLink: /apis/batch/v1/namespaces/new/jobs/pubsub-t-cre-default-kne-trigger-chan-topic-create
  uid: 52090e8e-f51f-11e9-b13c-42010aa80029
spec:
  backoffLimit: 3
  parallelism: 1
  selector:
    matchLabels:
      controller-uid: 52090e8e-f51f-11e9-b13c-42010aa80029
  template:
    metadata:
      annotations:
        sidecar.istio.io/inject: "false"
      creationTimestamp: null
      labels:
        action: create
        controller-uid: 52090e8e-f51f-11e9-b13c-42010aa80029
        job-name: pubsub-t-cre-default-kne-trigger-chan-topic-create
        resource-uid: ""
    spec:
      containers:
      - env:
        - name: GOOGLE_APPLICATION_CREDENTIALS
          value: /var/secrets/google/key.json
        - name: ACTION
          value: create
        - name: PROJECT_ID
        - name: PUBSUB_TOPIC_ID
          value: cre-chan-5205790d-f51f-11e9-b13c-42010aa80029
        image: gcr.io/harwayne2/topic-d3cc3ff42f0f973f29604fd499499c57@sha256:45536a88668b0469e7c07d9ca692cd99aaa9999c50f49b960925ad091715ba7c
        imagePullPolicy: Always
        name: job
        resources: {}
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /var/secrets/google
          name: google-cloud-key
      dnsPolicy: ClusterFirst
      restartPolicy: Never
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
      volumes:
      - name: google-cloud-key
        secret:
          defaultMode: 420
          secretName: google-cloud-key
status:
  conditions:
  - lastProbeTime: "2019-10-22T23:15:21Z"
    lastTransitionTime: "2019-10-22T23:15:21Z"
    message: Job has reached the specified backoff limit
    reason: BackoffLimitExceeded
    status: "True"
    type: Failed
  failed: 4
  startTime: "2019-10-22T22:57:28Z"
```

**Possible Fixes**
- When the Job is marked as FAILED, then delete the Job and create another. Hopefully with a relevant error message in the Topic's status.
- Have the Topic watch anything used by the Job, in this case just the Secret, and recreate the Job whenever those change.
- Have the Job retry an unlimited number of times. 
    - This may cause too many Pods to be created, leading to problems like #358.

### LINKED ISSUE #361: Storage - Bad keys don't self-heal
I created a `google-cloud-key` with a valid key. I used it with PubSub Channels and everything worked. Then I created a Storage object, *without reading the documentation*. The Storage has a failed status message:

```yaml
apiVersion: events.cloud.google.com/v1alpha1
kind: Storage
metadata:
  annotations:
    kubectl.kubernetes.io/last-applied-configuration: |
      {"apiVersion":"events.cloud.google.com/v1alpha1","kind":"Storage","metadata":{"annotations":{},"name":"storage","namespace":"new"},"spec":{"bucket":"storage-gcp","eventTypes":["com.google.cloud.storage.object.finalize"],"payloadFormat":"JSON_API_V1","sink":{"uri":"http://www.example.com/"}}}
  creationTimestamp: "2019-10-24T17:47:07Z"
  finalizers:
  - cloud-run-events-storage-source-controller
  generation: 1
  name: storage
  namespace: new
  resourceVersion: "682227"
  selfLink: /apis/events.cloud.google.com/v1alpha1/namespaces/new/storages/storage
  uid: 4bf4cae9-f686-11e9-9791-42010aa8000e
spec:
  bucket: storage-gcp
  eventTypes:
  - com.google.cloud.storage.object.finalize
  payloadFormat: JSON_API_V1
  secret:
    key: key.json
    name: google-cloud-key
  sink:
    uri: http://www.example.com/
status:
  conditions:
  - lastTransitionTime: "2019-10-24T17:47:26Z"
    message: 'Failed to create Storage notification: operation failed: googleapi:
      Error 403: knative-gcp-kn36@harwayne2.iam.gserviceaccount.com does not have
      storage.buckets.get access to storage-gcp., forbidden'
    reason: NotificationNotReady
    status: "False"
    type: NotificationReady
  - lastTransitionTime: "2019-10-24T17:47:23Z"
    status: "True"
    type: PullSubscriptionReady
  - lastTransitionTime: "2019-10-24T17:47:26Z"
    message: 'Failed to create Storage notification: operation failed: googleapi:
      Error 403: knative-gcp-kn36@harwayne2.iam.gserviceaccount.com does not have
      storage.buckets.get access to storage-gcp., forbidden'
    reason: NotificationNotReady
    status: "False"
    type: Ready
  - lastTransitionTime: "2019-10-24T17:47:17Z"
    status: "True"
    type: TopicReady
  observedGeneration: 1
  projectId: harwayne2
  sinkUri: http://www.example.com/
  topicId: storage-4bf4cae9-f686-11e9-9791-42010aa8000e
```

I **love** that the error message tells me _exactly_ what is wrong.

I went back and read the documentation and followed the command there to give my service acocunt storage admin.

**Expectation**:
Eventually the Storage would try again and become ready.

**Observation**:
The Storage is stuck forever, until I either delete it or delete the notification creation Job.

**Notes**:
I don't know what the best solution to this is. The change occurs outside of Kubernetes so there is nothing to watch.

### LINKED ISSUE #362: Channels - Don't self-heal deleted GCP Topics
I created a PubSub Channel. Then deleted the GCP Topic via gcloud. 

**Expectation**:
Eventually the Channel would notice and both mark itself as unhealthy, as well as self-heal.

**Observation**:
Nothing noticed. It will stay broken forever, until I manually delete either the Topic or the Job.

**Note**:
This is a general problem of how do we notice and fix issues with resources outside the K8s cluster itself.

### LINKED ISSUE #386: Expose "secret not found" error
## Problem

We rely on the `google-cloud-key` secret (or a custom one) to run Jobs to create topics/subscriptions. When the secret is not available, the job will stuck because the underlying pod cannot mount the volume. For example,

```
Name:           pubsub-t-cre-tstchan-5-chan-topic-create-jxzzs
Namespace:      pschan-5
Priority:       0
Node:           gke-knaive-3-default-pool-eee7931c-z35v/10.138.0.41
Start Time:     Tue, 29 Oct 2019 19:42:18 +0000
Labels:         action=create
                controller-uid=36fc42f3-fa84-11e9-b004-42010a8a0065
                job-name=pubsub-t-cre-tstchan-5-chan-topic-create
                resource-uid=
Annotations:    sidecar.istio.io/inject: false
Status:         Pending
IP:
IPs:            <none>
Controlled By:  Job/pubsub-t-cre-tstchan-5-chan-topic-create
Containers:
  job:
    Container ID:
    Image:          gcr.io/cshou-playground/topic-e7939f384cf950dd7b983054dc23c93b@sha256:f15c1f251044b5d07d5320fd487397789cd7133560dc0e6f76601af451835f9b
    Image ID:
    Port:           <none>
    Host Port:      <none>
    State:          Waiting
      Reason:       ContainerCreating
    Ready:          False
    Restart Count:  0
    Environment:
      GOOGLE_APPLICATION_CREDENTIALS:  /var/secrets/google/key.json
      ACTION:                          create
      PROJECT_ID:                      cshou-playground
      PUBSUB_TOPIC_ID:                 cre-chan-36f56d41-fa84-11e9-b004-42010a8a0065
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from default-token-pwsfk (ro)
      /var/secrets/google from google-cloud-key (rw)
Conditions:
  Type              Status
  Initialized       True
  Ready             False
  ContainersReady   False
  PodScheduled      True
Volumes:
  google-cloud-key:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  google-cloud-key
    Optional:    false
  default-token-pwsfk:
    Type:        Secret (a volume populated by a Secret)
    SecretName:  default-token-pwsfk
    Optional:    false
QoS Class:       BestEffort
Node-Selectors:  <none>
Tolerations:     node.kubernetes.io/not-ready:NoExecute for 300s
                 node.kubernetes.io/unreachable:NoExecute for 300s
Events:
  Type     Reason       Age                   From                                              Message
  ----     ------       ----                  ----                                              -------
  Normal   Scheduled    30m                   default-scheduler                                 Successfully assigned pschan-5/pubsub-t-cre-tstchan-5-chan-topic-create-jxzzs to gke-knaive-3-default-pool-eee7931c-z35v
  Warning  FailedMount  7m33s (x10 over 27m)  kubelet, gke-knaive-3-default-pool-eee7931c-z35v  Unable to mount volumes for pod "pubsub-t-cre-tstchan-5-chan-topic-create-jxzzs_pschan-5(372c56f6-fa84-11e9-b004-42010a8a0065)": timeout expired waiting for volumes to attach or mount for pod "pschan-5"/"pubsub-t-cre-tstchan-5-chan-topic-create-jxzzs". list of unmounted volumes=[google-cloud-key]. list of unattached volumes=[google-cloud-key default-token-pwsfk]
  Warning  FailedMount  3m29s (x21 over 30m)  kubelet, gke-knaive-3-default-pool-eee7931c-z35v  MountVolume.SetUp failed for volume "google-cloud-key" : secret "google-cloud-key" not found
```

As we can see from the Events, k8s was retrying to mount the volume (infinitely?).

## Fix

1. Before the controller spawns a job, it should check if the secret is available under the namespace.
2. We should add timeout for the job

### LINKED ISSUE #419: Remove Job-based Reconcilers
**Context** 
knative-gcp has decided to use a job-based framework to setup clients used to interact with different GCP services. The Pod running the Job is the one where the proper credentials are mounted and therefore is able to interact with the different GCP services. For example, in the case of Storage, whenever a Storage CR is instantiated, we create a Job in the same namespace, which setups the client to talk to GCS and creates the Storage notification. 
The Jobs framework was designed to run in the namespace in which the particular CR exists with the auth provided. The Job runs local to the resource to avoid copying service accounts or secrets into the cloud-run-events namespace. As correctly pointed out by @grantr,  this approach also allowed to manage different projects in the same cluster.

**Problem**
Although it’s a nice approach to avoid copying secrets, IMO it has gotten us more headaches than solutions (e.g., #386, #361, #358, #354, #327, #381, #281, etc, etc.). I would argue that the code is rather error-prone, and it won’t make it any easier for the addition of new sources. With this infra is also “harder” to surface error messages, as well as asserting logs in e2e tests.

**Proposal** 
We propose to move away from the Jobs-based reconcilers. The initial solution will involve having:
- A secret in the cloud-run-events namespace with elevated IAM Service Account permissions (or all of the finer-grained permissions needed to manage all of our Knative-GCP sources). 
- Secrets in user-defined namespaces for the data-plane pieces.
 
These approach should work both for GKE and non-GKE clusters. Later on, we might be able to introduce changes and integrate with Workload Identity for GKE clusters .

The control-plane piece is borrowed from ConfigConnector. Note that these approach also supports managing multiple projects.
Note that these dual-secret approach (for data and control plane) is what we used to have for the old Pub/Sub source (prior to 0.8). As correctly pointed out by @grantr, that was one of the annoyances of the previous source. I do agree, but I still think the old approach is better than what we currently have, and will solve many problems. 
For a (future) nicer UX, we might be able to auto-create default user-namespaced secrets upon annotating namespaces (either with the knative-eventing-injection annotation, or a new gcp one).

Once the security task force gets rolling, we can do another iteration on this and see if we can get away from using a single secret without jobs.

nachocano:
Fixes #419 
Fixes #112
Fixes #386
Fixes #444
Fixes #361
Fixes #358
Fixes #354
Fixes #362
Fixes #359
Fixes #325
...

## Proposed Changes

- Removing operations, i.e., job-based reconcilers.
- Mounting a "powerful" secret in the `cloud-run-events` controller to handle all interaction with GCP services. The secret does not need to be `roles/editor`, but rather have just the permissions required to interact with our current sources (e.g., for storage, I believe is `storage.admin`, for pubsub it should have `pubsub/editor`, and so on).
- Restructured the packages inside reconcilers, otherwise it was becoming too confusing (heard from multiple people, including myself). 
   - `events`  for the sources, i.e., storage, scheduler, pubsub
   - `messaging` for the channel
   - `pubsub` for topics and pullsubscriptions
- Updated control-plane stats reporter
- Removed unused code (e.g., decorator). If needed later on, we can add it back. There was no EventType registry, thus I also removed event type conditions. 
- Created google client libraries wrappers to be able to UT the interaction with outside services. Previously the only UTs done where for creation/deletion of jobs, but the `Run` methods of the operations (arguably the most important pieces), were not being unit tested.
- Added documentation to yamls. Removed deprecated attributes as well.
- Added proper k8s Events for easier debugging.
- Improved logging, used @Harwayne's logging guidelines.
- Added resync periods to re-reconcile in case there were outside cluster changes (this was done by @capri-xiyue )
- Fixed e2e tests (also done @capri-xiyue )
- Fixed sources tracking their sinks. Done with the destination resolver (also done by @capri-xiyue )

ACKs:

Special thanks to @capri-xiyue for fixing some very important bugs along the way. She ended up creating PRs to my branch because the fixes in master didn't quite work due to the job-based architecture...

Follow Ups:
- Update documentation with new installation procedures. If the secret doesn't exist the controller will crashloop.
- Remove pubSubSecret from all sources and just leave secret, which only needs `pubsub/subscriber` permissions for sources, and `pubsub/publisher` and `pubsub/subscriber` for the Channels. This will be the data-plane secrets.
- Create a namespace reconciler to watch for deletion of the `cloud-run-events` namespace, which should hold the deletion of the secret until all resources are deleted so that we do not leak pub/sub topics, subscriptions, etc (#327)

**Release Note**

```release-note
Follow the installation guide. In order to upgrade, you will need a secret called `google-cloud-key` in the `cloud-run-events` namespace, otherwise the control plane won't be able to initialize.
`Decorator` CRD has been removed.
```


test/e2e-common.sh
```diff
@@ -30,6 +30,12 @@ which gcloud &> /dev/null || gcloud() { echo "[ignore-gcloud $*]" 1>&2; }
 
 # Eventing main config.
 readonly E2E_TEST_NAMESPACE="default"
+readonly CONTROL_PLANE_NAMESPACE="cloud-run-events"
```
nachocano: @chizhg these are the changes I'm talking about... But there is an IF not PROW for the others. I want to see how we can add this to PROW?

chizhg: The service account we use to run Prow jobs already has `roles/editor`, so if the tests are run by Prow, there is no need to create another service account.
The current logic you have written is correct.

config/201-clusterrole.yaml
```diff
@@ -108,7 +106,7 @@ rules:
 - apiGroups:
     - ""
   resources:
-    - secrets # TODO: Does it really need global secrets access?
+    - secrets # For authentication with GCP.
```
nachocano: I may be able to remove this...

nachocano:
@grantr ready for a first pass! Thanks, and my apologies in advance for making this to be a huge PR. Should have modularized it better, but I got so carried away, and then it was too late :)

I'm still unsure why the Channel with the Broker is not working. The Channel alone does. And I don't think is a permissions thing.. Will keep on debugging.

test/e2e-common.sh
nachocano: Sounds good! Thanks @chizhg

nachocano:
> I'm still unsure why the Channel with the Broker is not working. The Channel alone does. And I don't think is a permissions thing.. Will keep on debugging.

Fixed the problem. Missed to set up a transformer. All E2Es should pass now

grantr:
Thanks @nachocano and @capri-xiyue for working on this huge PR! I'm starting the review now but it will probably take a while to get all the way through.

pkg/reconciler/events/scheduler/scheduler.go
```diff
@@ -0,0 +1,353 @@
...
+
+func (r *Reconciler) reconcileJob(ctx context.Context, scheduler *v1alpha1.Scheduler, topic, jobName string) (string, error) {
+	if scheduler.Spec.Project == "" {
+		project, err := metadata.ProjectID()
+		if err != nil {
+			logging.FromContext(ctx).Desugar().Error("Failed to find project id", zap.Error(err))
+			return "", err
+		}
+		scheduler.Spec.Project = project
+	}
```
nachocano: this is wrong everywhere. We shouldn't be "changing" the spec. We are actually not, but this is confusing... 
Should update in every place were we use this metadata server.
Will create a util. Further, this should be mocked.

nachocano: fixed

pkg/reconciler/pubsub/topic/topic.go
```diff
@@ -0,0 +1,441 @@
...
+	if svc.Status.IsReady() {
+		topic.Status.SetAddress(svc.Status.Address.URL)
+	}
+
+	return nil
+}
+
+func (r *Reconciler) reconcileTopic(ctx context.Context, topic *v1alpha1.Topic) error {
+	if topic.Status.ProjectID == "" {
+		projectID, err := utils.ProjectID(topic.Spec.Project)
```
grantr: Assuming:
- The project id cannot change over the lifetime of the cluster
- The webhook's project id is always the same as the controller's project id

then I suggest doing this defaulting in the webhook instead of the reconcilers.

Advantages:
- All defaulting is in one place (either in the per-kind APIs or shared across all kinds)
- If the project ID can't be defaulted (running outside GCE), the user sees that right away as a create error
- Fewer queries to the metadata server and decreased reliance on its availability. If metadata server goes down, existing resources can still be reconciled.

Disadvantages:
- Eliminates the ability to dump the resources from the cluster as yaml and apply them to a cluster in a different project without deleting the ProjectID field. This seems ok since (AFAIK) this is not generally possible in K8s anyway (metadata fields like uuid need to be removed)

This suggestion applies to all other reconcilers that have a similar defaulting.

cmd/controller/main.go
```diff
@@ -39,6 +38,5 @@ func main() {
 		pullsubscription.NewController,
 		topic.NewController,
 		channel.NewController,
-		decorator.NewController,
```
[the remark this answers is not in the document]
nachocano: Done!
BTW, there wasn't actually documentation about the decorator, so I'm not actually concerned about people using it. The author mentioned it introduced it as an "experiment". There weren't any UTs (in the adapter nor in the reconciler). 
I'm more of the opinion of not maintaining things that we do not need (at least for the time being). If we decide to maintain it, we should do it properly, with UTs, etc. 
Are you OK with removing it?

config/300-pullsubscription.yaml
```diff
@@ -18,11 +18,6 @@ metadata:
   labels:
     events.cloud.google.com/release: devel
     events.cloud.google.com/crd-install: "true"
-  annotations:
-    registry.knative.dev/eventTypes: |
-      [
-        { "type": "com.google.cloud.pubsub.topic.publish", "description": "This event is sent when a message is published to a Cloud Pub/Sub topic."}
-      ]
```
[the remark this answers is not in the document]
nachocano: None, this is actually not a Source, PullSubscription is an implementation detail of the PubSub Source, which do have this event type

config/transformers/pubsub-push.yaml
```diff
@@ -1,11 +0,0 @@
-apiVersion: serving.knative.dev/v1alpha1
```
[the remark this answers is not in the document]
nachocano: Similar with the decorator comment, I guess this was some other experiment. It pointed to an image that doesn't even exit.

config/500-controller.yaml
```diff
@@ -78,5 +72,8 @@ spec:
       - name: config-logging
         configMap:
           name: config-logging
+      - name: google-cloud-key
+        secret:
+          secretName: google-cloud-key
```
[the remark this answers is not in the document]
nachocano: Yes, it will have to be. I need to update the documentation. 
In order for the controller not to crash loop, we can provide a dummy "empty" secret, that the operator will then need to patch.

pkg/gclient/pubsub/testing/client.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/google/knative-gcp/pkg/gclient/pubsub"
+	"google.golang.org/api/option"
+)
+
+// TestClientCreator returns a pubsub.CreateFn used to construct the test Pub/Sub client.
+func TestClientCreator(value interface{}) pubsub.CreateFn {
+	var data TestClientData
+	var ok bool
+	if data, ok = value.(TestClientData); !ok {
```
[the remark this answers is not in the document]
nachocano: I guess I didn't want to cast it in the caller code, but I can certainly do it there. 
I introduced a generic `map[string]interface` field in [pkg's TableRow](https://github.com/knative/pkg/pull/933) to be able to pass arbitrary data to UTs. 
This is something we used to have in [0.8](https://github.com/knative/eventing-contrib/blob/fc84722bbae53d1b16e6b03687bb1b15f0f4302d/gcppubsub/pkg/reconciler/gcppubsubsource_test.go#L357), and I reintroduced here. 
Would you prefer me to do the assertions in the reconciler test code?

pkg/gclient/scheduler/doc.go
```diff
@@ -0,0 +1,18 @@
...
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// Package scheduler contains Scheduler client wrappers to be able to UT things.
+package scheduler
```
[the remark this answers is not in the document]
nachocano: I wasn't aware of mockery, and it seems pretty cool. Given that I already spent some time writing these, I'd rather leave that for future work, If that is OK with you.

Regarding writing more, I think in the near future we might just need only one. We can sync more offline

pkg/reconciler/pubsub/topic/topic.go
nachocano: This is another valid point and you are right, it will make things cleaner. I also like the fail fast approach.

I'm not entirely sure if it's valid the assumption that the project id given by the metadata server will always be the same throughout the lifetime of the cluster, thus we may be adding state to the webhook. I recall we had some of these conversations in the past, but don't actually remember the context then. Do you?

Even if the project id can change, do you still think it's better to have it on the webhook? 
Let me ask Scott why they didn't default it in the webhook, just in case we are missing something.

cmd/controller/main.go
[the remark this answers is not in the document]
grantr: I'm ok with removing it. Makes sense to me if it wasn't documented, tested, or used.

pkg/gclient/pubsub/testing/client.go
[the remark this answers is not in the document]
grantr: No, it's fine this way, I was just verifying the reasoning.

pkg/gclient/scheduler/doc.go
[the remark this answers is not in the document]
grantr: I'm not suggesting we replace the existing mocks, just pointing out that there might be another way. It's up to you whether you want to use a tool or not. :)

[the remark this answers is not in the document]
nachocano: Make sense, thanks for pointing it out. I think we can give it a try after this gets in. Will file an issue to investigate more about mockery.

pkg/reconciler/pubsub/topic/topic.go
grantr: If the project id can change, then I think we should leave the current reconciler defaulting. I'd be surprised if that's possible though.

Either way, this can be a followup. I created #450.

pkg/gclient/scheduler/doc.go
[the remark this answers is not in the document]
nachocano: filed: https://github.com/google/knative-gcp/issues/451
Thanks for the suggestion Grant!

pkg/reconciler/pubsub/topic/topic.go
nachocano: Hmm, isn't the project ID returned by the metadata server the default one? If you have multiple projects and you change the default, then wouldn't that return something different?
I might be wrong though

grantr: What do you mean by "change the default [project]"? Are you referring to a GCP setting?

nachocano: Let me take that back. I think you are right, and that cannot be changed. 
I was just talking with @yolocs and looked at https://cloud.google.com/compute/docs/storing-retrieving-metadata...
`Compute Engine defines a set of default metadata entries that provide information about your instance or project. Default metadata is always defined and set by the server. You can't manually edit any of these metadata pairs.`

nachocano:
> the little cleanups and doc additions.

Thank you so much @grantr !!! These thorough reviews will help us raise the bar of the repo.

pkg/reconciler/pubsub/topic/topic.go
nachocano: Resolving as it will be done in a follow up

config/500-controller.yaml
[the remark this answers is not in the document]
nachocano: Planning on adding documentation in follow ups if you are OK with it.

pkg/reconciler/events/pubsub/pubsub.go
```diff
@@ -63,89 +60,99 @@ type Reconciler struct {
...
+	if pubsub.DeletionTimestamp != nil {
 		// No finalizer needed, the pullsubscription will be garbage collected.
 		return nil
 	}
 
-	ps, err := r.reconcilePullSubscription(ctx, source)
+	ps, err := r.reconcilePullSubscription(ctx, pubsub)
 	if err != nil {
-		r.Logger.Infof("Failed to reconcile PubSub: %s", err)
+		pubsub.Status.MarkPullSubscriptionNotReady("PullSubscriptionCreateFailed", "Failed to reconcile PullSubscription: %s", err.Error())
```
grantr: Why is the reason `PullSubscriptionCreateFailed`? Can this error only happen on create?

pkg/reconciler/events/pubsub/pubsub.go
```diff
@@ -63,89 +60,99 @@ type Reconciler struct {
...
 		return err
 	}
-	source.Status.PropagatePullSubscriptionStatus(ps.Status.GetCondition(apis.ConditionReady))
+	pubsub.Status.PropagatePullSubscriptionStatus(ps.Status.GetCondition(apis.ConditionReady))
```
grantr: Should this only be executed in the success case? Does this overwrite the reason set in the error case at L127?

pkg/reconciler/events/pubsub/pubsub.go
```diff
@@ -63,89 +60,99 @@ type Reconciler struct {
...
+			"Failed to update status for PubSub %q: %v", pubsub.Name, uErr)
+		return uErr
+	} else if reconcileErr == nil {
+		// There was a difference and updateStatus did not return an error.
+		r.Recorder.Eventf(pubsub, corev1.EventTypeNormal, "Updated", "Updated PubSub %q", pubsub.Name)
+	}
 	if reconcileErr != nil {
-		// TODO: record the event (c.Recorder.Eventf(...
-		return reconcileErr
+		r.Recorder.Event(pubsub, corev1.EventTypeWarning, "InternalError", reconcileErr.Error())
```
grantr: Should we also log the error in controller logs or is that done at a higher level? (and if it is done at a higher level, should we remove the log at L101?)

pkg/reconciler/events/pubsub/pubsub.go
```diff
@@ -169,10 +176,10 @@ func (r *Reconciler) updateStatus(ctx context.Context, desired *v1alpha1.PubSub)
 
 	if err == nil && becomesReady {
 		duration := time.Since(src.ObjectMeta.CreationTimestamp.Time)
-		r.Logger.Infof("PubSub %q became ready after %v", source.Name, duration)
+		logging.FromContext(ctx).Desugar().Info("PubSub became ready", zap.Any("after", duration))
```
grantr: I might add an `Eventf` call here. The transition from unready to ready (and possibly vice versa) seems like it deserves a k8s event of its own since most of the time that's what users will be looking for.

pkg/reconciler/events/pubsub/pubsub.go
```diff
@@ -169,10 +176,10 @@ func (r *Reconciler) updateStatus(ctx context.Context, desired *v1alpha1.PubSub)
...
 
 		if err := r.StatsReporter.ReportReady("PubSub", source.Namespace, source.Name, duration); err != nil {
-			logging.FromContext(ctx).Infof("failed to record ready for Storage, %v", err)
+			logging.FromContext(ctx).Desugar().Info("Failed to record ready for PubSub", zap.Error(err))
```
grantr: Is this always unexpected? Seems like it should be Error level.

pkg/reconciler/events/pubsub/pubsub.go
```diff
@@ -63,89 +60,99 @@ type Reconciler struct {
...
-// Reconcile implements controller.Reconciler
+// Reconcile compares the actual state with the desired, and attempts to
+// converge the two. It then updates the Status block of the PubSub resource
+// with the current status of the resource.
 func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
 	// Convert the namespace/name string into a distinct namespace and name
 	namespace, name, err := cache.SplitMetaNamespaceKey(key)
 	if err != nil {
-		runtime.HandleError(fmt.Errorf("invalid resource key: %s", key))
+		logging.FromContext(ctx).Desugar().Error("Invalid resource key")
```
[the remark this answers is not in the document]
nachocano: Yes, I saw it [here](https://github.com/knative/pkg/blob/master/controller/controller.go#L356), so I removed it from everywhere...

pkg/pubsub/adapter/adapter.go
```diff
@@ -188,18 +186,22 @@ func (a *Adapter) receive(ctx context.Context, event cloudevents.Event, resp *cl
...
 		transformedCTX, transformedEvent, err := a.transformer.Send(ctx, event)
 		if err != nil {
 			logger.Errorf("error transforming cloud event %q", event.ID())
 			a.reporter.ReportEventCount(args, nethttp.StatusInternalServerError)
 			return err
 		}
 		if transformedEvent == nil {
+			// This doesn't mean there was an error. E.g., the Broker filter pod might not return a response.
+			// Report the returned Status Code and return.
 			logger.Warnf("cloud event %q was not transformed", event.ID())
```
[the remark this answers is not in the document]
nachocano: Agree, changed it to Debug. Also returned back the statusCode in the reply in both cases.

pkg/reconciler/events/pubsub/pubsub.go
nachocano: Currently, the reconcileError is logged within the reconcile method. In this case, the functions called from within reconcile. I think I've done this all over the reconcilers.
The error in L101 is when there is a failure updating the status of the object. The update can cause an error on its own. It can be the case that there was a reconcileErr, and we also failed to update the status. We log both. 
I filed an issue to later on add retries for UpdateStatus. See https://github.com/google/knative-gcp/issues/442. I've seen it happen quite often actually.

pkg/reconciler/events/pubsub/pubsub.go
nachocano: Done! and for the other reconcilers

pkg/reconciler/events/pubsub/pubsub.go
nachocano: Yes. This propagation will either mark the Subscription ready or not. If the status is not ready, then it will overwrite the reason above.

pkg/reconciler/events/pubsub/pubsub.go
nachocano: I guess the correct term would be `PullSubscriptionReconcileFailed`. It can happen during a reconciliation (e.g., when getting the resource, creating, in some cases updating, etc.)
I actually renamed most methods in the reconcilers to `reconcile*`, they used to be `create*`, and marked them `*ReconcileFailed` in case of errors. 
I later on went back to `*CreateFailed`, as IMO was more intuitive for users. My thought was that many folks don't know what a reconciliation is. 
I can go back to `*ReconcileFailed` everywhere if you prefer. I don't have a strong preference.

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -18,221 +18,145 @@ package pubsub
...
+	t, err := topics.Get(name, v1.GetOptions{})
 
-		job, err := c.KubeClientSet.BatchV1().Jobs(args.Owner.GetObjectMeta().GetNamespace()).Create(job)
-		if err != nil || job == nil {
-			c.Logger.Errorw("Failed to create Job.", zap.Error(err))
-			return ops.OpsJobCreateFailed, err
+	if err != nil {
+		if !apierrs.IsNotFound(err) {
+			logging.FromContext(ctx).Desugar().Error("Failed to get Topics", zap.Error(err))
+			return nil, nil, fmt.Errorf("failed to get Topics: %s", err.Error())
```
yolocs: A better verb now is `%w` https://golang.org/pkg/fmt/#Errorf
Similar for all other places.

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -18,221 +18,145 @@ package pubsub
...
+		return nil, nil, fmt.Errorf("nil pubsubable passed in")
+	}
+	namespace := pubsubable.GetObjectMeta().GetNamespace()
+	name := pubsubable.GetObjectMeta().GetName()
+	spec := pubsubable.PubSubSpec()
+	status := pubsubable.PubSubStatus()
 
-		job = operations.NewTopicOps(args)
+	topics := psb.pubsubClient.PubsubV1alpha1().Topics(namespace)
+	t, err := topics.Get(name, v1.GetOptions{})
```
yolocs: nit: seems no need to have two lines.

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -18,221 +18,145 @@ package pubsub
...
+	status.TopicID = t.Status.TopicID
+	status.ProjectID = t.Status.ProjectID
+	status.MarkTopicReady(cs)
 
-// UnmarshalJobTerminationMessage unmarshals the job pod termination message.
-func (c *PubSubBase) UnmarshalJobTerminationMessage(ctx context.Context, namespace, name string, v interface{}) error {
-	jobPod, err := ops.GetJobPodByJobName(ctx, c.KubeClientSet, namespace, name)
+	// Ok, so the Topic is ready, let's reconcile PullSubscription.
+	pullSubscriptions := psb.pubsubClient.PubsubV1alpha1().PullSubscriptions(namespace)
+	ps, err := pullSubscriptions.Get(name, v1.GetOptions{})
```
yolocs: nit: seems no need to have two lines.

pkg/reconciler/testing/storage.go
```diff
@@ -54,6 +54,12 @@ func WithStorageBucket(bucket string) StorageOption {
 	}
 }
 
+func WithStorageProject(project string) StorageOption {
+	return func(s *v1alpha1.Storage) {
+		s.Spec.Project = project
+	}
+}
+
 func WithStorageEventTypes(eventTypes []string) StorageOption {
```
yolocs: Is this used at all?

pkg/reconciler/events/pubsub/pubsub.go
grantr: I don't have a strong preference either, but `CreateFailed` seems wrong when the resource has already been created and the error is `get failed`. Not a big deal, can be a followup.

pkg/reconciler/events/pubsub/pubsub.go
grantr: Ignore this thread, I missed that the branch at L127 returns before this line is hit.

pkg/reconciler/pubsub/pullsubscription/pullsubscription.go
```diff
@@ -0,0 +1,556 @@
...
+	sinkURI, err := r.resolveDestination(ctx, ps.Spec.Sink, ps)
+	if err != nil {
+		ps.Status.MarkNoSink("InvalidSink", err.Error())
+		return err
+	} else {
+		ps.Status.MarkSink(sinkURI)
+	}
+
+	// Transformer is optional.
+	if ps.Spec.Transformer != nil {
```
grantr: Seems like if Transformer was originally set, then updated to be nil, the transformer URI will still be present. We should  clear the transformer URI if the Transformer field is nil.

config/500-controller.yaml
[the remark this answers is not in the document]
nachocano: That is not needed in OSS, documentation should say that the secret should be created first, and then the yamls should be applied.

pkg/apis/pubsub/v1alpha1/pull_subscription_lifecycle_test.go
```diff
@@ -383,14 +312,13 @@ func TestPubSubStatusGetCondition(t *testing.T) {
...
+		name: "mark sink and deployed and subscribed then no sink",
 		s: func() *PullSubscriptionStatus {
 			s := &PullSubscriptionStatus{}
 			s.InitializeConditions()
 			s.MarkSink("uri://example")
 			s.MarkDeployed()
-			s.MarkSubscribed()
-			s.MarkEventTypes()
+			s.MarkSubscribed("subID")
 			s.MarkNoSink("Testing", "hi%s", "")
```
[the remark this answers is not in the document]
nachocano: removed thanks! carried from before

pkg/apis/pubsub/v1alpha1/pull_subscription_lifecycle_test.go
```diff
@@ -402,33 +330,13 @@ func TestPubSubStatusGetCondition(t *testing.T) {
...
-		name: "mark sink and deployed and subscribed and event types then not deployed",
-		s: func() *PullSubscriptionStatus {
-			s := &PullSubscriptionStatus{}
-			s.InitializeConditions()
-			s.MarkSink("uri://example")
-			s.MarkDeployed()
-			s.MarkSubscribed()
-			s.MarkEventTypes()
+			s.MarkSubscribed("subID")
 			s.MarkNotDeployed("Testing", "hi%s", "")
```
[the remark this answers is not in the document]
nachocano: Done!

pkg/reconciler/events/pubsub/pubsub.go
nachocano: You are both right, finer-grained reasons would be better. And `CreateFailed` might be wrong. 
Let me go back to `ReconcileFailed` because it's an easier change than adding the finer-grain reasons. Otherwise I'll have to make changes in a bunch of places.
We can add those finer-grained reasons later on. I think with the detailed error message might be enough.

nachocano: Done!

pkg/reconciler/events/scheduler/scheduler.go
```diff
@@ -0,0 +1,348 @@
...
+	_, err = client.GetJob(ctx, &schedulerpb.GetJobRequest{Name: jobName})
+	if err != nil {
+		if st, ok := gstatus.FromError(err); !ok {
+			logging.FromContext(ctx).Desugar().Error("Failed from Scheduler client while retrieving Scheduler job", zap.String("jobName", jobName), zap.Error(err))
+			return err
+		} else if st.Code() == codes.NotFound {
+			// Create the job as it does not exist.
+			// For create we need a Parent, which from the jobName projects/PROJECT_ID/locations/LOCATION_ID/jobs/JOB_ID,
+			// is: projects/PROJECT_ID/locations/LOCATION_ID
+			parent := jobName[0:strings.LastIndex(jobName, "/jobs/")]
```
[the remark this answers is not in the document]
nachocano: Yes, cleaner! Done!

pkg/reconciler/events/scheduler/scheduler.go
```diff
@@ -0,0 +1,348 @@
...
+			_, err = client.CreateJob(ctx, &schedulerpb.CreateJobRequest{
+				Parent: parent,
+				Job: &schedulerpb.Job{
+					Name: jobName,
+					Target: &schedulerpb.Job_PubsubTarget{
+						PubsubTarget: &schedulerpb.PubsubTarget{
+							TopicName: fmt.Sprintf("projects/%s/topics/%s", scheduler.Status.ProjectID, topic),
```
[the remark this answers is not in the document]
nachocano: Done!

pkg/reconciler/pubsub/topic/topic.go
```diff
@@ -0,0 +1,441 @@
...
+	return reconcileErr
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, topic *v1alpha1.Topic) error {
+	ctx = logging.WithLogger(ctx, r.Logger.With(zap.Any("topic", topic)))
+
+	topic.Status.InitializeConditions()
+
+	if topic.DeletionTimestamp != nil {
+		logging.FromContext(ctx).Desugar().Debug("Deleting Pub/Sub topic")
```
grantr: Nit: Only log this if the topic will actually be deleted. Move the log statement inside the if block below.

pkg/reconciler/pubsub/topic/topic.go
```diff
@@ -0,0 +1,441 @@
...
+	exists, err := t.Exists(ctx)
+	if err != nil {
+		logging.FromContext(ctx).Desugar().Error("Failed to verify Pub/Sub topic exists", zap.Error(err))
+		return err
+	}
+
+	if !exists {
+		if topic.Spec.PropagationPolicy == v1alpha1.TopicPolicyNoCreateNoDelete {
+			logging.FromContext(ctx).Desugar().Error("Topic does not exist")
+			return fmt.Errorf("Topic %q does not exist", topic.Spec.Topic)
```
grantr: I'd expand the error message to clarify why the topic not existing is an unrecoverable error: `Topic %q does not exist and the topic policy doesn't allow creation`

pkg/reconciler/pubsub/topic/topic.go
```diff
@@ -0,0 +1,441 @@
...
+	existing, err := r.serviceLister.Services(topic.Namespace).Get(name)
+	if err != nil {
+		if !apierrors.IsNotFound(err) {
+			logging.FromContext(ctx).Desugar().Error("Unable to get an existing publisher", zap.Error(err))
+			return err, nil
+		}
+		existing = nil
+	} else if !metav1.IsControlledBy(existing, topic) {
+		p, _ := json.Marshal(existing)
+		logging.FromContext(ctx).Desugar().Error("Got a pre-owned publisher", zap.Any("publisher", p))
```
grantr: This error message is oddly worded. I suggest changing it to be similar to the message below: `Topic does not own publisher service`

pkg/reconciler/pubsub/topic/topic.go
```diff
@@ -0,0 +1,441 @@
...
+	delete(cfg.Data, "_example")
+
+	tracingCfg, err := tracingconfig.NewTracingConfigFromConfigMap(cfg)
+	if err != nil {
+		r.Logger.Warnw("failed to create tracing config from configmap", zap.String("cfg.Name", cfg.Name))
+		return
+	}
+	r.tracingConfig = tracingCfg
+	r.Logger.Debugw("Updated Tracing config", zap.Any("tracingCfg", r.tracingConfig))
+	// TODO: requeue all Topics.
```
grantr: This can be tracked by the same issue as in the other requeue TODOs

pkg/reconciler/events/scheduler/scheduler.go
```diff
@@ -0,0 +1,348 @@
...
+	}
+	becomesReady := desired.Status.IsReady() && !source.Status.IsReady()
+
+	// Don't modify the informers copy.
+	existing := source.DeepCopy()
+	existing.Status = desired.Status
+	src, err := r.RunClientSet.EventsV1alpha1().Schedulers(desired.Namespace).UpdateStatus(existing)
+
+	if err == nil && becomesReady {
+		duration := time.Since(src.ObjectMeta.CreationTimestamp.Time)
```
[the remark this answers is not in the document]
nachocano: Done! https://github.com/google/knative-gcp/issues/455
Good catch. This copy/pasting thing, my bad... We have this all over, even in eventing...

grantr:
Yep, there's an old issue about it somewhere. Never been high priority to
fix, but we can track it to make sure we don't forget.

pkg/utils/metadata.go
```diff
@@ -14,18 +14,22 @@ See the License for the specific language governing permissions and
 limitations under the License.
```
grantr: LOL github thinks this file is a rename but it really isn't. Git tracks renames by comparing how similar files are, and these were just similar enough it seems.

test/operations/job.go
```diff
@@ -30,15 +30,6 @@ import (
 
 type OpsJobStatus string
 
-const (
```
grantr: Do we need this file at all anymore?

pkg/reconciler/events/scheduler/scheduler.go
```diff
@@ -0,0 +1,348 @@
...
+		logging.FromContext(ctx).Desugar().Info("Scheduler became ready", zap.Any("after", duration))
+		r.Recorder.Event(source, corev1.EventTypeNormal, "ReadinessChanged", fmt.Sprintf("Scheduler %q became ready", source.Name))
+		if err := r.StatsReporter.ReportReady("Scheduler", source.Namespace, source.Name, duration); err != nil {
```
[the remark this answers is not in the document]
nachocano: Done, called it `metricErr` everywhere

pkg/reconciler/events/storage/storage.go
```diff
@@ -0,0 +1,387 @@
...
+
+	notifications, err := bucket.Notifications(ctx)
+	if err != nil {
+		logging.FromContext(ctx).Desugar().Error("Failed to fetch existing notifications", zap.Error(err))
+		return err
+	}
+
+	// This is bit wonky because, we could always just try to delete, but figuring out
+	// if an error returned is NotFound seems to not really work, so, we'll try
+	// checking first the list and only then deleting.
```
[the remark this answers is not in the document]
nachocano: We are ok with the current strategy. This is old code done by Ville, and it seems that he experienced this issue, thus this "workaround".
I had a somewhat similar experience when checking for existence of Pub/Sub topics. It was returning false, when it actually existed. So I was trying to create it again, and was failing with an AlreadyExists error. I did some wonky thing there as well..

pkg/reconciler/pubsub/reconciler.go
nachocano: not needed in the Delete. I just change it to have one line. Thanks!
Here is needed because afterwards it creates the topic with `topics`.

config/500-controller.yaml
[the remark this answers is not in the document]
grantr: I'm fine with doing docs as a followup, just create an issue for it and resolve this conversation.

pkg/reconciler/pubsub/reconciler.go
nachocano: needed here. But not in the delete. Thanks for the catch!

pkg/reconciler/testing/storage.go
nachocano: yes, this does not refer to the `EventType` CRD. Is the list of string event types you can configure in the Storage source

config/500-controller.yaml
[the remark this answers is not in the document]
nachocano: Done! https://github.com/google/knative-gcp/issues/456
thanks!

pkg/reconciler/pubsub/reconciler.go
nachocano: I'm getting this GoLand error: Unknown verb 'w' (%w).
Do you mind if we leave this for later?

pkg/reconciler/pubsub/pullsubscription/pullsubscription.go
nachocano: Done! updated UTs as well. Thanks!

pkg/reconciler/pubsub/pullsubscription/pullsubscription.go
```diff
@@ -0,0 +1,556 @@
...
+	// If the subscription doesn't exist, create it.
+	if !subExists {
+		// Create a new subscription to the previous topic with the given name.
+		sub, err = client.CreateSubscription(ctx, subID, subConfig)
+		if err != nil {
+			logging.FromContext(ctx).Desugar().Error("Failed to create subscription", zap.Error(err))
+			return "", err
+		}
+	}
+	// TODO update the subscription's config if needed.
```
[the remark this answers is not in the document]
nachocano: Done, added issue in the comment. https://github.com/google/knative-gcp/issues/457
I think we might be able to leverage @grac3gao work here.

pkg/reconciler/pubsub/pullsubscription/pullsubscription.go
```diff
@@ -0,0 +1,556 @@
...
+	}
+
+	logcfg, err := logging.NewConfigFromConfigMap(cfg)
+	if err != nil {
+		r.Logger.Warnw("Failed to create logging config from configmap", zap.String("cfg.Name", cfg.Name))
+		return
+	}
+	r.loggingConfig = logcfg
+	r.Logger.Debugw("Update from logging ConfigMap", zap.Any("loggingCfg", cfg))
+	// TODO: requeue all PullSubscriptions
```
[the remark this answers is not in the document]
nachocano: Done, added issue in the comment. https://github.com/google/knative-gcp/issues/457
I think we might be able to leverage @grac3gao work here.

pkg/reconciler/pubsub/topic/topic.go
nachocano: Done!

pkg/reconciler/pubsub/topic/topic.go
nachocano: Done! thanks!

pkg/reconciler/pubsub/topic/topic.go
nachocano: Done, thanks!

pkg/reconciler/pubsub/topic/topic.go
nachocano: Done! added a link to the issue

pkg/utils/metadata.go
nachocano: Oops!

pkg/reconciler/pubsub/reconciler.go
chizhg: FYI `%w` is new grammer supported in Go 1.13.
Prow is already running all jobs for this repo with Go 1.13, so if you update you local Go version to 1.13, you will be able to make this change.

test/operations/job.go
nachocano: Yes, this is the fix that @capri-xiyue did to e2e tests. I accidentally removed it, but this one in particular is being used for e2e tests. It's bad, I know :) I moved it inside `test/ `, to avoid confusions, and not used it anywhere else.
@chizhg and Peter will take care of refactoring all this stuff!

pkg/reconciler/pubsub/reconciler.go
nachocano: hmm I have 1.13... will sync with @yolocs tomorrow... I think this one is the last comment, and we might be ready to get this in! Thanks @chizhg !

nachocano:
OK folks, one more comment from @yolocs and I think we might be done! I'll sync with him tomorrow...

Many thanks to @grantr for the awesome review. And also to @yolocs! Without a doubt this kind of reviews will help to improve the quality bar of this repo by several X'es. But I promise not to create such big PRs in the future :)

pkg/reconciler/pubsub/reconciler.go
nachocano: Done, thanks @yolocs for the suggestion!

nachocano:
@grantr and @yolocs I think I addressed all the comments. Can you do a quick check, and if you agree we can get this in?

Issue #456, opened 2019-12-19
Document new installation procedure


nachocano:
**Problem**
Given that we are removing job-based reconcilers and will start using a control-plane secret, we need to update our install docs to reflect this.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
Docs in knative-gcp are updated

**Time Estimate (optional):** 
1d

PR #458, opened 2019-12-19
Updating install documentation

nachocano:
Fixes https://github.com/google/knative-gcp/issues/456

## Proposed Changes

- Updating installation of knative-gcp. Properly documenting about the creation of the control plane secret

**Release Note**

```release-note
Check out the new installation instructions. You will need a secret in the `cloud-run-events` namespace in order for the controller to start
```

### DIFF
--- a/docs/install/README.md
+++ b/docs/install/README.md
@@ -1,57 +1,93 @@
-## Installing Knative with GCP
+# Installing Knative with GCP
 
-To install Knative with GCP from master,
+1. Create a
+   [Google Cloud project](https://cloud.google.com/resource-manager/docs/creating-managing-projects)
+   and install the `gcloud` CLI and run `gcloud auth login`. This guide will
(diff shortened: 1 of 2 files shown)


docs/install/README.md
```diff
@@ -1,57 +1,93 @@
...
+   and [Eventing](https://knative.dev/docs/eventing/). The latter is only required if you want to use the Pub/Sub `Channel`.  
 
-```shell
-ko apply -f ./config
-```
+1. Create the `cloud-run-events` namespace. This is the namespace where our control plane pods run.
 
-Applying the `config` dir will:
+     ```shell
+     kubectl create namespace cloud-run-events
```
capri-xiyue: Do we need to create the namespace in advance?

nachocano: yes, because the secret needs to be created in that namespace, and that needs to be done before we apply the yamls.. I'd still like to keep the namespace yaml there, because this might change in the near future, and it doesn't hurt

PR #461, opened 2019-12-19
Updated PullSubscription and other docs

nachocano:
## Proposed Changes

- Updated PullSubscription docs.
- Created a couple of placeholders to add docs in other places.
- Updated event-display versions of examples.
- Added links, made a pass on Channel, Storage and Scheduler as well.

**Release Note**

```release-note
NONE
```


docs/install/pubsub-service-account.md
```diff
@@ -0,0 +1,59 @@
...
+
+        ```shell
+        gcloud iam service-accounts keys create cre-pubsub.json \
+        --iam-account=cre-pubsub@$PROJECT_ID.iam.gserviceaccount.com
+        ```
+
+    1.  Create a secret on the Kubernetes cluster with the downloaded key:
+
+        ```shell
+        kubectl --namespace default create secret generic google-cloud-key --from-file=key.json=cre-pubsub.json
```
capri-xiyue: The secret should get created in whatever namespace that data plane components reside, right? Namespace "default" is just an example, right?

docs/scheduler/README.md
```diff
@@ -3,38 +3,29 @@
 ## Overview
```
capri-xiyue: I remember when I tried to install scheduler, it asked me to create an app engine application first which is weird.

docs/install/pubsub-service-account.md
yolocs: +1

docs/pullsubscription/README.md
```diff
@@ -40,10 +37,10 @@ Pub/Sub topic, such as
...
    ```
 
-1. [Optional] If not using GKE, or want to use a Pub/Sub topic from another
-   project, uncomment and replace the
-   [`MY_PROJECT` placeholder](https://cloud.google.com/resource-manager/docs/creating-managing-projects)
-   in [`pullsubscription.yaml`](pullsubscription.yaml) and apply it.
+1. **[Optional]** If you are not using GKE, or want to use a Pub/Sub topic from another
+   project, uncomment and replace the `MY_PROJECT` placeholder in [`pullsubscription.yaml`](pullsubscription.yaml) and apply it. 
+   Note that the Service Account during the [installation](../install/README.md) step should be able to manage
+   [multiple projects](../install/managing-multiple-projects.md).
```
yolocs: It's better to also mention this in the `../install/README.md`.

docs/install/pubsub-service-account.md
```diff
@@ -0,0 +1,59 @@
...
+        ```
+
+        `google-cloud-key` and `key.json` are default values expected by our resources.
```
yolocs: Do you need a cleanup section for service account?

docs/install/pubsub-service-account.md
nachocano: Yes, you are right. Just added some clarifications. Thanks!

docs/scheduler/README.md
nachocano: do you remember that step? Do you know if there is a gcloud command for that? 
I think we will have to dogfood all of these guides. If things don't work, then we will have to add the missing steps...

docs/install/pubsub-service-account.md
nachocano: Done, added for deleting the secret and the service account! Thanks!

docs/scheduler/README.md
capri-xiyue: In my case, I just went to gcp console and created an app engine application manually. There is no gcloud command. I have not investigated why Scheduler needs an app engine application.

docs/pullsubscription/README.md
nachocano: Agree! Done, thanks!

docs/scheduler/README.md
nachocano: OK, I found the doc and updated it... Thanks for that catch @capri-xiyue

nachocano:
@yolocs @capri-xiyue ready for another look!

Issue #468, opened 2019-12-23
Controller pod cannot start because the google-cloud-key doesn't exist


yolocs:
**Describe the bug**
Controller pod cannot start because the google-cloud-key doesn't exist.

 **Expected behavior**
Applying yamls from `./confg` should at least have all control plane up and running.

 **To Reproduce**
`kubectl apply -f ./config` - controller pod won't start

**To solve**
Add a dummy secret for the controller pod to mount.

PR #469, opened 2019-12-23
Add dummy secret so that the controller pod can at least start

yolocs:
Fixes https://github.com/google/knative-gcp/issues/468

## Proposed Changes
Add a dummy secret for the controller pod to mount.

### DIFF
--- a/config/200-dummy-secret.yaml
+++ b/config/200-dummy-secret.yaml
@@ -1,0 +1,9 @@
+apiVersion: v1
+kind: Secret
+metadata:
+  name: google-cloud-key
+  namespace: cloud-run-events
+type: Opaque
+data:
+  # Base64 "dummy".
+  key.json: ZHVtbXkK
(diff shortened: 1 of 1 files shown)


config/200-dummy-secret.yaml
```diff
@@ -0,0 +1,9 @@
+apiVersion: v1
```
capri-xiyue: Can we add some comments which explains why we need the dummy secret?
(# This is a dummy secret, which will be mounted in the controller. It should be patched with credentials of a Service
# Account that can manipulate the GCP resources users are interested in.
# Data is set to dummy --> ZHVtbXkK)

capri-xiyue:
Just found https://github.com/google/knative-gcp/blob/master/docs/install/README.md mentions that user should create a secret before ko apply -f ./config. Maybe we don't need this PR.

PR #476, opened 2020-01-08
Added a converter for scheduler to set up its ce-type and ce-source

### LINKED ISSUE #475: Define a converter for scheduler, so that we can properly set up its ce-type and ce-source.
**Problem**
Define a converter for scheduler, so that we can properly set up its ce-type and ce-source. Will probably have to be done similar to the CAL source

**Exit Criteria**

**Time Estimate (optional):** 

capri-xiyue:
Added a converter for scheduler to set up its ce-type and ce-source

Fixes #475 

## Proposed Changes

- Added a converter for scheduler to set up its ce-type and ce-source
-
-

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/events/v1alpha1/scheduler_types.go
+++ b/pkg/apis/events/v1alpha1/scheduler_types.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
+	"fmt"
 	duckv1alpha1 "github.com/google/knative-gcp/pkg/apis/duck/v1alpha1"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/kmeta"
@@ -38,7 +39,12 @@ type Scheduler struct {
 	Spec   SchedulerSpec   `json:"spec"`
 	Status SchedulerStatus `json:"status"`
 }
-
+const (
+	// Scheduler CloudEvent type
+	SchedulerEventType = "com.google.cloud.sheduler.job.execute"
+	//Attributes jobName key
+	JobName = "jobName"
+)
 var (
 	_ apis.Validatable             = (*Scheduler)(nil)
 	_ apis.Defaultable             = (*Scheduler)(nil)
@@ -106,6 +112,11 @@ func (s *Scheduler) PubSubStatus() *duckv1alpha1.PubSubStatus {
 // ConditionSet returns the apis.ConditionSet of the embedding object
 func (s *Scheduler) ConditionSet() *apis.ConditionSet {
 	return &schedulerCondSet
+}
+
+// SchedulerEventSource returns the Cloud Scheduler CloudEvent source value.
+func SchedulerEventSource(jobName string) string {
+	return fmt.Sprintf("//cloudscheduler.googleapis.com/%s", jobName)
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/apis/events/v1alpha1/scheduler_types.go
+++ b/pkg/apis/events/v1alpha1/scheduler_types.go
@@ -21,7 +21,6 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
-	"fmt"
 	duckv1alpha1 "github.com/google/knative-gcp/pkg/apis/duck/v1alpha1"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/kmeta"
@@ -39,12 +38,16 @@ type Scheduler struct {
 	Spec   SchedulerSpec   `json:"spec"`
 	Status SchedulerStatus `json:"status"`
 }
+
 const (
-	// Scheduler CloudEvent type
-	SchedulerEventType = "com.google.cloud.sheduler.job.execute"
-	//Attributes jobName key
+	// CloudEvent types used by Scheduler.
+	SchedulerExecute = "com.google.cloud.sheduler.job.execute"
+	//Attributes jobName key.
 	JobName = "jobName"
+	//Attributes schedulerName key.
+	SchedulerName = "schedulerName"
 )
+
 var (
 	_ apis.Validatable             = (*Scheduler)(nil)
 	_ apis.Defaultable             = (*Scheduler)(nil)
@@ -112,11 +115,6 @@ func (s *Scheduler) PubSubStatus() *duckv1alpha1.PubSubStatus {
 // ConditionSet returns the apis.ConditionSet of the embedding object
 func (s *Scheduler) ConditionSet() *apis.ConditionSet {
 	return &schedulerCondSet
-}
-
-// SchedulerEventSource returns the Cloud Scheduler CloudEvent source value.
-func SchedulerEventSource(jobName string) string {
-	return fmt.Sprintf("//cloudscheduler.googleapis.com/%s", jobName)
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
--- a/pkg/pubsub/adapter/converters/scheduler_test.go
+++ b/pkg/pubsub/adapter/converters/scheduler_test.go
@@ -1,0 +1,151 @@
+/*
+Copyright 2019 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package converters
+
+import (
+	"context"
+	"strings"
+	"testing"
+
+	"cloud.google.com/go/pubsub"
+	cloudevents "github.com/cloudevents/sdk-go"
+	cepubsub "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub"
+	pubsubcontext "github.com/cloudevents/sdk-go/pkg/cloudevents/transport/pubsub/context"
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/knative-gcp/pkg/apis/events/v1alpha1"
+)
+
+func TestConvertScheduler(t *testing.T) {
+
+	tests := []struct {
+		name        string
+		message     *cepubsub.Message
+		source      string
+		sendMode    ModeType
+		wantEventFn func() *cloudevents.Event
+		wantErr     string
+	}{{
+		name: "valid attributes",
+		message: &cepubsub.Message{
+			Data: []byte("test data"),
+			Attributes: map[string]string{
+				"jobName":    "projects/knative-gcp-test/locations/us-east4/jobs/cre-scheduler-test",
+				"attribute1": "value1",
+				"attribute2": "value2",
+			},
+		},
+		sendMode: Binary,
+		wantEventFn: func() *cloudevents.Event {
+			return schedulerCloudEvent(map[string]string{
+				"jobName":    "projects/knative-gcp-test/locations/us-east4/jobs/cre-scheduler-test",
+				"attribute1": "value1",
+				"attribute2": "value2",
+			}, "//cloudscheduler.googleapis.com/projects/knative-gcp-test/locations/us-east4/jobs/cre-scheduler-test", )
+		},
+	}, {
+		name: "invalid upper case attributes",
+		message: &cepubsub.Message{
+			Data: []byte("test data"),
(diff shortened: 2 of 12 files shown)


capri-xiyue:
Now the events is as follows:
**☁️  cloudevents.Event 
Validation: valid
Context Attributes,
  specversion: 1.0
  type: com.google.cloud.sheduler.job.execute
  source: //cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4/jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c
  id: 441584502663166
  time: 2020-01-08T22:12:00.879Z
  datacontenttype: application/octet-stream
Extensions,
  knativecemode: binary
Data,
  my test data**

capri-xiyue:
The job "pull-google-knative-gcp-go-coverage" failed is because the test coverage of "pkg/apis/events/v1alpha1/scheduler_types.go" is below 50%. I checked the file,  it's because "pkg/apis/events/v1alpha1/scheduler_types.go" lot of boilerplate code without unit tests. And the code convention in knative-gcp doesn't test for such boilerplate code. I don't think the failed coverage is a blocker for this PR.

pkg/apis/events/v1alpha1/scheduler_types.go
```diff
@@ -108,6 +114,11 @@ func (s *Scheduler) ConditionSet() *apis.ConditionSet {
 	return &schedulerCondSet
 }
 
+// SchedulerEventSource returns the Cloud Scheduler CloudEvent source value.
+func SchedulerEventSource(jobName string) string {
+	return fmt.Sprintf("//cloudscheduler.googleapis.com/%s", jobName)
```
capri-xiyue: I'm wondering maybe I should use "//cloudscheduler.googleapis.com/v1beta1/" instead of "//cloudscheduler.googleapis.com/".  FYI: https://cloud.google.com/scheduler/docs/reference/rest/

pkg/apis/events/v1alpha1/scheduler_types.go
```diff
@@ -21,6 +21,7 @@ import (
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 
+	"fmt"
```
nachocano: import formatter

pkg/apis/events/v1alpha1/scheduler_types.go
```diff
@@ -38,7 +39,12 @@ type Scheduler struct {
 	Spec   SchedulerSpec   `json:"spec"`
 	Status SchedulerStatus `json:"status"`
 }
-
+const (
+	// Scheduler CloudEvent type
```
nachocano: nit: add a `.` at the end of the comments.

pkg/apis/events/v1alpha1/scheduler_types.go
nachocano: I'm more inclined to not include `v1beta1`. Let's leave it as is for now, and then we may revisit

pkg/apis/events/v1alpha1/scheduler_types.go
```diff
@@ -38,7 +39,12 @@ type Scheduler struct {
...
+	SchedulerEventType = "com.google.cloud.sheduler.job.execute"
```
nachocano: maybe rename the const to `SchedulerExecute` to match what we have in Storage?

pkg/apis/events/v1alpha1/scheduler_types.go
```diff
@@ -38,7 +39,12 @@ type Scheduler struct {
...
+	//Attributes jobName key
+	JobName = "jobName"
```
nachocano: does this needs to be here? Can it just be in the converter file?

pkg/reconciler/events/scheduler/scheduler.go
```diff
@@ -200,14 +200,18 @@ func (r *Reconciler) reconcileJob(ctx context.Context, scheduler *v1alpha1.Sched
 		} else if st.Code() == codes.NotFound {
 			// Create the job as it does not exist. For creation, we need a parent, extract it from the jobName.
 			parent := resources.ExtractParentName(jobName)
+			customAttributes := make(map[string]string)
+			// Add our own jobName here...
+			customAttributes[v1alpha1.JobName] = jobName
```
nachocano: this custom attribute will come in the PubSub message? If so, then we can just add an attribute to identify the converter, as Storage does... Otherwise, in the future, we might have a problem if we want to have a single receive adapter for multiple sources...

nachocano: do you know if we can do that with CAL as well?

pkg/pubsub/adapter/converters/sheduler.go
```diff
@@ -0,0 +1,70 @@
...
+func convertScheduler(ctx context.Context, msg *cepubsub.Message, sendMode ModeType) (*cloudevents.Event, error) {
+	tx := pubsubcontext.TransportContextFrom(ctx)
+	// Make a new event and convert the message payload.
+	event := cloudevents.NewEvent(cloudevents.VersionV1)
+	event.SetID(tx.ID)
+	event.SetTime(tx.PublishTime)
+	// We do not know the content type and we do not want to inspect the payload,
+	// thus we set this generic one.
+	event.SetDataContentType("application/octet-stream")
+	event.SetType(v1alpha1.SchedulerEventType)
```
nachocano: we should update the CRD yaml as well, to include this new information in the registry.... Without schema, but we can copy some description from the website.

pkg/pubsub/adapter/converters/scheduler_test.go
```diff
@@ -0,0 +1,151 @@
+/*
+Copyright 2019 Google LLC
```
nachocano: update the files of the new copyrights to be 2020

nachocano:
> Now the event is as follows:
> **cloud cloudevents.Event Validation: valid Context Attributes, specversion: 1.0 type: com.google.cloud.sheduler.job.execute source: //cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4/jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c id: 441584502663166 time: 2020-01-08T22:12:00.879Z datacontenttype: application/octet-stream Extensions, knativecemode: binary Data, my test data**

Having CE source as `//cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4/jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c`, will make it hard to be able to filter based on source, as the last part of dynamically generated... Might need to use `jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c` as part of subject... and keep source as `/cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4`... although the problem is that I'm not sure if the source/id tuple will be unique then...

pkg/reconciler/events/scheduler/scheduler.go
capri-xiyue: Yep. The custom attributes will come in the PubSub message, We can add an attribute to identify the converter as Storage does. However, we will also need an attribute to identify the cloud scheduler job.

capri-xiyue: For CAL, currently I don't find any way to set up custom attributes.

pkg/pubsub/adapter/converters/scheduler_test.go
```diff
@@ -0,0 +1,151 @@
...
+		name: "valid attributes",
+		message: &cepubsub.Message{
+			Data: []byte("test data"),
+			Attributes: map[string]string{
+				"jobName":    "projects/knative-gcp-test/locations/us-east4/jobs/cre-scheduler-test",
+				"attribute1": "value1",
+				"attribute2": "value2",
+			},
+		},
+		sendMode: Binary,
```
yolocs: Seems not necessary as every test case uses Binary

pkg/pubsub/adapter/converters/scheduler_test.go
```diff
@@ -0,0 +1,151 @@
...
+		wantErr:  "received event did not have jobName",
+	}}
+
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx := pubsubcontext.WithTransportContext(context.TODO(), pubsubcontext.NewTransportContext(
```
yolocs: `context.Background()`

capri-xiyue:
> > Now the event is as follows:
> > **cloud cloudevents.Event Validation: valid Context Attributes, specversion: 1.0 type: com.google.cloud.sheduler.job.execute source: //cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4/jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c id: 441584502663166 time: 2020-01-08T22:12:00.879Z datacontenttype: application/octet-stream Extensions, knativecemode: binary Data, my test data**
> 
> Having CE source as `//cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4/jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c`, will make it hard to be able to filter based on source, as the last part of dynamically generated... Might need to use `jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c` as part of subject... and keep source as `/cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4`... although the problem is that I'm not sure if the source/id tuple will be unique then...

Currently we use id of the pubsub message as the id of cloudevent. I checked the docs https://cloud.google.com/pubsub/docs/reference/rest/v1/PubsubMessage, it said "ID of this message, assigned by the server when the message is published. Guaranteed to be unique within the topic".  Since the id is unique within the topic, I don't think the souce/id tuple will be unique when we change the source to `/cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4`. How about we use  `/cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4` as the source, `jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c/+/id/pubsub message id` as the id to guanarntee the source/id tuple is unique? We can still use  `jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c` as part of subject.

pkg/reconciler/events/scheduler/scheduler.go
yolocs: nit: you can inline the map initialization with the job name.

nachocano: OK, we can leave CAL as is, but if we can set custom attributes here, then let's add it... both the job and the "adapter" type...

pkg/apis/events/v1alpha1/scheduler_types.go
capri-xiyue: Is there a package called "formatter"? I can't find it. And I also checked the code in eventing and knative-gcp, we usually use "fmt"

pkg/apis/events/v1alpha1/scheduler_types.go
capri-xiyue: These will be used in both reconciler and converter.

pkg/apis/events/v1alpha1/scheduler_types.go
nachocano: ohh no sorry. Just that you format the imports... the single ones, like "fmt" go in a group at the beginning.... For example: https://github.com/google/knative-gcp/blob/master/cmd/pubsub/publisher/main.go#L20

capri-xiyue:
> > `/cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4`. How about we use `/cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4` as the source, `jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c/+/id/pubsub message id` as the id to guanarntee the source/id tuple is unique? We can still use `jobs/cre-`
> 
> Hmm, I'm more inclined of not changing the id, although it might be the right thing to do... How about:
> 
> source: `//cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4/scheduler/<name-of-the-k8s-scheduler-object>`
> subject: `jobs/cre-scheduler-acf72800-e380-4d3b-bb2a-9ad44fae918c`
> 
> seems ugly as well, but I think with that we make sure of uniqueness... Cannot create two schedulers with the same name...

For the source, should it be source: `//cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4/schedulers/<name-of-the-k8s-scheduler-object>` instead of source: `//cloudscheduler.googleapis.com/projects/xiyue-knative-gcp/locations/us-east4/scheduler/<name-of-the-k8s-scheduler-object>` The difference is "schedulers" vs "scheduler"

pkg/pubsub/adapter/converters/scheduler_test.go
capri-xiyue: This is required when we call `Convert(ctx context.Context, msg *cepubsub.Message, sendMode ModeType, converterType string)`

PR #481, opened 2020-01-14
Standardize True/False/Unknown status throughout knative-gcp

### LINKED ISSUE #473: Standardize True/False/Unknown status throughout knative-gcp
**Problem**
 Standardize True/False/Unknown status throughout knative-gcp
 We have not kept it consistent in the knative-gcp repo. We should modify the code to standardize True/False/Unknown status throughout knative-gcp by following this decision table https://github.com/knative/eventing/issues/2187#issuecomment-562283838

**Exit Criteria**
The standard of True/False/Unknown status throughout knative-gcp is consistent.

**Time Estimate (optional):** 

capri-xiyue:
Fixes #473 

## Proposed Changes

- Standardize True/False/Unknown status throughout knative-gcp
-
-

**Release Note**

```release-note

```


pkg/apis/duck/v1alpha1/pubsub_lifecycle.go
```diff
@@ -20,23 +20,44 @@ import (
...
-// not ready and why.
-func (s *PubSubStatus) MarkPullSubscriptionNotReady(cs *apis.ConditionSet, reason, messageFormat string, messageA ...interface{}) {
+// MarkTopicNotConfigured sets the condition that the PubSub Topic has not yet been reconciled.
+func (s *PubSubStatus) MarkTopicNotConfigured(cs *apis.ConditionSet) {
+	cs.Manage(s).MarkUnknown(TopicReady, "TopicNotConfigured", "Topic has not yet been reconciled")
+}
+
+// MarkPullSubscriptionFalse sets the condition that the PubSub PullSubscription is
+// False and why.
+func (s *PubSubStatus) MarkPullSubscriptionFalse(cs *apis.ConditionSet, reason, messageFormat string, messageA ...interface{}) {
```
yolocs: Same here

pkg/apis/duck/v1alpha1/pubsub_lifecycle.go
```diff
@@ -20,23 +20,44 @@ import (
 	"knative.dev/pkg/apis"
 )
 
-// MarkTopicNotReady sets the condition that the PubSub Topic is not ready and why.
-func (s *PubSubStatus) MarkTopicNotReady(cs *apis.ConditionSet, reason, messageFormat string, messageA ...interface{}) {
+// MarkTopicFalse sets the condition that the PubSub Topic is not ready and why.
+func (s *PubSubStatus) MarkTopicFalse(cs *apis.ConditionSet, reason, messageFormat string, messageA ...interface{}) {
```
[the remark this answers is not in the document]
capri-xiyue: At first, we only need to differentiate between "ready" and "notReady". Now we need to differentiate among "True", "False" and "Unknown". I just followed the naming convention of `/knative.dev/pkg/apis/condition_set.go`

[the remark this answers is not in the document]
yolocs: I think the naming in `knative.dev/pkg/apis/condition_set.go` makes sense because the condition type is in the function parameters. It's pretty natural to say "mark condition type t to false". But it's not as quite natural to say "mark topic to false" because what you really do is "mark the topic ready condition to false".

An alternative is to do what eventing/serving does. E.g.

Use `MarkTopicReady` for ready=true, `MarkTopicFailed` for ready=false, `MarkTopicUnknown` for ready=unknown. Up to you.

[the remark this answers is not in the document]
capri-xiyue: In this discussion thread, https://github.com/knative/eventing/pull/2341#discussion_r361755665
. It seems like "Failed" is only for batch type resources and "False" is for living type resources. But currently in eventing/serving, it does not follow this standard.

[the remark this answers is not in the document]
yolocs: Interesting... I don't feel strongly about this. I simply prefer `MarkTopicNotReady` to `MarkTopicFalse`. I'm also fine with `MarkTopicFailed`. Will leave it to you to decide.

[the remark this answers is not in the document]
nachocano: hmm, I agree with @yolocs, I'd rather have `MarkTopicNotReady` instead of `MarkTopicFalse`... Similar with True conditions... `MarkTopicReady` instead of `MarkTopicTrue`... and I guess `MarkTopicUnknown` for the remaining sounds good...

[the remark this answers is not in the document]
capri-xiyue: I prefer `MarkTopicFailed` over `MarkTopicNotReady` since it's more clear. Is that ok?

nachocano:
Sounds good!

capri-xiyue:
@nachocano  Could you help review this PR?

pkg/apis/duck/v1alpha1/pubsub_lifecycle.go
```diff
@@ -20,23 +20,44 @@ import (
...
 func (s *PubSubStatus) MarkTopicReady(cs *apis.ConditionSet) {
 	cs.Manage(s).MarkTrue(TopicReady)
 }
 
-// MarkPullSubscriptionNotReady sets the condition that the PubSub PullSUbscription is
-// not ready and why.
-func (s *PubSubStatus) MarkPullSubscriptionNotReady(cs *apis.ConditionSet, reason, messageFormat string, messageA ...interface{}) {
+// MarkTopicNotConfigured sets the condition that the PubSub Topic has not yet been reconciled.
+func (s *PubSubStatus) MarkTopicNotConfigured(cs *apis.ConditionSet) {
+	cs.Manage(s).MarkUnknown(TopicReady, "TopicNotConfigured", "Topic has not yet been reconciled")
```
nachocano: Topic has not been configured instead of reconciled? Or is it OK as is? 
If not, update the comment as well

pkg/apis/events/v1alpha1/auditlogs_lifecycle.go
```diff
@@ -36,22 +41,33 @@ func (s *AuditLogsSourceStatus) InitializeConditions() {
 	auditLogsSourceCondSet.Manage(s).InitializeConditions()
 }
 
-// MarkPullSubscriptionNotReady sets the condition that the underlying PullSubscription
-// source is not ready and why.
-func (s *AuditLogsSourceStatus) MarkPullSubscriptionNotReady(reason, messageFormat string, messageA ...interface{}) {
+// MarkPullSubscriptionFailed sets the condition that the status of underlying PullSubscription
+// source is False and why.
```
nachocano: remove source

pkg/apis/events/v1alpha1/auditlogs_lifecycle.go
```diff
@@ -36,22 +41,33 @@ func (s *AuditLogsSourceStatus) InitializeConditions() {
...
+func (s *AuditLogsSourceStatus) MarkPullSubscriptionFailed(reason, messageFormat string, messageA ...interface{}) {
 	auditLogsSourceCondSet.Manage(s).MarkFalse(duckv1alpha1.PullSubscriptionReady, reason, messageFormat, messageA...)
 }
 
+// MarkPullSubscriptionUnknown sets the condition that the status of underlying PullSubscription
+// source is Unknown and why.
```
nachocano: remove source... PS is not a source but rather an implementation detail

pkg/apis/events/v1alpha1/auditlogs_lifecycle.go
```diff
@@ -26,6 +26,11 @@ func (s *AuditLogsSourceStatus) GetCondition(t apis.ConditionType) *apis.Conditi
 	return auditLogsSourceCondSet.Manage(s).GetCondition(t)
 }
 
+// GetTopLevelCondition returns the top level condition.
+func (s *AuditLogsSourceStatus) GetTopLevelCondition() *apis.Condition {
```
nachocano: sorry, I wasn't familiar with this TopLevelCondition... What it would be in this case? The AuditsLogReady condition?

pkg/apis/events/v1alpha1/auditlogs_lifecycle_test.go
```diff
@@ -27,63 +27,108 @@ import (
...
+				gotConditionStatus := test.s.GetTopLevelCondition().Status
+				if gotConditionStatus != test.wantConditionStatus {
+					t.Errorf("unexpected condition status: want %v, got %v", test.wantConditionStatus, gotConditionStatus)
+				}
+			}
 			got := test.s.IsReady()
-			if diff := cmp.Diff(test.want, got); diff != "" {
-				t.Errorf("%s: unexpected condition (-want, +got) = %v", test.name, diff)
+			if got != test.want {
+				t.Errorf("unexpected readiniess: want %v, got %v", test.want, got)
```
nachocano: nit: readiness

pkg/apis/events/v1alpha1/pubsub_lifecycle_test.go
```diff
@@ -17,61 +17,53 @@ limitations under the License.
 package v1alpha1
 
 import (
+	pubsubv1alpha1 "github.com/google/knative-gcp/pkg/apis/pubsub/v1alpha1"
```
nachocano: this import should go slightly below, not with the single word ones...

pkg/apis/duck/v1alpha1/pubsub_lifecycle.go
capri-xiyue: I follow the naming convention of serving https://github.com/knative/serving/blob/ec457adcb1466abfcbc1c0a069e60285e769d135/pkg/apis/serving/v1alpha1/route_lifecycle.go#L72. I will update the comments.

pkg/apis/events/v1alpha1/auditlogs_lifecycle.go
capri-xiyue: It will get the internal "happy" condition which is also the "ready" condition of AuditLogsSource.

pkg/apis/duck/v1alpha1/pubsub_lifecycle.go
nachocano: OK, I like the consistency... leave the reason as is then, and see if you can update the comments.
Thanks @capri-xiyue

pkg/apis/events/v1alpha1/pubsub_lifecycle.go
```diff
@@ -37,24 +43,42 @@ func (ps *PubSubStatus) InitializeConditions() {
...
+	case ready.Status == corev1.ConditionUnknown:
+		ps.MarkPullSubscriptionUnknown(ready.Reason, ready.Message)
+	case ready.Status == corev1.ConditionTrue:
 		ps.MarkPullSubscriptionReady()
-	case ready.IsFalse():
-		ps.MarkPullSubscriptionNotReady(ready.Reason, ready.Message)
+	case ready.Status == corev1.ConditionFalse:
+		ps.MarkPullSubscriptionFailed(ready.Reason, ready.Message)
+	default:
+		ps.MarkPullSubscriptionUnknown("PullSubscriptionUnknown", "The status of PullSubscription is invalid: %v", ready.Status)
```
nachocano: can it go in here? I guess is safer to have this default?

capri-xiyue: I'm confused with this comment. What do you mean by "it"?

capri-xiyue: Are you asking whether it's safer to have this default? I think so. Otherwise, we will mark the PullSubscriptionReady inaccurately or fail to mark it.

nachocano: yeah... I haven't seen that default in the serving example you sent. So I'm not sure if that condition status can be something different than ConditionFalse, ConditionTrue, or ConditionUnknown... If not, then it will never go through that default... 
But I think you are right, I'd rather leave it in case the user messes up with things there.

pkg/apis/events/v1alpha1/pubsub_lifecycle.go
```diff
@@ -37,24 +43,43 @@ func (ps *PubSubStatus) InitializeConditions() {
 	pubSubCondSet.Manage(ps).InitializeConditions()
 }
 
-// MarkPullSubscriptionNotReady sets the condition that the underlying PullSubscription
-// source is not ready and why.
-func (ps *PubSubStatus) MarkPullSubscriptionNotReady(reason, messageFormat string, messageA ...interface{}) {
+// MarkPullSubscriptionFailed sets the condition that the underlying PullSubscription
+// source is False and why.
```
nachocano: nit: remove source

pkg/apis/events/v1alpha1/test_helper.go
```diff
@@ -0,0 +1,48 @@
+/*
```
nachocano: I'm not very convinced of this test helper. I think it will be packaged along with non-test code due to its name (not ending in *_test.go)
But most importantly, we've been using the Option pattern for tests, for example, look at the reconciler tests https://github.com/google/knative-gcp/blob/master/pkg/reconciler/testing/pullsubscription.go...
I think we might need to move those out of pkg/reconciler, and put them in pkg/testing, and then you won't depend on anything within reconciler... But for now, I think you can just create these 3 functions as private helpers in the go file of the UT that uses them... I see only being used in a single place?

pkg/apis/messaging/v1alpha1/channel_lifecycle_test.go
```diff
@@ -17,6 +17,7 @@ limitations under the License.
 package v1alpha1
 
 import (
+	"github.com/google/knative-gcp/pkg/apis/pubsub/v1alpha1"
```
nachocano: put it below and separate from the single word imports

pkg/apis/messaging/v1alpha1/channel_lifecycle_test.go
```diff
@@ -176,14 +187,14 @@ func TestChannelIsReady(t *testing.T) {
...
+			cs.PropagateTopicStatus(test.topicStatus)
+			gotConditionStatus := cs.GetTopLevelCondition().Status
+			if test.wantConditionStatus != gotConditionStatus {
+				t.Errorf("unexpected condition status: want %v, got %v", test.wantConditionStatus, gotConditionStatus)
 			}
 			got := cs.IsReady()
-			if test.wantReady != got {
-				t.Errorf("unexpected readiness: want %v, got %v", test.wantReady, got)
+			if got != test.want {
+				t.Errorf("unexpected readiniess: want %v, got %v", test.want, got)
```
nachocano: nit: readiness

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -19,13 +19,15 @@ package pubsub
 import (
 	"context"
 	"fmt"
+	"github.com/google/knative-gcp/pkg/apis/duck/v1alpha1"
```
nachocano: same here, move this one down

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -19,13 +19,15 @@ package pubsub
...
 
 	pubsubsourcev1alpha1 "github.com/google/knative-gcp/pkg/apis/pubsub/v1alpha1"
```
nachocano: pubsubv1alpha1 alias might be more appropriate

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -19,13 +19,15 @@ package pubsub
...
 	pubsubsourceclientset "github.com/google/knative-gcp/pkg/client/clientset/versioned"
```
nachocano: just `clientset` ?

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -130,6 +112,63 @@ func (psb *PubSubBase) ReconcilePubSub(ctx context.Context, pubsubable duck.PubS
...
+	return nil
+}
+
+func propagateTopicStatus(t *pubsubsourcev1alpha1.Topic, status *v1alpha1.PubSubStatus, cs *apis.ConditionSet, topic string ) error {
+	tc := t.Status.GetTopLevelCondition()
+	if tc == nil {
+		status.MarkTopicNotConfigured(cs)
+		return fmt.Errorf("Topic %q has not yet been reconciled", t.Name)
+	}
+	if tc.Status == corev1.ConditionUnknown {
```
nachocano: switch statement, similar as the one above to be more consistent?

pkg/reconciler/pubsub/reconciler.go
```diff
@@ -130,6 +112,63 @@ func (psb *PubSubBase) ReconcilePubSub(ctx context.Context, pubsubable duck.PubS
...
+		status.MarkPullSubscriptionFailed(cs, pc.Reason, pc.Message)
+		return fmt.Errorf("the status of PullSubscription %q is False", ps.Name)
+	default:
+		status.MarkPullSubscriptionUnknown(cs, "PullSubscriptionUnknown", "The status of PullSubscription is invalid: %v", pc.Status)
+		return fmt.Errorf("the status of PullSubscription %q is invalid: %v", ps.Name, pc.Status)
+	}
+	return nil
+}
+
+func propagateTopicStatus(t *pubsubsourcev1alpha1.Topic, status *v1alpha1.PubSubStatus, cs *apis.ConditionSet, topic string ) error {
```
nachocano: `topic )` --> `topic)`, remove the whitespace

nachocano:
@capri-xiyue this looks great! Just a few nits if you can, and it should be ready to go...

pkg/reconciler/pubsub/reconciler.go
capri-xiyue: I checked the codebase. We usually use the alias "duckv1alpha1"

nachocano:
/lgtm
/approve

This looks great! Thanks @capri-xiyue

chizhg:
Please retest after https://github.com/google/knative-gcp/pull/487 gets merged.

PR #484, opened 2020-01-15
Scheduler E2E Test for Knative GCP

peterfeifanchen:
## Proposed Changes

-Add a scheduler e2e test that performs 2 checks:
1) A smoke test for checking that we can set up a scheduler source
2) A test that outputs events from a scheduler source to a bespoke sink
similar to docs/scheduler
**Release Note**

```release-note
NONE
```
/cc @chizhg @nachocano


test/e2e/e2e_test.go
```diff
@@ -158,3 +158,17 @@ func TestAuditLogsSource(t *testing.T) {
 	defer cancel()
 	AuditLogsSourceWithTestImpl(t)
 }
+
+// TestSmokeSchedulerSetup
```
nachocano: what does the test do? can we add more comments?

test/e2e/e2e_test.go
```diff
@@ -158,3 +158,17 @@ func TestAuditLogsSource(t *testing.T) {
...
+func TestSmokeSchedulerSetup(t *testing.T) {
+	cancel := logstream.Start(t)
+	defer cancel()
+	SmokeSchedulerSetup(t)
+}
+
+// TestSchedulerWithTargetTestImpl
```
nachocano: same here... not just the name of the function

test/e2e/test_scheduler.go
```diff
@@ -0,0 +1,107 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	kngcptesting "github.com/google/knative-gcp/pkg/reconciler/testing"
+	"github.com/google/knative-gcp/test/e2e/lib"
+	"github.com/google/knative-gcp/test/e2e/lib/resources"
+)
+
+// SmokeSchedulerSetup tests if a scheduler can be created
```
nachocano: add a `.` at the end of the comments. Here and other places

test/e2e/test_scheduler.go
```diff
@@ -0,0 +1,107 @@
...
+	client := lib.Setup(t, true)
+	defer lib.TearDown(client)
+
+	sName := "scheduler-test"
+
+	scheduler := kngcptesting.NewScheduler(sName, client.Namespace,
+		kngcptesting.WithSchedulerLocation("us-central1"),
+		kngcptesting.WithSchedulerData("my test data"),
+		kngcptesting.WithSchedulerSchedule("* * * * *"),
+		kngcptesting.WithSchedulerSink(metav1.GroupVersionKind{Version: "v1", Kind: "Service"},
```
nachocano: I think @chizhg created a gvk variable for a v1 Service that you can use

test/e2e/test_scheduler.go
```diff
@@ -0,0 +1,107 @@
...
+	client.CreateJobOrFail(job, lib.WithServiceForJob(targetName))
+
+	// Create a scheduler
+	sName := "scheduler-test"
+
+	scheduler := kngcptesting.NewScheduler(sName, client.Namespace,
+		kngcptesting.WithSchedulerLocation("us-central1"),
+		kngcptesting.WithSchedulerData("my test data"),
+		kngcptesting.WithSchedulerSchedule("* * * * *"),
+		kngcptesting.WithSchedulerSink(metav1.GroupVersionKind{Version: "v1", Kind: "Service"},
```
nachocano: same here

test/test_images/scheduler_target/main.go
```diff
@@ -0,0 +1,56 @@
...
+	}
+
+	if err := client.StartReceiver(context.Background(), r.Receive); err != nil {
+		log.Fatal(err)
+	}
+}
+
+type Receiver struct{}
+
+func (r *Receiver) Receive(event cloudevents.Event) {
```
nachocano: I think it would be better if we create a target image that validates more than just that the event was received. Take a look at the auditlogs target image, and there we are validating more of actual content of the event...
That will make the e2e test pretty robust...

nachocano:
Thanks @peterfeifanchen, this looks great! 
Few nits and one more important comment that will make the test much more robust!

test/test_images/scheduler_target/main.go
chizhg: +1
I think `SOURCE`, `TYPE`, `SUBJECT` and `TIME` are the parts we can validate, similar as what we have for `AuditLogsSource` - https://github.com/google/knative-gcp/tree/master/test/test_images/auditlogs_target.

peterfeifanchen: Do we need to do TYPE? The event isnt a special type like Audit has. Same with SOURCE, it doesn't have a set topic like audit logs does and there is already prefix check in SUBJECT. Also TIME is a timeout not a check.

test/e2e/lib/typemetas.go
```diff
@@ -55,6 +55,8 @@ var PubsubTypeMeta = eventsTypeMeta(resources.PubSubKind)
 
 var AuditLogsSourceTypeMeta = eventsTypeMeta(resources.AuditLogsSourceKind)
 
+var SchedulerSourceTypeMeta = eventsTypeMeta("Scheduler")
```
nachocano: "Scheduler" string in resources.SchedulerSourceKind ? as with others?

test/test_images/scheduler_target/main.go
```diff
@@ -0,0 +1,126 @@
...
+	fmt.Printf("Received event\n")
+	fmt.Printf("	Context: %v\n", event.Context.String())
+	fmt.Printf("	Data: %s\n", event.Data)
+	fmt.Printf("	Encoded: %v\n", event.DataEncoded)
+	fmt.Printf("	Binary: %v\n", event.DataBinary)
+
+	incorrectAttributes := make(map[string]interface{})
+
+	// Check subject prefix
+	subject := event.Context.GetSubject()
```
nachocano: `event.Subject()` should work

test/test_images/scheduler_target/main.go
```diff
@@ -0,0 +1,126 @@
...
+	if !strings.HasPrefix(subject, r.SubjectPrefix) {
+		incorrectAttributes[eventSubject] = subject
+	}
+
+	// Check knativecemode extension
+	if knativecemode, ok := event.Context.GetExtensions()["knativecemode"]; ok {
```
nachocano: `event.Context.GetExtension("knativecemode")`...  but I wouldn't care much about this knativecemode actually, that is something that we might remove anytime soon.
I'd rather check for the Source, Type and Subject, and the data if possible

test/test_images/scheduler_target/main.go
chizhg: I'm okay with not checking TYPE and SOURCE, if it's hard to do so. Besides we are not checking them either for `PullSubscription` and `Storage` either.

nachocano: hmm why is it hard to check the type or the source? I think it will be useful, so that we don't accidentally break it... 
but I'd rather have this in asap, so that we can mark Scheduler Source to be done. 
We can file an issue to do better assertions in e2e tests in general as a follow up. 
I think I saw some issue that @chizhg filed on how we should improve reading the logs...

Do you agree?

peterfeifanchen: I think I can add type since a change just seem to have gone in that actually set it for the event. The Source would be just a prefix check like Subject, so I don't know if it is adding anything. It may be better to leave the discussion for doing better asserts in e2e for a follow up since that is related to a discussion about what we want to be robust against.

test/test_images/scheduler_target/main.go
peterfeifanchen: I can remove knativecemode

chizhg:
> /retest

One quick tip here - whenever you make a new push, the jobs will be automatically retriggered, so you don't need to type `/retest` here. Also, `/retest` will only retest failed jobs.

Issue #480, opened 2020-01-13
CAL source misses the CloudAuditLogs which gets created shortly after CAL source becomes ready


chizhg:
From the comment in https://github.com/ian-mi/knative-gcp/pull/2#pullrequestreview-333515256.

If the AuditLog was created shortly after the CAL source becomes ready, the log will be missed. e.g. when the log sink got created and CAL source becomes ready at 1:42:35pm and one AuditLog got fired at 1:43:04pm, the AuditLog will get missed by the CAL source.

chizhg:
Also the e2e test for it is pretty flaky - https://testgrid.knative.dev/knative-gcp#continuous&exclude-non-failed-tests=

nachocano:
I don't see this as our issue. It seems to be related to CAL SLO. 
Closing it as I fixed the flakiness by increasing the timeout... 
We should properly document that users should wait a few seconds before they will start receiving events.

If you disagree, please re-open and we can revisit

PR #508, opened 2020-01-22
add retries/fallbacks for update status

### LINKED ISSUE #442: Add retries for Setting Status

nlopezgi:
Fixes #442 

## Proposed Changes

- Adds retires/fallbacks to update status similarly to knative/serving#6162

Note: I duplicated some code from knative/serving#6162 related to retry mechanism (+tests for retry mechanism), not sure if there is a way to properly reuse the code in knative/serving#6162, please let me know if so to make according changes.


nlopezgi:
Not entirely sure why the e2e tests are failing. Read through the error log, but I can seem to pin point exactly what/why the failure. If anyone has advice as to why this is failing it would be much appreciated! Thanks!

pkg/reconciler/events/auditlogs/auditlogs.go
```diff
@@ -263,32 +264,36 @@ func (c *Reconciler) removeFinalizer(s *v1alpha1.CloudAuditLogsSource) {
...
-	src, err := c.RunClientSet.EventsV1alpha1().CloudAuditLogsSources(desired.Namespace).UpdateStatus(existing)
-
-	if err == nil && becomesReady {
-		duration := time.Since(src.ObjectMeta.CreationTimestamp.Time)
-		logging.FromContext(ctx).Desugar().Info("CloudAuditLogsSource became ready", zap.Any("after", duration))
-		c.Recorder.Event(source, corev1.EventTypeNormal, "ReadinessChanged", fmt.Sprintf("CloudAuditLogsSource %q became ready", source.Name))
-		if err := c.StatsReporter.ReportReady("CloudAuditLogsSource", source.Namespace, source.Name, duration); err != nil {
-			logging.FromContext(ctx).Infof("failed to record ready for CloudAuditLogsSource, %v", err)
+func (c *Reconciler) updateStatus(ctx context.Context, existing *v1alpha1.CloudAuditLogsSource, desired *v1alpha1.CloudAuditLogsSource) error {
+	existing = existing.DeepCopy()
```
nachocano: this sounds confusing. Can you just pass `original`, and then do existing := original.DeepCopy() ?
here and in other places as well

pkg/reconciler/events/auditlogs/auditlogs.go
```diff
@@ -263,32 +264,36 @@ func (c *Reconciler) removeFinalizer(s *v1alpha1.CloudAuditLogsSource) {
...
+	return reconciler.RetryUpdateConflicts(func(attempts int) (err error) {
+		// The first iteration tries to use the informer's state, subsequent attempts fetch the latest state via API.
+		if attempts > 0 {
+			existing, err = c.auditLogsSourceLister.CloudAuditLogsSources(desired.Namespace).Get(desired.Name)
```
nachocano: here it should be using the client according to the comment, and the other folks code. 
attempt 0 uses the informer copy, which will be that original deepcopied to existing... 

but then if that fails, you query the ApiServer with the clientSet to get a new version, and try to perform the update.. does it make sense?

pkg/reconciler/retry.go
```diff
@@ -0,0 +1,30 @@
+/*
```
nachocano: I think this will be useful to add in pkg, as we might start using this in other repos as well, such as eventing..
Maybe under: https://github.com/knative/pkg/tree/master/reconciler

Can you check whether knative/eventing started doing this retries in their controllers. I think I saw an issue filed sometime back (similar timing as I filed this one)...

pkg/reconciler/retry_test.go
```diff
@@ -0,0 +1,77 @@
+/*
```
nachocano: same with the other comment, might be good to upstream this to pkg

nachocano:
Regarding the test failure, it's a test that has been flaky lately... If you click on Details next to where it says it fails, it will show you the logs, and which one did fail... 
We need to take care of that flakiness sometime soon. There is an issue filed for it: https://github.com/google/knative-gcp/issues/480#issuecomment-574846821

pkg/reconciler/events/auditlogs/auditlogs.go
nlopezgi: thanks for catching this; error on my part reading the other folk's code. Addressed this one and want to see if tests pass successfully before resolving suggestion you made above.

pkg/reconciler/retry.go
nlopezgi: looks like they did: https://github.com/knative/eventing/issues/2296. I'll set up a PR to add these to https://github.com/knative/pkg/tree/master/reconciler and ping this PR back once that is in (or if I get pushback)

nlopezgi: created https://github.com/knative/pkg/pull/1008

nachocano: sounds good! one thing that would be nice, if you can, is to add some UT that does at least one retry... not sure how hard it will be, if you need to mock things, etc. But I think it will also be good  so that you get familiar with the reconciler UTs (in case you are not). Maybe borrow something from serving.

nlopezgi: cool, I'll look at serving to see what they have wrt reconciler UTs and try to borrow something, otherwise if I cant figure it out I'll ask for help.

pkg/reconciler/events/auditlogs/auditlogs.go
nlopezgi: done

pkg/reconciler/retry_test.go
nlopezgi: as commented above, sent PR. Resolving this to keep only the comment above open.

pkg/reconciler/retry.go
nlopezgi: hi @nachocano ,
I spent some time today looking at the reconciler UTs. I feel like I understand a bit better now how they work. However, I cannot figure out a trivial way to create a test that does retries. Here is my understanding, and a half-baked potential way to achieve this kind of tests (though it seems very involved):
- In order to have a UT that performs retries, we would need to have a reconciler RunClientSet mock that can be customized to produce an error when updating a resource.
- The current testing infra relies on a MakeFactory that creates a reconciler using fake clients
- All these fake clients are generated by client-gen and injection-gen and cannot be customized to provide a mock like behavior to cause retries to be needed (afaiu)
- There seems to be some level of mocking in the reconcilers that could be mimicked for the desired behavior (this is my half baked idea):
  - Some of the reconcilers have a `createClientFn` used to create a client and inject a mock in UTs
  - We could provide a similar function to create the `RunClientSet` so that reconciler implementations would use that. But we would need to generate mocks for all the interfaces in `RunClientSet`, perhaps using some well known mocking library, as I could not find anything in the knative code about mocking.

So, my half baked idea sounds extremely involved, maybe you have other ideas as to how to achieve a test for this?

nlopezgi:
@nachocano 
https://github.com/knative/pkg/pull/1008 is in (but I'm not too familiar as to how releases / deps are managed in these projects to know when that PR will be ready to be used here). I left a comment about UTs after my exploration today, let me know if you have any ideas.

pkg/reconciler/retry.go
nachocano: OK, no worries about the UTs... 
I think that Table framework had a WithReactors or something, that you can "inject" failures when performing certain operations... But I haven't used it much... 
We can leave that out for the future...

nachocano:
So, you will have to do 
`dep ensure -update knative.dev/pkg`
and then run ./hack/update-deps.sh (which removes some OWNER files, does some patching etc..)

With that, you should get the latest pkg changes in master (taking into account that your `Gopkg.toml`  is pointing to knative.dev/pkg master, which I think it is...)

and then you also push those vendored files...

nlopezgi:
> So, you will have to do
> `dep ensure -update knative.dev/pkg`
> and then run ./hack/update-deps.sh (which removes some OWNER files, does some patching etc..)
> 
> With that, you should get the latest pkg changes in master (taking into account that your `Gopkg.toml` is pointing to knative.dev/pkg master, which I think it is...)
> 
> and then you also push those vendored files...

done

pkg/reconciler/retry.go
nlopezgi: @nachocano thanks for the advice. I had not noticed the `WithReactors` in Table, it does seem to work as intended. I added some basic UTs that include 1 retry to each of the modified resources, let me know if these tests look good as is (extending current tests) or if you want me to split off a separate test func just to validate the retries.

nlopezgi:
@nachocano this is ready for review, including UTs to validate retries via use of the `WithReactors`

nachocano:
you also need a 
./hack/update-codegen.sh

pkg/reconciler/retry.go
nachocano: They look great!! Thanks for adding them!
Can you just add a comment or update the name of the UTs you channged to say that they are also retrying?

nlopezgi: added "with retry" to UT names

nlopezgi:
@nachocano ready for review, let me know if anything is pending for this to go in. Thanks!

nlopezgi:
just ran hack/update-codegen.sh but CI still seems to complain. Let me know if you know what it is that I'm doing wrong.

PR #512, opened 2020-01-23
Move ConvertToPush method to pubsub.go

### LINKED ISSUE #497: CloudPubSubSource should use PushCompatible mode
**Problem**
CloudPubSubSource should be configured to send events in PushCompatible mode. 
Here is an example of a PubSub push message, which we should include as part of the CE data

```json
{
   "message": {
      "attributes": {
         "[KEY1]" : "[VALUE1]",
         "[KEY2]" : "[VALUE2]",
         …
      },
      "data" : "[DATA]",
      "messageId" : "[MESSAGE_ID]",
      "publishTime" : "[PUBLISH_TIME]"
   },
   "subscription": "[SUBSCRIPTION_ID]"
}

```

Developer

**Exit Criteria**
CloudPubSubSource emits CloudEvents with Push format

**Time Estimate (optional):** 
2 days

Currently there is a ConvertToPush method within pkg/adapter/push that does that. The problem is that the code goes through the pubsub converter first, which promotes attributes to extensions, and then performs this latter conversion to Push. During the first conversion, we might be missing attributes, as they might not be valid extensions. 
Given that CloudPubSubSource is the only one that currently supports Push, we can move the push.go code up to the pubsub.go converter, and do the appropriate changes there...
If users want the Pull format, they can simply use the PullSubscription.

nlopezgi:
Fixes #497

## Proposed Changes

- Following suggestion in #497 moving the ConvertToPush up to pubsub.go, making it possible for attributes that are not valid extensions to be preserved as part of conversion to Push

Details:
- Most code migrated from push.go to pubsub.go
- if Push mode, converter sets content type to "application/json" in order for convertToPush to decode message.
- All attrs that are part of the message (i.e., no valid extension filtering) are passed to convertToPush
- migrated tests from push_test.go to pubsub.go +
- added test that validates convert calls convertToPush appropriately. Note that message Data has to be passed with quotations for decoding to properly work.


pkg/pubsub/adapter/converters/pubsub.go
```diff
@@ -58,5 +62,74 @@ func convertPubSub(ctx context.Context, msg *cepubsub.Message, sendMode ModeType
 			}
 		}
 	}
+	// If send mode is Push, convert to Pub/Sub Push payload style.
+	if sendMode == Push {
+		// set the content type to something that can be handled by codec.go
+		event.SetDataContentType("application/json")
```
nachocano: use cloudevents.ApplicationJSON

pkg/pubsub/adapter/converters/pubsub.go
```diff
@@ -58,5 +62,74 @@ func convertPubSub(ctx context.Context, msg *cepubsub.Message, sendMode ModeType
...
+	// server for Messages obtained from a subscription.
+	// This field is read-only.
+	PublishTime time.Time `json:"publish_time,omitempty"`
+}
+
+// convertToPush convert an event to a Pub/Sub style Push payload.
+func convertToPush(ctx context.Context, event cloudevents.Event, attrs map[string]string) cloudevents.Event {
+	logger := logging.FromContext(ctx).With(zap.Any("event.id", event.ID()))
+
+	tx := pubsubcontext.TransportContextFrom(ctx)
```
nachocano: some of this things are already done in the calling code... 
I think we should try to do all the common stuff at the beginning, such as creating the event with the spec version, id, time, etc... and then have an if the sendMode is push... you set up the content type and the proper event.Data with this push payload.... if it's not push, then we promote the msg.Attributes to extensions, if they are valid...
does it make sense?

pkg/pubsub/adapter/converters/pubsub.go
```diff
@@ -58,5 +62,74 @@ func convertPubSub(ctx context.Context, msg *cepubsub.Message, sendMode ModeType
 			}
 		}
 	}
+	// If send mode is Push, convert to Pub/Sub Push payload style.
+	if sendMode == Push {
```
nachocano: this if should have the else clause, where you would have the msg.Attributes promoted to extensions as in the above lines... only promote attributes to extensions if it's not push...

nlopezgi: done, note as I mentioned below, tests used to check the event had the extensions as well in Push mode.

pkg/pubsub/adapter/converters/pubsub.go
nlopezgi: done

pkg/pubsub/adapter/converters/pubsub.go
nlopezgi: done. Note tests used to check the Push event also had the extensions, but probably not necessary to have them there.

nachocano: yeah, I think that was wrong... If we send push format, we shouldn't be promoting attributes to extensions also..

pkg/pubsub/adapter/converters/pubsub.go
```diff
@@ -57,6 +89,35 @@ func convertPubSub(ctx context.Context, msg *cepubsub.Message, sendMode ModeType
 				event.SetExtension(k, v)
 			}
 		}
+
 	}
 	return &event, nil
 }
+
+// pushMessage represents the format Pub/Sub uses to push events.
+type pushMessage struct {
```
nachocano: thanks for making it private!

nachocano:
/lgtm
/approve

Thanks @nlopezgi !!!

nachocano:
@nlopezgi can you cherry-pick it to release-0.12 branch... We will need this there, together with a follow up that I'll start now

nlopezgi:
> release-0.12 branch

Created https://github.com/google/knative-gcp/pull/520 to cherry pick change into release-0.12

PR #521, opened 2020-01-24
CloudPubSubSource to Push mode

nachocano:
Follow up from https://github.com/google/knative-gcp/pull/512

## Proposed Changes

- Setting the CloudPubSubSource to always emit Push-compatible format. If users want Pull format, then they should use PullSubscription.
- Adding a PullSubscriptionArgs struct, otherwise the method signature was getting to cumbersome.
- Updating docs

**Release Note**

```release-note
CloudPubSubSource now emits Push-based format events. If users want to keep on consuming Pull-based format, they should use PullSubscription. 
```


nachocano:
flaky CAL source

/test pull-google-knative-gcp-integration-tests

PR #537, opened 2020-02-05
adding CheckImmutableFields for CloudStorageSource

### LINKED ISSUE #207: #Bug Updating the event types of an GCS importer doesn't work and is not immutable
Create a GCS importer with just the "finalize" event type.
Then edit it to add "metadataupdate". Webhook accepts it and the object is ready.
However, gsutil notification list <bucketurl> doesn't show the new type. (also old ones stay forever, but I believe there is another bug tracking it).

I haven't tested for other fields. 

Expected behavior:
Either make the field immutable, or reconcile (ideal) it.

nlopezgi:
Fixes #207

## Proposed Changes

- Add checks for immutable fields in CloudStorageSource
- Modification of EventType, Bucket, ObjectNamePrefix and PayloadFormat, Secret, PubSubSecret and Project are not allowed
- Add UTs

### DIFF
--- a/pkg/apis/events/v1alpha1/cloudstoragesource_validation.go
+++ b/pkg/apis/events/v1alpha1/cloudstoragesource_validation.go
@@ -23,6 +23,9 @@ import (
 	"k8s.io/apimachinery/pkg/api/equality"
 	"knative.dev/pkg/apis"
 	duckv1 "knative.dev/pkg/apis/duck/v1"
+
+	"github.com/google/go-cmp/cmp"
+	"github.com/google/go-cmp/cmp/cmpopts"
 )
 
 func (current *CloudStorageSource) Validate(ctx context.Context) *apis.FieldError {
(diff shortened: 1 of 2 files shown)


pkg/apis/events/v1alpha1/cloudstoragesource_validation.go
```diff
@@ -73,3 +76,20 @@ func validateSecret(secret *corev1.SecretKeySelector) *apis.FieldError {
...
 }
+
+func (current *CloudStorageSource) CheckImmutableFields(ctx context.Context, original *CloudStorageSource) *apis.FieldError {
+	if original == nil {
+		return nil
+	}
+	// Modification of EventType, Secret, PubSubSecret, and Project are not allowed. Everything else is mutable.
+	if diff := cmp.Diff(original.Spec, current.Spec,
+		cmpopts.IgnoreFields(CloudStorageSourceSpec{},
+			"Sink", "CloudEventOverrides", "ServiceAccountName", "Bucket", "ObjectNamePrefix", "PayloadFormat")); diff != "" {
```
nachocano: let's do Bucket, ObjectNamePrefix and PayloadFormat also immutable...
Otherwise, if we change the bucket, we will be creating a new notification in that bucket, but leak the previous one... 
The reconciler should actually delete the previous, and create the new one...

I'd rather stay simple now, and make all of those immutable, and then we can improve the reconciler, and relax this condition... What do you think?

nachocano: BTW, would you mind adding this immutability to Scheduler as well? That can be in a separate PR, whatever works

nlopezgi: Thanks for the review. Added immutability checks + UTs for all the requested fields.
Will prepare a separate PR to add immutability to Scheduler. Let me know if there is anything else pending here. Thanks again.

nlopezgi: Created https://github.com/google/knative-gcp/pull/538 for immutability for Scheduler.

PR #542, opened 2020-02-10
Add labels for topics and pullsubscription

### LINKED ISSUE #509: Add labels to GCP resource to organize/identify them
**Problem**
After creating few sources and pubsub channel it becomes difficult to figure out which topic/subscription is created for which source and channels/subscriptions. 
This makes debugging difficult.
Could we add labels to these GCP resource such that users can identify which source or channel or broker the resource belongs to? This will also help someone administering the resource understand that they were created by Knative/CloudRun

End users

nlopezgi:
Fixes #509

## Proposed Changes

- Add labels as part of args for MakeTopic / MakePullSubscription
- Pass the name of the source as one of the labels
- refactor labels to include receive-adapter
- refactor MakeTopic to receive a TopicArgs as in https://github.com/google/knative-gcp/pull/521 for MakePullSubscription

Still have some questions about where the info for channel, if necessary at all, comes from.

**Release Note**

```release-note
Improved observability of which source or channel or broker a resource belongs to.
```

### DIFF
--- a/pkg/reconciler/pubsub/resources/labels.go
+++ b/pkg/reconciler/pubsub/resources/labels.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 Google LLC
+Copyright 2020 Google LLC
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
--- a/pkg/reconciler/pubsub/resources/labels.go
+++ b/pkg/reconciler/pubsub/resources/labels.go
@@ -1,0 +1,24 @@
+/*
+Copyright 2019 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package resources
+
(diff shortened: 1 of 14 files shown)


pkg/reconciler/pubsub/resources/labels.go
```diff
@@ -0,0 +1,24 @@
+/*
+Copyright 2019 Google LLC
```
nachocano: 2020

pkg/reconciler/pubsub/resources/labesl_test.go
```diff
@@ -0,0 +1,35 @@
+/*
+Copyright 2019 Google LLC
```
nachocano: same here

pkg/reconciler/pubsub/resources/topic_test.go
```diff
@@ -134,6 +155,7 @@ func TestMakeTopicWithCloudSchedulerSource(t *testing.T) {
 			Name:      "scheduler-name",
 			Labels: map[string]string{
 				"receive-adapter": "storage.events.cloud.google.com",
```
nachocano: seems an old copy/paste.. should be `scheduler.events.cloud.google.com`?

nachocano:
Looks great! a couple of nits and let's get it in...
Thanks for adding this!

pkg/reconciler/pubsub/resources/labels.go
nlopezgi: done

pkg/reconciler/pubsub/resources/labesl_test.go
nlopezgi: done

pkg/reconciler/pubsub/resources/topic_test.go
nlopezgi: done

nlopezgi:
done, ptal

PR #547, opened 2020-02-12
controller should watch for changes on its secret

### LINKED ISSUE #500: Controller should watch for changes on its secret
**Problem**
The controller should watch for patches of its secret, and "update" it accordingly.

Contributor

**Exit Criteria**
Whenever the controller secret is patched, we should be notified so that the Controller is able to read it again 

**Time Estimate (optional):** 
1d

Needed for initialization purposes

capri-xiyue:
Fixes #500 

## Proposed Changes

- Deployment `controller` of namespace `cloud-run-events` will watch for changes on its secret of `google-cloud-key `of `cloud-run-events`. Once secret `google-cloud-key`of `cloud-run-events` namespace gets changed, we will update the annotation of deployment `controller` in namespace `cloud-run-events` so that the deployment will `controller` will recreate the pods which will pick up the latest secret image.
- 
-

**Release Note**

```release-note

```


pkg/reconciler/deployment/controller.go
```diff
@@ -0,0 +1,78 @@
...
+	"k8s.io/client-go/tools/cache"
+	"knative.dev/pkg/client/injection/kube/informers/apps/v1/deployment"
+	"knative.dev/pkg/client/injection/kube/informers/core/v1/secret"
+	"knative.dev/pkg/configmap"
+	"knative.dev/pkg/controller"
+)
+
+const (
+	// ReconcilerName is the name of the reconciler
+	ReconcilerName = "Deployment" //
```
yolocs: Remove `//` in the end

pkg/reconciler/deployment/deployment.go
```diff
@@ -0,0 +1,94 @@
...
+package deployment
+
+import (
+	"context"
+	"fmt"
+	"github.com/google/knative-gcp/pkg/reconciler"
+	"k8s.io/client-go/tools/cache"
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/logging"
+	"strconv"
```
yolocs: nit: Ordering seems wrong.

pkg/reconciler/deployment/deployment.go
```diff
@@ -0,0 +1,94 @@
...
+
+	v1 "k8s.io/api/apps/v1"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	appsv1listers "k8s.io/client-go/listers/apps/v1"
+)
+
+const (
+	updatedSecretRevisionAnno = "api.v1.namespaces.cloud-run-events.secrets.google-cloud-key/updated-revision"
```
yolocs: Is this some convention we use? The order seems odd.

pkg/reconciler/deployment/deployment.go
```diff
@@ -0,0 +1,94 @@
...
+func (r *Reconciler) reconcile(ctx context.Context, d *v1.Deployment) error {
+	if d.DeletionTimestamp != nil {
+		return nil
+	}
+
+	anno := d.Spec.Template.GetObjectMeta().GetAnnotations()
+	revision, ok := anno[updatedSecretRevisionAnno]
+	if !ok {
+		anno = make(map[string]string)
+		anno[updatedSecretRevisionAnno] = "1"
```
yolocs: Probably better to use the secret generation?

capri-xiyue: I tried to check the generation of secrets with kubectl get google-cloud-key -n cloud-run-events -oyaml, I didn't see any generation info in the secret. Generation is a optional filed and it looks like we don't set the generation for secret.

nachocano:
starting to look at this... give me a couple of hours or so

pkg/reconciler/deployment/deployment.go
capri-xiyue: I'm not sure whether we have any convention for the annotation. The first part before `/`, I just copied the self link of secret which is `/api/v1/namespaces/cloud-run-events/secrets/google-cloud-key`. Do you mean the order of the self link is odd?

pkg/reconciler/deployment/deployment.go
capri-xiyue: I used goland to optimize imports. And It follows the lexicographical order

capri-xiyue: Talked offline. I will regroup the dependency

pkg/reconciler/deployment/deployment.go
yolocs: I see. It looks weird because the self link uses slashes in which case the bigger scope usually comes first (e.g. /google/cloud/knative). When you change the separator to dot, the convention usually is reversed - bigger scope comes after (e.g. knative.cloud.google). I'd suggest something like `google-cloud-key.cloud-run-events` but it's up to you.

pkg/reconciler/deployment/deployment.go
yolocs: You're right! `Revision` is misleading to me because it doesn't point to the revision of the secret. I'd suggest use timestamp instead, e.g. `google-cloud-key.cloud-run-events/lastObservedUpdateTime`.

config/201-clusterrole.yaml
```diff
@@ -110,6 +110,7 @@ rules:
   resources:
     - configmaps
     - pods # For reading termination messages
+    - secrets
```
nachocano: It seems that we can be even more specific here, and say only be able to list google-cloud-key... See here:
https://kubernetes.io/docs/concepts/configuration/secret/#clients-that-use-the-secret-api
So I think it will have to move secrets to a separate block, and add resourceNames to google-cloud-key... https://kubernetes.io/docs/reference/access-authn-authz/rbac/#referring-to-resources... 

Not a strong opinion, but if we see it works, I think it will be better to narrow down the permissions

pkg/reconciler/deployment/controller.go
```diff
@@ -0,0 +1,79 @@
...
+const (
+	// ReconcilerName is the name of the reconciler
+	ReconcilerName = "Deployment"
+
+	// controllerAgentName is the string used by this controller to identify
+	// itself when creating events.
+	controllerAgentName = "cloud-run-events-deployment-controller"
+
+	namespace      = "cloud-run-events"
+	secretName     = "google-cloud-key"
```
nachocano: there is already a constant in pkg/apis/duck/v1alpha1 google_cloud_secret_defaults... It's private, can you rename to DefaultSecretName, and maybe point it here... secretName = DefaultSecretName....
so that we don't mess up if we update it later on

pkg/reconciler/deployment/controller_test.go
```diff
@@ -0,0 +1,42 @@
+/*
+Copyright 2019 Google LLC
```
nachocano: 2020

pkg/reconciler/deployment/deployment.go
```diff
@@ -0,0 +1,96 @@
...
+// Check that our Reconciler implements controller.Reconciler
+var _ controller.Reconciler = (*Reconciler)(nil)
+
+// Reconciler implements controller.Reconciler
+// Reconciler get the deployment and then update the deployment's annotation.
+// Then the deployment will recreate the pods which will pick up the latest secret image immediately. Otherwise we will need to wait for 1 min for the deployment pods to pick up the update secret.
+func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
+	namespace, name, err := cache.SplitMetaNamespaceKey(key)
+	if err != nil {
+		logging.FromContext(ctx).Error("invalid resource key")
```
nachocano: logging.FromContext(ctx).Desugar().Error("Invalid resource key")

pkg/reconciler/deployment/deployment.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil
+	}
+	// Get the deployment resource with this namespace/name
+	original, err := r.deploymentLister.Deployments(namespace).Get(name)
+	if apierrs.IsNotFound(err) {
+		// The resource may no longer exist, in which case we stop processing.
+		logging.FromContext(ctx).Errorf("deployment key in work queue no longer exists %q %q", namespace, name)
```
nachocano: logging.FromContext(ctx).Desugar().Error("Deployment in work queue no longer exists")

the key is already in the context, no need to add it back

pkg/reconciler/deployment/deployment.go
```diff
@@ -0,0 +1,96 @@
...
+		return nil
+	} else if err != nil {
+		logging.FromContext(ctx).Errorf("Error fetching deployment with namespace %q name %q: error", namespace, name, err)
```
nachocano: let's remove this log line to be consistent with the other reconcilers.

pkg/reconciler/deployment/deployment.go
```diff
@@ -0,0 +1,96 @@
...
+	// Reconcile this copy of the Deployment.
+	return r.reconcile(ctx, d)
+}
+
+func (r *Reconciler) reconcile(ctx context.Context, d *v1.Deployment) error {
+	if d.DeletionTimestamp != nil {
+		return nil
+	}
+
+	anno := d.Spec.Template.GetObjectMeta().GetAnnotations()
```
nachocano: let's spell it out: `annotations`

pkg/reconciler/deployment/deployment.go
nachocano: +1

pkg/reconciler/deployment/deployment.go
```diff
@@ -0,0 +1,96 @@
...
+	if !ok {
+		anno = make(map[string]string)
+		anno[updatedSecretRevisionAnno] = "1"
+	} else {
+		i, err := strconv.Atoi(revision)
+		if err != nil {
+			return fmt.Errorf("error converting the %q of annotation %q of deployment of namespace %q, name %q from string to int", revision, updatedSecretRevisionAnno, d.Namespace, d.Name)
+		}
+		anno[updatedSecretRevisionAnno] = strconv.Itoa(i + 1)
+	}
```
nachocano: I think we can do one-liners here...

```
if revision, ok := annotations[SecretAnnotationKey]; !ok {
   
} else if i, err := strconv.Atoi(xxx); err != nil {
     
} else {

}
```

pkg/reconciler/deployment/deployment.go
nachocano: I think we can use `events.cloud.google.com/secret-update` or something like that

nachocano: and maybe call it `SecretUpdateAnnotation`

capri-xiyue:
The integration tests fail because there is something wrong with eventing. The trigger can't be ready even though all underlying resources are ready

nachocano:
/retest

FLAKY:

TestEventTransformationForTrigger/TestEventTransformationForTrigger-{Channel_messaging.cloud.google.com/v1alpha1}

PR #549, opened 2020-02-13
fixed propagate publisher status and fixed the pubsubsecret

capri-xiyue:
Fixes #

## Proposed Changes

- fixed propagate publisher status
- fixed empty pubsubsecret 
-
-

**Release Note**

```release-note

```

### DIFF
--- a/pkg/apis/events/v1alpha1/cloudschedulersource_validation.go
+++ b/pkg/apis/events/v1alpha1/cloudschedulersource_validation.go
@@ -66,9 +66,11 @@ func (current *CloudSchedulerSourceSpec) Validate(ctx context.Context) *apis.Fie
 	}
 
 	if current.PubSubSecret != nil {
-		err := validateSecret(current.PubSubSecret)
-		if err != nil {
-			errs = errs.Also(err.ViaField("pubsubSecret"))
+		if !equality.Semantic.DeepEqual(current.PubSubSecret, &corev1.SecretKeySelector{}) {
+			err := validateSecret(current.PubSubSecret)
+			if err != nil {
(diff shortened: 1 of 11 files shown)


nachocano:
/lgtm
/approve

Thanks for the quick fix!

nachocano:
BTW, have you tried gcloud with storage and scheduler with these fixes and they work?
If so, then we should cherry-pick to 0.12.

capri-xiyue:
> BTW, have you tried gcloud with storage and scheduler with these fixes and they work?
> If so, then we should cherry-pick to 0.12.

The gcloud create trigger works. 
However, when I tried to send event, it doesn't work for all four sources.
I checked the logs of default-broker-ingress, the error logs is as follows:
`{"level":"warn","ts":1581573877.5656204,"logger":"fallback","caller":"http/transport.go:510","msg":"got an error from receiver fn","error":"error sending cloudevent: 500 Internal Server Error"}
{"level":"warn","ts":1581573877.5656753,"logger":"fallback","caller":"http/transport.go:602","msg":"error returned from invokeReceiver","error":"error sending cloudevent: 500 Internal Server Error"}`

I don't think it's related to this PR.  Maybe because I use the latest serving and eventing. Will investigate further tomorrow.

nachocano:
Yeah, that seems something different. We should cherry-pick to 0.12, and
then test with that branch. We can do so tomorrow!

capri-xiyue:
It looks like we also need to cherry-pick this PR  https://github.com/google/knative-gcp/commit/cda1f5c8ed7fd05cd080838223885d0b264567df. My PR is based on it.

nachocano:
OK, that should be fine. And that will help with the debugging as well..

PR #551, opened 2020-02-13
Add KEDA support for GCP Sources

### LINKED ISSUE #423: Add KEDA-scaling for Sources
**Problem**
Our PubSub-based Sources are Pull-based. Given that our receive adapter polls for new messages, it’s hard to decide when to scale it out. In other words, the draining of the PubSub subscriptions will eventually become a throughput bottleneck. 

There are some initial talks in the Sources WG regarding Pull-based sources and how to scale them. @aslom suggested KEDA. Either with KEDA or something else, we will need to read metrics from Stackdriver to see how “congested” the subscription is and inform the scaling mechanism. See details on metrics [here](https://cloud.google.com/pubsub/docs/pull#scaling).

**Exit Criteria**
Being able to scale out our receive adapters.

**Time Estimate (optional):** 
2-3 weeks

nachocano:
Fixes #423

## Proposed Changes

- Using annotations to configure autoscaling, as done in serving. The annotation names might change.
- If you don't specify the keda autoscaling class (currently the only one supported), then the sources will be treated as non-scalable.
- Separated reconciler for Pullsubscription into two. A keda-based one, which understands about KEDA ScaledObject and checks whether KEDA is installed... And a k8s one, which reconciles a plain deployment. 
- Adding a Resource duck type to be able to track changes to KEDA ScaledObjects. Should be upstreamed to pkg. Also leveraging ListableTracker from eventing, should probably upstream it to pkg as well.
- Using Unstructured for ScaledObject. As KEDA still doesn't have client-go support for its types, and advises to use the dynamic client, I consider unnecessary to add the dependency for now. Also, it will bring a bunch of things like controller-runtime, etc.

TODOs:
- Docs (in another PR)
- E2E tests (in another PR)

**Release Note**

```release-note
CloudPubSubSource, CloudSchedulerSource, CloudStorageSource, and CloudAuditLogsSource now support Keda-based scheduling. They can scale all the way down to zero. Note that you will need to have Keda installed. Refer to the installation docs for more details.
```


pkg/reconciler/pubsub/pullsubscription/k8s/controller.go
```diff
@@ -14,28 +14,30 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package pullsubscription
+// TODO better package name?
+package k8s
```
ericlem: fixed?  static?  singleton?  It seems like a name like that may be somewhat more self-describing for what you care about (ie, the scaling strategy).  Don't care a lot either way (though a less conflicting package name would be nice)

pkg/reconciler/pubsub/pullsubscription/k8s/controller.go
```diff
@@ -74,24 +78,37 @@ func NewController(
...
+			FinalizerName:          finalizerName,
+		},
 	}
 
 	impl := controller.NewImpl(r, pubsubBase.Logger, reconcilerName)
 
 	pubsubBase.Logger.Info("Setting up event handlers")
-	pullSubscriptionInformer.Informer().AddEventHandlerWithResyncPeriod(controller.HandleAll(impl.Enqueue), reconciler.DefaultResyncPeriod)
+
+	onlyKedaScaler := pkgreconciler.AnnotationFilterFunc(duckv1alpha1.AutoscalingClassAnnotation, duckv1alpha1.KEDA, false)
```
ericlem: Is there any good way to generalize this and or pull it out into a helper somewhere with generic scaling code?  Not worth a lot of engineering effort now since it's unlikely we'll be adding a third option to pubsub anytime soon, but it would be nice if we didn't have one random piece of code to know about/fix that is hard coded to the existence of only a single other option (keda) rather than a potential family of other options.

pkg/apis/duck/v1alpha1/resource_types.go
```diff
@@ -0,0 +1,77 @@
...
+// Resource is an Implementable "duck type".
+var _ duck.Implementable = (*Resource)(nil)
+
+// +genduck
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+// Resource is a skeleton type wrapping all Kubernetes resources. It is typically used to watch
+// arbitrary other resources. This is not a real resource.
+// TODO upstream to pkg
+type Resource struct {
```
yolocs: Can we use https://github.com/knative/eventing/blob/92aa433f5934c416e56022ce5d2ca33130b37f16/vendor/knative.dev/pkg/apis/duck/v1alpha1/conditions_types.go#L129?

pkg/reconciler/pubsub/pullsubscription/k8s/controller.go
nachocano: Agreed, changed it to static. Seems better. Thanks for the suggestion.

pkg/reconciler/pubsub/pullsubscription/k8s/controller.go
nachocano: Valid point. Given that is unlikely we add a new strategy anytime soon, I just added a comment in the code. If things play out nice with this, I'll try to upstream some pieces to pkg, and revisit this again. Sounds good?

pkg/apis/duck/v1alpha1/resource_types.go
nachocano: The problem is that Keda ScaledObject doesn't have those Conditions in its status. 
Will try to upstream Resource to pkg, and make KResource use it... Resource would allow you to track any k8s object

yolocs: Sounds good. I think that will be useful.

pkg/reconciler/pubsub/pullsubscription/static/pullsubscription.go
```diff
@@ -0,0 +1,57 @@
...
+func (r *Reconciler) Reconcile(ctx context.Context, key string) error {
+	return r.Base.Reconcile(ctx, key)
+}
+
+func (r *Reconciler) ReconcileDeployment(ctx context.Context, ra *appsv1.Deployment, src *v1alpha1.PullSubscription) error {
+	existing, err := r.Base.GetOrCreateReceiveAdapter(ctx, ra, src)
+	if err != nil {
+		return err
+	}
+	if !equality.Semantic.DeepDerivative(ra.Spec, existing.Spec) {
```
capri-xiyue: Based on https://godoc.org/k8s.io/apimachinery/third_party/forked/golang/reflect#Equalities.DeepDerivative,
`DeepDerivative is similar to DeepEqual except that unset fields in a1 are ignored (not compared). This allows us to focus on the fields that matter to the semantic comparison.The unset fields include a nil pointer and an empty string.`. Should we use DeepEqual here, we should also check unset fields, right?

pkg/reconciler/pubsub/pullsubscription/k8s/controller.go
ericlem: I'm ok with that for now.

pkg/reconciler/pubsub/pullsubscription/static/pullsubscription.go
nachocano: Done, thanks!

PR #573, opened 2020-02-27
Fail fast for e2e test and performance test

chizhg:
## Proposed Changes
- If there is any error in setups for knative-gcp, the script should fail directly instead of continuing on running the tests.

/cc @grac3gao


chizhg:
/assign @nachocano 
for your approval power.

Issue #486, opened 2020-01-15
E2E tests for sending one event to Channel is flaky


chizhg:
**Describe the bug**
When we run the E2E test in eventing for Channel, `TestSingleBinaryEventForChannel` and `TestSingleStructuredEventForChannel` are flaky.

The error message when it happens:

channel_single_event_helper.go:74: Failed to send fake CloudEvent to the channel "e2e-singleevent-channel-binary": addressable's URL does not have a Host: &{TypeMeta:{Kind:Channel APIVersion:messaging.cloud.google.com/v1alpha1} ObjectMeta:{Name:e2e-singleevent-channel-binary GenerateName: Namespace:v1alpha1-t8bvk SelfLink:/apis/messaging.cloud.google.com/v1alpha1/namespaces/v1alpha1-t8bvk/channels/e2e-singleevent-channel-binary UID:5d955bfc-27bc-4a3b-96ac-b745a2c71c54 ResourceVersion:5634 Generation:2 CreationTimestamp:2020-01-15 20:15:42 +0000 UTC DeletionTimestamp:<nil> DeletionGracePeriodSeconds:<nil> Labels:map[] Annotations:map[] OwnerReferences:[] Initializers:nil Finalizers:[] ClusterName: ManagedFields:[]} Status:{Address:0xc001a6dd80}}

 **Expected behavior**
The two tests should be able to consistently pass.

nachocano:
@chizhg any updates on this? Do you know if we can enable them? 
Was the one with the default-ch-webhook configMap problem, that wasn't being updated?

BTW, that one got fixed?

chizhg:
I don't think they can be enabled as there was no change related to them. I can revisit them again today to see what's the exact reason for the failure.

The default-ch-webhook is a different issue and it hasn't been fixed. There are only two ways to fix it, one is changing the resync time for the informer, the other is recreating the webhook Pod in the tests.
I'm not sure which one is more preferable to you.

Harwayne:
Switching release as Channel is not in scope for `release/2`.

PR #590, opened 2020-03-03
Re-enabling channel tests

nachocano:
Fixes https://github.com/google/knative-gcp/issues/486

## Proposed Changes

- attempt to see if these have been fixed already.

**Release Note**

```release-note
NONE
```


nachocano:
@chizhg BTW, the channel defaulter singleton is gone, so those tests might be fixed as well...
Have you tried enabling them? I can give it a try here as well
https://github.com/knative/eventing-contrib/issues/627

chizhg:
Hmm, seems still flaky?

chizhg:
> @chizhg BTW, the channel defaulter singleton is gone, so those tests might be fixed as well...
> Have you tried enabling them? I can give it a try here as well
> [knative/eventing-contrib#627](https://github.com/knative/eventing-contrib/issues/627)

After this change, does it read the config from the configmap every time we create a new default channel? If not, I think these tests still cannot work..

nachocano:
> Hmm, seems still flaky?

yeah

/test pull-google-knative-gcp-integration-tests

nachocano:
> k these tests still cannot work..

It reads it from a "store" object that is attached to the context. It seems that it doesn't work.. Created a separate one for that...

chizhg:
> > Hmm, seems still flaky?
> 
> yeah
> 
> /test pull-google-knative-gcp-integration-tests

Some of the tests are also flaky when we run them against KafkaChannel, and Ville is trying to find out the root cause in https://github.com/knative/eventing-contrib/pull/920. Is it possible they are the same reason?

nachocano:
> > > Hmm, seems still flaky?
> > 
> > 
> > yeah
> > /test pull-google-knative-gcp-integration-tests
> 
> Some of the tests are also flaky when we run them against KafkaChannel, and Ville is trying to find out the root cause in [knative/eventing-contrib#920](https://github.com/knative/eventing-contrib/pull/920). Is it possible they are the same reason?

hmm not sure, these always fail though, they are not flaky AFAIK... or it seems that the parallelism has something to do?

chizhg:
Seems it does work for `parallel=1` 😮

nachocano:
yes, it's the same thing as kafka then...
the parallel thing just executes other tests in parallel right? although
all should be in different namespaces, right?
there is a concurrency problem somewhere...
Do you know if IMC and Natss work well with 12?

chizhg:
Yes, IMC and Natss works for parallel=12

nachocano:
> Yes, IMC and Natss works for parallel=12 On Mon, Mar 2, 2020, 9:45 PM Ignacio (Nacho) Cano <notifications@github.com> wrote:

hmm, ok there must be some weird stuff that we are doing in kafka and pubsub then. Will try to debug further tomorrow...

PR #599, opened 2020-03-04
Update the release script to use ARTIFACTS_TO_PUBLISH

chizhg:
## Proposed Changes
- The same change as https://github.com/google/knative-gcp/pull/598, but on master.

/cc @nachocano 

/hold
Let's merge it till we proves it works for release-0.13.


nachocano:
Sounds good

PR #600, opened 2020-03-04
Start knative gcp after creating secrets for perf tests

chizhg:
## Proposed Changes
- The change in https://github.com/google/knative-gcp/pull/573 broke the performance tests setup - the controller was unable to start due to missing google-cloud-key secret. We should create the secrets first, then start Knative GCP.

/cc @nachocano 
/cc @grac3gao


nachocano:
/lgtm
/approve

this is the problem that I think @nlopezgi was experiencing

Issue #557, opened 2020-02-19
Move to the GenReconciler 


nachocano:
**Problem**
Leverage the new genreconciler funcionality enabled in knative/pkg to reduce boilerplate reconciler code.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Contributor

**Exit Criteria**
Using the new genReconciler framework

**Time Estimate (optional):** 
1w

**Additional context (optional)**
This is a sensitive change that should be done in stages (one reconciler at a time). 
Preferably once we have all our e2e tests in place https://github.com/google/knative-gcp/issues/511.

nachocano:
Some examples here:

https://github.com/knative/eventing/pull/2540
https://github.com/knative/eventing/pull/2541
https://github.com/knative/eventing/pull/2543

Things to watch out for is the finalizers part. I think there is a new FinalizeKind method. There is an example in the Subscription controller.
Also some of the warning events that we were emitting won't be emitted anymore. We will have to update multiple UTs.
In order of priority I think I would do:
- Sources first (all the ones within the events package).  Can be done in multiple PRs.
- Channel 
- Topic
- and finally PullSubscription (this might require some more work as I added a base thing for KEDA or non-KEDA)

nachocano:
This was fixed by all the referenced PRs.
Thanks @nlopezgi !!!

PR #605, opened 2020-03-05
Migrate PubSubSource to use generated reconciler. 

nlopezgi:
1st part of # https://github.com/google/knative-gcp/issues/557

## Proposed Changes

- 🎁 Use reconciler gen to implement strongly typed reconcilers.

Issue #186, opened 2019-08-16
Storage creates multiple Notifications


Harwayne:
Occasionally I see that Storage creates two GCS notifications. Then, when the Storage CO is deleted, only the more recent notifications is deleted. This caused me to hit the limit of 10 notifications on a bucket.

```shell
$ gsutil notification list gs://eventing-demo
# Nothing

$ kn importers create storage desjani --parameters bucket=eventing-demo
# Finishes when the importer is ready.

$ gsutil notification list gs://eventing-demo
projects/_/buckets/eventing-demo/notificationConfigs/67
        Cloud Pub/Sub topic: projects/harwayne2/topics/storage-e711b0c4-f647-4723-be61-99ce59801287
        Custom attributes:
                knative-gcp: google.storage

projects/_/buckets/eventing-demo/notificationConfigs/68
        Cloud Pub/Sub topic: projects/harwayne2/topics/storage-b7cbfa63-a7ba-4f8f-8306-efc574daa7e4
        Custom attributes:
                knative-gcp: google.storage

$ ./kn importers generic delete storage desjani  
# Finishes successfully.

$ gsutil notification list gs://eventing-demo
projects/_/buckets/eventing-demo/notificationConfigs/67
        Cloud Pub/Sub topic: projects/harwayne2/topics/storage-e711b0c4-f647-4723-be61-99ce59801287
        Custom attributes:
                knative-gcp: google.storage
```

n3wscott:
@Harwayne can you try this flow again with the operations ?

vaikas-google:
as @n3wscott mentions, this should not happen again. The problem was that occasionally depending on which write (finalizer or the status update with the notificationId in Status) would go first we would indeed end up getting the duplicate notification. With the operations model, creating the subscription should be idempotent because the state is stored in the Job that creates the Subscription and not in the reconciler.

vaikas-google:
@Harwayne can you still repro this?

Harwayne:
Looking through the code, I think it should be much safer now. 

However, it is still possible if the Pod creating the notification dies at the wrong time. If the Pod dies after https://github.com/google/knative-gcp/blob/master/pkg/operations/storage/notification.go#L228, but before https://github.com/google/knative-gcp/blob/master/pkg/operations/storage/notification.go#L243, or writing the termination message fails for some reason, then we would still leak a Notification.

[...]

PR #607, opened 2020-03-05
Migrate StorageSource to use generated reconciler.

nlopezgi:
2nd PR part of # https://github.com/google/knative-gcp/issues/557

## Proposed Changes


pkg/reconciler/events/storage/controller.go
```diff
@@ -60,19 +61,24 @@ func NewController(
...
+	impl := cloudstoragesourcereconciler.NewImpl(ctx, r)
 
 	r.Logger.Info("Setting up event handlers")
 	cloudstoragesourceInformer.Informer().AddEventHandlerWithResyncPeriod(controller.HandleAll(impl.Enqueue), reconciler.DefaultResyncPeriod)
 
+	// Call GlobalResync on pubsubsource.
+	grCh := func(obj interface{}) {
+		impl.GlobalResync(cloudstoragesourceInformer.Informer())
+	}
+
```
nachocano: same with the other PR, no need for this

pkg/reconciler/events/storage/storage.go
```diff
@@ -345,52 +244,18 @@ func (r *Reconciler) updateStatus(ctx context.Context, original *v1alpha1.CloudS
...
-			"finalizers":      finalizers,
-			"resourceVersion": existing.ResourceVersion,
-		},
+func (r *Reconciler) FinalizeKind(ctx context.Context, storage *v1alpha1.CloudStorageSource) reconciler.Event {
+	logging.FromContext(ctx).Desugar().Debug("Deleting CloudStorageSource notification")
+	if err := r.deleteNotification(ctx, storage); err != nil {
+		storage.Status.MarkNotificationNotReady("NotificationDeleteFailed", "Failed to delete CloudStorageSource notification: %s", err.Error())
+		return reconciler.NewEvent(corev1.EventTypeWarning, "NotificationDeleteFailed", "Failed to delete CloudStorageSource notification: %s", err)
 	}
+	storage.Status.MarkNotificationNotReady("NotificationDeleted", "Successfully deleted CloudStorageSource notification: %s", storage.Status.NotificationID)
```
nachocano: I think we can actually remove this line... We shouldn't be updating the status of the object when is being deleted, otherwise the update status operation will fail...
If there is an error while trying to Delete, then we should, as in the previous if...

pkg/reconciler/events/storage/storage.go
```diff
@@ -345,52 +244,18 @@ func (r *Reconciler) updateStatus(ctx context.Context, original *v1alpha1.CloudS
...
-
-	mergePatch := map[string]interface{}{
-		"metadata": map[string]interface{}{
-			"finalizers":      finalizers,
-			"resourceVersion": existing.ResourceVersion,
-		},
+func (r *Reconciler) FinalizeKind(ctx context.Context, storage *v1alpha1.CloudStorageSource) reconciler.Event {
+	logging.FromContext(ctx).Desugar().Debug("Deleting CloudStorageSource notification")
+	if err := r.deleteNotification(ctx, storage); err != nil {
+		storage.Status.MarkNotificationNotReady("NotificationDeleteFailed", "Failed to delete CloudStorageSource notification: %s", err.Error())
```
nachocano: same as below... I think this MarkNotificationNotReady is not needed

pkg/reconciler/events/storage/storage.go
```diff
@@ -76,117 +75,29 @@ type Reconciler struct {
...
-	// Ensure that there's finalizer there, since we're about to attempt to
-	// change external state with the topic, so we need to clean it up.
-	addFinalizer(storage)
+	storage.Status.ObservedGeneration = storage.Generation
 
 	topic := resources.GenerateTopicName(storage)
 	_, _, err := r.PubSubBase.ReconcilePubSub(ctx, storage, topic, resourceGroup)
 	if err != nil {
-		return err
+		return reconciler.NewEvent(corev1.EventTypeWarning, "ReconcilePubSubFailed", "Failed to reconcile CloudStorageSource PubSub: %s", err)
```
nachocano: nit to maintain consistency with others:
ReconcilePubSubFailed -> PubSubReconcileFailed

pkg/reconciler/events/storage/storage.go
nachocano: this will apply for most of the other Sources as well...

nachocano:
In this case, it'd be great if you can also test that when deleting the CloudStorageSource object, we are properly removing the notification from GCS. 

I think you can with gsutil. See here: https://cloud.google.com/storage/docs/gsutil/commands/notification.

You can check after instantiating that we do create a notification, and then when deleting, that we are actually removing it (as well as the pubsub topic and subscription)... 
Just to make sure we are not leaking things

pkg/reconciler/events/storage/storage.go
```diff
@@ -63,6 +56,12 @@ var (
 	}
 )
 
+// newReconciledNormal makes a new reconciler event with event type Normal, and
+// reason StorageSourceReconciled.
+func newReconciledNormal(namespace, name string) reconciler.Event {
+	return reconciler.NewEvent(corev1.EventTypeNormal, "Updated", "Updated CloudStorageSource: \"%s/%s\"", namespace, name)
```
yolocs: tip: use ` to avoid escapes.

pkg/reconciler/events/storage/controller.go
nlopezgi: done

pkg/reconciler/events/storage/storage.go
nlopezgi: done

pkg/reconciler/events/storage/storage.go
nlopezgi: done

pkg/reconciler/events/storage/storage.go
nlopezgi: done

pkg/reconciler/events/storage/storage.go
nlopezgi: done

nlopezgi:
> In this case, it'd be great if you can also test that when deleting the CloudStorageSource object, we are properly removing the notification from GCS.
> 
> I think you can with gsutil. See here: https://cloud.google.com/storage/docs/gsutil/commands/notification.
> 
> You can check after instantiating that we do create a notification, and then when deleting, that we are actually removing it (as well as the pubsub topic and subscription)...
> Just to make sure we are not leaking things

Do you mean to create some unit tests that does this check or are you referring to some manual test? In either case, I'm a bit lost as to how to perform the test that you want me to build, could you provide some more pointers?
Addressed all other review comments,
Thanks, ptal.

pkg/reconciler/events/storage/storage.go
```diff
@@ -43,7 +43,12 @@ import (
 )
 
 const (
-	resourceGroup = "cloudstoragesources.events.cloud.google.com"
```
nachocano: same comment as other PR. Let's group the reasons together, and let finalizerName and resourceGroup be separate

pkg/reconciler/events/storage/storage.go
```diff
@@ -82,22 +81,22 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, storage *v1alpha1.CloudS
 	ctx = logging.WithLogger(ctx, r.Logger.With(zap.Any("storage", storage)))
 
 	storage.Status.InitializeConditions()
-	storage.Status.ObservedGeneration = storage.Generation
```
nachocano: as with other review. I think this is correct and we should add it back

pkg/reconciler/events/storage/storage.go
```diff
@@ -82,22 +81,22 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, storage *v1alpha1.CloudS
...
 
 	notification, err := r.reconcileNotification(ctx, storage)
 	if err != nil {
-		storage.Status.MarkNotificationNotReady("NotificationReconcileFailed", "Failed to reconcile CloudStorageSource notification: %s", err.Error())
-		return reconciler.NewEvent(corev1.EventTypeWarning, "NotificationReconcileFailed", "Failed to reconcile CloudStorageSource notification: %s", err)
+		storage.Status.MarkNotificationNotReady(reconciledNotificationFailed, "Failed to reconcile CloudStorageSource notification: %s", err.Error())
+		return reconciler.NewEvent(corev1.EventTypeWarning, reconciledNotificationFailed, "Failed to reconcile CloudStorageSource notification: %s", err)
 	}
 	storage.Status.MarkNotificationReady(notification)
+	storage.Status.ObservedGeneration = storage.Generation
```
nachocano: remove

pkg/reconciler/events/storage/storage_test.go
```diff
@@ -195,7 +194,7 @@ func TestAllCases(t *testing.T) {
 		},
 		WantEvents: []string{
 			Eventf(corev1.EventTypeNormal, "FinalizerUpdate", "Updated %q finalizers", storageName),
-			Eventf(corev1.EventTypeWarning, "ReconcilePubSubFailed", fmt.Sprintf("%s: Topic %q has not yet been reconciled", failedToReconcilePubSubMsg, storageName)),
+			Eventf(corev1.EventTypeWarning, "PubSubReconcileFailed", fmt.Sprintf("%s: Topic %q has not yet been reconciled", failedToReconcilePubSubMsg, storageName)),
```
nachocano: can use the consts, here and elsewhere

nachocano:
> a bit lost as to how to perform the test that you want me to build, could you provide some more pointers?
> Addressed all other review comments,

It's just a manual test (no unit nor e2e). Just to quickly check if the `FinalizeKind` callback is properly being called and we are not leaking resources... It's optional, no need to do it.
In case you want to do it, the way is just to create a `CloudStorageSource` pointing to some bucket. With`gsutil` you can list notifications in that bucket... You should be able to see one, which was created by the reconcilers... Then you delete the source, and you query the notifications with gsutil again, and there shouldn't be any... We used to have a bug that we were leaking notifications (https://github.com/google/knative-gcp/issues/186), I couldn't repro last time but I don't want to regress it...

pkg/reconciler/events/storage/storage.go
nlopezgi: If I remove it from here and move it up top as you suggest I end up having to modify lots of failure test cases that do not pass as is. Specifically, I end up having to add lots of:
`WithCloudStorageSourceStatusObservedGeneration(generation),`
to these failure test cases. I moved this as suggested in https://github.com/google/knative-gcp/pull/605#discussion_r389199573. Lets talk with @ian-mi to see what the right place for this should be.

pkg/reconciler/events/storage/storage.go
nlopezgi: see comment below.

pkg/reconciler/events/storage/storage_test.go
nlopezgi: done

nlopezgi:
> > a bit lost as to how to perform the test that you want me to build, could you provide some more pointers?
> > Addressed all other review comments,
> 
> It's just a manual test (no unit nor e2e). Just to quickly check if the `FinalizeKind` callback is properly being called and we are not leaking resources... It's optional, no need to do it.
> In case you want to do it, the way is just to create a `CloudStorageSource` pointing to some bucket. With`gsutil` you can list notifications in that bucket... You should be able to see one, which was created by the reconcilers... Then you delete the source, and you query the notifications with gsutil again, and there shouldn't be any... We used to have a bug that we were leaking notifications (#186), I couldn't repro last time but I don't want to regress it...

Thanks for the clarification, will run manual test and report back if I have issues.

pkg/reconciler/events/storage/storage.go
nlopezgi: done

pkg/reconciler/events/storage/storage.go
nlopezgi: done

pkg/reconciler/events/storage/storage.go
```diff
@@ -43,12 +43,13 @@ import (
 )
 
 const (
-	reconciledSuccessReason      = "CloudStorageSourceReconciled"
-	reconciledNotificationFailed = "NotificationReconcileFailed"
-	reconciledPubSubFailed       = "PubSubReconcileFailed"
+	resourceGroup = "cloudstoragesources.events.cloud.google.com"
```
nachocano: this made me wonder if we needed the finalizerName in the other PR. Seems that we don't anymore. 
If you remember, when you create the next PR for a new reconciler, try to remove that finalizerName from the CloudPubSubSource. Otherwise no big deal, we can clean those up later on...
No need to do anything on this PR

nlopezgi: will do, sorry I did not notice that was no longer needed.

nachocano:
we used to have an updateStatus method IIRC, that one is not needed anymore right?
Can you remove it if not... Missed that from the previous PR, we used to have it as well there. Can you remove it if we are not using it anymore?

pkg/reconciler/events/storage/storage.go
```diff
@@ -255,5 +256,5 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, storage *v1alpha1.CloudSt
 	}
 
 	// ok to remove finalizer.
-	return nil
+	return reconciler.NewEvent(corev1.EventTypeNormal, deleteNotificationSuccess, "Successfully deleted CloudStorageSource notification: %s", storage.Status.NotificationID)
```
nachocano: I'd rather leave this one as before, just return nil... The object will be deleted after this, so IMO no need to emit an deleteNotificationSuccess... We do emit, when there was a failure deleting, so that users can just describe the object and see what happened...
I think if we care about deletion events, we can use the ApiServerSource, listening on `CloudStorageSource` objects delete, and that should suffice... What do you think?

nlopezgi: done

nachocano:
/lgtm
/approve

As part of the scheduler PR, can you check and remove the following if not needed anymore?

we used to have an updateStatus method IIRC, that one is not needed anymore right?
Can you remove it if not... Missed that from the previous PR, we used to have it there as well. Can you remove it if we are not using it anymore?

It should probably be removed for all the stuff that has been migrated already (CloudPubSubSource and this one)

nlopezgi:
> /lgtm
> /approve
> 
> As part of the scheduler PR, can you check and remove the following if not needed anymore?
> 
> we used to have an updateStatus method IIRC, that one is not needed anymore right?
> Can you remove it if not... Missed that from the previous PR, we used to have it there as well. Can you remove it if we are not using it anymore?
> 
> It should probably be removed for all the stuff that has been migrated already (CloudPubSubSource and this one)

Removed it from this PR as that was something I missed here but had indeed removed in the CloudPubSubSource PR.

nachocano:
> > /lgtm
> > /approve
> > As part of the scheduler PR, can you check and remove the following if not needed anymore?
> > we used to have an updateStatus method IIRC, that one is not needed anymore right?
> > Can you remove it if not... Missed that from the previous PR, we used to have it there as well. Can you remove it if we are not using it anymore?
> > It should probably be removed for all the stuff that has been migrated already (CloudPubSubSource and this one)
> 
> Removed it from this PR as that was something I missed here but had indeed removed in the CloudPubSubSource PR.

Perfect! Thanks!

/lgtm

PR #623, opened 2020-03-10
Migrate Channel to use generated reconciler.

nlopezgi:
5th PR part of # https://github.com/google/knative-gcp/issues/557

## Proposed Changes


pkg/reconciler/messaging/channel/channel.go
```diff
@@ -53,77 +58,20 @@ type Reconciler struct {
...
-		return nil
-	}
+	channel.Status.ObservedGeneration = channel.Generation
 
 	// 1. Create the Topic.
 	topic, err := r.reconcileTopic(ctx, channel)
 	if err != nil {
 		channel.Status.MarkTopicFailed("TopicReconcileFailed", "Failed to reconcile Topic: %s", err.Error())
-		return err
+		return pkgreconciler.NewEvent(corev1.EventTypeWarning, reconciledTopicFailedReason, "Reconcile Topic failed with: %s", err.Error())
```
yolocs: nit: usually I just use `%v` for `err`.
But also wonder if you have any special reason of doing what're doing now.

nlopezgi: thanks, no specific reason why I'm using this. I'll make one final pass to fix all of these, once all PRs are in, to leave them all in a consistent state

PR #624, opened 2020-03-10
Migrate Topic to use generated reconciler

nlopezgi:
6th PR part of # https://github.com/google/knative-gcp/issues/557

## Proposed Changes


pkg/reconciler/pubsub/topic/topic.go
```diff
@@ -444,3 +250,14 @@ func (r *Reconciler) UpdateFromTracingConfigMap(cfg *corev1.ConfigMap) {
...
 	// TODO: requeue all Topics. See https://github.com/google/knative-gcp/issues/457.
 }
+
+func (r *Reconciler) FinalizeKind(ctx context.Context, topic *v1alpha1.Topic) reconciler.Event {
+	if topic.Spec.PropagationPolicy == v1alpha1.TopicPolicyCreateDelete {
+		logging.FromContext(ctx).Desugar().Debug("Deleting Pub/Sub topic")
+		if err := r.deleteTopic(ctx, topic); err != nil {
+			return reconciler.NewEvent(corev1.EventTypeWarning, deleteTopicFailed, "Failed to delete Pub/Sub topic: %s", err.Error())
+		}
+		topic.Status.TopicID = ""
```
nachocano: I think this can be removed, because it will try to update the status otherwise, and it's not needed when it it will be deleted...

pkg/reconciler/pubsub/topic/topic.go
```diff
@@ -413,21 +234,6 @@ func (r *Reconciler) reconcilePublisher(ctx context.Context, topic *v1alpha1.Top
 	return nil, svc
 }
 
-func (r *Reconciler) getPublisher(ctx context.Context, topic *v1alpha1.Topic) (*serving.Service, error) {
```
nachocano: aren't we using this?

nachocano: seems we are not, we just do everything from the reconcilePublisher, discard the comment

pkg/reconciler/pubsub/topic/topic.go
nachocano: @grac3gao I'll get this one in, can you address this cosmetic comment in your PR once you rebase?

Issue #544, opened 2020-02-11
Add Documentation for CE PubSub binding


nachocano:
**Problem**
As with other [bindings](https://github.com/cloudevents/spec), we need to define the PubSub binding for CloudEvents.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
The CE PubSub binding pushed into this repo

**Time Estimate (optional):** 
2d

**Additional context (optional)**
See https://github.com/google/knative-gcp/pull/1

PR #633, opened 2020-03-12
Adding PubSub v1 spec

nachocano:
Fixes https://github.com/google/knative-gcp/issues/544

## Proposed Changes

- 🎁Added the v1 PubSub binding

**Release Note**

```release-note
CloudEvents PubSub binding spec is now available
```


nachocano:
/assign @grantr 

per CE folks [feedback](https://github.com/cloudevents/spec/pull/587#issuecomment-598234431), we should add it to our repo and then link it from there...

nachocano:
@grantr ready for another look

nachocano:
fyi @n3wscott

docs/spec/pubsub-protocol-binding.md
```diff
@@ -153,7 +156,7 @@ Examples:
     * `id` maps to `ce-id`
     * `specversion` maps to `ce-specversion`
 
-Attribute names are case-insensitive.
+Attribute names are lower-case.
```
grantr: If all attribute names are lower case, it seems a bit odd to use `contentType` as the only attribute with an uppercase character.

nachocano:
🤦🏻‍♂️ My bad, let's do content-type then, which seems to match binary
better

grantr:
The `contentType` change came from https://github.com/cloudevents/spec/pull/587#discussion_r391700876, where it was suggested alongside `content-type`. I think `content-type` makes sense.

nachocano:
fyi @nlopezgi for when you start with the sdk refactor. Some things changed, such as `Content-Type` for `content-type` and lower-casing. We can sync later on and I can show you where that is in the code.

^^^ @hrasadi -> this is the latest version so far

nachocano:
/hold cancel
let's get this one in! need one more lgtm @grantr

docs/spec/pubsub-protocol-binding.md
nachocano: changed it!

Issue #575, opened 2020-02-27
Knative policy API definition


yolocs:
**Problem**
Add Knative policy API as proposed in https://docs.google.com/document/d/1Nymrlp__dXKsp02Dpif0GN_Pskm6YXv6WjeF8trYzzk/edit#
It will be experimental initially.

Tasks

- [x] HTTP policy API
- [x] Event policy API
- [x] HTTP policy binding API
- [x] Event policy binding API
- [ ] Mutation check

**Exit Criteria**
Policy API added.

PR #659, opened 2020-03-16
Update policy binding validation and immutable fields check

yolocs:
Fixes https://github.com/google/knative-gcp/issues/575

## Changes

- Make policy binding spec immutable. This is needed because otherwise the controller cannot properly cleanup resources from previous subject/policy.
- Better subject validation. A policy binding subject can either be a concrete object reference or label selector. If specified with label selector, then the label selector will directly be used to select workload to bind the policy.

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/apis/security/v1alpha1/common_validation.go
+++ b/pkg/apis/security/v1alpha1/common_validation.go
@@ -20,7 +20,13 @@ import (
 	"context"
 
 	"knative.dev/pkg/apis"
+	"knative.dev/pkg/kmp"
 )
+
+var subjectRefErr = &apis.FieldError{
+	Paths:   []string{"apiVersion", "kind", "name"},
+	Message: `"apiVersion", "kind" and "name" must be specified altogether`,
+}
 
 // Validate validates a StringMatch.
 func (m *StringMatch) Validate(ctx context.Context) *apis.FieldError {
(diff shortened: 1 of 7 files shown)


pkg/apis/security/v1alpha1/common_validation.go
```diff
@@ -112,9 +118,19 @@ func (pbs *PolicyBindingSpec) Validate(ctx context.Context, parentNamespace stri
 	if pbs.Subject.Namespace != parentNamespace {
 		errs = errs.Also(apis.ErrInvalidValue(pbs.Subject.Namespace, "namespace").ViaField("subject"))
 	}
+	// Specify subject either by an object reference or by label selector.
+	// Cannot be both.
+	if pbs.Subject.APIVersion != "" || pbs.Subject.Kind != "" || pbs.Subject.Name != "" {
+		if pbs.Subject.APIVersion == "" || pbs.Subject.Kind == "" || pbs.Subject.Name == "" {
```
nachocano: this confuses me a bit, don't find it very readable. Anyway we can revisit this logic? 
Maybe some inspiration: https://github.com/knative/eventing/blob/8ce1b8084c9055ae24c7b0d8ca22f99d6907558d/pkg/apis/messaging/v1beta1/subscribable_channelable_validation.go#L37

nachocano:
/lgtm
/approve

Thanks for adding this checks @yolocs!!!

Issue #660, opened 2020-03-16
Broker ConfigMap data and common utility to read it


yolocs:
**Problem**

From https://github.com/google/knative-gcp/issues/654, we have agreed that we want to use ConfigMap for tracking triggers and likely for other broker configs.

- We need to decide what data and how to store them in ConfigMap.
- It's likely the same ConfigMap will be loaded in multiple broker workloads and the controller. We should have a common lib to manipulate the ConfigMap so that those components can focus on implementing the broker business logic.

**Time Estimate (optional):** 
1w

**Additional**

During my previous prototype, I found https://github.com/micro/go-micro/tree/master/config pretty useful. If the license permits, we can consider using it.

yolocs:
Once the *contract* of the config is settled, we can parallel the work:

1. Controller (https://github.com/google/knative-gcp/issues/635) - reconciles Pubsub resources and update the config
2. Fanout (https://github.com/google/knative-gcp/issues/637) - it loads the config and reconciles itself
3. Retry (https://github.com/google/knative-gcp/issues/638) - it loads the config and reconciles itself

grantr:
Go micro is apache licensed, so probably fine: https://github.com/micro/go-micro/blob/master/LICENSE
I'm fine with that, but if it doesn't work out an alternative with similar features is viper, by the author of cobra: https://github.com/spf13/viper (MIT license).

I haven't thought about it too much, but maybe it would make sense to have a single ConfigMap with one key (file) per handler (ingress, fanout, retry). A single ConfigMap allows the controller to transactionally update the configs for all handlers at once, plus makes it slightly easier to run all workloads in the same pod (since there's only one volume mount). Separate files allows each handler to focus only on the configs that are relevant to its function.

yolocs:
I was thinking have a single config file. I will put several options in this issue soon so that we can vote.

grantr:
Something to keep in mind is that we want to minimize delay between when the Trigger is created and the data plane is ready to service the Trigger.

A separate but related issue is that there's no obvious way for the controller to know when the Trigger has processed an updated configmap. But this is a general issue for all data plane components so we don't have to solve it here.

yolocs:
Some initial thoughts on the config.

```
type TargetsConfig struct {
	// Namespace => name => target.
	Targets map[string]map[string]Target
}

type Target struct {
	// UID of the trigger resource.
	// Can be used for naming topic and subscription.
	ID        string
	Namespace string
	Name      string
[...]

PR #681, opened 2020-03-20
Add broker config type and utils to manipulate it

yolocs:
Fixes https://github.com/google/knative-gcp/issues/660

## Proposed Changes

- Defined broker config with protobuf
- Used fsnotify to watch config file changes and refresh memory cache

**Release Note**

```release-note

```

**Docs**


yolocs:
@grantr @ericlem @liu-cong @grac3gao 

This PR is based on the idea from the referenced issue. Please take a look and let me know what you think.

pkg/broker/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+// Bytes serializes all the targets.
+func (bt *BaseTargets) Bytes() ([]byte, error) {
+	cfg := bt.Internal.Load().(*TargetsConfig)
+	return proto.Marshal(cfg)
+}
+
+// String returns the text format of all the targets.
+func (bt *BaseTargets) String() string {
+	cfg := bt.Internal.Load().(*TargetsConfig)
+	fmt.Println(cfg.GetNamespaces()["ns1"].GetNames()["name1"].State)
```
ericlem: Is this just debug (the fmt.Print inside String())?

pkg/broker/config/volume/volume.go
```diff
@@ -0,0 +1,147 @@
...
+	"github.com/fsnotify/fsnotify"
+	"github.com/golang/protobuf/proto"
+	"github.com/google/knative-gcp/pkg/broker/config"
+)
+
+const (
+	defaultPath = "/var/run/cloud-run-events/broker/targets"
+)
+
+// Targets implementas config.ReadOnlyTargets with data
```
ericlem: nit: implements

pkg/broker/config/config.go
```diff
@@ -0,0 +1,122 @@
...
+	EqualsString(string) bool
+}
+
+// Targets provides "read" and "write" functions for broker targets.
+type Targets interface {
+	ReadOnlyTargets
+
+	// Union adds the given targets.
+	Union(...Target) Targets
+	// Except removes the give targets.
```
grantr: This (and Union) should document whether it returns a new set or modifies the set in place. I think it's the latter, looking at the code.

Also, check out https://pkg.go.dev/k8s.io/apimachinery/pkg/util/sets for an existing set interface. It might be a good idea to be consistent. In the generated k8s sets, `Insert` adds items and `Delete` removes items. `Union` is a method for merging two sets.

pkg/broker/config/volume/volume.go
```diff
@@ -0,0 +1,147 @@
...
+				if ok {
+					log.Printf("watcher error: %v\n", err)
+				}
+				done <- struct{}{}
+				return
+			}
+		}
+	}()
+
+	watcher.Add(configDir)
```
ericlem: Nitpick.  It feels like if this watcher.Add() was moved above the go func() above, the defer watcher.Close() could be moved into the go func() above as a simple call to watcher.Close() (after the for loop), and this whole function could be run synchronously by it's caller get rid of initialized and done (since as far as i can tell done is just used to wake up to run the defer).

ericlem: actually, rather than a simple call to watcher.Close() after the for loop, just a defer water.Close() inside that function.  If you did a call you'd have to use named breaks instead of returns

ericlem:
Overall looks good.  Not sure if you meant to keep that fmt.Println() in the code.  I also added some comments on how i think you could simplify the watching goroutine pattern, but it works as is and isn't that complicated so I'm also fine with it going in as is if you want to get it in quickly for sharing.
/lgtm

pkg/broker/config/config.go
yolocs: Yes. Forgot to remove.

pkg/broker/config/volume/volume.go
yolocs: Love it. I need more such comments :)

pkg/broker/config/volume/volume.go
```diff
@@ -0,0 +1,147 @@
...
+	initialized <- struct{}{}
```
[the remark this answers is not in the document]
yolocs: Used what Eric proposed and removed these channels.

pkg/broker/config/volume/volume.go
```diff
@@ -72,27 +72,23 @@ func NewTargetsFromFile(opts ...Option) (config.ReadOnlyTargets, error) {
...
+
 	go func() {
 		for {
 			select {
 			case event, ok := <-watcher.Events:
 				if !ok {
 					// 'Events' channel is closed.
-					done <- struct{}{}
-					return
+					break
```
ericlem: you'll actually need a named break statement here (or do the return with the defer inside the function) since just doing the bear break just breaks you out of the select

yolocs: ah, right! No wonder it gave me warning.

PR #683, opened 2020-03-20
[release-0.13] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign grantr nachocano
/cc grantr nachocano


nachocano:
I don't think we want a point release

PR #690, opened 2020-03-20
Add AdapterLifecycle field to Topic and PullSubscription

grantr:
This field is intended to control whether the Topic and PullSubscription receive adapters are created/reconciled by the controller. Just the types for now, no controller support yet.

@ericlem @yolocs @liu-cong Creating this PR for early feedback on the approach. I think it might be quickest to use the existing resources to create (and display status on) GCP objects, but skip creating the receive adapters. If you agree this makes sense, I'll go ahead and plumb support for these fields through the controllers.

## Proposed Changes

- Add AdapterLifecycle field to Topic and PullSubscription (type only)

(No user impact yet because it's not hooked up)

### DIFF
--- a/pkg/apis/pubsub/v1alpha1/pullsubscription_defaults.go
+++ b/pkg/apis/pubsub/v1alpha1/pullsubscription_defaults.go
@@ -58,4 +58,8 @@ func (ss *PullSubscriptionSpec) SetDefaults(ctx context.Context) {
 		// Default is CloudEvents Binary Mode.
 		ss.Mode = ModeCloudEventsBinary
 	}
+
+	if ss.AdapterLifecycle == "" {
+		ss.AdapterLifecycle = AdapterLifecycleReconcile
+	}
 }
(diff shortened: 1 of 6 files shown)


yolocs:
I haven’t looked into details. But I remember there is a common lib for pub sub operations. It’s used by the existing controllers. Maybe it won’t be that much trouble to direct reconcile pubsub?

ericlem:
If it can be done cleanly i'm ok with it...  My only question is whether it makes it any harder to find this info and feed it into the configmaps to have another controller doing this.  I don't have a strong opinion (or enough experience) to say which choice is best.

grantr:
Both good questions. Most of the code I've seen assumes that every time a topic is created, a receive adapter is also created, but since this would need to be refactored for this change anyway, it is worth seeing what that refactor looks like before altering the api.

/hold

grantr:
Closing this since the broker controller doesn't need it yet. We can bring it back if desired.

PR #692, opened 2020-03-21
Adding a status for workload identity

### LINKED ISSUE #668: Adding a status condition for workload identity
**Problem**
We need  a status condition for objects for workload identity. When workload identity wasn't properly set, it will be marked.
If the user is not using workload identity (using secret), the status type WorkloadIdentityReady should be unknown. This status will be not included in condset, so that the unknown status will not affect the whole condition to be ready.
This condition should be set in channel and all sources.

**Exit Criteria**
Adding a status condition for workload identity

grac3gao:
Fixes #668 

## Proposed Changes

- Add a status for workload identity
-
-

**Release Note**

```release-note

```

**Docs**


pkg/apis/duck/v1alpha1/identity.go
```diff
@@ -0,0 +1,75 @@
...
+	Enabled string `json:"enabled,omitempty"`
+
+	// Ready represents whether workload identity reconciles successfully or not. If it is not enabled, Ready will be unknown.
+	Ready string `json:"ready,omitempty"`
+
+	// Short reason for the status.
+	Reason string `json:"reason,omitempty"`
+
+	// A human readable message indicating details about the failure.
+	Message string `json:"message,omitempty"`
```
nachocano: it seems that we should be able to reuse Condition from pkg for these?
See https://github.com/knative/pkg/blob/0840da9555a3a75f801abc1d654fb00dfe9a687a/apis/condition_types.go#L58?
And you will get more type-safety.
The duckv1.SourceStatus has a Status, which has a list of Conditions. 
I'm not entirely sure we should be adding these ones here...

pkg/reconciler/events/auditlogs/auditlogs.go
```diff
@@ -70,6 +70,7 @@ func (c *Reconciler) ReconcileKind(ctx context.Context, s *v1alpha1.CloudAuditLo
 	ctx = logging.WithLogger(ctx, c.Logger.With(zap.Any("auditlogsource", s)))
 
 	s.Status.InitializeConditions()
```
nachocano: it feels that the WorkloadIdentityCondition should be added to this Status...
And if you have workload identity enabled, but not working, then you can mark that condition failed, and also Mark the ready false...
not entirely sure though

pkg/apis/duck/v1alpha1/identity.go
nachocano: I think I'd rather keep it simple for now to see how this workload identity places out. 
And would just add a condition to the types called something like "WorkloadIdentityConfigured" or something like that, not part of the NewLivingConditionSet...
and provide methods to MarkWorkloadIdentityConfigured and not, with proper reasons....
We can use the Unknown when users don't specify a serviceAccount... nothing will be set, but that won't affect the readiness.
We can use True when users specified the serviceAccount and it configured correctly.
We can use False when users specified the serviceAccount and it wasn't configured properly. In this latter case, we should also Mark Ready = false to the object.

I don't think we need to save Enabled in status. But it would be nice if we save the k8s service account, what you are currently doing. 

I think this will simplify things quite a bit... but want to hear your thoughts as you have done all the work here.

grac3gao: I'll change that, and save the k8s service account name. I'll combine the new commit with change of k8s name.

grac3gao: It seems I can only set the condition in every reconciler, rather than combine it to the identity reconciler if we use the Conditions for WI.

The probelm is still for the status. If I want to mark the condition status in the identity reconciler, I need to get sources/channel `status` and `api.condtionSet`. Then I can use `sourceStatus.Manager(api.condtionSet).MarkTrue` to mark the condition.

I can get the condSet, but I cannot create an identifiable interface function for getting all sources and channel’s status, the only common thing shared by sources and channel’s status is they all use duck.status. Then I have to write things like (s *duck.status)MarkworkloadIdentityReconciled. I am not sure if we can do that.

nachocano:
@grac3gao I created a sample PR here for you to take a look: https://github.com/google/knative-gcp/pull/703. It was more involved than I initially thought. 
There are a bunch of cleanups that are needed. Especially in the lifecycles of Sources, we have unnecessary methods...

fyi @Harwayne in case you pick up the v1beta1 task, there are a bunch of cleanups to do first..

grac3gao:
@nachocano Thank you for adding that sample PR! By including the pubsubStatus in the identifiable, I can mark the WI condition in the identity reconciler now! It seems I also need to change Channel status to use pubsubStatus? 

For serviceAccount, I am fine with `gsa`, but I am worried about that it might be too brief for user to understand? Perhaps  `gServiceAccount` would be a choice?

I wrote a rough change for the status condition last night, but that is about marking the status in every reconciler. I think it might be better to change all the things (adding status condition, change name) together in a later PR, after determining the name and add the E2E test. 

I'll close this PR and focus on E2E test now. After finishing that, I'll go back to condition.

nachocano:
Sounds like a plan!
Yeah gsa might not be correct as well, and gServiceAccount sounds better. I
think we need a third opinion, maybe Adam?

I think Channel status shouldn't use pubsubStatus, might have missed that
in the code, not sure.

Agree let's do that in other PRs. E2E tests are top priority now.

Thanks grace!

Issue #595, opened 2020-03-03
Policy binding implementation with Istio


yolocs:
**Problem**
After https://github.com/google/knative-gcp/issues/575, we will need at least one policy binding implementation for policy enforcement.
We will start with an Istio implementation.

Tasks

- [x] HTTP policy binding implementation with Istio
- [ ] Event policy binding implementation with HTTP policy

PR #696, opened 2020-03-23
Add controller for EventPolicyBinding

yolocs:
Fixes https://github.com/google/knative-gcp/issues/595

## Proposed Changes

The controller will interpret the referenced EventPolicy to a HTTPPolicy and create a HTTPPolicyBinding to bind it to the same subject.

**Release Note**

```release-note
Added controller for EventPolicyBinding which leverages HTTPPolicy and HTTPPolicyBinding.
```

**Docs**


pkg/reconciler/security/istio/eventpolicybinding/controller.go
```diff
@@ -0,0 +1,90 @@
...
+		httpPolicyLister:        httpPolicyInformer.Lister(),
+		httpPolicyBindingLister: httpPolicyBindingInformer.Lister(),
+	}
+
+	impl := bindingreconciler.NewImpl(ctx, r)
+
+	r.Logger.Info("Setting up event handlers")
+
+	eventPolicyBindingInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: pkgreconciler.AnnotationFilterFunc(securityapi.PolicyBindingClassAnnotationKey, securityapi.IstioPolicyBindingClassValue, false),
```
nachocano: make sure to check for immutability of this annotations, otherwise the proper controller won't be called if changed

pkg/reconciler/security/istio/eventpolicybinding/controller.go
```diff
@@ -0,0 +1,90 @@
...
+	})
+
+	httpPolicyInformer.Informer().AddEventHandler(cache.FilteringResourceEventHandler{
+		FilterFunc: controller.FilterGroupKind(securityv1alpha1.Kind("EventPolicyBinding")),
+		Handler:    controller.HandleAll(impl.EnqueueControllerOf),
+	})
+
+	r.policyTracker = tracker.New(impl.EnqueueKey, controller.GetTrackerLease(ctx))
+	eventPolicyInformer.Informer().AddEventHandler(controller.HandleAll(
+		controller.EnsureTypeMeta(r.policyTracker.OnChanged, eventPolicyGVK),
```
nachocano: nit: can the eventPolicyGVK be inlined here for easier readability, or you are using it from somewhere else?

nachocano: NVM saw you are using it later

pkg/reconciler/security/istio/eventpolicybinding/eventpolicybinding.go
```diff
@@ -0,0 +1,146 @@
...
+	b.Status.ObservedGeneration = b.Generation
+
+	p, err := r.eventPolicyLister.EventPolicies(b.Spec.Policy.Namespace).Get(b.Spec.Policy.Name)
+	if err != nil {
+		logging.FromContext(ctx).Error("Problem getting EventPolicy", zap.Any("EventPolicy", b.Spec.Policy), zap.Error(err))
+		b.Status.MarkBindingFailure("GetPolicyFailure", "%v", err)
+		return fmt.Errorf("failed to get EventPolicy: %w", err)
+	}
+	// Track referenced policy.
+	r.policyTracker.TrackReference(tracker.Reference{
```
nachocano: does this return an error? I think it used to, unless they changed the API. If it does, then we should check for the error...
I've seen cases where we don't have proper permissions to track things and this can fail

pkg/reconciler/security/istio/eventpolicybinding/eventpolicybinding.go
```diff
@@ -0,0 +1,146 @@
...
+
+	// Copy the HTTP policy binding status.
+	b.Status = *hpb.Status.DeepCopy()
+	return pkgreconciler.NewEvent(corev1.EventTypeNormal, "EventPolicyBindingReconciled", "EventPolicyBinding reconciled: \"%s/%s\"", b.Namespace, b.Name)
+}
+
+func (r *Reconciler) reconcileHTTPPolicyBinding(
+	ctx context.Context,
+	b *v1alpha1.EventPolicyBinding,
+	p *v1alpha1.HTTPPolicy) (*v1alpha1.HTTPPolicyBinding, pkgreconciler.Event) {
```
nachocano: I would just return error instead of pkgreconciler.Event here...

pkg/reconciler/security/istio/eventpolicybinding/eventpolicybinding.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+	}
+
+	return existing, nil
+}
+
+func (r *Reconciler) reconcileHTTPPolicy(
+	ctx context.Context,
+	b *v1alpha1.EventPolicyBinding,
+	p *v1alpha1.EventPolicy) (*v1alpha1.HTTPPolicy, pkgreconciler.Event) {
```
nachocano: same here

pkg/reconciler/security/istio/eventpolicybinding/resources/httppolicy.go
```diff
@@ -0,0 +1,77 @@
...
+	"github.com/google/knative-gcp/pkg/apis/security/v1alpha1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/kmeta"
+)
+
+// MakeHTTPPolicy generates a HTTPPolicy based on the given EventPolicy.
+func MakeHTTPPolicy(b *v1alpha1.EventPolicyBinding, p *v1alpha1.EventPolicy) *v1alpha1.HTTPPolicy {
+	hp := &v1alpha1.HTTPPolicy{
+		ObjectMeta: metav1.ObjectMeta{
+			// TODO: Consider use random suffix.
```
nachocano: yeah, maybe you can just do "http-%s-%s" and have p.Name and p.UUID or something...
And use kmeta.ChildName to make sure is a valid name

pkg/reconciler/security/istio/httppolicybinding/httppolicybinding.go
```diff
@@ -82,8 +82,8 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, b *v1alpha1.HTTPPolicyBi
 	}
 	// Track referenced policy.
 	r.policyTracker.TrackReference(tracker.Reference{
```
nachocano: same thing with the error

pkg/reconciler/security/istio/eventpolicybinding/eventpolicybinding.go
```diff
@@ -0,0 +1,146 @@
...
+
+	hpb, err := r.reconcileHTTPPolicyBinding(ctx, b, hp)
+	if err != nil {
+		logging.FromContext(ctx).Error("Problem reconciling HTTPPolicyBinding", zap.Error(err))
+		b.Status.MarkBindingFailure("HTTPPolicyBindingReconcileFailure", "%v", err)
+		return err
+	}
+
+	// Copy the HTTP policy binding status.
+	b.Status = *hpb.Status.DeepCopy()
```
nachocano: do you need this DeepCopy here?

pkg/reconciler/security/istio/eventpolicybinding/resources/httppolicy.go
yolocs: I will address this separately

pkg/reconciler/security/istio/eventpolicybinding/eventpolicybinding.go
yolocs: I think it might be dangerous to deep copy. So now I changed to explicitly set status.

pkg/reconciler/security/istio/eventpolicybinding/eventpolicybinding.go
```diff
@@ -82,15 +86,20 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, b *v1alpha1.EventPolicyB
...
 
-	// Copy the HTTP policy binding status.
-	b.Status = *hpb.Status.DeepCopy()
+	// Propagate the status.
+	if hpb.Status.IsReady() {
+		b.Status.MarkBindingAvailable()
+	} else {
+		st := hpb.Status.GetTopLevelCondition()
+		b.Status.MarkBindingFailure(st.Reason, st.Message)
+	}
```
nachocano: nit: might be cleaner to just create a `b.Status.PropagateBindingStatus(hpb)`, and you put this logic there. 
That is what we've been doing for propagating deployments statuses, and other objects as well.
See for example: https://github.com/knative/eventing/blob/1c825f9a6478582a9f803e445d1989272515421c/pkg/reconciler/apiserversource/apiserversource.go#L138

pkg/apis/security/v1alpha1/common_lifecycle.go
```diff
@@ -69,3 +69,20 @@ func (pbs *PolicyBindingStatus) MarkBindingFailure(reason, messageFormat string,
 func (pbs *PolicyBindingStatus) MarkBindingAvailable() {
 	policybindingCondSet.Manage(pbs).MarkTrue(PolicyBindingConditionReady)
 }
+
+// PropagateBindingStatus propagates the current binding status from another binding status.
+func (pbs *PolicyBindingStatus) PropagateBindingStatus(other *PolicyBindingStatus) {
+	if other == nil {
+		return
+	}
+	pbs.InitializeConditions()
```
nachocano: I think this shouldn't be done here. This initializeConditions is typically done in the reconciler. And I think you are doing it here already: https://github.com/google/knative-gcp/blob/c8af8ac3b3723180be52aa94d7487f4675ef8233/pkg/reconciler/security/istio/eventpolicybinding/eventpolicybinding.go#L54

Let's get this in, and you remove it in a follow up.

yolocs: Ack

Issue #667, opened 2020-03-17
Determine pubsub encoding format for decoupling topic


grantr:
If the data plane must be namespace-aware, the event format used in the pubsub messages may need to encode the namespace or other metadata along with the event.

We should version (separately from CloudEvents version) this internal format to allow for evolution.

yolocs:
As we decided to adopt the shared broker data-plane, it becomes obvious that we need to attach additional info to each event, e.g. hostname, request path, etc. The current cloudevents pubsub binding doesn't allow us to do that easily (there is no interface to do that with the current sdk).

The current cloudevents binding logic:
With structured encoding - it directly writes the event bytes to the pubsub `message.Data`
With binary encoding - it writes standard cloudevent attributes to pubsub `message.Attribute` and data to `message.Data`

Some options I can think of:

Option 1:
Extend cloudevents pubsub binding to have additional function to "label" events. These functions won't be part of the common `protocol` interface. These "labels" will be set to pubsub `message.Attributes` with a special prefix (anything different from "ce-").

Option 2:
Develop a lib in knative-gcp for publish and pull pubsub message for gcp broker.
Publish: events -> convert to structured encoding -> set to `message.Data` and labels -> set to `message.Attributes`
Pull: the reverse of what's above.

ericlem:
I'd tend to to favor something in the Option 1 case.   We almost certainly don't want to be having to use structured encoding if possible.  And 1 leaves the most fields still in interpretable form in pubsub without decoding.

grantr:
We don't actually have to use structured encoding for option 2. Since we control both ends of the pubsub topics, we can use whatever encoding we want in pubsub messages, cloudevents compatible or not.

yolocs:
Option 3:
WE can directly use cloudevents extension with a special prefix.

grantr:
> it becomes obvious that we need to attach additional info to each event, e.g. hostname, request path, etc

If we have a topic per broker, do we need any additional fields? Or are these fields only needed if we have a single shared topic for all brokers?

yolocs:
I'm thinking a single shared topic. I thought we're going to have a single shared topic?

grantr:
If single topic requires us to do additional work to partition the events, what is the advantage of that single topic? But I created #699 to discuss that.

I think I'm leaning toward option 3 (using extension attributes) since it seems like the least work for now. As has been pointed out by others, we can change this later if it proves to be an issue.

Would this be sufficient?

```
[...]

PR #700, opened 2020-03-24
Add a lib to label/delabel events as cloudevents extensions

yolocs:
Fixes https://github.com/google/knative-gcp/issues/667

## Proposed Changes

- Also added broker name to the config proto. We need it for proper routing.

**Release Note**

```release-note
Add a lib to label/delabel events as cloudevents extensions
```

**Docs**


pkg/broker/eventutil/label.go
```diff
@@ -0,0 +1,74 @@
...
+	return le
+}
+
+// GetLabels gets all the labels as a map.
+func (le *LabeledEvent) GetLabels() map[string]string {
+	m := make(map[string]string)
+	exts := le.event.Extensions()
+	for k, v := range exts {
+		if strings.HasPrefix(strings.ToLower(k), LabelPrefix) {
+			m[k] = v.(string)
```
Harwayne: Should this remove the prefix too? So that it is symmetric with `WithLabel`.

pkg/broker/eventutil/label.go
```diff
@@ -0,0 +1,74 @@
...
+
+import (
+	"strings"
+
+	cloudevents "github.com/cloudevents/sdk-go/v2"
+)
+
+const (
+	// LabelPrefix is the prefix for label keys.
+	LabelPrefix = "kgcp"
```
Harwayne: I would make this private unless you expect other code to use it too.

pkg/broker/eventutil/label.go
yolocs: Ah, I forgot to test this.

pkg/broker/eventutil/label.go
```diff
@@ -0,0 +1,78 @@
...
+	return le
+}
+
+// GetLabels gets all the labels as a map.
+func (le *LabeledEvent) GetLabels() map[string]string {
+	m := make(map[string]string)
+	exts := le.event.Extensions()
+	for k, v := range exts {
+		if strings.HasPrefix(strings.ToLower(k), labelPrefix) {
+			m[strings.TrimPrefix(k, labelPrefix)] = v.(string)
```
Harwayne: This doesn't `strings.ToLower(k)`, which means it doesn't _quite_ match the conditional above.

yolocs: Good point.

PR #703, opened 2020-03-25
[Draft] PoC for Workload Identity status condition

nachocano:
## Proposed Changes

- @grac3gao this is an example of how it might look like. We had a bunch of unnecessary stuff. Probably we have more, but I haven't had the time to clean it up. Take a look and let me know if it makes sense. 
- No need to do it, I'd rather have the e2e tests working first. But if you have cycles, might be worth giving it a shot.

**Release Note**

```release-note

```

**Docs**

/hold won't submit this, it's just for demonstration purposes...

### DIFF
--- a/pkg/apis/duck/v1alpha1/identity_types.go
+++ b/pkg/apis/duck/v1alpha1/identity_types.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2019 The Knative Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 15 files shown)


pkg/apis/duck/v1alpha1/identity_types.go
```diff
@@ -0,0 +1,102 @@
...
+
+	Spec   PubSubSpec   `json:"spec"`
+	Status PubSubStatus `json:"status"`
+}
+
+type IdentitySpec struct {
+	// ServiceAccount is the GCP service account which has required permissions to poll from a Cloud Pub/Sub subscription.
+	// If not specified, defaults to use secret.
+	// +optional
+	ServiceAccount string `json:"serviceAccount,omitempty"`
```
nachocano: @grac3gao I think you were right. I think we should call this something different, otherwise it might be confusing with a k8s service account. I'm inclined to call it `gsa` not to be too verbose with `googleServiceAccount`, but I don't have strong opinion... 
This is lower priority, but something to keep in mind, because it changes a bunch of places..

pkg/apis/duck/v1alpha1/identity_types.go
```diff
@@ -0,0 +1,102 @@
...
+// +genduck
+// +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
+
+type Identity struct {
+	metav1.TypeMeta   `json:",inline"`
+	metav1.ObjectMeta `json:"metadata,omitempty"`
+
+	Spec   PubSubSpec   `json:"spec"`
+	Status PubSubStatus `json:"status"`
+}
```
nachocano: this duck is not even needed.

Issue #658, opened 2020-03-16
 E2E tests for workload identity


grac3gao:
**Problem**
We need to have e2e tests to verify workload identity. When enabled workload identity, the control plane should work without secret, and the data plane should work with a qualified GCP serviceAccount inputted (without secret). 

**Exit Criteria**
e2e tests to verify workload identity.

grac3gao:
Put some thoughts here. 
Compare with the existing secret method:
1. In control plane, we have a different controller deployment for workload identity, which means the init set up for the e2e testing cluster needs to change (apply different yams for controller deployment, and running command to enable WI in control plane)
2. In data plane, we need to have a GSA, and when running the e2e test, instead of duplicate secret, set the GSA under spec.serviceAccount

grac3gao:
Sync with @chizhg offline, and Chi listed what needs to be done for this:
1. After https://github.com/kubernetes/test-infra/pull/16888 is merged, update prow-tests image to get the latest kubetest
2. Add a new script in knative-gcp/test, namely e2e-wi-tests.sh. In this script, source e2e-common.sh and override the setup and teardown functions to install wi-enabled control plane components, call initialize $@ --cluster-creation-flag "--workload-pool=\${PROJECT}.svc.id.goog" to create a cluster with Workload Identity enabled.
3. Update Prow config to run e2e-wi-tests.sh and e2e-tests.sh separately in two Prow jobs.

He will take step1 and step3, and I will take step2, and update test lib to make it compatible with WI.

grac3gao:
After running e2e test in PROW, there is an issue. In PROW, the control plane's Google Service Account doesn't belong to the current project, which will cause error like: `NOT_FOUND: Service account projects/knative-boskos-96/serviceAccounts/prow-job@knative-tests.iam.gserviceaccount.com does not exist`, this makes the GSA unable to set iam policy binding for Workload Identity.

grac3gao:
E2E test for Workload Identity works. Thanks for your help! @chizhg. Currently, testing for if two e2e test scripts can properly run in the same boskos project in PROW.

ericlem:
Done.

PR #709, opened 2020-03-25
E2E tests for workload identity

grac3gao:
related on #658 

## Proposed Changes

- add e2e-wi-test.sh
- add flags to pass workloadIdentity and service account information
- change E2E test to accept workloadIdentity and service account.

Have some problems when dealing with Broker and Channel E2E test. Details are in comments...

**Release Note**

```release-note

```

**Docs**


test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,141 @@
...
+    gcloud projects add-iam-policy-binding ${E2E_PROJECT_ID} \
+      --member=serviceAccount:${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
+      --role roles/cloudscheduler.admin
+  fi
+  # Give iam.serviceAccountAdmin role to the Google service account.
+  gcloud projects add-iam-policy-binding ${E2E_PROJECT_ID} \
+    --member=serviceAccount:${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
+    --role roles/iam.serviceAccountAdmin
+  # Allow the Kubernetes service account to use Google service account.
+  MEMBER="serviceAccount:${E2E_PROJECT_ID}.svc.id.goog[${CONTROL_PLANE_NAMESPACE}/${CONTROLLER_SERVICE_ACCOUNT}]"
```
grac3gao: Should I also add `\` here for the MEMBER string?

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,141 @@
...
+    # Remove iam.serviceAccountAdmin role to the Google service account.
+    gcloud projects remove-iam-policy-binding ${E2E_PROJECT_ID} \
+      --member=serviceAccount:${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
+      --role roles/iam.serviceAccountAdmin
+    gcloud iam service-accounts delete -q ${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com
+  fi
+}
+
+#create a cluster with Workload Identity enabled.
+initialize $@ --cluster-creation-flag "--workload-pool=\${E2E_PROJECT_ID}.svc.id.goog"
```
grac3gao: I simply tried it in the terminal, running echo, only work when there is no `\`, it that expected？

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,141 @@
...
+
+# Add annotation to Kubernetes service account.
+kubectl annotate serviceaccount ${CONTROLLER_SERVICE_ACCOUNT} iam.gke.io/gcp-service-account=${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
+--namespace ${CONTROL_PLANE_NAMESPACE}
+
+# Channel related e2e tests we have in Eventing is not running here.
+go_test_e2e -timeout=20m -parallel=12 ./test/e2e -workloadIdentity=true -pubsubServiceAccount=${PUBSUB_SERVICE_ACCOUNT} || fail_test
```
grac3gao: Remove the pubsub channel tests related on eventing. Eventing test lib only accepts channel with ChannelTypeMeta, but we also need `Spec.serviceAccount` to make Workload Identity work for channel. Is there any way to also include channel's `Spec.serviceAccount` when creating broker?

test/e2e/e2e_test.go
```diff
@@ -163,105 +172,120 @@ func TestChannelTracing(t *testing.T) {
...
 	t.Skip("See issues https://github.com/google/knative-gcp/issues/317 and https://github.com/cloudevents/sdk-go/pull/234")
 	cancel := logstream.Start(t)
 	defer cancel()
-	CloudPubSubSourceWithTargetTestImpl(t, true /*assertMetrics */)
+	CloudPubSubSourceWithTargetTestImpl(t, true /*assertMetrics */, workloadIdentity, pubsubServiceAccount)
 }
 
 // TestBrokerWithPubSubChannel tests we can knock a Knative Service from a broker with PubSub Channel.
 func TestBrokerWithPubSubChannel(t *testing.T) {
+	if workloadIdentity {
```
grac3gao: Broker test will create the Broker and use `client.Core.CreateBrokerOrFail(brokerName, eventingtestresources.WithChannelTemplateForBroker(lib.ChannelTypeMeta))` to create channel, `WithChannelTemplateForBroker` only accept `ChannelTypeMeta`. https://github.com/knative/eventing/blob/93707e9f7466a729d16e5ad3a20fd2c48490c54b/test/lib/resources/eventing.go#L181-L188

Looks like we need to change eventing's `WithChannelTemplateForBroker` method to include channel's `Spec.serviceAccount` in channeltemplate?

test/e2e-wi-tests.sh
grac3gao: Also, for the installation of the cluster, workload identity and secret have some differences. currently, using `ko apply -f config`  will install secret way, you have to apply core/deployment/controller-gke.yaml after that to make workload identity configured. Not sure if I should also include that here?

test/e2e-wi-tests.sh
chizhg: No, you don't need to.

test/e2e-wi-tests.sh
chizhg: 1. Yes, it's expected.
2. I would suggest you add an arg for https://github.com/google/knative-gcp/blob/master/test/lib.sh#L23 and https://github.com/google/knative-gcp/blob/master/test/lib.sh#L31 to determine if we install knative-gcp that has workload identity configured, and pass it as a flag somewhere else.

test/e2e-wi-tests.sh
chizhg: Yeah it seems we need to do some refactor to the existing e2e tests. Instead of only including `Spec.serviceAccount`, I think we can include the whole `Spec`, and set it in https://github.com/knative/eventing/blob/master/test/lib/duck/resource_creators.go#L65.

test/e2e/e2e_test.go
chizhg: The Broker v1beta1 spec seems to have deprecated the `ChannelTemplate`, and is using `Config` instead. But the same suggestion as the above, we should pass `Spec` instead of `Spec.serviceAccount`, the code to set it should be at https://github.com/knative/eventing/blob/master/test/lib/creation.go#L176 (I don't know when it's added but it's pretty ugly...)

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,141 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
grac3gao: Like `chmod +x #!/usr/bin/env bash`? I have a dumb question...why we need chmod +x here, but not have it in e2e-tests.sh?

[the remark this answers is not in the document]
grac3gao: or add the `chmod +x e2e-wi-tests.sh` in next line? I think I know why we need chmod + x here, otherwise, it does not have permission to run the file.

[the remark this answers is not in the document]
grac3gao: `chmod +x e2e-wi-tests.sh` works

test/e2e-wi-tests.sh
chizhg: Please replace `${E2E_PROJECT_ID}` with `${PROJECT}` in this line.

grac3gao:
Have tried the `e2e-wi-test.sh` locally by creating a local cluster, and it works now. Let's see if it can work in PROW once the issue in test-infra (for image update to allow passing env var from GKE create command) is solved @chizhg.  I skip tests related on channel with broker, and eventing's test. I'll create a issue for that to remove the skip after making the eventing lib support to pass `Spec` for channel `config`

test/e2e-wi-tests.sh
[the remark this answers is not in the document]
Harwayne: I'm curious how this works. If the file isn't executable, then how is the `chmod` line run? And if it is executable, why is the line needed?

test/e2e_flags.go
```diff
@@ -0,0 +1,48 @@
...
+// EventingEnvironmentFlags holds the e2e flags needed only by the eventing repo.
+type EnvironmentFlags struct {
+	WorkloadIdentityEnabled bool
+	PubsubServiceAccount    string
+}
+
+// InitializeFlags registers flags used by e2e tests, calling flag.Parse() here would fail in
+// go1.13+, see https://github.com/knative/test-infra/issues/1329 for details
+func InitializeFlags() {
+	flag.BoolVar(&Flags.WorkloadIdentityEnabled, "workloadIdentityEnabled", false, "Indicating whether the workload identity is enabled or not.")
```
Harwayne: My preference would be to name this `workloadIndentity`, as I feel 'enabled' is redundant with the value of the flag.

test/e2e/lib/auditlogs.go
```diff
@@ -26,14 +26,15 @@ import (
 )
 
 func MakeAuditLogsOrDie(client *Client,
-	auditlogsName, methodName, project, resourceName, serviceName, targetName string,
+	auditlogsName, methodName, project, resourceName, serviceName, targetName, pubsubServiceAccount string,
```
Harwayne: Not needed in this PR, just want to make sure people think about this (maybe make an issue)?

This is a lot of strings in a row. It will be hard/impossible for someone reading the code to see that everything is in the correct position. I recommend creating a struct instead of passing seven strings in a row.

A similar problem exists on most of the `MakeFoo(...)` style methods in this change.

test/e2e-wi-tests.sh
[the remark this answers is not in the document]
grac3gao: I am curious too...as I checked the E2E test result, actually it would show things like `file not exist` when running this `e2e-wi-tests.sh`, and the then the file will continue to run for the next lines. But somehow, it works... (before adding it, would show error like `doesn't have permission to run the file`, and `e2e-wi-tests.sh` file cannot run). Let me remove the this command to try again. There might be other way under it.

grac3gao:
> /test pull-google-knative-gcp-integration-tests

Integration test is currently blocked by https://github.com/knative/test-infra/issues/1863, @chizhg is working on that.

test/e2e/test_pullsubscription.go
```diff
@@ -34,20 +34,21 @@ import (
...
-	client := lib.Setup(t, true)
+	client := lib.Setup(t, true, authConfig.WorkloadIdentityEnabled)
 	defer lib.TearDown(client)
 
 	// Create PullSubscription.
 	pullsubscription := kngcptesting.NewPullSubscription(psName, client.Namespace,
 		kngcptesting.WithPullSubscriptionSpec(v1alpha1.PullSubscriptionSpec{
-			Topic: topic,
+			Topic:          topic,
+			ServiceAccount: authConfig.PubsubServiceAccount,
```
Harwayne:
```suggestion
			IdentitySpec: duckv1alpha1.IdentitySpec{
				ServiceAccount: authConfig.PubsubServiceAccount,
			},
```

Something like this. Also needed on line 81.

test/e2e_flags.go
grac3gao: @chizhg mentioned to change the name (https://github.com/google/knative-gcp/pull/709#discussion_r398316896) from` workloadIndentity` to `WorkloadIdentityEnabled `, that is for the name after parsing the flag to the e2e-test. But I changed all, for flag and for the name after parsing it. Do you think we need to use two different names for this, or `workloadIndentity` can be used in both place?

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,150 @@
...
+      --role roles/cloudscheduler.admin
+  fi
+  # Give iam.serviceAccountAdmin role to the Google service account.
+  gcloud projects add-iam-policy-binding ${E2E_PROJECT_ID} \
+    --member=serviceAccount:${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
+    --role roles/iam.serviceAccountAdmin
+  # Allow the Kubernetes service account to use Google service account.
+  MEMBER="serviceAccount:${E2E_PROJECT_ID}.svc.id.goog[${CONTROL_PLANE_NAMESPACE}/${CONTROLLER_SERVICE_ACCOUNT}]"
+  gcloud iam service-accounts add-iam-policy-binding \
+  --role roles/iam.workloadIdentityUser \
```
chizhg: Please add two spaces ahead of this line.

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,150 @@
...
+    gcloud iam service-accounts delete -q ${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com
+  fi
+}
+
+#create a cluster with Workload Identity enabled.
+initialize $@ --cluster-creation-flag "--workload-pool=\${PROJECT}.svc.id.goog"
+
+# Add annotation to Kubernetes service account.
+kubectl annotate serviceaccount ${CONTROLLER_SERVICE_ACCOUNT} iam.gke.io/gcp-service-account=${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
+--namespace ${CONTROL_PLANE_NAMESPACE}
```
chizhg: Two spaces ahead of this line.

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,150 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source $(dirname $0)/e2e-common.sh
+
+# Override the setup and teardown functions to install wi-enabled control plane components.
+
+readonly CONTROLLER_SERVICE_ACCOUNT="controller"
```
chizhg: To avoid confusion, could you please rename it to `K8S_CONTROLLER_SERVICE_ACCOUNT`?

grac3gao:
After running e2e test in PROW, there is a new issue. https://github.com/google/knative-gcp/issues/658#issuecomment-606743086

test/e2e-wi-tests.sh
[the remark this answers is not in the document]
grac3gao: Looks like the file is already executable...I removed that line

grac3gao:
Looks like boskos project needs to enable IAM Service Account Credentials API, otherwise it is unable to generate token. https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_on_a_new_cluster

test/e2e/lib/auditlogs.go
grac3gao: The MakeFoo(…) style method looks like:

https://github.com/google/knative-gcp/blob/01e5d59d820e7918fe4f5d557dc669bdfac45eea/test/e2e/lib/auditlogs.go#L28-L42

Those strings inside the MakeFoo(…) will be transformed to source options, and it accepts other options. 
I feel I should not add a new input variable, but using format like following instead.
`lib.MakeAuditLogsOrDie(client,
   auditlogsName, 
   methodName, 
   project, 
   resourceName, 
   serviceName, 
   targetName,
   kngcptesting.WithCloudStorageSourceGCPServiceAccount(pubsubServiceAccount)))`
(I was wondering if the existing input strings are the basic/required part for making a source). I am fine to make them as a struct. But should I include the pubsubServiceAccount in the struct? Or use it just like `kngcptesting.WithCloudStorageSourceGCPServiceAccount(pubsubServiceAccount)`?

grac3gao:
ready to review

pkg/reconciler/identity/reconciler.go
```diff
@@ -165,7 +166,9 @@ func setIamPolicy(ctx context.Context, action, projectID string, gServiceAccount
 		return fmt.Errorf("failed to get project id: %w", err)
 	}
 
-	resource := fmt.Sprintf("projects/%s/serviceAccounts/%s", projectId, gServiceAccount)
+	// Extract gServiceAccount's project name.
+	gsaProject := strings.Split(strings.Split(gServiceAccount, "@")[1], ".")[0]
```
chizhg, line 170: Can we add an example service account name here, so it's clearer how the project name is extracted?

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,186 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+source $(dirname $0)/e2e-common.sh
+
+# Override the setup and teardown functions to install wi-enabled control plane components.
+readonly K8S_CONTROLLER_SERVICE_ACCOUNT="controller"
+readonly PROW_SERVICE_ACCOUNT=$(gcloud config list account --format "value(core.account)")
```
chizhg, line 22: `gcloud config get-value core/account` also works and it's shorter.

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,186 @@
...
+        # Give iam.serviceAccountAdmin role to the Google service account.
+    gcloud projects add-iam-policy-binding ${E2E_PROJECT_ID} \
+      --member=serviceAccount:${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
+      --role roles/iam.serviceAccountAdmin
+    gcloud iam service-accounts add-iam-policy-binding \
+      --role roles/iam.workloadIdentityUser \
+      --member ${MEMBER} ${AUTHENTICATED_SERVICE_ACCOUNT}
+  else
+    # If the tests are run on Prow, clean up the member for roles/iam.workloadIdentityUser before running it.
+    members=$(gcloud iam service-accounts get-iam-policy --project=${PROW_PROJECT_NAME} ${AUTHENTICATED_SERVICE_ACCOUNT} --format="value(bindings.members)" --filter="bindings.role:roles/iam.workloadIdentityUser" --flatten="bindings[].members")
```
chizhg, line 79: Please add line breaks for this long line..

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,186 @@
...
+          --member ${member_name} \
+          --project ${PROW_PROJECT_NAME} ${AUTHENTICATED_SERVICE_ACCOUNT}
+      fi
+    done <<< "$members"
+    gcloud iam service-accounts add-iam-policy-binding \
+      --role roles/iam.workloadIdentityUser \
+      --member ${MEMBER} \
+      --project ${PROW_PROJECT_NAME} ${AUTHENTICATED_SERVICE_ACCOUNT}
+  fi
+  # Allow the Kubernetes service account to use Google service account.
```
chizhg, line 95: This comment should be at line 90?

test/e2e-wi-tests.sh
```diff
@@ -0,0 +1,186 @@
...
+      --role roles/iam.workloadIdentityUser \
+      --member ${MEMBER} \
+      --project ${PROW_PROJECT_NAME} ${AUTHENTICATED_SERVICE_ACCOUNT}
+  fi
+}
+
+# Create a cluster with Workload Identity enabled.
+initialize $@ --cluster-creation-flag "--workload-pool=\${PROJECT}.svc.id.goog"
+
+kubectl annotate serviceaccount ${K8S_CONTROLLER_SERVICE_ACCOUNT} iam.gke.io/gcp-service-account=${AUTHENTICATED_SERVICE_ACCOUNT} \
```
chizhg, line 180: Move this command call to `control_plane_setup`?

grac3gao: These two comand? annotate and initialize?

chizhg: No just this annotate command (please note `control_plane_setup` is also called after `initialize`)

grac3gao: I see, so I need to add initialize at the beginning of control_plane_setup? and annotate at the end?

grac3gao: Sorry, ignore my last commend...

chizhg:
/lgtm
Defer to @Harwayne for approval.

test/e2e/lib/auditlogs.go
grac3gao: I opened https://github.com/google/knative-gcp/issues/764 for this

test/e2e-wi-tests.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
-chmod +x e2e-wi-tests.sh
+#chmod +x e2e-wi-tests.sh
```
Harwayne, line 2: Remove.

test/e2e/lib/auditlogs.go
Harwayne: My intention was something more like:

```go
struct AuditLogsConfig {
  // Possibly include Client too.
  auditlogsName string
  methodName string
  project string
  resourceName string
  serviceName string
  targetName string
  pubsubServiceAccount string
  options []kngcptesting.CloudAuditLogsSourceOption
}

func MakeAuditLogsOrDie(client *Client, cfg AuditLogsConfig) { ... }

// Then calling it would be something like:
MakeAuditLogsOrDie(client, AuditLogsConfig {
  auditlogsName: auditlogsName,
  methodName: methodName,
  project: project,
  resourceName: resourceName,
  serviceName: serviceName,
  targetName: targetName,
  pubsubServiceAccount: pubsubServiceAccount,
  options: options,
})
```

That way the strings are named, so it is easier to see if something is out of order.

grac3gao: Yeah, so I was wondering if I need to do like this? (not including `pubsubServiceAccount` as a part of that struct, as I was wondering if the rest input strings are the basic/required part for making a source, and the `pubsubServiceAccount` input I added in this PR is not required.)
```
MakeAuditLogsOrDie(client, AuditLogsConfig {
  auditlogsName: auditlogsName,
  methodName: methodName,
  project: project,
  resourceName: resourceName,
  serviceName: serviceName,
  targetName: targetName,
  options: []kngcptesting.CloudAuditLogsSourceOption{kngcptesting.WithCloudStorageSourceGCPServiceAccount(pubsubServiceAccount),}
})```

File: test/e2e-wi-tests.sh [hunk 18]
```diff
@@ -0,0 +1,188 @@
...
+
+# Create resources required for the Control Plane setup.
+function knative_setup() {
+  control_plane_setup || return 1
+  start_knative_gcp "workloadIdentityEnabled" || return 1
+  kubectl annotate serviceaccount ${K8S_CONTROLLER_SERVICE_ACCOUNT} iam.gke.io/gcp-service-account=${AUTHENTICATED_SERVICE_ACCOUNT} \
+    --namespace ${CONTROL_PLANE_NAMESPACE}
+}
+
+function control_plane_setup() {
```
liu-cong, line 43: This file seems very similar to `test/e2e-common.sh`, is there any chance you can reuse(or refactor and reuse) some logic there such as `control_plane_setup`?

grac3gao: e2e-common.sh is for non-wi script. We will finally use the init script to replace those gcloud command with the issue #738. After that, they will not be similar. The similarity are all for the gcloud command to grant permission in non-PROW environment. Keep it as it is for now.

pkg/reconciler/identity/reconciler.go
```diff
@@ -175,7 +180,14 @@ func setIamPolicy(ctx context.Context, action, projectID string, gServiceAccount
 	currentMember := fmt.Sprintf("serviceAccount:%s.svc.id.goog[%s/%s]", projectId, kServiceAccount.Namespace, kServiceAccount.Name)
 	rb := makeSetIamPolicyRequest(resp.Bindings, action, currentMember)
 
-	if _, err := iamService.Projects.ServiceAccounts.SetIamPolicy(resource, rb).Context(ctx).Do(); err != nil {
+	err = fmt.Errorf("%s", concurrencyError)
+	retryCount := 0
+	// If the setIamPolicy error is caused by concurrency, retry it until it reaches max retry time.
+	for err != nil && strings.Contains(err.Error(), concurrencyError) && retryCount < concurrencyMaxRetryTime {
+		_, err = iamService.Projects.ServiceAccounts.SetIamPolicy(resource, rb).Context(ctx).Do()
```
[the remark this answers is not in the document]
grantr: We should probably use a standard error retry and backoff mechanism like [`wait.ExponentialBackoff`](https://pkg.go.dev/k8s.io/apimachinery@v0.18.0/pkg/util/wait?tab=doc#ExponentialBackoff). There's an example of this in the client-go code at https://github.com/kubernetes/client-go/blob/master/util/retry/util.go along with default backoff settings. The genreconciler code uses this to retry status updates ([example](https://github.com/google/knative-gcp/blob/52be239f88268699510ef6400a45e426f568b7b0/pkg/client/injection/reconciler/pubsub/v1alpha1/topic/reconciler.go#L206)).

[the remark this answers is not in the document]
grac3gao: Thank you! @grantr I'll add that backoff func according to the examples.

Issue #636, opened 2020-03-12
Implement Broker ingress basic functionality


grantr:
* Accept events via HTTP and publish them to decoupling topic
* Unit tests

Future improvements to think about but not implement for this issue:
* Filtering at ingress
* Batching by filter expression/consumer

grantr:
This code has mostly been written for the Topic data plane. Handling multiple namespaces is additional work.

While scaling isn't necessary to close this issue, it would be a good idea to keep in mind that this should be compatible with Knative Serving. Could start with a Serving object with min/max scale 2 (to verify that scaling >1 is possible also).

PR #710, opened 2020-03-25
Basic broker ingress.

liu-cong:
Part of #636

## Proposed Changes

- Basic ingress functionality: receive events from HTTP and send events to pubsub, no routing
-
-

**Release Note**

```release-note

```

**Docs**

cc @yolocs @grantr @ericlem


cmd/broker/ingress/main.go
```diff
@@ -0,0 +1,48 @@
+/*
```
liu-cong: How to add the `kodata` folder?

yolocs: What I learned - just copy it from another folder.

cmd/broker/ingress/main.go
```diff
@@ -0,0 +1,48 @@
...
+	"knative.dev/pkg/logging"
+)
+
+func main() {
+	logCfg := zap.NewProductionConfig() // TODO: to replace with a dynamically updating logger.
+	logCfg.EncoderConfig.EncodeTime = zapcore.ISO8601TimeEncoder
+	logger, err := logCfg.Build()
+	if err != nil {
+		log.Fatalf("Unable to create logger: %v", err)
+	}
```
yolocs: You can also try

```
cfg, _ := logging.NewConfigFromMap(nil)
logger, _ = logging.NewLoggerFromConfig(cfg, "broker-ingress")
```

cmd/broker/ingress/main.go
```diff
@@ -0,0 +1,48 @@
...
+
+	ingress, err := ingress.NewHandler(ctx)
+	if err != nil {
+		logger.Fatal("Unable to create ingress handler: ", zap.Error(err))
+	}
+
+	logger.Info("Starting ingress.", zap.Any("ingress", ingress))
+	if err := ingress.Start(ctx); err != nil {
+		logger.Fatal("failed to start ingress: ", zap.Error(err))
+	}
```
yolocs: I believe it needs to handle term signal for graceful shutdown. So either add a TODO or just do it (we have example everywhere).

pkg/broker/ingress/handler.go
```diff
@@ -0,0 +1,101 @@
...
+
+func (h *handler) Start(ctx context.Context) error {
+	return h.inbound.StartReceiver(ctx, h.receive)
+}
+
+// receive receives events from inbound and sends them to decouple.
+func (h *handler) receive(ctx context.Context, event event.Event) protocol.Result {
+	// Code below is for ce v1.
+	// TODO(liu-cong) figure out tracing support for ce v2
+	// event = tracing.AddTraceparentAttributeFromContext(ctx, event)
```
yolocs: @ian-mi

cmd/broker/ingress/main.go
liu-cong: Can you share an example? I am not sure what else needs to be handled here

liu-cong: Sorry not familiar with this code base.  I tried to look for examples but didn't find one

yolocs: I think you can simply use `signals.NewContext()`, e.g. https://github.com/knative/eventing/blob/master/cmd/broker/ingress/main.go#L86

liu-cong: Done. Thanks.

cmd/broker/ingress/kodata/HEAD
```diff
@@ -0,0 +1 @@
+ref: refs/heads/brokeringress
```
grantr: The files in kodata should be symlinks, not actual files. Take a look at the other directories to see where they should link.

liu-cong: Oh didn't know that. Just did a `cp` without preserving links...

Please take a look now.

pkg/broker/ingress/handler.go
ian-mi: In order to add tracing create a NewObserved client using the WithTracePropagation option.

cmd/broker/ingress/kodata/HEAD
grantr: Looks good now, thanks!

pkg/broker/ingress/handler.go
```diff
@@ -0,0 +1,101 @@
...
+		pubsub.WithProjectIDFromDefaultEnv(),
+		pubsub.WithTopicIDFromDefaultEnv())
+	if err != nil {
+		return nil, err
+	}
+
+	// Use the pubsub prototol to make a new CloudEvents client.
+	return client.NewObserved(p,
+		client.WithUUIDs(),
+		client.WithTimeNow(),
```
ian-mi: Add WithTracePropagation to this client as well

pkg/broker/ingress/handler.go
```diff
@@ -0,0 +1,101 @@
...
+	inbound client.Client
+	// decouple is the client to send events to a decouple sink.
+	decouple DecoupleSink
+}
+
+func NewHandler(ctx context.Context, options ...Option) (*handler, error) {
+	h := &handler{}
+	var err error
+	// Create a default HTTP client for ingress.
+	if h.inbound, err = client.NewDefault(); err != nil {
```
ian-mi: Default clients should only be created if a client was not specified in options.

pkg/broker/ingress/handler.go
liu-cong: Thanks. Let me know what else needs to be done for tracing. I am also not sure what's the best way to test tracing before we have an e2e test.

pkg/broker/ingress/handler.go
liu-cong: Done.

pkg/broker/ingress/handler.go
liu-cong: Good call. Thanks.

liu-cong:
> Please add tests as well.

Yeah I was thinking about adding an integration test but haven't found a good example (like set up pubsub topics, pull msgs from pubsub, etc.). Existing components (e.g., pubsub publisher) seem to have e2e tests only (not too many interesting unit tests to add for now).

ian-mi:
Maybe start with a test which stubs the inbound and outbound clients?

liu-cong:
> Maybe start with a test which stubs the inbound and outbound clients?

@ian-mi 
I am not sure if this is a good investment of time because currently this code currently does nothing but just plumbs in the inbound and outbound clients. I think it's more useful to have an integration test. Unless you think otherwise, I prefer to spend time to add integ tests.

ian-mi:
> > Maybe start with a test which stubs the inbound and outbound clients?
> 
> @ian-mi
> I am not sure if this is a good investment of time because currently this code currently does nothing but just plumbs in the inbound and outbound clients. I think it's more useful to have an integration test. Unless you think otherwise, I prefer to spend time to add integ tests.

I don't think unit tests are wasted effort. Right now the behavior of ingress is fairly trivial, but in the future functionality may be added. Having testing in place makes it easier to add tests as features are added and ensures that changes don't break the basic functionality of ingress.

liu-cong:
> I don't think unit tests are wasted effort. Right now the behavior of ingress is fairly trivial, but in the future functionality may be added. Having testing in place makes it easier to add tests as features are added and ensures that changes don't break the basic functionality of ingress.

Agree it's not a waste of effort. I was just trying to say that given the behavior is trivial for now, it's more useful to have an integ test. But definitely as we add more logic, we need unit tests.

pkg/broker/ingress/options.go
```diff
@@ -0,0 +1,35 @@
...
+
+package ingress
+
+import (
+	"github.com/cloudevents/sdk-go/v2/client"
+)
+
+type Option func(*handler)
+
+func WithInboundClient(c client.Client) Option {
```
yolocs: nit: add comments in this file

liu-cong: done

cmd/broker/ingress/main.go
```diff
@@ -0,0 +1,42 @@
...
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/signals"
+)
+
+func main() {
+	// Since we pass nil, a default config with no error will be returned.
+	cfg, _ := logging.NewConfigFromMap(nil)
+	logger, _ := logging.NewLoggerFromConfig(cfg, "broker-ingress")
+	ctx := signals.NewContext()
+	logging.WithLogger(ctx, logger)
```
Harwayne:
```suggestion
	ctx = logging.WithLogger(ctx, logger)
```

cmd/broker/ingress/main.go
```diff
@@ -0,0 +1,42 @@
...
+
+	ingress, err := ingress.NewHandler(ctx)
+	if err != nil {
+		logger.Fatal("Unable to create ingress handler: ", zap.Error(err))
```
Harwayne:
```suggestion
		logger.Fatalw("Unable to create ingress handler: ", zap.Error(err))
```

The alternative to these is using `knative.dev/eventing/pkg/logging`, instead of `knative.dev/pkg/logging`, but that would require changing lines 28-29.

Or adding `Desugar()` to each call.

cmd/broker/ingress/main.go
```diff
@@ -0,0 +1,42 @@
...
+	}
+
+	logger.Info("Starting ingress.", zap.Any("ingress", ingress))
+	if err := ingress.Start(ctx); err != nil {
+		logger.Fatal("failed to start ingress: ", zap.Error(err))
```
Harwayne:
```suggestion
		logger.Fatalw("failed to start ingress: ", zap.Error(err))
```

cmd/broker/ingress/main.go
```diff
@@ -0,0 +1,42 @@
...
+	logger, _ := logging.NewLoggerFromConfig(cfg, "broker-ingress")
+	ctx := signals.NewContext()
+	logging.WithLogger(ctx, logger)
+
+	ingress, err := ingress.NewHandler(ctx)
+	if err != nil {
+		logger.Fatal("Unable to create ingress handler: ", zap.Error(err))
+	}
+
+	logger.Info("Starting ingress.", zap.Any("ingress", ingress))
```
Harwayne:
```suggestion
	logger.Infow("Starting ingress.", zap.Any("ingress", ingress))
```

pkg/broker/ingress/handler.go
```diff
@@ -0,0 +1,114 @@
...
+
+import (
+	"context"
+	"fmt"
+
+	cev2 "github.com/cloudevents/sdk-go/v2"
+	"github.com/cloudevents/sdk-go/v2/protocol"
+	"github.com/cloudevents/sdk-go/v2/protocol/pubsub"
+	"go.uber.org/zap"
+	"knative.dev/pkg/logging"
```
Harwayne:
```suggestion
	"knative.dev/eventing/pkg/logging"
```

pkg/broker/ingress/handler_test.go
```diff
@@ -0,0 +1,79 @@
...
+		t.Fatalf("Failed to create HTTP client: %+v", err)
+	}
+	for _, event := range events {
+		ce.Send(context.Background(), event)
+	}
+}
+
+// createAndStartIngress creates an ingress and calls its Start() method in a goroutine.
+func createAndStartIngress(t *testing.T, inbound cloudevents.Client, decouple DecoupleSink) *handler {
+	ctx := context.Background()
```
Harwayne: Does this need to get closed? If so, I recommend returning a 'cleanup' goroutine. So the function's return would become something like  `(*hander, func())` and the caller defers calling the clean up function.

ian-mi: [T.Cleanup](https://golang.org/pkg/testing/#T.Cleanup) makes this easy.

liu-cong: I couldn't find any `cleanUp` methods in the underlying clients so I guess no need to do so.

pkg/broker/ingress/handler.go
```diff
@@ -0,0 +1,114 @@
...
+}
+
+func (h *handler) Start(ctx context.Context) error {
+	return h.inbound.StartReceiver(ctx, h.receive)
+}
+
+// receive receives events from inbound and sends them to decouple.
+func (h *handler) receive(ctx context.Context, event cev2.Event) protocol.Result {
+	// TODO(liu-cong) add metrics
+	// TODO(liu-cong) route events based on broker.
```
grantr: I think routing to different broker topics is unblocked now that #725 is in

cmd/broker/ingress/main.go
grantr: I prefer the Desugared logger interface myself. You could just add this line after calling logging.WithLogger:

```go
logger = logger.Desugar()
```

liu-cong: This is in the setup path and performance is not critical, so I guess it's OK to use the sugared logger? I am also OK to change it to desugar though.

With regards to `Fatalw`, does that require key-value pairs? I am new to zap and not sure how does it look like.

liu-cong: Changed to `Desugar`.

pkg/broker/ingress/handler.go
liu-cong: OK will send a follow up PR for that. Trying to keep this one small.

pkg/broker/ingress/handler.go
```diff
@@ -73,7 +73,7 @@ func (h *handler) receive(ctx context.Context, event cev2.Event) protocol.Result
 	// TODO(liu-cong) route events based on broker.
 
 	if err := h.decouple.Send(ctx, event); err != nil {
-		logging.FromContext(ctx).Desugar().Error("Error publishing to PubSub", zap.String("event", event.String()), zap.Error(err))
+		logging.FromContext(ctx).Error("Error publishing to PubSub", zap.String("event", event.String()), zap.Error(err))
```
grantr: Shouldn't this also have Desugar()? Or is this logger already desugared coming from context?

pkg/broker/ingress/handler_test.go
Harwayne: You're right, looking through the code it doesn't need to happen because the context is only used to create the decoupler, which is passed in as an option here. 

My preference is to add a comment or create the clean up method anyway, but I won't hold up the PR for it.

pkg/broker/ingress/handler.go
liu-cong: I switched to `knative.dev/eventing/pkg/logging` which provides a desugared logger.

Harwayne:
I think I'm the last person who had open comments, if so feel free to cancel the hold.

pkg/broker/ingress/handler_test.go
liu-cong: SG. Added a placeholder cleanup function.

Issue #686, opened 2020-03-20
Add release artifact(s)


yolocs:
The installation should be optional and decoupled from the "core" installation.

PR #711, opened 2020-03-25
Add mains and yamls for security apis and controllers

yolocs:
Fixes https://github.com/google/knative-gcp/issues/686

## Proposed Changes

- Added yamls under `./config/security` so that it's separated from other stuff
- Fixed a couple of small issues
- Will add doc soon

**Release Note**

```release-note
Add mains and yamls for security apis and controllers
```


config/security/201-clusterrole.yaml
```diff
@@ -0,0 +1,107 @@
...
+- apiGroups:
+    - ""
+  resources:
+    - services
+  verbs:
+    - get
+    - list
+
+- apiGroups:
+    - "eventing.knative.dev"
```
Harwayne:
```suggestion
    - eventing.knative.dev
```

config/security/201-clusterrole.yaml
```diff
@@ -0,0 +1,107 @@
...
+  resources:
+    - "brokers"
```
Harwayne:
```suggestion
    - brokers
```

config/security/400-webhook-service.yaml
```diff
@@ -0,0 +1,28 @@
...
+    role: webhook
+    security.knative.dev/release: devel
+  name: webhook
+  namespace: knative-security
+spec:
+  ports:
+    - port: 443
+      targetPort: 8443
+  selector:
+    role: webhook
```
Harwayne:
```suggestion
    role: security-webhook
```
?

That might make it easier to put this in the same namespace as other things later.

config/security/500-controller.yaml
```diff
@@ -0,0 +1,61 @@
...
+  replicas: 1
+  selector:
+    matchLabels:
+      app: knative-security
+      role: security-controller
+  template:
+    metadata:
+      labels:
+        app: knative-security
+        role: security-controller
```
Harwayne:
```suggestion
    matchLabels: &labels
      app: knative-security
      role: security-controller
  template:
    metadata:
      labels: *labels
```

config/security/500-webhook.yaml
```diff
@@ -0,0 +1,61 @@
...
+  selector:
+    matchLabels:
+      app: knative-security
+      role: webhook
+  template:
+    metadata:
+      labels:
+        app: knative-security
+        role: webhook
+        security.knative.dev/release: devel
```
Harwayne: This is not present on the webhook Deployment's Pods. Should it be?

config/security/config-logging.yaml
```diff
@@ -0,0 +1,71 @@
...
+          "timeEncoder": "iso8601",
+          "durationEncoder": "",
+          "callerEncoder": ""
+        }
+      }
+
+    # Log level overrides
+    # For all components except the autoscaler and queue proxy,
+    # changes are be picked up immediately.
+    # For autoscaler and queue proxy, changes require recreation of the pods.
```
Harwayne: Not really needed here, but we can keep just to stay in lock step with Knative Serving if you want.

cmd/security-webhook/main.go
```diff
@@ -0,0 +1,130 @@
+/*
+Copyright 2020 The Knative Authors
```
Harwayne:
```suggestion
Copyright 2020 Google LLC
```

I think...

cmd/security-webhook/main.go
```diff
@@ -0,0 +1,130 @@
...
+func NewDefaultingAdmissionController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+	// Decorate contexts with the current state of the config.
+	ctxFunc := func(ctx context.Context) context.Context {
+		return ctx
+	}
+
+	return defaulting.NewAdmissionController(ctx,
+
+		// Name of the default webhook.
+		"webhook.security.knative.dev",
```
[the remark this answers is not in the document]
yolocs: It's in a weird state now so I'm open to suggestions. Here is a little background.

Security policy started as something for Eventing but gradually evolved to have a bigger scope - between security boundaries events are transported with http; to provide events policies problem essentially we would need to rely on http policies; and http policy is generic. So in the design we proposed the feature track to be knative-scope rather than eventing-scope. That's why I'm putting `security.knative.dev` here and as the API group name.

However, the progress of the feature track has been really slow in the community. So we decided to put it in knative-gcp first so that we can at least start coding and provide it for google users. That's why the code ended up in knative-gcp.

I struggled with the naming because ideally it would be `security.knative.dev` but it's not something knative officially supports today. Naming it with `events.cloud.google.com` also seems not appropriate because it's not meant to be just for eventing.

config/security/500-webhook.yaml
yolocs: I copied it over from the other webhook yaml. I think the idea is to have `*/release: devel` label on every resource we create?

yolocs:
@Harwayne If you don't mind, I'd like to keep the `security.knative.dev` in the PR and do a batch update in the follow-up one.

cmd/security-webhook/main.go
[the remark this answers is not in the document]
Harwayne: I don't know how we are allowed to use the `knative.dev` namespace if this isn't within the knative project. I would lean towards something else, especially if we know we want to change it later anyway (e.g. to remove `experimental`). How about `security.experimental.events.cloud.google.com`? My reasoning being that we are allowed to use `events.cloud.google.com`. I agree that it implies that this security is only for eventing, but as we intend to change the name in the future anyway, I don't see that as a problem while we are experimenting.

config/security/100-namespace.yaml
```diff
@@ -0,0 +1,20 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: knative-security
```
Harwayne: Similar thoughts, should this be named something else? E.g. `cloud-run-events-security`.

config/security/500-webhook.yaml
Harwayne: That makes sense to me. What I meant to point out is that the Webhook and Controller's Pods are inconsistent. The Webhook Pods have this label. The Controller Pods do not. I would prefer we either have the label in both or neither.

cmd/security-webhook/main.go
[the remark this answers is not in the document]
yolocs: Track in https://github.com/google/knative-gcp/issues/726

config/security/100-namespace.yaml
yolocs: Track in https://github.com/google/knative-gcp/issues/726

cmd/security-webhook/main.go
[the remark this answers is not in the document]
nachocano: one extra thing... we are using `messaging.cloud.google.com` and `pubsub.cloud.google.com` for Channel, and Topic/PullSubscription respectively... Based on @Harwayne 's point, I think we also need to change it to `messaging.events.cloud.google.com` and `pubsub.events.cloud.google.com`... 
@yolocs if you loop in Ron, can you ask about this as well? We rather do the whole renaming now in v1alpha1 before we move all to v1beta1. 
And I think I'd vote for `security.events.cloud.google.com`, without the `experimental`, but no strong opinions.

yolocs:
/unhold

Track naming change in https://github.com/google/knative-gcp/issues/726

Issue #712, opened 2020-03-25
Topic reconciler checks for existence with the wrong name


grantr:
**Describe the bug**
This call should be using the topic name given in the Topic spec as argument, but it's using the ProjectID instead:
https://github.com/google/knative-gcp/blob/45d9dbd44358730aa34474225fa2969c12ff1d43/pkg/reconciler/pubsub/topic/topic.go#L127

Since the existence check is checking for the wrong topic, the reconciler will always attempt to create the topic. This works because we have error handling for this case:

https://github.com/google/knative-gcp/blob/45d9dbd44358730aa34474225fa2969c12ff1d43/pkg/reconciler/pubsub/topic/topic.go#L142-L144

Maybe this comment can be removed and the logic simplified when this bug is fixed.

 **Expected behavior**
The topic reconciler checks for the existence of a topic before trying to create it.

 **To Reproduce**

 **Knative-GCP release version**
0.13.0

**Additional context**
This bug was introduced in #419. The previous operation-based reconciler was correct:

https://github.com/google/knative-gcp/blob/d02c888d059ec547a3812538214aa19deb321b35/pkg/operations/pubsub/topic.go#L130

/cc @nachocano

nachocano:
oops! Thanks Grant, let me fix this and clean up some unnecessary stuff as well

/assign

PR #713, opened 2020-03-25
Fixing TopicId when creating Topic

nachocano:
Fixes https://github.com/google/knative-gcp/issues/712

## Proposed Changes

- Didn't update the error logic below as I cannot test it. There were some weird things in storage as well, so I'd rather play it safe. 

**Release Note**

```release-note
NONE
```

**Docs**

### DIFF
--- a/pkg/reconciler/pubsub/topic/topic.go
+++ b/pkg/reconciler/pubsub/topic/topic.go
@@ -124,7 +124,7 @@ func (r *Reconciler) reconcileTopic(ctx context.Context, topic *v1alpha1.Topic) 
 	}
 	defer client.Close()
 
-	t := client.Topic(topic.Status.ProjectID)
+	t := client.Topic(topic.Status.TopicID)
 	exists, err := t.Exists(ctx)
 	if err != nil {
 		logging.FromContext(ctx).Desugar().Error("Failed to verify Pub/Sub topic exists", zap.Error(err))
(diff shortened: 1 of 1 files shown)


Harwayne:
From the test failure:

```
"message": "Failed to reconcile Pub/Sub topic: rpc error: code = InvalidArgument desc = Invalid resource name given (name=projects/knative-boskos-18/topics/). Refer to https://cloud.google.com/pubsub/docs/admin#resource_names for more information.",
            "reason": "TopicReconcileFailed",
```

Is it possible that the `topic.Status.TopicID` is an empty string?

pkg/reconciler/pubsub/topic/topic.go
```diff
@@ -124,7 +124,7 @@ func (r *Reconciler) reconcileTopic(ctx context.Context, topic *v1alpha1.Topic)
 	}
 	defer client.Close()
 
-	t := client.Topic(topic.Status.ProjectID)
+	t := client.Topic(topic.Status.TopicID)
```
Harwayne:
```suggestion
	t := client.Topic(topic.Spec.TopicID)
```

I think.

nachocano: that's right, had the wrong fix...
Thanks @Harwayne 
giving it another try now

nachocano:
There is a problem with a caching internal image from serving nightly. Same
happened in the sockpuppet PRs

Issue #726, opened 2020-03-27
Update API group name and related naming


yolocs:
Currently `security.knative.dev` is used. Since it's not official yet, we need something more appropriate.

PR #732, opened 2020-03-30
Rename security.knative.dev -> policy.run.google.cloud.com

yolocs:
Fixes https://github.com/google/knative-gcp/issues/726

## Proposed Changes

- API group name and where it's used
- Go package name
- Installation namespace and labels
- (Piggybacking) Allow controller service account to update k8s Event

**Release Note**

```release-note
Rename security.knative.dev -> policy.run.google.cloud.com
```


yolocs:
FYI @Harwayne

config/policy/201-clusterrole.yaml
```diff
@@ -61,6 +61,13 @@ rules:
   verbs:
     - get
 
+- apiGroups: [“”]
```
Harwayne: This looks unrelated to the renaming. Is it here intentionally?

yolocs: It is intentionally. During debugging I found the controller needs to have Event permission. So piggybacking here :D

Harwayne: For changes of this size (almost three hundred files...), I prefer to not piggy back things like this. 

If you still want to, make sure to mention it in the PR's description. My reasoning is that if we need to roll this back (probably not for this particular PR, but somewhat common for large renaming changes), we don't want a functionality regression at the same time as a large name change. And based on the current description, I wouldn't expect any functionality changes, just name changes. So someone might do a straight rollback and not realize the problem.

yolocs: Ack. Description updated.

PR #737, opened 2020-03-30
Updating workload identity install docs 

### LINKED ISSUE #719: Issues with workload identity install docs
**Describe the bug**
Following the readme, I ran into these issues:

### Paths in commands
`controller-gke.yaml` is in `config/core/deployments` so the command `ko apply -f controller-gke.yaml` should include the directory from the repo root.

The `init_control_plane*` scripts are in the `hack` directory, so the commands to run them should be prefixed with `hack/` rather than `./`.

### gcloud configuration dependency

The `init_control_plane_gke.sh` script fails if the cluster is not in the gcloud configured zone:

```sh
 $ hack/init_control_plane_gke.sh
Operation "operations/acf.10358169-34c2-481f-85cc-e874e790cc93" finished successfully.
Operation "operations/acf.88e79cc0-eea7-4ea9-b197-c55b589c5a5e" finished successfully.
WARNING: Problem loading details of cluster to update:

ResponseError: code=404, message=Not found: projects/grantrodgers-knative-dev/zones/us-central1-b/clusters/gcpbroker-1.
Could not find [gcpbroker-1] in [us-central1-b].
Did you mean [gcpbroker-1] in [us-west1-c]?
```
In this case gcloud was configured for us-central1-b, but my cluster was in us-west1-c. I had to change the gcloud config to be us-west1-c. This is fine but it should be mentioned in the docs.

### pre-existing service account

I already had a `cloud-run-events` service account in my project. This generated errors in the `init_control_plane_gke.sh` script, which may be safe to ignore, but the script doesn't tell me. Ideally the script would be idempotent so this wouldn't be an issue, but maybe that's difficult.

```
ERROR: (gcloud.iam.service-accounts.create) Resource in project [grantrodgers-knative-dev] is the subject of a conflict: Service account cloud-run-events already exists within project projects/grantrodgers-knative-dev.

...

error: --overwrite is false but found the following declared annotation(s): 'iam.gke.io/gcp-service-account' already has a value (cloud-run-events@grantrodgers-knative-dev.iam.gserviceaccount.com)
```

I got a similar error on running `init_control_plane.sh`, which would presumably be encountered by everyone who runs both of these scripts:
```
ERROR: (gcloud.iam.service-accounts.create) Resource in project [grantrodgers-knative-dev] is the subject of a conflict: Service account cloud-run-events already exists within project projects/grantrodgers-knative-dev.
```

 **Expected behavior**
No errors when following the instructions, or documentation of common errors encountered and what to do about them.

 **To Reproduce**
Create a cluster and follow the workload identity instructions.

HEAD

 **Additional context**

grac3gao:
Fixes #719 

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


docs/install/install-knative-gcp.md
```diff
@@ -72,16 +76,19 @@
 
         ```shell
         chmod +x init_control_plane_gke.sh
```
Harwayne: Make the file executable in this PR, then remove this line.

docs/install/install-knative-gcp.md
```diff
@@ -72,16 +76,19 @@
...
-        ./init_control_plane_gke.sh
+        ./hack/init_control_plane_gke.sh
         ```
 
       - Export service account keys and store them as **Kubernetes Secrets**.
         Apply [init_control_plane](../../hack/init_control_plane.sh) to install
         all the configuration by running:
         ```shell
         chmod +x init_control_plane.sh
```
Harwayne: I think this line is unneeded. It looks like the file is already executable.

docs/install/install-knative-gcp.md
grac3gao: This file looks like already executable too, as I run the chmod +x locally, nothing in HEAD changed.

docs/install/install-knative-gcp.md
```diff
@@ -60,28 +64,29 @@
         In order to make controller compatible with Workload Identity, use
         [ko](http://github.com/google/ko) to apply
         [controller-gke](../../config/core/deployments/controller-gke.yaml)
-        first.
+        first. It is in `config/core/deployments`.
```
yolocs, line 67: nit: since you already have the full ko command below, do we really state it here again?

docs/install/install-knative-gcp.md
```diff
@@ -60,28 +64,29 @@
...
 
         ```shell
-        ko apply -f controller-gke.yaml
+        ko apply -f config/core/deployments/controller-gke.yaml
         ```
 
         Then you can apply
         [init_control_plane_gke](../../hack/init_control_plane_gke.sh) to
         install all the configuration by running:
```
yolocs, line 75: Missing some context here... Do we have need some brief explanation on what the script does?

grac3gao: The doc includes two methods to configure it, initialization scripts and manual steps. What the script does is the same as the manual step. In the script, it also has comments for explanation. I feel explanation is redundant here...

PR #753, opened 2020-04-01
Add recordevents as Required Test Image

ian-mi:
This will ensure that the image remains vendored after update-deps


yolocs:
Thanks for fixing!

Issue #637, opened 2020-03-12
Implement Broker fanout basic functionality


grantr:
* Receive events from decoupling topic
* Publish to matching consumers (with timeout) executing filters to determine match
* If consumer timeouts expire, publish to correct retry topics then ack to decoupling topic
* If reply arrives, publish to decoupling topic or ingress (efficiency improvement if ingress is implemented in same process)
* Unit tests

/area broker

yolocs:
In proker, this is where it does the fanout https://github.com/yolocs/proker/blob/master/pkg/fanout/handler.go
It doesn't handle timeout or graceful shutdown properly.

Harwayne:
It seems like this has been accomplished. Can this issue be closed?

yolocs:
There are a couple of minor things left. For proper planning, I will close this one and create separate issues for them.

PR #761, opened 2020-04-02
Initial fanout service handlers

yolocs:
Ref https://github.com/google/knative-gcp/issues/637

## Proposed Changes

- `./pkg/broker/handler/processors` attempts to build reusable modules for both fanout and retry (e.g. filter and deliver are similar for both)
- `./pkg/broker/handler/pool` defines a sync pool where it will watch `TargetsConfig` change and reconcile the pool of handlers 

**Release Note**

```release-note
Initiated fanout service handler
```


yolocs:
@grantr @ericlem @grac3gao @liu-cong

pkg/broker/handler/pool/fanout/pool.go
```diff
@@ -0,0 +1,134 @@
...
+			logging.FromContext(ctx).Error("failed to start event handler", zap.String("broker", bk), zap.Error(err))
+			errs++
+			return true
+		}
+
+		p.pool.Store(bk, h)
+		return true
+	})
+
+	p.pool.Range(func(key, value interface{}) bool {
```
liu-cong, line 121: Run this before the `RangeBrokers` to potentially save a few cycles

pkg/broker/handler/processors/fanout/processor.go
```diff
@@ -0,0 +1,120 @@
...
+		defer close(out)
+		for target := range tc {
+			ctx = handlerctx.WithTarget(ctx, target)
+			out <- &fanoutResult{err: p.Next().Process(ctx, event)}
+		}
+	}()
+	return out
+}
+
+func (p *Processor) mergeResults(ctx context.Context, resChs []<-chan *fanoutResult) error {
```
liu-cong, line 92: Not sure if knative has similar utilities but k8s has a workqueue utility that helps parallelize workers:

https://github.com/kubernetes/kubernetes/blob/a10021787b2d037e3c1d44a205980fd48ac23eff/staging/src/k8s.io/client-go/util/workqueue/parallelizer.go#L30

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,97 @@
...
+	// cancel is function to stop pulling messages.
+	cancel context.CancelFunc
+}
+
+// Start starts the handler.
+func (h *Handler) Start(ctx context.Context) error {
+	ctx, h.cancel = context.WithCancel(ctx)
+	var err error
+	go func() {
+		err = h.PubsubEvents.OpenInbound(ctx)
```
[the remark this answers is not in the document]
yolocs: OpenInbound is a blocking call so it won't return. But you're right I need a better way to handle the err.

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,97 @@
...
+	"go.uber.org/zap"
+	"knative.dev/eventing/pkg/logging"
+)
+
+// Handler pulls Pubsub messages as events and process them
+// with chain of processors.
+type Handler struct {
+	// PubsubEvents is the CloudEvents Pubsub protocol to pull
+	// messages as events.
+	PubsubEvents *pubsub.Protocol
```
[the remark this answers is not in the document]
liu-cong: I just don't want to be confused with `event from pusbsub` as this is not the event itself.

pkg/broker/handler/processors/filter/processor.go
```diff
@@ -0,0 +1,89 @@
...
+func (p *Processor) Process(ctx context.Context, event *event.Event) error {
+	target := handlerctx.GetTarget(ctx)
+	if target.FilterAttributes == nil {
+		return p.Next().Process(ctx, event)
+	}
+
+	if p.passFilter(ctx, target.FilterAttributes, event) {
+		return p.Next().Process(ctx, event)
+	}
+	logging.FromContext(ctx).Info("event is filtered for target", zap.Any("target", target))
```
Harwayne, line 46:
```suggestion
	logging.FromContext(ctx).Info("event does not pass filter for target", zap.Any("target", target))
```

pkg/broker/handler/processors/filter/processor_test.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/cloudevents/sdk-go/v2/event"
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/google/knative-gcp/pkg/broker/config"
+	handlerctx "github.com/google/knative-gcp/pkg/broker/handler/context"
+	"github.com/google/knative-gcp/pkg/broker/handler/processors"
+)
+
+func TestFilterProcessor(t *testing.T) {
```
Harwayne, line 31: My preference would be to rewrite this as a table test, my guess is that the test struct would look something like:

```golang
[]struct {
    filterAttributes *map[string]string
    event ce.Event
    shouldPass bool
}
```

pkg/broker/handler/processors/filter/processor_test.go
```diff
@@ -0,0 +1,100 @@
...
+				},
+			},
+		)
+		pass := event.New()
+		pass.SetID("id")
+		pass.SetSubject("foo")
+		pass.SetType("bar")
+		notpass := event.New()
+
+		go func() {
```
Harwayne, line 83:
```suggestion
		defer func() {
```

I don't think you want to use `go` here. I think you want `defer`. 

```golang
func TestGo(t *testing.T) {
	go func() {
		t.Fatal("hi")
	}()
}

func TestDefer(t *testing.T) {
	defer func() {
		t.Fatal("hi")
	}()
}
```
Only `TestDefer` fails, `TestGo` succeeds. My guess is that the test's success is determined immediately after the test returns. Any subsequent calls to `t.Fatal` are ignored (such as those from the goroutine).

This will apply to all tests that have a goroutine that call `t.Fatal` or `t.Error` without some other form of concurrency control.

pkg/broker/handler/processors/filter/processor_test.go
```diff
@@ -0,0 +1,100 @@
...
+			&config.Target{
+				Name:      "name",
+				Namespace: "namespace",
+			},
+		)
+		e := event.New()
+		e.SetID("id")
+		e.SetSource("source")
+
+		go func() {
```
Harwayne, line 50:
```suggestion
		defer func() {
```

pkg/broker/handler/processors/processor.go
```diff
@@ -0,0 +1,84 @@
...
+
+	// WithNext sets the next Processor to pass the event.
+	WithNext(ChainableProcessor) ChainableProcessor
+}
+
+// BaseProcessor provoides implementation to set and get
+// next processor. It can gracefully handle the case where the next
+// processor doesn't exist.
+type BaseProcessor struct {
+	N ChainableProcessor
```
Harwayne, line 46: Should this be a public field? I would expect it to be private.

pkg/broker/handler/processors/processor.go
```diff
@@ -0,0 +1,84 @@
...
+var noop = &noOpProcessor{}
+
+type noOpProcessor struct{}
+
+func (p noOpProcessor) Process(_ context.Context, _ *event.Event) error {
+	return nil
+}
+
+func (p noOpProcessor) Next() Processor {
+	return noop
```
Harwayne, line 83: Perhaps this should return `nil` instead? Or a failing processor? My reasoning being that it would be fairly easy to get into an infinite loop as-is, by continuously calling Next().

pkg/broker/handler/context/broker.go
```diff
@@ -0,0 +1,56 @@
...
+// WithBrokerKey sets a broker key in the context.
+func WithBrokerKey(ctx context.Context, key string) context.Context {
+	return context.WithValue(ctx, brokerKeyKey{}, key)
+}
+
+// GetBrokerKey gets the broker key from the context.
+func GetBrokerKey(ctx context.Context) string {
+	untyped := ctx.Value(brokerKeyKey{})
+	if untyped == nil {
+		logging.FromContext(ctx).Panic("Unable to fetch Broker key from context.")
```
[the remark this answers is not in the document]
yolocs: This is a pattern widely used in informer injection code. But I think you're right. Our case is not quite the same. Will change to return zero value instead.

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// Stop stops the handlers.
+func (h *Handler) Stop() {
+	h.cancel()
+}
+
+func (h *Handler) handle(ctx context.Context) {
+	for {
+		// TODO(yolocs): we need to update dep for fix:
```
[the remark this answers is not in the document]
yolocs: I don't feel the need at the moment because there isn't too much logic here. We can always change if we find it's confusing.

pkg/broker/handler/pool/options.go
```diff
@@ -0,0 +1,117 @@
...
+	if err != nil {
+		panic(err)
+	}
+	c, err := cev2.NewClientObserved(p,
+		ceclient.WithUUIDs(),
+		ceclient.WithTimeNow(),
+		ceclient.WithTracePropagation(),
+	)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
yolocs: I now changed it to be in `init()` func. All I want is a default value. If init panics, then let's abort.

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -0,0 +1,59 @@
...
+	// Requester is the cloudevents client to send events.
+	Requester ceclient.Client
+}
+
+// Process delivers the event based on the broker/target in the context.
+func (p *Processor) Process(ctx context.Context, event *event.Event) error {
+	broker := handlerctx.GetBroker(ctx)
+	target := handlerctx.GetTarget(ctx)
+
+	resp, res := p.Requester.Request(cecontext.WithTarget(ctx, target.Address), *event)
```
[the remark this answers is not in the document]
yolocs: My idea it to have a called of Processor(s) to set the timeout in the context. Like the one in `handler.go`

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -0,0 +1,59 @@
...
+	if !protocol.IsACK(res) {
+		// TODO(yolocs): Have option to send to retry.
+		return fmt.Errorf("deliver event failed with error: %v", res.Error())
+	}
+	if resp == nil {
+		return nil
+	}
+
+	if res := p.Requester.Send(cecontext.WithTarget(ctx, broker.Address), *resp); !protocol.IsACK(res) {
+		return fmt.Errorf("deliver replied event failed with error: %v", res.Error())
```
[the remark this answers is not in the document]
yolocs: res.Error() is `string` not `error`

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -0,0 +1,59 @@
...
+
+// Process delivers the event based on the broker/target in the context.
+func (p *Processor) Process(ctx context.Context, event *event.Event) error {
+	broker := handlerctx.GetBroker(ctx)
+	target := handlerctx.GetTarget(ctx)
+
+	resp, res := p.Requester.Request(cecontext.WithTarget(ctx, target.Address), *event)
+	if !protocol.IsACK(res) {
+		// TODO(yolocs): Have option to send to retry.
+		return fmt.Errorf("deliver event failed with error: %v", res.Error())
```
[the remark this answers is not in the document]
yolocs: res.Error() is `string` not `error`

pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -0,0 +1,165 @@
...
+		if err := resp(ctx, binding.ToMessage(&reply), protocol.ResultACK); err != nil {
+			t.Errorf("unexpected error from target responding event: %v", err)
+		}
+		defer msg.Finish(nil)
+		gotEvent, err := binding.ToEvent(ctx, msg)
+		if err != nil {
+			t.Errorf("target received message cannot be converted to an event: %v", err)
+		}
+		// Force the time to be the same so that we can compare easier.
+		gotEvent.SetTime(origin.Time())
```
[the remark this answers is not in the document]
yolocs: The problem is event is an interface rather than a struct.

pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -0,0 +1,165 @@
...
+		msg, err := ingressClient.Receive(ctx)
+		if err != nil {
+			t.Errorf("unexpected error from ingress receiving event: %v", err)
+		}
+		defer msg.Finish(nil)
+		gotEvent, err := binding.ToEvent(ctx, msg)
+		if err != nil {
+			t.Errorf("ingress received message cannot be converted to an event: %v", err)
+		}
+		// Force the time to be the same so that we can compare easier.
```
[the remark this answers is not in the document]
yolocs: The problem is event is an interface rather than a struct.

pkg/broker/handler/processors/fake_test.go
```diff
@@ -0,0 +1,124 @@
...
+		gotEvent := <-ch
+		if diff := cmp.Diff(&event, gotEvent); diff != "" {
+			t.Errorf("processed event (-want,+got): %v", diff)
+		}
+	}()
+
+	if err := p.Process(context.Background(), &event); err == nil {
+		t.Error("expected error when AlwaysErr is set to true")
+	}
+	close(ch)
```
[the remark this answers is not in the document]
yolocs: I remember I stuck with some goroutine if I don't do this. Let me see if I can reproduce.

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,95 @@
...
+		}
+
+		event, err := binding.ToEvent(ctx, msg)
+		if err != nil {
+			logging.FromContext(ctx).Error("failed to convert received message to an event", zap.Error(err))
+			continue
+		}
+
+		// Don't block new events.
+		go func() {
```
grantr, line 80: Is this really necessary? Are we sure that handle() isn't already called in its own goroutine?

pkg/broker/handler/processors/fanout/processor.go
```diff
@@ -0,0 +1,120 @@
...
+	return p.mergeResults(ctx, resChs)
+}
+
+func (p *Processor) fanoutEvent(ctx context.Context, event *event.Event, tc <-chan *config.Target) <-chan *fanoutResult {
+	out := make(chan *fanoutResult)
+	go func() {
+		defer close(out)
+		for target := range tc {
+			ctx = handlerctx.WithTarget(ctx, target)
+			out <- &fanoutResult{err: p.Next().Process(ctx, event)}
```
grantr, line 86: Looks like this processor has no timeout handling of its own, so it will wait indefinitely for all iterations of the next processor in the chain. Is that intentional? Will the next processor always time out with the context?

pkg/broker/handler/processors/fake_test.go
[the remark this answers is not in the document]
yolocs: I removed most `close`. For those ones left I need them so that `for e := range ch` in the defer func can finish.

pkg/broker/handler/processors/filter/processor_test.go
yolocs: Good to know!

pkg/broker/handler/handler.go
yolocs: I believe so. Looking at the pubsub protocol code, even though the pubsub client pulls messages with multiple goroutines, they all end up in the same channel. The `Receive` func simply reads from the channel.

pkg/broker/handler/processors/fanout/processor.go
yolocs: That is intentional. The idea is that if a processor can rely on its successor(s) for timeout, then it shouldn't explicitly handle timeout. If a processor has some internal long running processing, then it must handle timeout by itself. Is that reasonable?

grantr: That sounds reasonable. Can you add a comment explaining that here?

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Handler pulls Pubsub messages as events and processes them
+// with chain of processors.
+type Handler struct {
+	// PubsubEvents is the CloudEvents Pubsub protocol to pull
+	// messages as events.
+	PubsubEvents *pubsub.Protocol
+
+	// Processor is the processor to process events.
+	Processor processors.Processor
```
liu-cong, line 38: nit: `processors.Processor` reads weirdly. Better to come up with a different name if you can.

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,95 @@
...
+		// https://github.com/cloudevents/sdk-go/pull/430
+		msg, err := h.PubsubEvents.Receive(ctx)
+		if err != nil {
+			logging.FromContext(ctx).Error("failed to receive the next message from Pubsub", zap.Error(err))
+			continue
+		}
+
+		event, err := binding.ToEvent(ctx, msg)
+		if err != nil {
+			logging.FromContext(ctx).Error("failed to convert received message to an event", zap.Error(err))
```
liu-cong, line 75: Helpful to also log the msg (Not sure if the err contains the msg).

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,95 @@
...
+				var cancel context.CancelFunc
+				pctx, cancel = context.WithTimeout(ctx, h.Timeout)
+				defer cancel()
+			}
+			err := h.Processor.Process(pctx, event)
+			if err != nil {
+				logging.FromContext(pctx).Error("failed to process event", zap.Any("event", event), zap.Error(err))
+			}
+			// This will ack/nack the message.
+			msg.Finish(err)
```
liu-cong, line 92: Should we log a message if `Finish` returns an error?

pkg/broker/handler/handler.go
liu-cong: `handle()` is called in its goroutine, but here in `handle`, we also want to parallelize processing of events, so I also think the goroutine is needed here.

However, I was wondering if we should control the total number of goroutines here. Suppose we can pull events from pubsub very quickly while the `Processors` are slow, we may end up with OOM.

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,95 @@
...
+	go h.handle(ctx)
+}
+
+// Stop stops the handlers.
+func (h *Handler) Stop() {
+	h.cancel()
+}
+
+func (h *Handler) handle(ctx context.Context) {
+	for {
```
liu-cong, line 64: I think we should terminate the loop if ctx is cancelled.

pkg/broker/handler/handler.go
```diff
@@ -0,0 +1,95 @@
...
+	// Timeout is the timeout for processing each individual event.
+	Timeout time.Duration
+
+	// cancel is function to stop pulling messages.
+	cancel context.CancelFunc
+}
+
+// Start starts the handler.
+// done func will be called if the pubsub inbound is closed.
+func (h *Handler) Start(ctx context.Context, done func(error)) {
```
liu-cong, line 49: What do you expect the `done` func to do? If it just cancels the context, maybe just add the `done` logic directly here so caller doesn't need to care about it.

pkg/broker/handler/handler_test.go
```diff
@@ -0,0 +1,137 @@
...
+func TestHandler(t *testing.T) {
+	ctx := context.Background()
+	c, close := testPubsubClient(ctx, t, "test-project")
+	defer close()
+
+	topic, err := c.CreateTopic(ctx, "test-topic")
+	if err != nil {
+		t.Fatalf("failed to create topic: %v", err)
+	}
+	if _, err := c.CreateSubscription(ctx, "test-sub", pubsub.SubscriptionConfig{
```
liu-cong, line 63: Do we need to create the topic and subscription here? Won't `cepubsub.New` create them?

pkg/broker/handler/pool/fanout/pool.go
```diff
@@ -0,0 +1,139 @@
...
+		}
+		// Start the handler with broker key in context.
+		h.Start(handlerctx.WithBrokerKey(ctx, bk), func(err error) {
+			if err != nil {
+				logging.FromContext(ctx).Error("handler for broker has stopped with error", zap.String("broker", bk), zap.Error(err))
+			} else {
+				logging.FromContext(ctx).Info("handler for broker has stopped", zap.String("broker", bk))
+			}
+			// Make sure the handler is deleted from the pool.
+			p.pool.Delete(h)
```
liu-cong, line 127: I think `h.Stop()` should be called too.

pkg/broker/handler/pool/fanout/pool.go
```diff
@@ -0,0 +1,139 @@
...
+			return
+		case <-p.options.SyncSignal:
+			if err := p.sync(ctx); err != nil {
+				logging.FromContext(ctx).Error("failed to sync handlers pool on watch signal", zap.Error(err))
+			}
+		}
+	}
+}
+
+func (p *SyncPool) sync(ctx context.Context) error {
```
liu-cong, line 76: nit: syncOnce

pkg/broker/handler/pool/fanout/pool_test.go
```diff
@@ -0,0 +1,124 @@
...
+	"time"
+
+	"github.com/google/go-cmp/cmp"
+
+	"github.com/google/knative-gcp/pkg/broker/config"
+	"github.com/google/knative-gcp/pkg/broker/config/memory"
+	"github.com/google/knative-gcp/pkg/broker/handler/pool"
+)
+
+func TestSync(t *testing.T) {
```
liu-cong, line 32: nit: `TestWatchAndSync`  because `sync` only syncs once.

pkg/broker/handler/pool/options.go
```diff
@@ -0,0 +1,119 @@
...
+)
+
+var (
+	defaultMaxConcurrency = 1
+	defaultTimeout        = 10 * time.Minute
+)
+
+var defaultCeClinet ceclient.Client
+
+func init() {
```
liu-cong, line 34: If the CeClient can be provided via options, then probably shouldn't always initialize it.

pkg/broker/handler/pool/options.go
```diff
@@ -0,0 +1,119 @@
...
+	cev2 "github.com/cloudevents/sdk-go/v2"
+	ceclient "github.com/cloudevents/sdk-go/v2/client"
+)
+
+var (
+	defaultMaxConcurrency = 1
+	defaultTimeout        = 10 * time.Minute
+)
+
+var defaultCeClinet ceclient.Client
```
liu-cong, line 32: typo: defaultCeClient

pkg/broker/handler/handler.go
yolocs: Pretty bad at naming. The other one I can think of `processors.Interface`

pkg/broker/handler/handler.go
yolocs: I don't think that's necessary. IIUC `Receive` will cancel itself if context is done.

pkg/broker/handler/handler.go
yolocs: For proper error handling. See example in the fanout/pool.go

pkg/broker/handler/handler_test.go
yolocs: I don't want to use the cloudevents sdk to create pubsub stuff because in reality we won't reach that code path.

pkg/broker/handler/pool/fanout/pool.go
yolocs: To reach the code here, it means the handler has already stopped.

pkg/broker/handler/pool/options.go
yolocs: What's the problem with that?

pkg/broker/handler/handler.go
yolocs: Good point. Let me add that.

pkg/broker/handler/processors/filter/processor.go
```diff
@@ -0,0 +1,92 @@
...
+		return err
+	}
+	if target.FilterAttributes == nil {
+		return p.Next().Process(ctx, event)
+	}
+
+	if p.passFilter(ctx, target.FilterAttributes, event) {
+		return p.Next().Process(ctx, event)
+	}
+	logging.FromContext(ctx).Info("event does not pass filter for target", zap.Any("target", target))
```
liu-cong, line 49: This may cause excessive logging. May be `Debug`? And probably also log the event itself

pkg/broker/handler/handler.go
liu-cong: Yeah `Interface` sounds better to me.

pkg/broker/handler/pool/options.go
liu-cong: Avoid creating unused resource. Another issue is that this code can panic, while it doesn't have to if the caller already provides a `Requester`.

pkg/broker/handler/pool/fanout/pool.go
liu-cong: I am talking more about the `err !=  nil` case. If for some reason the code fails to `OpenInbound` for pubsub, we should stop the handler I believe.

pkg/broker/handler/pool/options.go
yolocs: I think such unused resource is ok since the cost is close to zero.

pkg/broker/handler/processors/fanout/processor.go
```diff
@@ -0,0 +1,126 @@
...
+			tc <- target
+		}
+	}()
+
+	curr := len(broker.Targets)
+	if curr > p.MaxConcurrency {
+		curr = p.MaxConcurrency
+	}
+
+	var resChs []<-chan *fanoutResult
```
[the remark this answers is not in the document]
yolocs: Why? I feel append is easier.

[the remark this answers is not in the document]
liu-cong: Pre-allocate memory if you know the length. You can still use append if you do `make([]<-chan *fanoutResult, 0,  curr)`

yolocs:
@capri-xiyue this integ seems keeping failing on cloud build source?

pkg/broker/handler/pool/fanout/pool.go
liu-cong: Reopening this as I still think there is a chance the handler is not stopped.

pkg/broker/handler/handler.go
liu-cong: It's up to you but I would still recommend doing it.  That way even if the dependencies don't handle context properly, we can terminate as early as we could.

pkg/broker/handler/pool/fanout/pool.go
yolocs: If `OpenInbound` returns with error instantly, then it *should* mean that the pull was not started. Or at least the handler itself should only call `done` when it's already stopped. For safety, I will call `stop` in the handler once OpenInbound returns. This shouldn't have any side effect.

pkg/broker/handler/handler.go
yolocs: Sorry I still don't think it's necessary. Notice `Receive` is a blocking call (it's not a channel). We would need to have a goroutine to call it and pass the message to another channel and do something like

```
for {
  select {
  case msg := <-ch:
    // handle msg
  case <-ctx.Done():
   // stop
  }
}
```

But if you look at the code of cloudevents sdk, it's already what it does. That's for pubsub cancelling.

For processor(s) cancelling, I have commented on the processor interface. A processor shouldn't do any explicit timeout if it can fully rely on its successor(s) for timeout. Unless a processor itself does some long running processing. We should assume processor(s) can handle timeout correctly. It very much simplifies processor code because most of them won't need explicit timeout handling.

pkg/broker/handler/processors/fanout/processor.go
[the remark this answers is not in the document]
yolocs: Done

pkg/broker/handler/pool/options.go
liu-cong: The `panic` worries me more but I guess it 's low risk in practice. 

/lgtm

PR #768, opened 2020-04-03
Small improvement on broker memory config for easier testing.

liu-cong:
Fixes #

## Proposed Changes
- Replace `NewTargetsFromBytes` with `NewTargets` that takes in a proto struct. It seems we will never need NewTargetsFromBytes for the in memory config.
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/broker/config/memory/memory.go
+++ b/pkg/broker/config/memory/memory.go
@@ -17,7 +17,6 @@ limitations under the License.
 package memory
 
 import (
-	"fmt"
 	"sync"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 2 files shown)


pkg/broker/config/memory/memory.go
```diff
@@ -81,22 +80,14 @@ var _ config.Targets = (*memoryTargets)(nil)
...
-	t := &memoryTargets{
-		mux: sync.Mutex{},
-	}
-	t.Store(&config.TargetsConfig{Brokers: make(map[string]*config.Broker)})
-	return t
+	return NewTargets(&config.TargetsConfig{Brokers: make(map[string]*config.Broker)})
 }
 
-// NewTargetsFromBytes creates a mutable Targets in memory.
-func NewTargetsFromBytes(b []byte) (config.Targets, error) {
```
yolocs, line 92: This could be useful when reading bytes from configmap to form a memory config. I don't see it's used by the controller though. I'm fine replacing (vs. adding a new func) if it's not used. @grantr

yolocs:
@Harwayne 
 
> Out of curiosity, what are these used for?

We decided to use configmap to pass control plane data for data plane use. This lib helps manipulating this data.

pkg/broker/config/memory/memory.go
grantr: The broker should be reading the configuration from the configmap on start. It doesn't do that yet but it will need to soon. Otherwise the configmap will be inaccurate while the initial resync happens.

grantr: In other words, the controller will need some way to load an in-memory targets config from a configmap (not a volume).

yolocs: I guess option 1: keep `FromBytes` and the controller can pass the bytes from configmap; or option 2: replace this `FromBytes` with `FromConfigMap` which accepts a configmap object.

grantr: I don't like option 2. It requires this code to be aware that the config comes from a configmap and what filename the data is in. That seems like the wrong layering.

I'm fine with option 1 or this PR. I think option 1 is slightly easier since it removes any proto marshaling from the controller code and tests.

grantr:
/unhold

I'm fine with this. We can bring back NewTargetsFromBytes if needed.

PR #774, opened 2020-04-06
Route events based on path

### LINKED ISSUE #636: Implement Broker ingress basic functionality

liu-cong:
Fixes #636

## Proposed Changes

- Route events to corresponding decouple topic based on path `/<ns>/<broker>`
- The decouple topic is configured by the controller and mounted as a volume in the ingress pod
-

**Release Note**

```release-note

```

**Docs**

cc @Harwayne @yolocs @ericlem @grantr @grac3gao


pkg/broker/ingress/handler.go
```diff
@@ -18,97 +18,131 @@ package ingress
...
 }
 
 // handler receives events and persists them to storage (pubsub).
+// TODO(liu-cong) add metrics
+// TODO(liu-cong) add tracing
+// TODO(liu-cong) support event TTL
 type handler struct {
-	// inbound is the cloudevents client to receive events.
-	inbound cev2.Client
+	// httpReceiver starts an HTTP server to receive events.
```
Harwayne, line 54:
```suggestion
	// httpReceiver is an HTTP server to receive events.
```

pkg/broker/ingress/handler.go
```diff
@@ -18,97 +18,131 @@ package ingress
...
+	h.logger.Debug("Serving http", zap.Any("request", request))
+	if request.Method != nethttp.MethodPost {
+		response.WriteHeader(nethttp.StatusMethodNotAllowed)
+		return
+	}
 
-	if res := h.decouple.Send(ctx, event); !cev2.IsACK(res) {
-		logging.FromContext(ctx).Error("Error publishing to PubSub", zap.String("event", event.String()), zap.Error(res))
-		return res
+	// Path should be in the form of "/<ns>/<broker>"
```
Harwayne, line 103:
```suggestion
	// Path should be in the form of "/<ns>/<broker>".
```

I also recommend including the Broker's UID if you can. It just makes it less likely someone sends to the wrong Broker by mistake.

pkg/broker/ingress/handler.go
```diff
@@ -18,97 +18,131 @@ package ingress
...
+		response.WriteHeader(statusCode)
+		return
+	}
 
-// newDefaultHTTPClient provides good defaults for an HTTP client with tracing.
-func newDefaultHTTPClient() (cev2.Client, error) {
-	p, err := cev2.NewHTTP()
-	if err != nil {
-		return nil, err
+	ctx, cancel := context.WithTimeout(context.Background(), decoupleSinkTimeout)
```
Harwayne, line 118:
```suggestion
	ctx, cancel := context.WithTimeout(request.Context(), decoupleSinkTimeout)
```

pkg/broker/ingress/handler_test.go
```diff
@@ -1,72 +1,143 @@
...
 }
 
-// TestHandlerDefaultHTTPInbound uses default HTTP inbound client and a channel based mock decouple
-// sink.
-func TestHandlerDefaultHTTPInbound(t *testing.T) {
-	decouple, dc := test.NewMockSenderClient(t, 1)
-	_, cleanup := createAndStartIngress(t, nil, decouple)
-	defer cleanup()
+func TestHandler(t *testing.T) {
+	tests := []testCase{
```
Harwayne, line 50: I recommend inling the `testCase` struct here, rather than defining it outside the function (unless you intend to reuse it somewhere else):

```suggestion
	tests := []struct{
		name    string
		options []HandlerOption
		// If port is different than the default(8080), a matching WithPort option should be provided.
		port int
		// in happy case, path should match the /<ns>/<broker> in the brokerConfig.
		path  string
		event *cloudevents.Event
		// If method is empty, POST will be used as default.
		method string
		// body and header can be specified if the client is making raw HTTP request instead of via cloudevents.
		body     map[string]string
		header   nethttp.Header
		wantCode int
	}{
```

pkg/broker/ingress/handler_test.go
```diff
@@ -1,72 +1,143 @@
...
+		{
+			name:     "request is not an event",
+			path:     "/ns1/broker1",
+			port:     defaultPort,
+			wantCode: nethttp.StatusBadRequest,
+			header:   nethttp.Header{},
+		},
+		{
+			name:     "wrong path - broker doesn't exist in config",
+			path:     "/ns-1/broker2",
```
Harwayne, line 90: Have distinct tests for incorrect namespace and incorrect broker, so that we know both cases are caught. Rather than a single test that has both incorrect at the same time.

pkg/broker/ingress/handler_test.go
```diff
@@ -1,72 +1,143 @@
...
+			event:    createTestEvent("test-event"),
+			port:     defaultPort,
+			wantCode: nethttp.StatusInternalServerError,
```
Harwayne, line 93: Shouldn't this 404?

pkg/broker/ingress/handler_test.go
```diff
@@ -1,72 +1,143 @@
...
-		WithInboundClient(inbound),
-		WithDecoupleSink(decouple))
-	if err != nil {
-		t.Fatalf("Failed to create ingress handler: %+v", err)
+	h, err2 := NewHandler(ctx, opts...)
+	if err2 != nil {
+		t.Fatalf("Failed to create ingress handler: %+v", err2)
 	}
 	go h.Start(ctx)
 	cleanup = func() {
```
Harwayne, line 143: Should this cancel `ctx` so that the handler is stopped?

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -0,0 +1,104 @@
...
+func NewMultiTopicDecoupleSink(ctx context.Context, options ...MultiTopicDecoupleSinkOption) (*multiTopicDecoupleSink, error) {
+	sink := &multiTopicDecoupleSink{
+		logger: logging.FromContext(ctx),
+	}
+
+	for _, opt := range options {
+		opt(sink)
+	}
+
+	// Aplly defaults
```
Harwayne, line 29:
```suggestion
	// Apply defaults.
```

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -0,0 +1,104 @@
...
+	if sink.client == nil {
+		client, err := newDefaultPubSubClient(ctx)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create pubsub client: %w", err)
```
Harwayne, line 33: Just for my own knowledge: as I haven't used `errors.Wrap` yet, when do we generally wrap errors vs using `Errorf`?

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -0,0 +1,104 @@
...
+		// There is an propagation delay between the controller reconciles the broker config and
+		// the config being pushed to the configmap volume in the ingress pod.
+		// TODO(liu-cong) Should we wait (with a timeout) for the configmap volume update instead of
+		// returning error immediately?
+		msg := fmt.Sprintf("config is not found for %v/%v", ns, broker)
+		m.logger.Warn(msg)
+		return "", errors.New(msg)
+	}
+	if brokerConfig.DecoupleQueue == nil || brokerConfig.DecoupleQueue.Topic == "" {
+		msg := fmt.Sprintf("DecouplQueue or topic missing for broker, this should NOT happen. BrokerConfig: %+v", brokerConfig)
```
Harwayne, line 83:
```suggestion
		msg := fmt.Sprintf("DecoupleQueue or topic missing for broker, this should NOT happen. BrokerConfig: %+v", brokerConfig)
```

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -0,0 +1,211 @@
...
+	topic       string
+	clientErrFn func(client *fakePubsubClient)
+	wantErr     bool
+}
+
+func TestMultiTopicDecoupleSink(t *testing.T) {
+	tests := []struct {
+		name         string
+		brokerConfig *config.TargetsConfig
+		cases        []brokerTestCase
```
Harwayne, line 32: Similar to above, I would probably define the struct inline, unless you plan to use it outside this function.

pkg/broker/ingress/handler.go
liu-cong: @grantr Any concern from the controller perspective?

pkg/broker/ingress/handler_test.go
liu-cong: Yes the `createRequest` helper function needs it.

pkg/broker/ingress/multi_topic_decouple_sink.go
liu-cong: I found `errors.Wrap` to be very convenient as it adds context information by recording a stack trace. 

> // Wrap returns an error annotating err with a stack trace
// at the point Wrap is called, and the supplied message.
// If err is nil, Wrap returns nil.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
liu-cong: No I don't need it outside. However it seems that when I define `cases` I have to repeat the struct definition again like so:
```
	tests := []struct {
		name         string
		brokerConfig *config.TargetsConfig
		cases        []struct{
			ns          string
			broker      string
			topic       string
			clientErrFn func(client *fakePubsubClient)
			wantErr     bool
		}
	}{
		{
			name: "happy path single broker",
			brokerConfig: &config.TargetsConfig{
				Brokers: map[string]*config.Broker{
					"test_ns_1/test_broker_1": {DecoupleQueue: &config.Queue{Topic: "test_topic_1"},},
				},
			},
			cases: []struct{
				ns          string
				broker      string
				topic       string
				clientErrFn func(client *fakePubsubClient)
				wantErr     bool
			}{
				{
					ns:     "test_ns_1",
					broker: "test_broker_1",
					topic:  "test_topic_1",
				},
			},
		},
}
```

Let me know if there is a more concise way to do it. I will be very happy to change to that  :)

pkg/broker/ingress/handler_test.go
liu-cong: This one is tricky. The request can fail due to bad request, or due to the configmap volume propagation delay (which should be 500s). 

I am open to suggestion here.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
Harwayne: Good point, that is quite ugly. How about defining the struct in the function?

```golang
func TestMultiTopicDecoupleSink(t *testing.T) {
	type brokerTestCase struct{
		ns          string
		broker      string
		topic       string
		clientErrFn func(client *fakePubsubClient)
		wantErr     bool
	}
	tests := []struct {
		name         string
		brokerConfig *config.TargetsConfig
		cases        []brokerTestCase
	}{
		{
			name: "happy path single broker",
			brokerConfig: &config.TargetsConfig{
				Brokers: map[string]*config.Broker{
					"test_ns_1/test_broker_1": {DecoupleQueue: &config.Queue{Topic: "test_topic_1"},},
				},
			},
			cases: []brokerTestCase{
				{
					ns:     "test_ns_1",
					broker: "test_broker_1",
					topic:  "test_topic_1",
				},
			},
		},
	}
...
}
```

pkg/broker/ingress/handler_test.go
Harwayne: My personal preference is 404, as that is the correct steady state error in my opinion. @grantr thoughts?

pkg/broker/ingress/multi_topic_decouple_sink_test.go
liu-cong: oh yes I can certainly do this

pkg/broker/ingress/handler.go
grantr: I have concerns from the UX perspective. If we use a UID in the path then:
- The path to produce to the Broker is no longer predictable
- There is no way to configure the producer with a broker URI in advance of the Broker being created
- If the Broker is deleted and recreated, the producer must update its broker URI

None of these are issues for producers that have a watch on the Broker and can resolve its address, but watching the Broker requires permissions, it requires a control loop, it requires a lot of lifecycle code just to get that URI. I'd like to keep open the option to start a random pod and send to a known URI.

If the concern about sending to the wrong broker by mistake is about the "delete and recreate the Broker" case, then I'd argue that it makes sense for the new Broker to have the same URI because the producer's directive is still "send an event to the Broker with this name".

if the concern is about sending to the wrong broker due to a misconfiguration, then I don't think including the UID helps, because the user specifies the Broker by name anyway in the reference (for CRD sources).

So my inclination is to not use the URI in the path.

pkg/broker/ingress/handler_test.go
grantr: I agree it should be 404. Technically 404 may be temporary so this seems like a correct response. Same with the above broker-not-exist test.

pkg/broker/ingress/multi_topic_decouple_sink.go
ericlem: I haven't used wrapping much yet either.  You probaby want to avoid using Wrap() if you're at a level where your code completely understands the error and is capable of given a full description of the error.   This is particularly true if you're in a path where upper layers may want to be able to do differing things depending on the type of the error.  In particular, if you thing the error your function is returning may be something that users will want to detect and deal with(but one of the errors you are wrapping is quite generic), you may want to use your own error type and not wrap.  With unwrapping you can use Is() on the error to match an error chain where any element on the chain unwraps to that error type.  If you have an error path that can return multiple "interesting" error types (sql not found errors for multiple databases, etc), or something generic like EOF (but where the EOF isn't really logically your error) you might just want to use your own type and not wrap the sql error to avoid exposing your internal implementation.

cmd/broker/ingress/main.go
```diff
@@ -17,12 +17,17 @@ limitations under the License.
...
+
+	"github.com/google/knative-gcp/pkg/broker/ingress"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/signals"
 )
 
+// main creates and starts an ingress handler using default options.
+// 1. it listens on port 8080
+// 2. it reads "GOOGLE_CLOUD_PROJECT" env var for pubsub project.
+// 3. it expects broker configmap mounted at "/var/run/cloud-run-events/broker/targets"
```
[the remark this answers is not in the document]
liu-cong: I think the config lib should be able to handle this.

We can make it configurable, for now I am just using the default path that the config lib expects.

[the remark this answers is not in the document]
liu-cong: @yolocs

cmd/broker/ingress/main.go
```diff
@@ -17,12 +17,17 @@ limitations under the License.
...
 	"go.uber.org/zap"
+
+	"github.com/google/knative-gcp/pkg/broker/ingress"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/signals"
 )
 
+// main creates and starts an ingress handler using default options.
+// 1. it listens on port 8080
+// 2. it reads "GOOGLE_CLOUD_PROJECT" env var for pubsub project.
```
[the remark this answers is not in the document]
liu-cong: I am expecting the controller to set this, since the controller needs this to talk to pubsub anyway. Let me know if this works for you.

pkg/broker/ingress/handler.go
Harwayne: I retract my suggestion. The predictability of the Broker's URI was a conscious design choice of the Broker. I did not think through my suggestion, which would have broken that design.

cmd/broker/ingress/main.go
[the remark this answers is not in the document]
liu-cong: I was assuming that the controller creates the ingress but that may not be true. How is the controller going to get the project?

pkg/broker/ingress/multi_topic_decouple_sink.go
liu-cong: Sorry I didn't read the line carefully. You are right I don't need to wrap the pubsusb error here.

cmd/broker/ingress/main.go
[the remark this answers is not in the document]
yolocs: For the pubsub channel API we have today, you can provide a project ID. If missing, I believe somewhere (either in our code or in the pubsub sdk) it will use the project ID in the cluster metadata (which is the project hosting the gke cluster).

cmd/broker/ingress/main.go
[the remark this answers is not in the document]
yolocs: The default path (`/var/run/cloud-run-events/broker/targets`) assumes the filenanme is `targets`. We can change it to `targets.pb` if you prefer.

The option to set path is at: https://github.com/google/knative-gcp/blob/master/pkg/broker/config/volume/options.go#L23

cmd/broker/ingress/main.go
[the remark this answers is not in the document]
liu-cong: There is a `utils` package which reads project id from metadata. Going to use that.

cmd/broker/ingress/main.go
```diff
@@ -17,12 +17,18 @@ limitations under the License.
...
 	"go.uber.org/zap"
+
+	"github.com/google/knative-gcp/pkg/broker/ingress"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/signals"
 )
 
+// main creates and starts an ingress handler using default options.
+// 1. It listens on port 8080
+// 2. It reads "GOOGLE_CLOUD_PROJECT" env var for pubsub project. If the env var is empty, it retrieves project ID from
```
grantr, line 29: We've previously used `PROJECT_ID` for this env var in other receive adapters. We should be consistent here.

cmd/broker/ingress/main.go
```diff
@@ -17,12 +17,18 @@ limitations under the License.
...
-	"github.com/google/knative-gcp/pkg/broker/ingress"
 	"go.uber.org/zap"
+
+	"github.com/google/knative-gcp/pkg/broker/ingress"
 	"knative.dev/pkg/logging"
 	"knative.dev/pkg/signals"
 )
 
+// main creates and starts an ingress handler using default options.
+// 1. It listens on port 8080
```
grantr, line 28: It should listen on `PORT` if that env var is present, since that's the env var serving uses.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+func newFakePubsubClient(t *testing.T) *fakePubsubClient {
+	return &fakePubsubClient{
+		t:             t,
+		topics:        map[string]<-chan cloudevents.Event{},
+		topicToClient: map[string]cloudevents.Client{},
+		topicToErr:    map[string]error{},
+	}
+}
```
yolocs, line 182: Would it be possible to use pubsub test client `pstest` package. I think it would be more close to real case.

pkg/broker/ingress/errors.go
```diff
@@ -0,0 +1,11 @@
+package ingress
+
+import "errors"
+
+// ErrNotFound is the error when a broker doesn't exist in the configmap.
+// This can happen if the clients specifies invalid broker in the path, or the configmap volume hasn't been updated.
+var ErrNotFound = errors.New("not found")
+
+// ErrIncomplete is the error when a broker entry exists in the configmap but its decouple queue is nil or empty.
+// This should never happen unless there is a bug in the controller.
```
grantr, line 10: This is just a comment, not a suggestion. Looking ahead to future broker and trigger lifecycle, it's possible that a broker entry could have a nil decouple queue if it's been deleted but its triggers still exist and have retry queues. In this case we should probably have a DELETED status for the broker in the targets config.

@yolocs for thoughts. I plan to investigate how PubSub handles deletion of topics and pull subscriptions and write that down (though if someone else wants to do it that's fine too).

liu-cong: In that case the code here will have to check the broker status first. For now let's assume the decouple queue is never nil and see what you finds out how pubsub handles deletion. We can change the ingress logic accordingly. SG?

pkg/broker/ingress/handler.go
```diff
@@ -18,97 +18,140 @@ package ingress
...
+	// If encoding is unknown, the message is not an event.
+	if message.ReadEncoding() == binding.EncodingUnknown {
+		h.logger.Debug("Encoding is unknown. Not a cloud event?", zap.Any("request", request))
+		return nil, nethttp.StatusBadRequest
+	}
+	event, err := binding.ToEvent(request.Context(), message, transformer.AddTimeNow)
 	if err != nil {
-		return nil, err
+		h.logger.Error("Failed to convert request to event", zap.Error(err))
+		return nil, nethttp.StatusBadRequest
```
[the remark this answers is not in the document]
liu-cong: You mean the log msg here or the response code?

For the log msg, I don't have a reason to hide this.  Is there anything you are concerned in particular?

pkg/broker/ingress/handler_test.go
liu-cong: For the empty queue case, I am using 500s for now, since we assume the controller always fill the queue for a broker.

As Grant mentioned, we may mark the broker as `DELETED` with empty queue. If that happens, I will refactor the code to check broker status and return proper status code.

pkg/broker/ingress/handler.go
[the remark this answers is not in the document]
yolocs: I meant reply the error message back to the caller.

pkg/broker/ingress/errors.go
yolocs: @grantr sgtm. It's easy to add additional states.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
liu-cong: I gave it a try and it turned out there is a lot of code required to set up the client, topics, subscriptions, and retrieve events from topics. Essentially we have to do many jobs that the controller does, plus plumbing the test client to cloudevents client.

I agree it's more "real" to use that but meanwhile the added complexity makes the test harder to read. Moreover, the decouple sink relies on a cloudevent client abstraction, not necessarily pubsub. So I prefer to keep the test as is.

yolocs: It's really not that much code. I have a one written in my fanout service PR. It's like 15 lines more? Due to the fact we lack test coverage in the cloudevents sdk for pubsub, I still highly recommend using pstest :)

pkg/broker/ingress/handler.go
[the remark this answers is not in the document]
liu-cong: Sorry I misunderstood. Good idea, I added error msg to the response.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
liu-cong: There is also an issue with cloudevents v2 client. If you cancel the context, it panics. There is a TODO to handle error properly in ce v2. Due to that, it gets ugly to receive events back from pubsub. I guess there are workarounds to use the binding instead of the client, but then there is more logic not related to the test itself.

That said, I agree we need some more real testing. In fact, it's on my list to write some integration test using real pubsub (or pstest, which can be swapped easily). The test will be more like a blackbox test for the ingress as a whole. Does that sound good to you?

pkg/broker/ingress/handler.go
```diff
@@ -18,97 +18,152 @@ package ingress
...
+	if request.Method != nethttp.MethodPost {
+		response.WriteHeader(nethttp.StatusMethodNotAllowed)
+		return
+	}
 
-	if res := h.decouple.Send(ctx, event); !cev2.IsACK(res) {
-		logging.FromContext(ctx).Error("Error publishing to PubSub", zap.String("event", event.String()), zap.Error(res))
-		return res
+	// Path should be in the form of "/<ns>/<broker>".
+	pieces := strings.Split(request.URL.Path, "/")
```
ian-mi, line 114: I think it would be better to use something like gorilla/mux for routing, especially since we might also want to serve zpages on the same port.

liu-cong: I'd like to try that. Is it OK address this as a follow up https://github.com/google/knative-gcp/issues/785?

pkg/broker/ingress/multi_topic_decouple_sink_test.go
liu-cong: Created an issue to follow up on this: https://github.com/google/knative-gcp/issues/786

yolocs: Yes. Just a note. If we're going to use pstest, it's better to replace this one here.

pkg/broker/ingress/handler.go
ian-mi: Sounds good

Issue #614, opened 2020-03-06
Create v1beta1 types


nachocano:
**Problem**
We need to start promoting stable APIs to v1beta1. This issue only involves adding the new types.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
Stable APIs have v1beta1 types

**Additional context (optional)**
https://github.com/knative/eventing/pull/2414
https://github.com/knative/eventing/pull/2407
https://github.com/knative/eventing/pull/2466

nachocano:
For workload identity we added a `spec.serviceAccount` field to all constructs. That refers to a google service account. @grac3gao proposed to use `spec.googleServiceAccount`, which is more accurate. I'm wondering if she was right. `serviceAccount` might be confusing with a typical k8s service account.

Might be worth thinking of what we should use `spec.serviceAccount`, `spec.gsa` or `spec.googleServiceAccount`. I'm currently more inclined for the second one `spec.gsa`

nachocano:
my apologies, I was too optimistic, won't be able to make this in the three days I have remaining.
/unassign

grac3gao:
> For workload identity we added a `spec.serviceAccount` field to all constructs. That refers to a google service account. @grac3gao proposed to use `spec.googleServiceAccount`, which is more accurate. I'm wondering if she was right. `serviceAccount` might be confusing with a typical k8s service account.
> 
> Might be worth thinking of what we should use `spec.serviceAccount`, `spec.gsa` or `spec.googleServiceAccount`. I'm currently more inclined for the second one `spec.gsa`

Have talked with Nacho under other thread. https://github.com/google/knative-gcp/pull/692#issuecomment-603946288. 

[Grace]
I am fine with `gsa`, but I am worried about that it might be too brief for user to understand? Perhaps `gServiceAccount` would be a choice?

[Nacho]
Yeah gsa might not be correct as well, and gServiceAccount sounds better. I
think we need a third opinion, maybe Adam?

@Harwayne would like to hear your opinions :)

Harwayne:
> > For workload identity we added a `spec.serviceAccount` field to all constructs. That refers to a google service account. @grac3gao proposed to use `spec.googleServiceAccount`, which is more accurate. I'm wondering if she was right. `serviceAccount` might be confusing with a typical k8s service account.
[...]

PR #779, opened 2020-04-07
v1beta1 - Duck and Pubsub

Harwayne:
Helps with #614
## Proposed Changes

- Create v1beta1 go types for:
    - Duck types
    - Pubsub types
- All code is copied from the v1alpha1 types. There is only one change, in `topic_lifecycle.go` because the v1beta1 duck version of `Addressable` no longer has `hostname`, so its support was dropped.
    - This also caused a changed status condition when there is no addressable URL. It changed from `emptyHostname` to `emptyUrl`.

---
This was generated using:

```shell
foo=pubsub
```

**Release Note**

```release-note
NONE
```


Harwayne:
> /hold
> In case you need to make change. This is just copying from v1alpha1 API right?

There was a single change, now mentioned in the PR description. Please take a look to make sure you are OK with it.

pkg/apis/pubsub/v1beta1/topic_lifecycle.go
```diff
@@ -0,0 +1,109 @@
...
+func (ts *TopicStatus) SetAddress(url *apis.URL) {
+	if ts.Address == nil {
+		ts.Address = &v1beta1.Addressable{}
+	}
+	if url != nil {
+		ts.Address.URL = url
+		topicCondSet.Manage(ts).MarkTrue(TopicConditionAddressable)
+	} else {
+		ts.Address.URL = nil
+		topicCondSet.Manage(ts).MarkFalse(TopicConditionAddressable, "emptyHostname", "hostname is the empty string")
```
yolocs, line 56: nit: since hostname is gone, should reword the reason?

Harwayne: Good idea. Done. I changed both the v1alpha1 and v1beta1 branches, so that we don't suddenly change error messages when our reconcilers start using v1beta1.

PR #780, opened 2020-04-07
Fix flaky volume config test with atomic file write

yolocs:
As reported by Cong: https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/774/pull-google-knative-gcp-unit-tests/1247559720701530115

**Release Note**

```release-note

```


pkg/broker/config/volume/volume_test.go
```diff
@@ -154,12 +158,31 @@ func TestSyncConfigFromFile(t *testing.T) {
...
+		t.Errorf("updated targets got=%+v, want=%+v", gotTargets, data)
+	}
+}
+
+func atomicWriteFile(t *testing.T, tempDir, file string, bytes []byte) {
+	t.Helper()
+	// In order to more closely replicate how K8s writes ConfigMaps to the file system, we will
+	// atomically swap out the file by writing it to a temp directory, then renaming it into the
+	// directory we are watching.
+	tmpFile, err := ioutil.TempFile(tempDir, "temp-*")
```
Harwayne, line 176: My preference would be to write a different directory, as that will more closely match what k8s does. In particular, depending on how the file system watcher is setup, this approach will cause multiple change notifications (ie temp file created, temp file written, overwrite config file).

Harwayne:
/hold
Holding for comments.

pkg/broker/config/volume/volume_test.go
```diff
@@ -168,12 +168,16 @@ func TestSyncConfigFromFile(t *testing.T) {
...
+func atomicWriteFile(t *testing.T, file string, bytes []byte) {
 	t.Helper()
 	// In order to more closely replicate how K8s writes ConfigMaps to the file system, we will
 	// atomically swap out the file by writing it to a temp directory, then renaming it into the
 	// directory we are watching.
-	tmpFile, err := ioutil.TempFile(tempDir, "temp-*")
+	dir, err := ioutil.TempDir("", "temp-*")
+	if err != nil {
+		t.Fatalf("unexpected error from creating temp dir: %v", err)
+	}
```
Harwayne, line 179:
```suggestion
	}
	defer os.RemoveAll(tempDir)
```

Then, if you want, you can remove the defer to delete `tmpFile`.

Harwayne:
/hold cancel

I think my comments were the last ones to be addressed.

Issue #615, opened 2020-03-06
Conversion, store and serving of v1beta1 APIs


nachocano:
**Problem**
Once we have the v1beta1 APIs https://github.com/google/knative-gcp/issues/614, we need to add the proper conversion mechanism, and start serving them.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
Serving v1beta1 APIs and provide conversions if needed

**Additional context (optional)**
https://github.com/knative/eventing/pull/2508

nachocano:
When we start serving v1beta1, I think we will break gcloud integration. 
We should make sure we are not. We should have the gcloud E2E tests in place before doing so.
https://github.com/google/knative-gcp/issues/556

nachocano:
my apologies, I was too optimistic, won't be able to make this in the three days I have remaining.
/unassign

PR #784, opened 2020-04-07
Pubsub v1beta1 Conversions

Harwayne:
Helps with #615.

## Proposed Changes

- Adds the v1beta1 types PullSubscription and Topic to the API server.
- Adds a conversion webhook that converts between v1alpha1 and v1beta1.

**Release Note**

```release-note
PullSubscriptions and Topics now exist in both v1alpha1 and v1beta1 versions. The versions are almost identical. All existing objects will continue to work. All valid configurations for the objects will continue to be valid. 

If you wish, you may start using the v1beta1 API surface. If you choose not to, then everything will continue to work as-is.
```

**Docs**


cmd/webhook/main.go
```diff
@@ -123,6 +126,41 @@ func NewConfigValidationController(ctx context.Context, cmw configmap.Watcher) *
...
+			pubsubv1beta1.Kind("Topic"): {
+				DefinitionName: pubsub.TopicsResource.String(),
+				HubVersion:     pubsubv1alpha1_,
+				Zygotes: map[string]conversion.ConvertibleObject{
+					pubsubv1alpha1_: &pubsubv1alpha1.Topic{},
+					pubsubv1beta1_:  &pubsubv1beta1.Topic{},
+				},
+			},
+		},
+		func(ctx context.Context) context.Context {
```
grantr, line 159: Nit: add a comment about the purpose of this argument, as in https://github.com/knative/eventing/blob/1e68f1cfe5bb4f8c67aca7b09b36b973cde278dc/cmd/webhook/main.go#L333

pkg/apis/pubsub/v1alpha1/pullsubscription_conversion_test.go
```diff
@@ -0,0 +1,222 @@
...
+				Type:   "type",
+				Source: "source",
+			},
+		},
+		ProjectID:      "projectID",
+		TopicID:        "topicID",
+		SubscriptionID: "subscriptionID",
+	}
+
+	// completePullSubscription is a PullSubscription with every filled in, except TypeMeta.
```
grantr, line 148: nit: `every field filled in`

Harwayne: https://github.com/google/knative-gcp/pull/813

cmd/webhook/main.go
Harwayne: https://github.com/google/knative-gcp/pull/813

PR #794, opened 2020-04-08
Structural schema for Topic and PullSubscription CRDs.

Harwayne:
Helps with #615.

## Proposed Changes

- Make Topic and PullSubscription CRD's schema structural.
  - This is required when serving multiple versions of a CRD.

I am making this a distinct PR from #784 because I made a mistake there and it took me a long time to understand the mistake was in the schema, not all the other changes. By moving this to a separate PR, it will make sure the schema changes are not overlooked amongst the many other changes in #784.

**Release Note**

```release-note
Topic and PullSubscription will aggressively remove unknown fields.
```


config/core/resources/pullsubscription.yaml
```diff
@@ -57,46 +64,57 @@ spec:
...
-                        name:
-                          type: string
-                          minLength: 1
+                    kind:
+                      type: string
+                      minLength: 1
+                    name:
+                      type: string
+                      minLength: 1
+              anyOf:
```
grantr, line 100: I don't think we need this anyOf. It's not specified inhttps://github.com/knative/eventing/blob/master/config/core/resources/trigger.yaml

config/core/resources/topic.yaml
```diff
@@ -46,10 +46,17 @@ spec:
     - name: Age
       type: date
       JSONPath: .metadata.creationTimestamp
+  versions:
+    - name: v1alpha1
+      served: true
+      storage: true
+  # All versions happen to have the same schema today. They will likely diverge in the future.
```
grantr, line 53: I'd probably remove this and the other similar one. Can't guarantee it will be removed when it goes out of date.

config/core/resources/pullsubscription.yaml
Harwayne: This `anyOf` is for the `spec.sink` field, which is present on PullSubscriptions, but not Topics.

config/core/resources/topic.yaml
Harwayne: If they ever diverge, we will actually end up removing the `validation` field one line down and adding the individual schemas in the `versions` field. I tried to do that initially, but the CRD was rejected if the schema of all versions was identical and specified in the `versions` field.

config/core/resources/pullsubscription.yaml
grantr: Not sure why it matters that it's only on PullSubscription. `sink` is a Destination, similar to Trigger's `subscriber`. In particular, because sink has two properties and neither are required, I believe the semantics are already the same as what this anyOf describes.

PR #798, opened 2020-04-09
Dep update eventing to fix integ tests

yolocs:
Integ test is broker of a missing file introduced in eventing yesterday. Updating eventing to see if it fixes the problem.

## Proposed Changes

`dep ensure -update knative.dev/eventing && ./hack/update-deps.sh`

**Release Note**

```release-note

```


yolocs:
I think this is actually caused by the eventing nightly release yaml hasn't yet had `config-br-defaults` configmap. So close this one.

Issue #804, opened 2020-04-09
The broker config bytes output might not always work with configmap


yolocs:
**Describe the bug**
The broker config currently directly outputs raw proto marshaled bytes via `Bytes()`. Converting that to a string and put in the configmap might not work because of encoding problem.

**Solution**
Use base64 encoding to output encoded config.

grantr:
We can also use the `BinaryData` field in the configmap. This seems to work in #739.

PR #805, opened 2020-04-09
Use base64 encoding to encode broker config bytes

yolocs:
Fixes https://github.com/google/knative-gcp/issues/804

## Proposed Changes

- drop `Bytes()` since it probably won't be used (when putting into configmap, we use string)
- use base64 encoding to encode raw proto bytes to avoid encoding problem in configmap

**Release Note**

```release-note
Use base64 encoding to encode broker config bytes
```

### DIFF
--- a/pkg/broker/config/cache.go
+++ b/pkg/broker/config/cache.go
@@ -17,6 +17,7 @@ limitations under the License.
 package config
 
 import (
+	"encoding/base64"
 	"sync/atomic"
 
 	"github.com/golang/protobuf/proto"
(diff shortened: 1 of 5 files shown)


yolocs:
@grantr - this would affect the controller

grantr:
@yolocs why is this PR necessary? The controller is using `Bytes()` to put the raw proto bytes into the configmap's `BinaryData` field. Seems to work fine.

yolocs:
> @yolocs why is this PR necessary? The controller is using `Bytes()` to put the raw proto bytes into the configmap's `BinaryData` field. Seems to work fine.

Ah, I see. I didn't realize that. Closing.

PR #812, opened 2020-04-10
v1beta1 - Sources

Harwayne:
Helps with #614

## Proposed Changes

- Create v1beta1 go types for all the sources ready for promotion:
  - CloudAuditLogsSource
  - CloudPubSubSource
  - CloudSchedulerSource
  - CloudStorageSource
- All code is copied from the v1alpha1 types. 

---

This was generated using:
```shell
foo=events
```


Harwayne:
This is ready for review.

Issue #802, opened 2020-04-09
pool_test fails without Google default credentials


ian-mi:
**Describe the bug**
go test ./pkg/broker/handler/pool/fanout/ fails with errors such as:
```
{"level":"error","ts":1586458631.452828,"logger":"fallback","caller":"fanout/pool.go:105","msg":"failed to create pubsub protocol","broker":"ns-3/broker-2","error":"pubsub: google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.","stacktrace":"github.com/google/knative-gcp/pkg/broker/handler/pool/fanout.(*SyncPool).syncOnce.func2\n\t/home/ian/go/src/github.com/google/knative-gcp/pkg/broker/handler/pool/fanout/pool.go:105\ngithub.com/google/knative-gcp/pkg/broker/config.(*CachedTargets).RangeBrokers\n\t/home/ian/go/src/github.com/google/knative-gcp/pkg/broker/config/cache.go:84\ngithub.com/google/knative-gcp/pkg/broker/handler/pool/fanout.(*SyncPool).syncOnce\n\t/home/ian/go/src/github.com/google/knative-gcp/pkg/broker/handler/pool/fanout/pool.go:87\ngithub.com/google/knative-gcp/pkg/broker/handler/pool/fanout.(*SyncPool).watch\n\t/home/ian/go/src/github.com/google/knative-gcp/pkg/broker/handler/pool/fanout/pool.go:69"}
```

 **Expected behavior**
The test passes

Issue #638, opened 2020-03-12
#3 Implement Broker retry basic functionality


grantr:
_Can be combined with fanout implementation in #637_

* Receive events from all retry topics
* Publish to matching consumers (but don't requeue on failure)
* Unit tests

/area broker

grac3gao:
- [ ] basic implementation
- [ ] semi e2e with with some refactoring

grac3gao:
#883

PR #817, opened 2020-04-13
Basic Broker Retry Implementation

grac3gao:
related #638

Using reusable modules defined in #761 for sync pool and processor.
Reuse delivery processor in fanout.
UT test for pool still need modification as it might also cause issue like #802

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


pkg/broker/config/config.go
```diff
@@ -74,3 +85,13 @@ type Targets interface {
 func BrokerKey(namespace, name string) string {
 	return namespace + "/" + name
 }
+
+// TriggerKey returns the key of a trigger. Format is namespace/brokerName/targetName.
+func TriggerKey(namespace, broker, target string) string {
+	return fmt.Sprintf("%s/%s/%s", namespace, broker, target)
+}
+
+func SplitTriggerKey(key string) (string, string, string) {
```
yolocs, line 94: nit: comment
Also, could you check if `SplitBrokerKey` would be useful?

pkg/broker/handler/pool/retry/pool.go
```diff
@@ -0,0 +1,139 @@
...
+
+		projectIDOpt := pubsub.WithProjectIDFromDefaultEnv()
+		if p.options.ProjectID != "" {
+			projectIDOpt = pubsub.WithProjectID(p.options.ProjectID)
+		}
+		ps, err := pubsub.New(ctx,
+			projectIDOpt,
+			pubsub.WithTopicID(t.RetryQueue.Topic),
+			pubsub.WithSubscriptionID(t.RetryQueue.Subscription),
+		)
```
yolocs, line 104: Also honor pubsub client override: https://github.com/google/knative-gcp/blob/master/pkg/broker/handler/pool/fanout/pool.go#L107-L109

This also helps unit test.

pkg/broker/handler/pool/retry/pool.go
```diff
@@ -0,0 +1,139 @@
...
+		h := &handler.Handler{
+			Timeout:      p.options.TimeoutPerEvent,
+			PubsubEvents: ps,
+			Processor: processors.ChainProcessors(
+				// TODO filter processor may be added in the future, but need more discussion for that.
+				&deliver.Processor{Requester: p.options.EventRequester},
+			),
+		}
+		// Start the handler with target in context.
+		h.Start(handlerctx.WithTarget(ctx, t), func(err error) {
```
yolocs, line 120: The deliver processor also needs the broker for reply: https://github.com/google/knative-gcp/blob/master/pkg/broker/handler/processors/deliver/processor.go#L44

yolocs: This also reminds me that we would need a test to test reply with pool

cmd/broker/retry/kodata/LICENSE
```diff
@@ -0,0 +1,202 @@
+
```
yolocs, line 1: There should be more stuff in kodata? And they should be symlinks. See my PR https://github.com/google/knative-gcp/commit/acedadbe224d707f0a3b3fcd34d9d9643e734c94

pkg/broker/handler/pool/retry/pool.go
grac3gao: I'll have second PR for adding a semi-e2e for retry

pkg/broker/config/config.go
grac3gao: Checked, it seems there is no need to add `SplitBrokerKey` now. TargetsConfig uses brokeKey as the key to store broker, so brokerName itself is not that needed. Targets are stored with key of target name in the map under a broker, so `SplitTriggerKey` would help to get the targetName for that.

pkg/broker/handler/pool/retry/pool.go
```diff
@@ -0,0 +1,150 @@
...
+			Timeout:      p.options.TimeoutPerEvent,
+			PubsubEvents: ps,
+			Processor: processors.ChainProcessors(
+				// TODO filter processor may be added in the future, but need more discussion for that.
+				&deliver.Processor{Requester: p.options.EventRequester},
+			),
+		}
+
+		// Deliver processor needs the broker in the context for reply.
+		tctx := handlerctx.WithBrokerKey(ctx, bk)
```
yolocs, line 129: Currently the deliver processor requires a broker (not broker key) in the context. But maybe using keys consistently in the context makes more sense? Let's put the broker in the context for now. And let's discuss offline.

pkg/broker/handler/pool/retry/pool.go
```diff
@@ -0,0 +1,150 @@
...
+		if p.options.ProjectID != "" {
+			opts = append(opts, pubsub.WithProjectID(p.options.ProjectID))
+		} else {
+			opts = append(opts, pubsub.WithProjectIDFromDefaultEnv())
+		}
+
+		if p.options.PubsubClient != nil {
+			opts = append(opts, pubsub.WithClient(p.options.PubsubClient))
+		}
+		ps, err := pubsub.New(ctx, opts...)
```
yolocs, line 112: I forgot to mention. For retry, maybe you want to override the default pubsub `ReceiveSettings` - 1) you probably want to set `Synchronous` to `true` 2) you probably want to limit the `NumGoroutines`, `MaxOutstandingMessages` and `MaxOutstandingBytes`

pkg/broker/config/cache.go
```diff
@@ -56,6 +56,24 @@ func (ct *CachedTargets) RangeAllTargets(f func(*Target) bool) {
...
+	t, ok := Broker.Targets[targetName]
+	return t, ok
+}
+
+// GetTargetByKey returns a target by its trigger key. The format of trigger key is namespace/brokerName/targetName.
+// Do not modify the returned Target copy.
+func (ct *CachedTargets) GetTargetByKey(key string) (*Target, bool) {
+	namespace, brokerName, targetName := SplitTriggerKey(key)
+	return ct.GetTarget(namespace, brokerName, targetName)
+}
```
yolocs, line 75: Although trivial, better to add unit tests.

cmd/broker/retry/main.go
```diff
@@ -0,0 +1,192 @@
...
+	}
+
+	cfg, err := sharedmain.GetConfig(*masterURL, *kubeconfig)
+	if err != nil {
+		log.Fatal("Error building kubeconfig", err)
+	}
+
+	var env envConfig
+	if err := envconfig.Process("", &env); err != nil {
+		log.Fatal("Failed to process env var", zap.Error(err))
```
Harwayne, line 83:
```suggestion
		log.Fatalf("Failed to process env var: %v", err)
```

Or something similar as this isn't a zap logger.

cmd/broker/retry/main.go
```diff
@@ -0,0 +1,192 @@
...
+	if err != nil {
+		log.Fatal("Error loading/parsing logging configuration:", err)
+	}
+
+	sugaredLogger, atomicLevel := logging.NewLoggerFromConfig(loggingConfig, component)
+	defer flush(sugaredLogger)
+	ctx = logging.WithLogger(ctx, sugaredLogger)
+
+	configMapWatcher := configmap.NewInformedWatcher(kubeClient, system.Namespace())
+	// Watch the observability config map
```
Harwayne, line 110:
```suggestion
	// Watch the observability config map.
```

pkg/broker/config/config_test.go
```diff
@@ -0,0 +1,48 @@
...
+
+func TestSplitTriggerKey(t *testing.T) {
+	wantNS, wantBroker, wantTarget := "namespace", "broker", "target"
+	gotNS, gotBroker, gotTarge := SplitTriggerKey("namespace/broker/target")
+	if gotNS != wantNS || gotBroker != wantBroker || gotTarge != wantTarget {
+		t.Errorf("unexpected readiness: want %v, got %v",
+			fmt.Sprintf("%s/%s/%s", wantNS, wantBroker, wantTarget),
+			fmt.Sprintf("%s/%s/%s", gotNS, gotBroker, gotTarge))
+	}
+}
```
Harwayne, line 48: Add a test that round trips through both functions, something like:

```golang
x, y, z := SplitTriggerKey(TriggerKey(a, b, c))
if a != x || b != y || c != z {
  fail
}
```

pkg/broker/handler/pool/retry/pool.go
```diff
@@ -0,0 +1,150 @@
...
+package retry
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"go.uber.org/zap"
+
+	"knative.dev/pkg/logging"
```
Harwayne, line 26:
```suggestion
	"knative.dev/eventing/pkg/logging"
```

pkg/broker/handler/pool/retry/pool.go
grac3gao: When `Synchronous` is true, then no more than `MaxOutstandingMessages` will be in memory at one time, `MaxOutstandingBytes` still refers to the total bytes processed, rather than in memory and `NumGoroutines` is ignored. Do you think I need to set the Synchronous as ENV in the main to check whether it is set to true or false, or directly override it to true and set `MaxOutstandingMessages` and `MaxOutstandingBytes` when building `buildPoolOptions`?

grac3gao: Currently, I only set the `Synchronous` to `true`, but didn't change `MaxOutstandingMessages` and `MaxOutstandingBytes`

yolocs: Please add a TODO and we can revisit this topic.

cmd/broker/retry/main.go
```diff
@@ -0,0 +1,197 @@
...
+	}()
+	return ch
+}
+
+func buildPoolOptions(env envConfig, syncSignal chan struct{}) []pool.Option {
+	rs := pubsub.DefaultReceiveSettings
+	// If Synchronous is true, then no more than MaxOutstandingMessages will be in memory at one time.
+	// MaxOutstandingBytes still refers to the total bytes processed, rather than in memory.
+	// NumGoroutines is ignored.
+	// TODO For the case when synchronous is true, default value of MaxOutstandingMessages and MaxOutstandingBytes might need to override.
```
yolocs, line 175: We definitely need to revisit this. For now I think a TODO is ok.

pkg/broker/handler/pool/retry/pool.go
```diff
@@ -0,0 +1,156 @@
...
+
+	p.targets.RangeAllTargets(func(t *config.Target) bool {
+		tk := config.TriggerKey(t.Namespace, t.Broker, t.Name)
+		bk := config.BrokerKey(t.Namespace, t.Broker)
+		broker, ok := p.targets.GetBrokerByKey(bk)
+		if !ok {
+			// If corresponding broker doesn't exist, no need to process retry.
+			logging.FromContext(ctx).Error(fmt.Sprintf("event broker %q doesn't exist in config", bk), zap.String("trigger", t.Name))
+			return true
+		}
```
yolocs, line 96: Add a TODO here and we need to revisit this.

@grantr I remember we briefly discussed the desired behavior of retry if the broker gets deleted. If we keep retrying for a trigger whose broker was deleted, one problem is "reply" doesn't work anymore.

yolocs:
/lgtm
/hold
In case for other reviewers to take another round of look. Feel free to unhold.

Harwayne:
/hold cancel

I think I was the only other reviewer.

PR #824, opened 2020-04-14
Structural schemas for all the sources we intend to include in release/1

Harwayne:
Helps with #614.

## Proposed Changes

- Create structural schemas for the sources that we intend to include in release/1.
  - This is a prerequisite for serving multiple versions of the same CRD.

**Release Note**

```release-note
`CloudAuditLogsSource`, `CloudPubSubSource`, `CloudSchedulerSource`, 
and `CloudStorageSource` will aggressively remove unknown fields.
```


Harwayne:
This is ready for review.

PR #833, opened 2020-04-15
Add IAMPolicyManager to serialize IAM policy changes

### LINKED ISSUE #657: Have a mock client for iam policy binding unit test
**Describe the bug**
It is hard to write UT for iam policy binding because we are not mocking it. 

 **Expected behavior**
Having a mock client for iam policy binding to write UT

### LINKED ISSUE #821: Flaky test in Workload Identity with concurrency issue
**Describe the bug**
Using this issue to track.
E2E tests for Workload Identity are flaky and it will show errors as follows: adding iam policy binding failed with: failed to set iam policy: googleapi: Error 409: There were concurrent policy changes. Please retry the whole read-modify-write with exponential backoff.

Currently, an exponential backoff is added, retry 5 times, but sometime test will still be flaky. 
Reduce parallel running tests now to temporally avoid that flakiness. 

 **Expected behavior**
E2E tests won't fail for concurrency problem.

 **To Reproduce**

 **Additional context**

### LINKED ISSUE #857: Identity Reconciler Makes Conflicting Calls to SetIAMPolicy
**Describe the bug**
The Identity Reconciler makes concurrent conflicting calls to SetIAMPolicy for the same service account.

 **Expected behavior**
SetIAMPolicy calls to a single service account should be serialized and ideally batched.

 **To Reproduce**
Run multiple parallel workload identity tests.

0.14

ian-mi:
Fixes #821, fixes #857, and fixes #657

## Proposed Changes

- Adds an IAMPolicyManager process which serializes updates to google service account IAM policy updates. Updates are batched by the manager process to ensure that reconcilers are not starved by other updates. Changes are submitted to the policy manager and batched until the currently executing get or set IAM policy call for the given service account returns. Once a service account's policy is retrieved and available for modification, all batched changes are applied and a call to set IAM policy is initiated. If additional changes are batched during that call, those changes are directly applied against the returned policy without making a new call to GetIAMPolicy.
- Replaces use of the autogenerated google.golang.org/api IAM client with the higher level cloud.google.com/go client.
- Adds a fake IAM client for better testing.
- Removes IAM retry and conflict handling behavior from the identity reconciler.
- Reverts reduced concurrency on workload identity e2e test.


pkg/reconciler/identity/iam_policy.go
```diff
@@ -0,0 +1,238 @@
...
+	"cloud.google.com/go/iam/admin/apiv1"
+	gclient "github.com/google/knative-gcp/pkg/gclient/iam/admin"
+	iampb "google.golang.org/genproto/googleapis/iam/v1"
+)
+
+type action = int
+
+const (
+	Add = iota
+	Remove
```
Harwayne, line 33:
```suggestion
	ActionAdd action = iota
	ActionRemove
```

In my experience these are normally prefixed with their 'type', e.g. `ActionAdd` and `ActionRemove`.

pkg/reconciler/identity/iam_policy.go
```diff
@@ -0,0 +1,238 @@
...
+	serviceAccount GServiceAccount
+	role           iam.RoleName
+	member         string
+	action         action
+	respCh         chan error
+}
+
+type roleModification struct {
+	addMembers    map[string]bool
+	removeMembers map[string]bool
```
Harwayne, line 52: Assuming these are logical sets, I would suggest using K8s' [sets.String](https://godoc.org/k8s.io/apimachinery/pkg/util/sets#String).

pkg/reconciler/identity/reconciler.go
```diff
@@ -42,33 +41,38 @@ import (
...
-}
+// TODO remove global instance
+var (
+	policyManager       *IAMPolicyManager
+	createPolicyManager sync.Once
+)
 
 func NewIdentity(ctx context.Context) *Identity {
+	createPolicyManager.Do(func() {
+		c, err := admin.NewIamClient(context.Background())
```
Harwayne, line 57:
```suggestion
		c, err := admin.NewIamClient(ctx)
```

pkg/reconciler/identity/reconciler.go
```diff
@@ -42,33 +41,38 @@ import (
...
+		if err != nil {
+			panic(err)
+		}
+		m, err := NewIAMPolicyManager(context.Background(), c)
```
Harwayne, line 61:
```suggestion
		m, err := NewIAMPolicyManager(ctx, c)
```

pkg/reconciler/identity/reconciler.go
```diff
@@ -42,33 +41,38 @@ import (
...
-		KubeClient: kubeclient.Get(ctx),
+		KubeClient:    kubeclient.Get(ctx),
+		PolicyManager: policyManager,
 	}
 }
 
 type Identity struct {
-	KubeClient kubernetes.Interface
+	KubeClient    kubernetes.Interface
+	PolicyManager *IAMPolicyManager
```
Harwayne, line 75: Can this be private?

pkg/reconciler/identity/reconciler.go
```diff
@@ -155,87 +159,35 @@ func (i *Identity) createServiceAccount(ctx context.Context, namespace, gService
...
-}
-
-func setIamPolicy(ctx context.Context, action, projectID string, gServiceAccount string, kServiceAccount *corev1.ServiceAccount) error {
-	iamService, err := iam.NewService(ctx)
-	if err != nil {
-		return fmt.Errorf("failed to set google iam service: %w", err)
-	}
-
+func (i *Identity) addIamPolicyBinding(ctx context.Context, projectID, gServiceAccount string, kServiceAccount *corev1.ServiceAccount) error {
 	projectId, err := utils.ProjectID(projectID)
```
Harwayne, line 163: `projectId` is almost identical to `projectID` (an argument), change the name of one of them.

pkg/reconciler/identity/reconciler.go
```diff
@@ -155,87 +159,35 @@ func (i *Identity) createServiceAccount(ctx context.Context, namespace, gService
...
 	if err != nil {
 		return fmt.Errorf("failed to get project id: %w", err)
 	}
 
 	// Extract gServiceAccount's project name. The format of gServiceAccountName is service-account-name@project-id.iam.gserviceaccount.com.
 	gsaProject := strings.Split(strings.Split(gServiceAccount, "@")[1], ".")[0]
```
Harwayne, line 169: Can any malformed string get this far in? If so, I would prefer it returns errors on a bad string, rather than crash.

pkg/reconciler/identity/reconciler.go
```diff
@@ -155,87 +159,35 @@ func (i *Identity) createServiceAccount(ctx context.Context, namespace, gService
...
+		return fmt.Errorf("failed to get project id: %w", err)
 	}
-	return nil
-}
 
-// isConflict determines if the err is an error which indicates concurrency issue.
-func isConflict(err error) bool {
-	return strings.Contains(err.Error(), concurrencyError)
+	// Extract gServiceAccount's project name. The format of gServiceAccountName is service-account-name@project-id.iam.gserviceaccount.com.
+	gsaProject := strings.Split(strings.Split(gServiceAccount, "@")[1], ".")[0]
```
Harwayne, line 185: Same question as above. Can a malformed string get here?

pkg/reconciler/identity/reconciler_test.go
```diff
@@ -123,57 +102,76 @@ func TestCreates(t *testing.T) {
...
-		if (tc.expectedErr != "" && err == nil) ||
-			(tc.expectedErr == "" && err != nil) ||
-			(tc.expectedErr != "" && err != nil && tc.expectedErr != err.Error()) {
-			t.Errorf("Test case %q, Error mismatch, want: %q got: %q", tc.name, tc.expectedErr, err)
-		}
-		if diff := cmp.Diff(tc.expectedServiceAccount, kserviceAccount, ignoreLastTransitionTime); diff != "" {
-			t.Errorf("Test case %q, unexpected topic (-want, +got) = %v", tc.name, diff)
-		}
+			arl := pkgtesting.ActionRecorderList{cs}
+			kserviceAccount, err := identity.ReconcileWorkloadIdentity(context.Background(), projectID, identifiable)
```
Harwayne, line 130: How about using `ctx` instead? So that we are sure everything gets cancelled eventually.

pkg/reconciler/identity/iam_policy.go
```diff
@@ -0,0 +1,238 @@
...
+				for _, listener := range batched.listeners {
+					listener <- getPolicy.err
+				}
+				delete(m.pending, getPolicy.account)
+				break
+			}
+			m.pending[getPolicy.account] = &batchedModifications{
+				roleModifications: make(map[iam.RoleName]*roleModification),
+			}
+			go m.applyBatchedModifications(ctx, getPolicy.account, getPolicy.policy, batched)
```
Harwayne, line 154: `batched` is read in this new go routine. Do we need any concurrency control to makes sure `manage` and `makeModificationRequest` don't modify `batched` at the same time as this go routine is reading it?

pkg/reconciler/identity/iam_policy.go
```diff
@@ -0,0 +1,238 @@
...
+	case err := <-req.respCh:
+		return err
+	case <-ctx.Done():
+		return ctx.Err()
+	}
+}
+
+func (m *IAMPolicyManager) manage(ctx context.Context) {
+	for {
+		select {
```
Harwayne, line 133: Please add some documentation about how this is expected to work. E.g. `AddIAMPolicyBinding` is called, which puts something into the `requestCh`, which ...

pkg/reconciler/identity/iam_policy.go
Harwayne: Also, can they be private?

pkg/reconciler/identity/reconciler.go
ian-mi: The manager will be shared by other controller so I don't think it should the controller's ctx

pkg/reconciler/identity/iam_policy.go
ian-mi: Done, although it feels odd depending on apimachinery for a generic data structure.

pkg/reconciler/identity/iam_policy.go
ian-mi: No, batched is replaced with a new empty batch in m.pending immediately prior to this so ownership of batched is effectively handed off to the applyBatchedModifications goroutine.

pkg/reconciler/identity/iam_policy.go
ian-mi: Done.

pkg/reconciler/identity/reconciler.go
ian-mi: I'm not sure, I didn't modify this part of the code. IAM is able to infer project from the service account which seems to be what we're trying to do here, so I will try modifying the policy manager to use a project wildcard.

pkg/reconciler/identity/reconciler.go
ian-mi: I've left this public so that tests can construct instances using a fake client. What would probably be better would be to add it to the NewIdentity constructor and inject it somehow, but that requires figuring out how to do injection.

pkg/reconciler/identity/reconciler.go
ian-mi: Done.

pkg/reconciler/identity/reconciler.go
ian-mi: Same as above.

pkg/reconciler/identity/reconciler_test.go
ian-mi: Done, although I don't think this will call will leave anything running when it returns.

pkg/reconciler/identity/reconciler.go
ian-mi: I was able to refactor things to make this and KubeClient private

pkg/reconciler/identity/iam/iam_policy.go
```diff
@@ -0,0 +1,276 @@
...
+			removeMembers: sets.NewString(),
+		}
+		if batched != nil {
+			batched.roleModifications[req.role] = mod
+		}
+	}
+	switch req.action {
+	case actionAdd:
+		if mod.removeMembers.Has(req.member) {
+			return fmt.Errorf("conflicting remove of member %s", req.member)
```
liu-cong, line 219: Does this mean this was an `remove` request and then an `add`? Seems like a valid sequence of requests. Or is this a race condition? I wonder whether we should return an error here.

pkg/reconciler/identity/iam/iam_policy.go
```diff
@@ -0,0 +1,276 @@
...
+		}
+		mod.addMembers.Insert(req.member)
+	case actionRemove:
+		if mod.addMembers.Has(req.member) {
+			return fmt.Errorf("conflicting add of member %s", req.member)
+		}
+		mod.removeMembers.Insert(req.member)
+	}
+	if batched == nil {
```
liu-cong, line 228: It seems like if you move this block to the beginning, the logic is more clear and can hopefully simplify the `if mod==nil` and `if bached !=nil` checks. I understand that you may not need to run this if there is an error.

pkg/reconciler/identity/iam/iam_policy.go
```diff
@@ -0,0 +1,276 @@
...
+	return globalManager
+}
+
+// manager is an IAMPolicyManager which serializes and batches IAM policy changes to a Google
+// Service Account to avoid conflicting changes.
+type manager struct {
+	iam         gclient.IamClient
+	requestCh   chan *modificationRequest
+	pending     map[GServiceAccount]*batchedModifications // a non-nil batch indicates an outstanding request
+	getPolicyCh chan *getPolicyResponse
```
liu-cong, line 101: It seems you use this to serialize policy updates for an SA since there is one policy associated with an SA? If so, can you add a comment here?

pkg/reconciler/identity/iam/iam_policy.go
ian-mi: This is unlikely to be a valid sequence of requests since it means an add request was made while another process is still blocking on a remove request. It also seems wrong in this case to return  success to the remove if it was cancelled by a concurrent add.

pkg/reconciler/identity/iam/iam_policy.go
ian-mi: Done.

pkg/reconciler/identity/iam/iam_policy.go
liu-cong: I see. Hopefully this doesn't happen in our code.

pkg/reconciler/identity/iam/iam_policy.go
ian-mi: I would say that it's the manage process that serializes updates. This channel is used to communicate responses from GetIAMPolicy calls. I've added additional documentation to manage, let me know if anything is still unclear.

pkg/reconciler/identity/iam/iam_policy.go
ian-mi: Hopefully not, this would imply that two reconcilers are contending over the same ksa. In this case I think returning an error would be appropriate behavior.

pkg/reconciler/identity/iam/iam_policy.go
liu-cong: The updated documentation really helps! Thanks!

PR #841, opened 2020-04-15
Get default project ID from metadata

liu-cong:
Fixes #

## Proposed Changes

- Give a default value to TARGETS_CONFIG_PATH. Alternatively we can mark it as `required`
- If `PROJECT_ID` is not set, we get it from metadata.
- 

**Release Note**

```release-note

```

**Docs**

/assign @yolocs

### DIFF
--- a/pkg/broker/handler/pool/fanout/pool.go
+++ b/pkg/broker/handler/pool/fanout/pool.go
@@ -33,6 +33,7 @@ import (
 	"github.com/google/knative-gcp/pkg/broker/handler/processors/deliver"
 	"github.com/google/knative-gcp/pkg/broker/handler/processors/fanout"
 	"github.com/google/knative-gcp/pkg/broker/handler/processors/filter"
+	"github.com/google/knative-gcp/pkg/utils"
 )
 
 // SyncPool is the sync pool for fanout handlers.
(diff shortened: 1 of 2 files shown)


pkg/broker/handler/pool/fanout/pool.go
```diff
@@ -33,23 +33,31 @@ import (
...
+	projectID, err := utils.ProjectID(options.ProjectID)
+	if err != nil {
+		return nil, err
+	}
 	p := &SyncPool{
-		targets: targets,
-		options: pool.NewOptions(opts...),
+		targets:   targets,
+		options:   options,
+		projectID: projectID,
```
yolocs, line 60: There is already an option for project id: https://github.com/google/knative-gcp/blob/master/pkg/broker/handler/pool/options.go#L95-L99

pkg/broker/handler/pool/fanout/pool.go
```diff
@@ -33,23 +33,31 @@ import (
...
+	options   *pool.Options
+	targets   config.ReadonlyTargets
+	pool      sync.Map
+	projectID string
 }
 
 // StartSyncPool starts the sync pool.
 func StartSyncPool(ctx context.Context, targets config.ReadonlyTargets, opts ...pool.Option) (*SyncPool, error) {
+	options := pool.NewOptions(opts...)
+	projectID, err := utils.ProjectID(options.ProjectID)
```
yolocs, line 53: I'm thinking maybe do that in main.go and pass in as an option?

liu-cong: We already do that in `main`: https://github.com/google/knative-gcp/blob/master/cmd/broker/fanout/main.go#L184

This change is to handle the case where the PROJECT_ID env var is absent.

liu-cong: Or do you mean `main` should call `utils.ProjectID` and pass it as option?

I think we should do it here because projectID is optional. The pool should be able to find a good default if option is absent. Otherwise we should make project id mandatory for the pool.

yolocs: Oh I misunderstood what `utils.ProjectID` does. You're right. An alternative is to reuse `options.ProjectID` instead of adding a new struct field.

liu-cong: Using a struct field because `options.ProjectID` can be empty. We can of course override it with the default but I think let's keep them separate to avoid surprises. So `options` always contains original values that the caller provides.

yolocs: `options.ProjectID` being empty means no project ID is set and it's ok to assign default value to it. In fact, I would rather it be in `options.go` but that requires `NewOptions` to return error (which I now think is a good idea because we can also add validation there). You can leave that to me if you just want to keep this change small.

Issue #835, opened 2020-04-15
#2 Broker fanout/retry processors should directly read targets config


yolocs:
**Problem**
Currently some processors expect `Broker` and/or `Target` in context while some other expect broker key. To make it more consistent, we can consistently only store broker/target key in context and have each processor read TargetsConfig by keys.

PR #870, opened 2020-04-16
Cleanups for broker handler

yolocs:
## Proposed Changes

- Add `Key()` function to `config.Broker` and `config.Target` for easier access
- For consistency, now only put keys in context and have each processor use the key to find broker/trigger.
- Provide default project ID via `pool.Options`.

Refs:
- https://github.com/google/knative-gcp/issues/835
- https://github.com/google/knative-gcp/pull/841

### DIFF
--- a/pkg/broker/handler/pool/options_test.go
+++ b/pkg/broker/handler/pool/options_test.go
@@ -26,7 +26,10 @@ import (
 
 func TestWithProjectID(t *testing.T) {
 	want := "random"
-	opt := NewOptions(WithProjectID(want))
+	opt, err := NewOptions(WithProjectID(want))
+	if err != nil {
+		t.Errorf("NewOptions got unexpected error: %v", err)
+	}
 	if opt.ProjectID != want {
 		t.Errorf("options project id got=%s, want=%s", opt.ProjectID, want)
 	}
@@ -34,7 +37,11 @@ func TestWithProjectID(t *testing.T) {
 
 func TestWithHandlerConcurrency(t *testing.T) {
 	want := 10
-	opt := NewOptions(WithHandlerConcurrency(want))
+	// Always add project id because in some env the default value can't be retrieved.
+	opt, err := NewOptions(WithHandlerConcurrency(want), WithProjectID("pid"))
+	if err != nil {
+		t.Errorf("NewOptions got unexpected error: %v", err)
+	}
 	if opt.HandlerConcurrency != want {
 		t.Errorf("options handler concurrency got=%d, want=%d", opt.HandlerConcurrency, want)
 	}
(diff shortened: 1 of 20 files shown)


pkg/broker/handler/processors/filter/processor_test.go
```diff
@@ -242,19 +243,23 @@ func TestFilterProcessor(t *testing.T) {
 
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
+			testTarget := &config.Target{
+				Name:             "target",
+				Broker:           "broker",
+				Namespace:        "ns",
+				FilterAttributes: tc.filter,
+			}
+			testTargets := memory.NewEmptyTargets()
```
liu-cong, line 252: nit: `use memory.NewTargets(testTarget)`

pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -139,12 +146,17 @@ func TestDeliverFailure(t *testing.T) {
 	}
 	targetSvr := httptest.NewServer(targetClient)
 	defer targetSvr.Close()
-	p := &Processor{Requester: requester}
 
 	broker := &config.Broker{Namespace: "ns", Name: "broker"}
 	target := &config.Target{Namespace: "ns", Name: "target", Broker: "broker", Address: targetSvr.URL}
-	ctx := handlerctx.WithBroker(context.Background(), broker)
-	ctx = handlerctx.WithTarget(ctx, target)
+	testTargets := memory.NewEmptyTargets()
```
liu-cong, line 152: nit: use memory.NewTargets(testTarget)

pkg/broker/handler/pool/options_test.go
```diff
@@ -61,25 +76,41 @@ func TestWithReceiveSettings(t *testing.T) {
...
+	// Always add project id because in some env the default value can't be retrieved.
+	opt, err := NewOptions(WithProjectID("pid"))
+	if err != nil {
+		t.Errorf("NewOptions got unexpected error: %v", err)
+	}
 	if opt.PubsubClient != nil {
 		t.Errorf("options PubsubClient got=%v, want=nil", opt.PubsubClient)
 	}
-	opt = NewOptions(WithPubsubClient(&pubsub.Client{}))
+	// Always add project id because in some env the default value can't be retrieved.
```
liu-cong, line 98: nit: in some env the default value can't be retrieved -> the default value can only be retrieved if running in a GKE cluster

yolocs: Apparently it also works on my cloudtop :) I think it's a GCE thing.

liu-cong: yes you are right

pkg/broker/handler/processors/filter/processor_test.go
yolocs: `NewTargets` accepts a raw `TargetsConifg` right? I didn't do that because I would have to know the key/value mapping in the test (which causes a little leaky abstraction but mostly fine). But it also turns out using `MutateBroker` doesn't introduce more code so I chose it.

pkg/broker/handler/pool/options_test.go
yolocs: Done

Issue #785, opened 2020-04-07
Investigate using gorilla/mux for broker ingress routing


liu-cong:
**Problem**
A short explanation of the problem, including relevant restrictions.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

liu-cong:
An example: 
https://github.com/cloudevents/sdk-go/blob/master/v2/cmd/samples/http/receiver-gorilla/main.go

PR #880, opened 2020-04-17
Add YAMLs for gcp broker

### LINKED ISSUE #808: #1 Initial deployment strategy for broker data plane components
**Problem**
Once we have the broker data plane images, we need to create deployments for them in the cluster.

**Exit Criteria**
A cluster with the GCP broker enabled or present has a running data plane that can route events.

liu-cong:
Fixes #808

## Proposed Changes

- Deploy broker data plane with static yamls

If `broker-targets` confimap doesn't exist, it will be created by the broker controller. Based on my experience, the controller usually starts faster than the data plane components, and the configmap is available when data plane components starts up. Sometimes if the controllers starts up slower, we may need to wait a few seconds for data plane pods to try the remount. But I have never seen it taking longer than a few seconds. 

**Release Note**

```release-note
Add deployment yamls for GCP broker data plane components.
```

**Docs**

/assign @grantr


config/core/deployments/broker.yaml
```diff
@@ -0,0 +1,157 @@
+# Copyright 2019 Google LLC
```
yolocs, line 1: 2020

config/core/deployments/broker.yaml
```diff
@@ -0,0 +1,157 @@
...
+  name: broker-ingress
+  namespace: cloud-run-events
+  labels:
+    events.cloud.google.com/release: devel
+spec:
+  replicas: 1
+  selector:
+    matchLabels: &labels
+      app: cloud-run-events
+      role: broker-ingress
```
yolocs, line 27: Maybe do similar things as https://github.com/knative/eventing/blob/master/pkg/reconciler/broker/resources/ingress.go#L155-L160

config/core/deployments/broker.yaml
```diff
@@ -0,0 +1,157 @@
...
+      containers:
+      - name: broker-ingress
+        image: github.com/google/knative-gcp/cmd/broker/ingress
+        env:
+          - name: PORT
+            value: "8080"
+        volumeMounts:
+        - mountPath: /var/run/cloud-run-events/broker
+          name: broker-config
+        resources: # TODO(issue #876): determine good values for resource requests/limits
```
yolocs, line 41: Do we need parameters for metrics/tracing/logging?

config/core/deployments/broker.yaml
```diff
@@ -0,0 +1,157 @@
...
+  replicas: 1
+  selector:
+    matchLabels: &labels
+      app: cloud-run-events
+      role: broker-ingress
+  template:
+    metadata:
+      labels: *labels
+    spec:
+      containers:
```
yolocs, line 32: Since ingress is a service, better to have readiness and liveness probes?

config/core/deployments/broker.yaml
```diff
@@ -0,0 +1,157 @@
...
+  name: broker-fanout
+  namespace: cloud-run-events
+  labels:
+    events.cloud.google.com/release: devel
+spec:
+  replicas: 1
+  selector:
+    matchLabels: &labels
+      app: cloud-run-events
+      role: broker-fanout
```
yolocs, line 67: Similarly, maybe we should do more knative-style label keys

config/core/deployments/broker.yaml
```diff
@@ -0,0 +1,157 @@
...
+  selector:
+    matchLabels: &labels
+      app: cloud-run-events
+      role: broker-ingress
+  template:
+    metadata:
+      labels: *labels
+    spec:
+      containers:
+      - name: broker-ingress
```
yolocs, line 33: Does ingress need its own service account?

config/core/deployments/broker.yaml
```diff
@@ -0,0 +1,157 @@
...
+  name: broker-retry
+  namespace: cloud-run-events
+  labels:
+    events.cloud.google.com/release: devel
+spec:
+  replicas: 1
+  selector:
+    matchLabels: &labels
+      app: cloud-run-events
+      role: broker-retry
```
yolocs, line 120: Similar to my other two comments.

config/core/deployments/broker.yaml
yolocs, line 32: Also specify `ports` here with name `http`. I believe some mesh (e.g. istio) requires proper port naming.

config/core/services/broker.yaml
```diff
@@ -0,0 +1,30 @@
...
+  name: broker-ingress
+  namespace: cloud-run-events
+  labels:
+    events.cloud.google.com/release: devel
+spec:
+  selector:
+    app: cloud-run-events
+    role: broker-ingress
+  ports:
+    - name: ingress
```
yolocs, line 27: Better using `http` as the name.

config/core/deployments/broker.yaml
liu-cong: No, not at the moment. I heard that eventing may stop supporting the configmap feature for metrics/tracing/logging. Need to follow up on that.

yolocs: Hmm. What's the replacement? Operator?

config/core/deployments/broker.yaml
liu-cong: We will be serving those zpages in https://github.com/google/knative-gcp/issues/785, also created an issue to add the probes:https://github.com/google/knative-gcp/issues/882

config/core/deployments/broker.yaml
liu-cong: No, it doesn't at the moment.

config/core/deployments/broker.yaml
liu-cong: I am using the same style as other deployments in `knative-gcp`. I think we should be consistent with `knative-gcp`. What do you tink?

config/core/deployments/broker.yaml
liu-cong: Same as https://github.com/google/knative-gcp/pull/880#discussion_r410309018

config/core/deployments/broker.yaml
liu-cong: Same as https://github.com/google/knative-gcp/pull/880#discussion_r410309018

config/core/deployments/broker.yaml
liu-cong: I didn't follow up on that. I heart from @grantr .

Maybe just stop supporting dynamic config.

config/core/deployments/broker.yaml
yolocs: I didn't realize that. I agree keeping it consistent at the moment is a better idea.

liu-cong:
/hold 

Should support both secret and workload identity

config/core/deployments/controller.yaml
```diff
@@ -78,5 +78,4 @@ spec:
         secret:
           secretName: google-cloud-key
           optional: true
-      serviceAccount: controller
```
liu-cong, line 81: We already have `serviceAccountName` specified. `serviceAccount` is deprecated in favor of `serviceAccountName`

config/core/deployments/broker.yaml
grantr: This was a conversation related to the dissolution of the Knative Observability working group. Evan questioned if the complexity of the dynamic configuration was worth maintaining. No concrete guidance yet.

config/core/deployments/broker.yaml
```diff
@@ -0,0 +1,190 @@
...
+  template:
+    metadata:
+      labels: *labels
+    spec:
+      # Only needed if running with workload identity.
+      serviceAccountName: broker
+      containers:
+      - name: broker-ingress
+        image: ko://github.com/google/knative-gcp/cmd/broker/ingress
+        env:
```
grantr, line 38: The other two deployments have `SYSTEM_NAMESPACE`, `POD_NAME`, and metrics-related env vars. Why does the ingress container not have them? Seems like we'll need them when we start adding metrics and tracing.

liu-cong: The other two currently reports memstats and therefore need metrics domain env var. 

Yes once we add metrics to ingress we will need those env vars as well.  I can add those now or wait until we have metrics added.

grantr: Either way is fine with me. Feel free to unhold if you'd like to add the env vars later.

liu-cong: Sounds good. The binary should fail to start if we forget to add the required env var. So we we will find out easily if we forget to add them

liu-cong:
@grantr @yolocs 

I made some updates and now it's ready for review:
- Use the `cre-pubsub` GSA instead of the `cloud-run-events` GSA which is used by the controller.
- Moved broker yamls to `config/broker` so they are not installed by default. Currently GCP broker is an optional installation.
- Added an initial doc for how to install the GCP broker.

liu-cong:
cc @capri-xiyue

capri-xiyue:
I tested the installation guide with secret, it works.

PR #884, opened 2020-04-17
Remove status.ServiceAccountName from last reconcile circle in Identity reconciler

grac3gao:
Fixes #

## Proposed Changes

- Remove `status.ServiceAccountName` from last reconcile cycle in Identity reconciler at the beginning of the current reconcile cycle to avoid case: resources used to reconcile WorkloadIdentity successfully, but later failed, and needed to be deleted. `status.ServiceAccountName` from laster reconcile cycle will cause the deletion failed for unable to remove iam policy binding.
- Pubsub Source, Build Source and Pullsubscription don't have DefaultReSyncPeriod, which makes it hard to detect Workload Identity change in the Google Service Account side for them. Thus, add the DefaultReSyncPeriod.
-

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/events/build/controller.go
```diff
@@ -62,7 +64,8 @@ func NewController(
 	impl := cloudbuildsourcereconciler.NewImpl(ctx, r)
 
 	r.Logger.Info("Setting up event handlers")
-	cloudbuildsourceInformer.Informer().AddEventHandler(controller.HandleAll(impl.Enqueue))
+	cloudbuildsourceInformer.Informer().AddEventHandlerWithResyncPeriod(
```
Harwayne, line 67: I'm surprised this is needed, I don't see it eventing at all. Are we sure there is nothing else in the former or the controller that is doing the resync?

Harwayne: Talked over IM. Sounds like this is adding a 5 min re-reconcile, as opposed to the controller's more standard 10 hours. 

If we ever find this is reconciling too often, we can alter the time as appropriate.

Issue #834, opened 2020-04-15
#2 Send events to retry queue if the intial delivery fails


yolocs:
On initial deliver failure, send events to retry queue

PR #889, opened 2020-04-17
Send events to retry queue on initial delivery failure

yolocs:
Fixes https://github.com/google/knative-gcp/issues/834

## Proposed Changes

- Modify deliver processor to also support retry

**Release Note**

```release-note
Broker: send events to retry queue on initial delivery failure
```

**Docs**

### DIFF
--- a/pkg/broker/handler/processors/deliver/processor.go
+++ b/pkg/broker/handler/processors/deliver/processor.go
@@ -20,10 +20,12 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/cloudevents/sdk-go/v2/binding"
 	ceclient "github.com/cloudevents/sdk-go/v2/client"
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
 	"github.com/cloudevents/sdk-go/v2/event"
 	"github.com/cloudevents/sdk-go/v2/protocol"
+	"github.com/cloudevents/sdk-go/v2/protocol/pubsub"
 	"go.uber.org/zap"
 	"knative.dev/pkg/logging"
 
@@ -41,6 +43,14 @@ type Processor struct {
 
 	// Targets is the targets from config.
 	Targets config.ReadonlyTargets
+
+	// RetryOnFailure if set to true, the processor will send the event
+	// to the retry topic if the delivery fails.
+	RetryOnFailure bool
+
+	// RetryEvents is the Pubsub protocol to send events
+	// to the retry topic.
+	RetryEvents *pubsub.Protocol
 }
 
 var _ processors.Interface = (*Processor)(nil)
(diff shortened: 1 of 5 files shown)


pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -41,6 +43,14 @@ type Processor struct {
...
 	// Targets is the targets from config.
 	Targets config.ReadonlyTargets
+
+	// RetryOnFailure if set to true, the processor will send the event
+	// to the retry topic if the delivery fails.
+	RetryOnFailure bool
+
+	// RetryEvents is the Pubsub protocol to send events
+	// to the retry topic.
+	RetryEvents *pubsub.Protocol
```
liu-cong, line 53: Can we use CloudEvents pubsub Client instead of protocol?

pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -135,7 +140,7 @@ func TestDeliverSuccess(t *testing.T) {
 	}
 }
 
-func TestDeliverFailure(t *testing.T) {
+func TestDeliverFailureNoRetry(t *testing.T) {
```
liu-cong, line 143: I suggest consolidating these tests to a table test

pkg/broker/handler/processors/deliver/processor.go
yolocs: Why? I prefer protocol because it's more direct.

liu-cong: My understanding is that Client is a higher level interface that allows you to only deal with Events while with protocol you still need to deal with `message`.

pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -140,223 +140,105 @@ func TestDeliverSuccess(t *testing.T) {
...
-		// it's not possible to use Receive to easily return error.
-		if err := resp(ctx, nil, &cehttp.Result{StatusCode: http.StatusInternalServerError}); err != nil {
-			t.Errorf("unexpected error from target responding event: %v", err)
-		}
-		defer msg.Finish(nil)
-	}()
-
-	if err := p.Process(ctx, &origin); err == nil {
-		t.Error("expected error from processing")
+func TestDeliverFailure(t *testing.T) {
```
liu-cong, line 143: Further, it seems you can even merge the success case in the table test.  I will lgtm and it's up to you whether to do that or not

pkg/broker/handler/processors/deliver/processor.go
yolocs: I don't have a strong preference actually. How about let's defer this a bit and get this PR in first?

pkg/broker/handler/processors/deliver/processor_test.go
yolocs: I wanted to. But the success case also involves an event reply which doesn't exist in the failure case. Separating them is probably a little bit clearer?

pkg/broker/handler/processors/deliver/processor.go
liu-cong: It should be as simple as replace the protocol with client and call `client.Send(ctx, event)`.  If we don't do it now, I am afraid we may never get to it.

Examples in ingress:

https://github.com/google/knative-gcp/blob/master/pkg/broker/ingress/multi_topic_decouple_sink.go#L88-L89

https://github.com/google/knative-gcp/blob/master/pkg/broker/ingress/multi_topic_decouple_sink.go#L110-L123

yolocs: > we may never get to it

Maybe it's an indicator it doesn't make any difference?

liu-cong: I think it's better to not get to the `bindings` concept if you don't need to. This makes the logic much easier to follow without understanding additional concepts. 

So both approach has the same effect (I believe the client implementation internally just converts events to message and calls protocol). But it's easier for someone else to read and maintain using a higher level abstraction.

PR #890, opened 2020-04-17
Use new google.golang.org/protobuf API with targets.proto

ian-mi:
## Proposed Changes

- Compile targets.proto using the google.golang.org/protobuf compiler.
- Use google.golang.org/protobuf packages in targets related code.
- Update github.com/golang/protobuf version to 1.4 to ensure compability with the new proto API.


pkg/broker/config/cache.go
```diff
@@ -114,7 +115,7 @@ func (ct *CachedTargets) Bytes() ([]byte, error) {
 // String returns the text format of all the targets.
 func (ct *CachedTargets) String() string {
 	val := ct.Load()
-	return val.String()
+	return prototext.MarshalOptions{}.Format(val)
```
grantr, line 118: I suggest specifying `MarshalOptions{Multiline: false}` to clarify why this is used instead of `prototext.Format`. Arguably it's a bug that [`prototext.Format()` doesn't use the zero-value `MarshalOptions`](https://github.com/protocolbuffers/protobuf-go/blob/3b9eee12916ce611400f93d6c2fed2fc2911b0ad/encoding/prototext/encode.go#L32-L34).

PR #891, opened 2020-04-17
Use standard pubsub client in Broker controller

### LINKED ISSUE #886: Verify correct pubsub calls made by broker controller
**Describe the bug**
@ian-mi points out in https://github.com/google/knative-gcp/pull/739#discussion_r410380685 that the pubsub mock testing client is insufficient because it can't check whether a call to pubsub was actually made.

 **Expected behavior**
Unit tests verify that pubsub calls were made correctly by the controller.

 **Additional context**
This is probably an issue for all pubsub-related controller code. We can develop a solution for the broker tests, then use it for other controllers.

grantr:
Broker controller no longer uses the gclient interface.

Tests now use the pstest fake pubsub server instead of the gclient mock.
Tests verify pubsub resource creation and deletion.

Fixes #886

## Proposed Changes

- Use standard pubsub client in Broker controller
- Use pstest as the testing interface for Broker controller


grantr:
There's a bug in pstest keeping this from being merged and passing unit tests: https://github.com/googleapis/google-cloud-go/issues/1938. It looks like we can comment out https://github.com/googleapis/google-cloud-go/blob/ba65210b25f32592a049d54e2266957f4b8ebe8c/pubsub/pstest/fake.go#L536 and all tests pass, but I'm not sure if it's worth patching a dependency yet.

grantr:
/hold
Holding until the vendor patch is resolved.

pkg/reconciler/broker/broker.go
```diff
@@ -320,12 +321,15 @@ func (r *Reconciler) deleteDecouplingTopicAndSubscription(ctx context.Context, b
...
-		logger.Error("Failed to create Pub/Sub client", zap.Error(err))
-		return err
+	client := r.pubsubClient
+	if client == nil {
+		client, err := pubsub.NewClient(ctx, projectID)
+		if err != nil {
+			logger.Error("Failed to create Pub/Sub client", zap.Error(err))
+			return err
+		}
+		defer client.Close()
```
yolocs, line 331: Curious why creating client here rather than in the controller.go and reuse the same client over and over? Similar question for several other places.

yolocs:
> There's a bug in pstest keeping this from being merged and passing unit tests: [googleapis/google-cloud-go#1938](https://github.com/googleapis/google-cloud-go/issues/1938). It looks like we can comment out [googleapis/google-cloud-go:pubsub/pstest/fake.go@`ba65210`#L536](https://github.com/googleapis/google-cloud-go/blob/ba65210b25f32592a049d54e2266957f4b8ebe8c/pubsub/pstest/fake.go#L536) and all tests pass, but I'm not sure if it's worth patching a dependency yet.

Could you also point out which part of the test is failing because of this bug? Thanks

grantr:
https://github.com/googleapis/google-cloud-go/issues/1938 is now fixed in master.

pkg/reconciler/broker/broker.go
grantr: If project id ever changes between reconciles, then a single client wouldn't work, but we don't support that today. Can we properly close the client if it's created by the controller?

grantr:
> Could you also point out which part of the test is failing because of this bug? Thanks

The delete test fails when the reconciler tries to delete a topic, then delete a subscription. Deleting the subscription panics due to a double close in pstest.

grantr:
/hold cancel
Ready for review.

This PR contains a temporary replace to upgrade the pubsub client module to master. It can be removed when the next pubsub client release happens.

pkg/reconciler/broker/broker.go
yolocs: It seems it's recommended to reuse pubsub client: https://pkg.go.dev/cloud.google.com/go/pubsub?tab=doc#Client
Maybe better to only recreate the client if the project is changed. But I don't know how to handle client close properly if we reuse the pubsub client.

yolocs: One thing I can think of is to override https://github.com/google/knative-gcp/blob/12200f7088cd728b7e1a683fe5b6db80b7a36468/vendor/knative.dev/pkg/controller/controller.go#L354

grantr: Just pushed a change to use a single pubsub client for all reconciles if possible. If not possible the reconcilers fall back to the original behavior. PTAL.

pkg/reconciler/broker/controller.go
```diff
@@ -59,14 +61,29 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
...
+	if err != nil {
+		logging.FromContext(ctx).Error("Failed to create controller-wide Pub/Sub client", zap.Error(err))
+	}
+
+	if client != nil {
+		go func() {
+			<-ctx.Done()
+			client.Close()
+		}()
+	}
```
yolocs, line 77: Nice.

pkg/reconciler/testing/pstest.go
```diff
@@ -0,0 +1,169 @@
+/*
+Copyright 2019 The Knative Authors
```
yolocs, line 2: nit: Copyright 2020 Google LLC

PR #897, opened 2020-04-20
Channels v1beta1

### LINKED ISSUE #614: Create v1beta1 types

Harwayne:
Fixes #614.

## Proposed Changes

- Create v1beta1 go types for the Channel.
- All code is copied from the v1alpha1 type. 
  - There is an upstream change to the Channel duck type:
      1. status.subscribableTypesStatus -> status.subscribableStatus
        - Remvoes a layer of JSON: `status.subscribableStatus.subscribers` is now `status.subscribers`.

---

This was generated using:
```shell
foo=messaging
```

### Release Note
```release-notes
NONE
```


pkg/apis/messaging/v1beta1/channel_defaults.go
```diff
@@ -0,0 +1,50 @@
+/*
+Copyright 2019 The Knative Authors
```
yolocs, line 2: Maybe change all to 2020?

Harwayne: I'm not really sure what it should be, but because the other PRs doing this didn't change the years, I'll leave this one unchanged too.

PR #907, opened 2020-04-21
Update broker data plane pod annotation to trigger configmap volume refresh

### LINKED ISSUE #851: #2 Make sure data plane configmap propagation delay is acceptable
**Problem**
* Investigate configmap volume mount delay for broker, it it takes longer than a few seconds, try to implement the "annotation hack" in controller
* 

**Exit Criteria**

**Time Estimate (optional):** 
2

/area broker

liu-cong:
Fixes #851

## Proposed Changes

- After broker-targets configmap is updated, also update annotation on data plane pods to trigger volume refresh
- 
-

**Release Note**

```release-note
NONE
```

**Docs**

cc @grantr @yolocs @ian-mi


pkg/reconciler/broker/resources/deployments.go
```diff
@@ -0,0 +1,54 @@
...
+		return fmt.Errorf("error converting deployment selector to label selector for %v/%v: %w", ns, name, err)
+	}
+	pods, err := pl.Pods(ns).List(selector)
+	if err != nil {
+		return fmt.Errorf("error listing pods for deployment %v/%v: %w", ns, name, err)
+	}
+
+	for _, pod := range pods {
+		err = multierr.Append(err, UpdateVolumeGeneration(kubeClient, pod))
+	}
```
yolocs, line 52: Curious - why not update deployment (`spec.template.metadata.annotations`) instead?

pkg/reconciler/broker/resources/pods.go
```diff
@@ -0,0 +1,48 @@
...
+// trigger volume mount refresh (for configmap).
+func UpdateVolumeGeneration(kubeClientSet kubernetes.Interface, p *corev1.Pod) error {
+	pod := p.DeepCopy()
+	annotations := pod.GetAnnotations()
+	if annotations == nil {
+		annotations = make(map[string]string, 1)
+	}
+	gen, _ := strconv.Atoi(annotations[volumeGenerationKey])
+	annotations[volumeGenerationKey] = strconv.Itoa(gen + 1)
+	pod.SetAnnotations(annotations)
```
yolocs, line 42: Is it possible (or does it matter) if this generation differs from each other? An alternative is to have the controller to write the config generation as a configmap annotation and use it for pod annotation (instead of reading each pod for the previous value). This might also appear useful when check if a pod has the latest config loaded or not.

pkg/reconciler/broker/resources/deployments.go
liu-cong: Good question! Updating this actually causes the pod to be recreated. Added some comments for this.

yolocs: Good to know!

pkg/reconciler/broker/resources/pods.go
liu-cong: We only need an update on the annotation so actual value doesn't matter. But I like the other point that we can compare pod annotation with configmap annotation.

Added an issue https://github.com/google/knative-gcp/issues/913

liu-cong: I almost tried to do it in this PR. But not quite sure how to test the configmap generation logic with the way the controller tests are setup. Need some more investigation. I will do it as a follow up ASAP.

yolocs: Sg.

pkg/reconciler/broker/resources/deployments.go
```diff
@@ -0,0 +1,58 @@
...
+func UpdateVolumeGenerationForDeployment(kubeClient kubernetes.Interface, dl appsv1listers.DeploymentLister, pl corev1listers.PodLister, ns, name string) error {
+	d, err := dl.Deployments(ns).Get(name)
+	if err != nil {
+		if apierrs.IsNotFound(err) {
+			return fmt.Errorf("deployment %v/%v doesn't exist: %w", ns, name, err)
+		}
+		return fmt.Errorf("error getting deployment %v/%v: %w", ns, name, err)
+	}
+
+	selector, err := metav1.LabelSelectorAsSelector(d.Spec.Selector)
```
yolocs, line 45: Out of curiosity, it's us creating the deployments and we know the labels we put on the pods. Do we really need to get the label selector from the deployments?

pkg/reconciler/broker/resources/pods.go
yolocs: I don't know how we should handle the annotation for new pods though.

pkg/reconciler/broker/resources/deployments.go
liu-cong: I am doing this so we don't have to make sure whatever we use here is consistent with the yamls, or in case the users change the yamls. 

Anyways, the idea is that once we start to implement BrokerCell, the the BrokerCell controller will provide the labels,  so yes we won't need to get the labels from the deployment.

pkg/reconciler/broker/resources/pods.go
liu-cong: I don't think it's very important for new pods to have the annotation. New pods should get the latest configmap anyway.

For debugging purpose, we can put the generation number as part of the targets config. So we can still find the generation number from within the configmap volume of a new pod.

pkg/reconciler/broker/broker.go
```diff
@@ -425,6 +433,9 @@ func (r *Reconciler) updateTargetsConfig(ctx context.Context) error {
 		if err != nil {
 			return fmt.Errorf("error creating targets ConfigMap: %w", err)
 		}
+		if err := r.reconcileDataPlaneDeployments(); err != nil {
+			r.Logger.Warnf("Error reconciling data plane deployments: %w", err)
```
yolocs, line 437: I'm not really sure if `%w` can work with formatter other than `fmt.Errorf`. Do you know?

pkg/reconciler/broker/resources/deployments.go
```diff
@@ -0,0 +1,58 @@
...
+	appsv1listers "k8s.io/client-go/listers/apps/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+// UpdateVolumeGenerationForDeployment updates the volume generation annotation
+// for all pods in this deployment. This technique is used to trigger immediate
+// configmap volume refresh without restarting the pod, see
+// https://kubernetes.io/docs/tasks/configure-pod-container/configure-pod-configmap/#mounted-configmaps-are-updated-automatically
+// NOTE: update spec.template.metadata.annotations on the deployment resource
+// causes pods to bre recreated.
```
yolocs, line 35: Typo?

pkg/reconciler/broker/resources/deployments.go
yolocs: Add a TODO would be helpful.

pkg/reconciler/broker/broker.go
grantr: I think that's true. This should probably be `%v`.

pkg/reconciler/broker/resources/deployments.go
liu-cong: good catch. fixed.

pkg/reconciler/broker/resources/deployments.go
liu-cong: Done

pkg/reconciler/broker/broker.go
liu-cong: I thought I was doing Errorf... Good catch, thanks

liu-cong:
/unhold

Thanks @yolocs and @grantr for the review!

PR #919, opened 2020-04-22
CRD Generation

### LINKED ISSUE #920: Generate CRD Schemas
**Problem**
Most of our CRDs reuse the same underlying structs for common functionality. For example, most sources contain `spec.pubSubSpec`. However, each CRD's YAML must contain its own OpenAPIV3Schema. 

We should have a tool that makes it easy to use a common schema 'snippet' for these shared types. 

Developer

**Exit Criteria**
- A developer can add a field or extra OpenAPI schema validation to an existing field that is used by multiple types without having to manually alter the YAML for every CRD that uses that type.
  - E.g. I add extra schema validation to `spec.pubSubSpec.foo`. That field is used by a dozen sources. I do not have to update a dozen YAML files. Instead I only need to update one.

This need became apparent while working on #614 and #615. While working on those, I noticed that almost every Source had a _slightly_ different schema for the commonly shared types. Even more fun was that they sometimes were at different levels of nesting, so copy-pasting was difficult and error prone.

Harwayne:
Fixes #920.

## Proposed Changes

- Create a tool to generate CRD YAMLs using templates.

**Release Note**

```release-note
NONE
```


Harwayne:
Having seen inconsistencies in the CRDs pop-up again, I would like to get this submitted. Please take a look and see if you think this is a reasonable solution.

Harwayne:
/close

This tool was not adopted. Instead there seems to be mostly manual changes with YAML anchors when possible. A variant of https://github.com/Harwayne/knative-gcp/blob/sources-structural/hack/crdgeneration/generate_crds.sh was used to generate CRDs in knative/eventing, with minor manual tweaking.

Issue #781, opened 2020-04-07
Investigate switching pubsub testing to pstest


grantr:
We currently have a copy of the pubsub client interface so we can mock methods for unit testing (https://github.com/google/knative-gcp/tree/master/pkg/gclient/pubsub). The pubsub client now has a `pstest` package that might make this unnecessary.

grantr:
Should also look into this in the CloudEvents SDK.

PR #923, opened 2020-04-22
Use pstest to Test Broker Ingress

### LINKED ISSUE #786: Add integration test for broker ingress
**Problem**
Prepare pubsub topics, start ingress, send events to it, and pull events from pubsub to verify.

**Exit Criteria**

**Time Estimate (optional):** 

ian-mi:
This allows the handler test to effectively function as an integration test connecting over GRPC to the test pubsub server.

Fixes #786
See #781

### DIFF
--- a/pkg/broker/ingress/handler_test.go
+++ b/pkg/broker/ingress/handler_test.go
@@ -136,17 +136,15 @@ func TestHandler(t *testing.T) {
 	}
 
 	client := nethttp.Client{}
-	t.Cleanup(client.CloseIdleConnections)
+	defer client.CloseIdleConnections()
 	for _, tc := range tests {
 		t.Run(tc.name, func(t *testing.T) {
 			ctx := logging.WithLogger(context.Background(), logtest.TestLogger(t))
 			ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
-			t.Cleanup(cancel)
+			defer cancel()
 
 			psSrv := pstest.NewServer()
-			t.Cleanup(func() {
-				psSrv.Close()
-			})
+			defer psSrv.Close()
 
 			url := createAndStartIngress(ctx, t, psSrv)
 			rec := setupTestReceiver(ctx, t, psSrv)
--- a/pkg/broker/ingress/handler_test.go
+++ b/pkg/broker/ingress/handler_test.go
@@ -227,12 +227,12 @@ func setupTestReceiver(ctx context.Context, t *testing.T, psSrv *pstest.Server) 
 // createAndStartIngress creates an ingress and calls its Start() method in a goroutine.
 func createAndStartIngress(ctx context.Context, t *testing.T, psSrv *pstest.Server) (string, func()) {
 	p, cancel := createPubsubClient(ctx, t, psSrv)
-	decouple, err1 := NewMultiTopicDecoupleSink(ctx,
+	decouple, err := NewMultiTopicDecoupleSink(ctx,
 		WithBrokerConfig(memory.NewTargets(brokerConfig)),
 		WithPubsubClient(p))
-	if err1 != nil {
+	if err != nil {
 		cancel()
-		t.Fatalf("Failed to create decouple sink: %v", err1)
+		t.Fatalf("Failed to create decouple sink: %v", err)
 	}
 
 	receiver := &testHttpMessageReceiver{urlCh: make(chan string)}
--- a/pkg/broker/ingress/handler_test.go
+++ b/pkg/broker/ingress/handler_test.go
@@ -23,16 +23,28 @@ import (
 	nethttp "net/http"
 	"net/http/httptest"
 	"testing"
+	"time"
 
+	"cloud.google.com/go/pubsub"
+	"cloud.google.com/go/pubsub/pstest"
 	cloudevents "github.com/cloudevents/sdk-go/v2"
 	"github.com/cloudevents/sdk-go/v2/binding"
+	cecontext "github.com/cloudevents/sdk-go/v2/context"
 	"github.com/cloudevents/sdk-go/v2/protocol/http"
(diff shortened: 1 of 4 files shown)


pkg/broker/ingress/handler_test.go
```diff
@@ -124,11 +136,20 @@ func TestHandler(t *testing.T) {
 	}
 
 	client := nethttp.Client{}
-	defer client.CloseIdleConnections()
+	t.Cleanup(client.CloseIdleConnections)
```
Harwayne, line 139: Prow isn't using go 1.14 yet.

pkg/broker/ingress/options.go
```diff
@@ -40,28 +40,41 @@ func WithPort(port int) HandlerOption {
...
+	}
+}
 
 // WithPubsubClient specifies the pubsub client to use.
-func WithPubsubClient(client cloudevents.Client) MultiTopicDecoupleSinkOption {
-	return func(sink *multiTopicDecoupleSink) {
-		sink.client = client
+func WithPubsubClient(client *pubsub.Client) MultiTopicDecoupleSinkOption {
+	return func(opts *multiTopicDecoupleSinkOptions) {
+		opts.pubsub = client
```
Harwayne, line 71:
```suggestion
func WithPubsubClient(ps *pubsub.Client) MultiTopicDecoupleSinkOption {
	return func(opts *multiTopicDecoupleSinkOptions) {
		opts.pubsub = ps
```

Mostly because it feels like an error to see `opts.pubsub = client` when `opts.client` exists too. I know its not an error, but just feels like it on first glance.

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -36,36 +37,40 @@ import (
 const projectEnvKey = "PROJECT_ID"
 
 // NewMultiTopicDecoupleSink creates a new multiTopicDecoupleSink.
-func NewMultiTopicDecoupleSink(ctx context.Context, options ...MultiTopicDecoupleSinkOption) (*multiTopicDecoupleSink, error) {
-	sink := &multiTopicDecoupleSink{
-		logger: logging.FromContext(ctx),
-	}
-
+func NewMultiTopicDecoupleSink(ctx context.Context, options ...MultiTopicDecoupleSinkOption) (sink *multiTopicDecoupleSink, err error) {
```
Harwayne, line 40: I don't think I've seen us use the named return variables in the past. My preference would be to not use them.

pkg/broker/ingress/handler_test.go
ian-mi: Ah darn, I can no longer live without T.Cleanup. I will try bumping the go version: https://github.com/knative/test-infra/pull/1953

Harwayne: It looks like the only place in this PR that we want `t.Cleanup` rather than `defer` is line 195, which could just return a function for the caller to defer for now.

Actually, I think we would prefer `defer` on lines 144 and 147, so things are shutdown more quickly.

ian-mi: It would certainly be possible to thread through a cleanup function, but it would complicate the test functions a bit. I can do this if prow can't be switched to go 1.14 yet, but if it can this seems like a good time to make the switch. I believe the plan is for the rest of knative to switch to go 1.14  before the next release as well.

pkg/broker/ingress/multi_topic_decouple_sink.go
ian-mi: OK, I've removed them.

pkg/broker/ingress/handler_test.go
ian-mi: I've gone ahead and removed use of T.Cleanup for now, but it's not pretty. Hopefully I'll be able to revert it shortly.

pkg/broker/ingress/handler_test.go
```diff
@@ -140,52 +161,99 @@ func TestHandler(t *testing.T) {
...
+			select {
+			case <-ctx.Done():
+				t.Fatalf("test cancelled or timed out: %v", ctx.Err())
+			default:
+			}
 		})
 	}
 }
 
+func createPubsubClient(ctx context.Context, t *testing.T, psSrv *pstest.Server) (*pubsub.Client, func()) {
```
liu-cong, line 190: Can we move this to a util package so other tests can use it too?

Harwayne:
/hold

Holding for other's comments.

pkg/broker/ingress/handler_test.go
Harwayne: Ooph, that is uglier than I thought it would be. `t.Cleanup` will be much nicer.

pkg/broker/ingress/options.go
```diff
@@ -40,28 +40,41 @@ func WithPort(port int) HandlerOption {
...
 		}
-		h.decouple, err = NewMultiTopicDecoupleSink(context.Background(), WithPubsubClient(client))
+		h.decouple, err = NewMultiTopicDecoupleSink(context.Background(), WithPubsubClient(p))
 		return err
 	}
 }
 
 // MultiTopicDecoupleSinkOption is the option to configure multiTopicDecoupleSink.
-type MultiTopicDecoupleSinkOption func(sink *multiTopicDecoupleSink)
+type MultiTopicDecoupleSinkOption func(sink *multiTopicDecoupleSinkOptions)
```
liu-cong, line 53: I am curious to know how is using a new `options` struct compared to functional option directly applied to the main struct. I have seen both used but not sure what's the difference.

ian-mi: The options struct is used to accumulate options that aren't members of the target struct such as the pubsub client.

pkg/broker/ingress/handler_test.go
ian-mi: I'll move these out once I can revert the cleanup change.

Issue #838, opened 2020-04-15
#2 Broker fanout fails for recreated brokers


liu-cong:
**Describe the bug**
For a recreated broker, fanout fails with 
```
{"level":"error","ts":"2020-04-14T19:00:58.790Z","logger":"broker-fanout","caller":"fanout/pool.go:132","msg":"handler for broker has stopped with error","commit":"5f3eb60","broker":"default/test-broker","error":"rpc error: code = NotFound desc = Resource not found (resource=projects/conliu-kgcp-dev/subscriptions/cre-bkr_default_test-broker_a217e2d9-65e3-4a54-bcb3-d663d500158e).","stacktrace":"github.com/google/knative-gcp/pkg/broker/handler/pool/fanout.(*SyncPool).syncOnce.func2.1\n\tgithub.com/google/knative-gcp/pkg/broker/handler/pool/fanout/pool.go:132\ngithub.com/google/knative-gcp/pkg/broker/handler.(*Handler).Start.func1\n\tgithub.com/google/knative-gcp/pkg/broker/handler/handler.go:58"}
```

 **Expected behavior**
Fanout should deliver events.

 **To Reproduce**
* Create a broker, create a trigger to this broker and make sure events can be delivered
* Delete the broker and recreate it, create a trigger to the broker
* Send an event to the broker and fanout won't deliver the event to the consumers

 **Knative-GCP release version**

**Time Estimate (optional):**
2

 **Additional context**
After a broker is recreated, its decouple queue topic/subscription are updated. I looked at the targets config file in the fanout pod. It was updated with new topic/subscription for the decouple queue. It seems that fanout doesn't pick up the new DecoupleQueue config.

capri-xiyue:
+1 I encountered the same issue when I did manual test

yolocs:
It would be really helpful if you can provide the following context.

1. By "recreate", do you mean 1) create a broker 2) delete it 3) create the broker again with the same name?
2. Was the broker in TargetsConfig ever deleted?
3. Did the 2nd broker get new decouple topic/sub (different from the 1st broker)?
4. Was the decouple topic/sub of the 1st broker ever deleted?

I suspect the problem lies at:

https://github.com/google/knative-gcp/blob/c7e834b0b446177475915c025a5b01f38a35690f/pkg/broker/handler/pool/fanout/pool.go#L114-L117

[...]

PR #149, opened 2019-08-09
Decorator takes an input and arguments it with extensions.

n3wscott:
This is a POC.

PR #934, opened 2020-04-23
Reconcile handlers on decouple/retry subscription updates

yolocs:
Fixes https://github.com/google/knative-gcp/issues/838

## Proposed Changes

- Reconcile handlers on decouple/retry subscription updates
- Stop deleting the handler from pool on handler exit. This is dangerous because such a deletion could remove a newly created handler for the same broker. Now instead, store/delete handlers always happen in the same thread.

**Release Note**

```release-note
Reconcile handlers on decouple/retry subscription updates
```

**Docs**

### DIFF
--- a/pkg/broker/handler/handler.go
+++ b/pkg/broker/handler/handler.go
@@ -75,8 +75,6 @@ func (h *Handler) Stop() {
 
 func (h *Handler) handle(ctx context.Context) {
 	for {
-		// TODO(yolocs): we need to update dep for fix:
-		// https://github.com/cloudevents/sdk-go/pull/430
 		msg, err := h.PubsubEvents.Receive(ctx)
 		// It doesn't seem like that these errors will even happen.
 		if err == io.EOF {
--- a/pkg/broker/handler/handler.go
+++ b/pkg/broker/handler/handler.go
@@ -48,6 +48,7 @@ type Handler struct {
 
 	// cancel is function to stop pulling messages.
 	cancel context.CancelFunc
+	alive  bool
 }
 
 // Start starts the handler.
@@ -55,7 +56,7 @@ type Handler struct {
 func (h *Handler) Start(ctx context.Context, done func(error)) {
 	ctx, h.cancel = context.WithCancel(ctx)
 	go func() {
-		defer h.cancel()
+		defer h.Stop()
 		done(h.PubsubEvents.OpenInbound(ctx))
 	}()
 
@@ -66,11 +67,18 @@ func (h *Handler) Start(ctx context.Context, done func(error)) {
 	for i := 0; i < curr; i++ {
 		go h.handle(ctx)
 	}
+	h.alive = true
 }
 
 // Stop stops the handlers.
 func (h *Handler) Stop() {
 	h.cancel()
+	h.alive = false
+}
+
+// IsAlive indicates whether the handler is alive.
+func (h *Handler) IsAlive() bool {
+	return h.alive
 }
 
 func (h *Handler) handle(ctx context.Context) {
--- a/pkg/broker/handler/handler.go
+++ b/pkg/broker/handler/handler.go
@@ -19,6 +19,7 @@ package handler
 import (
 	"context"
 	"io"
+	"sync/atomic"
 	"time"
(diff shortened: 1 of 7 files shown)


yolocs:
cc @grac3gao

pkg/broker/handler/pool/retry/pool.go
```diff
@@ -50,6 +50,25 @@ type SyncPool struct {
 	deliverClient ceclient.Client
 }
 
+type handlerCache struct {
```
liu-cong, line 53: I guess there is already a plan to consolidate fanout pool and retry pool since they share a lot in common?

pkg/broker/handler/handler.go
```diff
@@ -48,14 +49,18 @@ type Handler struct {
 
 	// cancel is function to stop pulling messages.
 	cancel context.CancelFunc
+	alive  atomic.Value
```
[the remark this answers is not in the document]
liu-cong: Actually I misunderstood.

pkg/broker/handler/pool/fanout/pool.go
```diff
@@ -54,6 +54,25 @@ type SyncPool struct {
...
+type handlerCache struct {
+	handler.Handler
+	b *config.Broker
+}
+
+func (hc *handlerCache) requiresRestart(b *config.Broker) bool {
+	if !hc.IsAlive() {
+		return true
+	}
+	if b == nil || b.DecoupleQueue == nil {
```
liu-cong, line 66: I suggest moving this to a `shouldDelete` function because in this case we don't want to `restart` but simply `delete` the handler.

yolocs: We never really reuse a handler. Restart here means delete and recreate.

pkg/broker/handler/handler.go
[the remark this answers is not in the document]
yolocs: In practice it’s really fine. But because it’s accessed by multiple goroutine, the test complains race condition. So for safety...

pkg/broker/handler/pool/fanout/pool.go
liu-cong: It seems the code will fail if b.DecoupeQueue is nil when we try to use it to create the handler.

yolocs: Let me rename it to `shouldReconcile`.

`b.DecoupleQueue.Topic` empty shouldn't really happen in this case. If it really happens, then let the handler creation fail. If the data eventually gets correct, at that point a handler will be created.

pkg/broker/handler/pool/retry/pool.go
yolocs: We have extracted some common code to pkg/broker/handler/pool/pool.go. It's not very obvious to me how to share the remaining code. Feel free to share your thoughts.

For `handlerCache` it may seem that you can just store topic/sub (vs. broker and target). But that may not be the only thing we check to modify handlers. E.g. we may need to check state to stop some handlers.

pkg/broker/handler/pool/fanout/pool.go
yolocs: Also added a comment.

liu-cong: It doesn't look we will return an error gracefully if `b.DecoupleQueue == nil`, see line #149 in RangeBroker. This probably doesn't happen in practice but also it's hard to ensure the behavior of the controller won't change. 

It doesn't need to be solved in this PR but maybe we should have a `Validate` method on the config and the controller only persists validated results.

liu-cong: Created an issue: https://github.com/google/knative-gcp/issues/939

pkg/broker/handler/pool/retry/pool.go
liu-cong: Just at a glance there are some similarities but haven't read carefully. I will give it a try and see if it makes sense to consolidate more :)

Let's keep this PR as is.

PR #942, opened 2020-04-24
Truncate pubsub names if greater than 255 chars

### LINKED ISSUE #831: #1 Broker reconciler doesn't validate length of pubsub resource names
**Describe the bug**
Pubsub resource names can't be longer than 255 characters. The resource names generated for brokers and triggers can be longer than that if their K8s names are especially long.

Similar to kmeta.ChildName, we need a strategy to truncate those names safely if they are greater than the max length. We probably shouldn't use kmeta.ChildName directly because its max length is too short.

 **Expected behavior**
Generated names are never longer than 255 characters.

 **To Reproduce**
Namespace max length: 63
Broker and trigger name max length: 253
UID length: 36
Prefix and separators: 10

HEAD

 **Additional context**

grantr:
Uses similar hashing strategy as kmeta.ChildName. Can't use kmeta.ChildName directly because the concatenation is different and the max length is greater.

Fixes #831

## Proposed Changes

- Truncate pubsub resource names deterministically if they would be greater than 255 characters

### DIFF
--- a/pkg/reconciler/broker/resources/names.go
+++ b/pkg/reconciler/broker/resources/names.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 
 	brokerv1beta1 "github.com/google/knative-gcp/pkg/apis/broker/v1beta1"
+	"k8s.io/apimachinery/pkg/types"
 )
 
 // For reference, the minimum number of characters available for a name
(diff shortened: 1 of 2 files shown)


grantr:
@ian-mi I believe you mentioned this in your review of #739

pkg/reconciler/broker/resources/names.go
```diff
@@ -30,34 +43,60 @@ const (
...
+	nameOver := len(s) - pubsubMax
+	nameMax := len(n) - nameOver
+
+	return fmt.Sprintf("%s_%s_%s_%s", prefix, ns, hashedTruncate(n, nameMax), uid)
+}
+
+// hashedTruncate truncates a string longer than l by replacing the end of the
+// string with a hash of its contents. Panics if l is less than the length of
+// the hash value (currently 32 characters).
+func hashedTruncate(s string, max int) string {
```
liu-cong, line 90: Why do we use md5 and why not just simply truncate?

pkg/reconciler/broker/resources/names.go
```diff
@@ -30,34 +43,60 @@ const (
...
 // GenerateRetrySubscriptionName generates a deterministic subscription name
-// for a Trigger.
+// for a Trigger. If the subscription name would be longer than allowed by
+// PubSub, the Trigger name is truncated to fit.
 func GenerateRetrySubscriptionName(t *brokerv1beta1.Trigger) string {
-	return fmt.Sprintf("cre-tgr_%s_%s_%s", t.Namespace, t.Name, string(t.UID))
+	return truncatedPubsubResourceName("cre-tgr", t.Namespace, t.Name, string(t.UID))
+}
+
+func truncatedPubsubResourceName(prefix, ns, n, uid string) string {
```
Harwayne, line 74: Pass `uid` as `type.UID` instead of string.

pkg/reconciler/broker/resources/names_test.go
```diff
@@ -153,3 +212,46 @@ func trigger(ns, n, uid string) *brokerv1beta1.Trigger {
...
+	checkPanic(t, func() { hashedTruncate("abcde", 4) })
+
+	// panic when string is shorter than max
+	checkPanic(t, func() { hashedTruncate("", 31) })
+
+	// no panic when max == 32
+	hashedTruncate("", 32)
+}
+
+func genString(p string, l int) string {
```
Harwayne, line 243: Can we use [`strings.Repeat`](https://golang.org/pkg/strings/#Repeat) instead?

pkg/reconciler/broker/resources/names.go
grantr: Hashing the full string ensures that the pubsub resource names are unique for all kubernetes resource names. Just truncating means that two long broker names with the same first 200 characters map to the same topic name.

liu-cong: I think we also have the uid of the broker as the topic name. Is that not enough?

pkg/reconciler/broker/resources/names_test.go
grantr: Good idea; didn't know about that one.

grantr:
RFAL @Harwayne @liu-cong

pkg/reconciler/broker/resources/names.go
grantr: That's a good point :) I'll try to rework this a bit.

/hold

grantr:
Replacing this with #965 which eliminates the hash from the truncated string.

Issue #862, opened 2020-04-15
BrokerCell reconciler: static resources


grantr:
**Problem**
Reconcile BrokerCell static resources like deployments, ingress service, etc for #852

**Time Estimate (optional):** 
3

grantr:
See #880. The BrokerCell controller replaces those yamls (and creates identical versions of the resources).

PR #943, opened 2020-04-24
Add brokercell controller and reconciler skeleton

grantr:
No behavior yet, just the generated stubs and a controller test.

Related to #862.


grantr:
100% coverage! Easy when the reconciler doesn't actually do anything :)

PR #944, opened 2020-04-25
Use wire to inject IAMPolicyManager

ian-mi:
This allows the global default IAMPolicyManager to be removed as dependency injection is used to
build controllers using a single shared IAMPolicyManager instance. In order to interface with informer injection, injection must happen in two stages. First wire injects non-informer dependencies, returning injection.ControllerConstructors. Informer injection then uses those constructors to create the controllers. Ideally informer injection would be integrated with wire so that controllers could be injected directly, but this would be a more significant undertaking.

## Proposed Changes

- Adopt wire as an injection framework for controllers. This allows controllers to more easily be constructed with shared dependencies and can help to remove constructor boilerplate.


ian-mi:
I'd like to add support for wire at some point to knative/pkg. I don't think that ctx based injection is a viable alternative in the meantime. Adding and extracting dependencies to ctx is error prone and not checked at compile-time and adds a lot of toil compared with using a dependency injection package.

Harwayne:
Now that we use wire, should this be rebased and submitted?

Harwayne:
> Now that we use wire, should this be rebased and submitted?

I have tried to do so in #1241.

PR #948, opened 2020-04-27
Fix fanout and ingress due to duplicated metric views

liu-cong:
Fixes #

## Proposed Changes

- remove mem stats from main since they are already reported in observability.SetDynamicConfig
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/cmd/broker/fanout/main.go
+++ b/cmd/broker/fanout/main.go
@@ -52,6 +52,7 @@ var (
 )
 
 type envConfig struct {
+	PodName                string        `envconfig:"POD_NAME" required:"true"`
 	ProjectID              string        `envconfig:"PROJECT_ID"`
 	TargetsConfigPath      string        `envconfig:"TARGETS_CONFIG_PATH" default:"/var/run/cloud-run-events/broker/targets"`
 	HandlerConcurrency     int           `envconfig:"HANDLER_CONCURRENCY"`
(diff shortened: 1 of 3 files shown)


cmd/broker/fanout/main.go
```diff
@@ -54,7 +52,6 @@ var (
 )
 
 type envConfig struct {
-	PodName                string        `envconfig:"POD_NAME" required:"true"`
```
liu-cong, line 57: This is not used

Harwayne:
> they are already reported in observability.SetDynamicConfig

Where is this called?

liu-cong:
> > they are already reported in observability.SetDynamicConfig
> 
> Where is this called?

This is called in the line way below: https://github.com/google/knative-gcp/blob/afaa7f8a909e64511a215c53643f63eafaec48d6/cmd/broker/fanout/main.go#L97

cmd/broker/fanout/main.go
ian-mi: We may want access to POD_NAME to the future. Maybe leave this out since this change is orthogonal?

liu-cong: I prefer to add things when we need them, otherwise they just add to the noises. (I can send a separate PR to remove this if you prefer that way)

liu-cong: Or if there will be a need for PodName soon, then I will leave it as is :)

ian-mi: Normally I would agree, but I believe it would be useful to tag metrics with pod name as discusses on the broker ingress PR and I don't think it makes sense to remove this if we will need to immediately add it back.

liu-cong: Added podName back as it may be needed soon.

PR #949, opened 2020-04-27
Add broker ingress metrics

### LINKED ISSUE #840: Add metrics to broker ingress
**Problem**
Add metrics; at least test it e2e with SD manually;
May require extra work to write a e2e test, which can be part of the broker e2e test.

**Exit Criteria**

**Time Estimate (optional):** 
2

liu-cong:
Fixes #840 

## Proposed Changes

- Add broker metrics: https://cloud.google.com/monitoring/api/metrics_knative. Verified in Stackdriver.
-
-

**Release Note**

```release-note
Add broker metrics.
```

**Docs**


pkg/broker/ingress/stats_reporter.go
```diff
@@ -0,0 +1,153 @@
...
+		"The time spent dispatching an event to the decouple topic",
+		stats.UnitMilliseconds,
+	)
+
+	// Create the tag keys that will be used to add tags to our measurements.
+	// Tag keys must conform to the restrictions described in
+	// go.opencensus.io/tag/validate.go. Currently those restrictions are:
+	// - length between 1 and 255 inclusive
+	// - characters are printable US-ASCII
+	namespaceKey         = tag.MustNewKey(metricskey.LabelNamespaceName)
```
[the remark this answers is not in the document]
liu-cong: Yes I am able to see those metrics, and the pod_name and container_name are of course empty.

pkg/broker/ingress/stats_reporter.go
```diff
@@ -0,0 +1,153 @@
...
+func register() {
+	tagKeys := []tag.Key{
+		namespaceKey,
+		brokerKey,
+		eventTypeKey,
+		responseCodeKey,
+		responseCodeClassKey,
+	}
+
+	// Create view to see our measurements.
```
ian-mi, line 104: It is not necessary to use separate measures for event count and latency, these should be different aggregations of the same mesasure.

pkg/broker/ingress/stats_reporter.go
```diff
@@ -0,0 +1,153 @@
...
+type StatsReporter interface {
+	ReportEventCount(args *ReportArgs, responseCode int) error
+	ReportEventDispatchTime(args *ReportArgs, responseCode int, d time.Duration) error
+}
+
+var _ StatsReporter = (*reporter)(nil)
+var emptyContext = context.Background()
+
+// Reporter holds cached metric objects to report ingress metrics.
+type reporter struct{}
```
ian-mi, line 88: I'm not sure we gain much by defining this reporter type vs. simply adding an event measure to handler and using the opencensus APIs directly w/ helper function if needed. This type being empty is also a clue the functionality could be expressed just as easily with helper functions or inlined. In my opinion this is an area where less abstraction would make the code easier to understand and modify.

liu-cong: Sure I will revisit this. I simply copied the stats_reporter code from eventing without thinking too much about it.

pkg/broker/ingress/stats_reporter.go
liu-cong: I am not quite familiar with opencensus. But shouldn't event_count and latency have different type (int vs. float) and units? How can I have a "generic" measure for both?

pkg/broker/ingress/handler.go
```diff
@@ -167,3 +177,17 @@ func (h *handler) toEvent(request *nethttp.Request) (event *cev2.Event, msg stri
 	}
 	return event, "", nethttp.StatusOK
 }
+
+func (h *handler) reportMetrics(ns, broker string, event *cev2.Event, start time.Time, statusCode int) {
+	eventType := "Unknown"
+	if event != nil {
```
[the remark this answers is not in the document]
liu-cong: The event can be nil if the request cannot be converted to an event.

pkg/broker/ingress/stats_reporter.go
liu-cong: I don't feel strongly about this but I prefer having the `StatsReporter` interface and the `reporter` type which implements the interface. This allows us to switch to other implementations easily, e.g.,
* if we want to switch metrics to something else (OpenTelemetry in the future?)
* use a different implementation in tests (although I don't see a need for this yet).

Let me know your thoughts.

ian-mi: If this was a general purpose interface then I would agree but since this interface is specific to this particular metric it doesn't actually make switching implementations any easier. So I think the relevant question is whether the interface that the stats reporter provides is more useful than the opencensus / otel APIs and I am not convinced that they are. The metrics APIs should also already be designed to be testable out of the box, see go.opencensus.io/metric/test.

liu-cong: Changed this to a helper function, please take a look!

pkg/broker/ingress/stats_reporter.go
ian-mi: Check out the example on opencensus.io: https://opencensus.io/stats/view/

pkg/broker/ingress/stats_reporter.go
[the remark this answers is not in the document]
liu-cong: I added podName and containerName back. Thanks for pointing this out! PTAL.

pkg/broker/ingress/stats_reporter.go
liu-cong: Done.

pkg/broker/ingress/handler.go
```diff
@@ -55,21 +61,21 @@ type HttpMessageReceiver interface {
...
-// TODO(liu-cong) add metrics
-// TODO(liu-cong) add tracing
 // TODO(liu-cong) support event TTL
 type handler struct {
 	// httpReceiver is an HTTP server to receive events.
 	httpReceiver HttpMessageReceiver
 	// decouple is the client to send events to a decouple sink.
 	decouple DecoupleSink
 	logger   *zap.Logger
+	ctx      context.Context
```
ian-mi, line 71: Context should always use the current request context and should not be stored in a struct. I recommend applying the tag mutators to the incoming request context. This has the advantage preserving any incoming tags.

pkg/broker/ingress/handler.go
```diff
@@ -167,3 +175,16 @@ func (h *handler) toEvent(request *nethttp.Request) (event *cev2.Event, msg stri
 	}
 	return event, "", nethttp.StatusOK
 }
+
+func (h *handler) reportMetrics(ns, broker string, event *cev2.Event, start time.Time, statusCode int) {
+	eventType := "Unknown"
+	if event != nil {
+		eventType = event.Type()
+	}
+	ctxWithTag, err := generateTag(h.ctx, ns, broker, eventType, statusCode)
```
ian-mi, line 184: I would apply most tag mutators at the beginning of each request and simply use the request ctx + additional tags to report metrics. I would also inline the generateTag helper, since this helper has too many string type arguments and I think directly manipulating the tags is a little more readable even if it is more verbose.

pkg/broker/ingress/stats_reporter_test.go
```diff
@@ -0,0 +1,69 @@
...
+	"context"
+	"net/http"
+	"testing"
+	"time"
+
+	"knative.dev/pkg/metrics/metricskey"
+	"knative.dev/pkg/metrics/metricstest"
+)
+
+func TestStatsReporter(t *testing.T) {
```
ian-mi, line 29: I think this behavior is sufficiently tested in the handler test.

pkg/broker/ingress/stats_reporter.go
```diff
@@ -0,0 +1,134 @@
...
+
+type reportArgs struct {
+	namespace    string
+	broker       string
+	eventType    string
+	responseCode int
+}
+
+func init() {
+	register()
```
ian-mi, line 70: Failure to register metrics should be fatal

pkg/broker/ingress/stats_reporter.go
```diff
@@ -0,0 +1,134 @@
...
+			Aggregation: view.Distribution(metrics.Buckets125(1, 10000)...), // 1, 2, 5, 10, 20, 50, 100, 500, 1000, 5000, 10000
+			TagKeys:     tagKeys,
+		},
+	)
+	if err != nil {
+		log.Printf("failed to register opencensus views, %s", err)
+	}
+}
+
+func reportEventDispatchTime(ctxWithTag context.Context, d time.Duration) {
```
ian-mi, line 106: I might inline this function as well.

pkg/broker/ingress/handler.go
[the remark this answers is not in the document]
ian-mi: I'm not sure we should include invalid events in the event count and latency metrics.

pkg/broker/ingress/stats_reporter_test.go
liu-cong: We need to test the latency conversion (to milliseconds) in these tests.

pkg/broker/ingress/handler.go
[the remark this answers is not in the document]
liu-cong: I am open to suggestions here. I believe we should somehow surface this info to the user, I am not sure if using this metric is the best approach though.

pkg/broker/ingress/stats_reporter.go
liu-cong: Is error better? I think we should still allow the application to run.

pkg/broker/ingress/handler.go
liu-cong: The ctx is used to store the common pod and container name tags. Brought back StatsReporter to hold the pod and container names.

pkg/broker/ingress/handler.go
liu-cong: Consolidated to a single function in StatsReporter, which also takes in the request context.

pkg/broker/ingress/handler.go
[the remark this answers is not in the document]
ian-mi: Perhaps we should also export HTTP metrics?

pkg/broker/ingress/stats_reporter.go
ian-mi: This should not happen unless views are misconfigured. In this case it's better for the issue to be fatal so it is caught right away.

liu-cong: SGTM.

pkg/broker/ingress/handler.go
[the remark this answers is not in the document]
liu-cong: I don't know a good answer on top of my head. Let's not report it for now, which is consistent with existing ChannelBasedBroker.

Opened an issue to come up with a solution for this case:
https://github.com/google/knative-gcp/issues/982

pkg/broker/ingress/stats_reporter.go
```diff
@@ -0,0 +1,136 @@
...
+}
+
+// StatsReporter reports ingress metrics.
+type StatsReporter struct {
+	podName       string
+	containerName string
+}
+
+func (r *StatsReporter) reportEventDispatchTime(ctx context.Context, args reportArgs, d time.Duration) error {
+	tag, err := tag.New(
```
ian-mi, line 120: If the tags don't need to be stored in ctx, just use RecordWithTags instead.

pkg/broker/ingress/handler.go
```diff
@@ -120,29 +126,29 @@ func (h *handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
...
 		response.WriteHeader(statusCode)
 		response.Write([]byte(msg))
 		return
 	}
 
+	event.SetExtension(EventArrivalTime, cev2.Timestamp{Time: time.Now()})
+
 	ctx, cancel := context.WithTimeout(request.Context(), decoupleSinkTimeout)
 	defer cancel()
+	defer func() { h.reportMetrics(request.Context(), ns, broker, event, statusCode, time.Now()) }()
```
ian-mi, line 140: Start time will be incorrect if it is calculated inside the defer. Instead make it an argument to the defer func.

liu-cong: This is indeed a regression. Thanks!

pkg/broker/ingress/stats_reporter.go
liu-cong: This is using `"knative.dev/pkg/metrics"`, which doesn't have `RecordWithTags`. `"knative.dev/pkg/metrics"` manages the metric config.

PR #950, opened 2020-04-27
Include cluster name in k8s Service Account name for Workload Identity

### LINKED ISSUE #715: Include cluster name in k8s Service Account name for Workload Identity
**Describe the bug**
When enable Workload Identity, reconciler will create a k8s Service Account, the k8s SA's name is the same as corresponding Google Cloud Service Account. 
https://github.com/google/knative-gcp/blob/45d9dbd44358730aa34474225fa2969c12ff1d43/pkg/reconciler/identity/resources/service_account.go#L31

More details about reconciler logic for workload identity: https://github.com/google/knative-gcp/blob/master/docs/install/pubsub-service-account.md

It is better to make k8s SA use name like googleServiceAccount + cluster name.
We can use `cluster, _ := metadata.InstanceAttributeValue("cluster-name")` to get cluster name, and clusterName is also an optional value in `ObjectMeta`.

 **Expected behavior**
Include cluster name in k8s Service Account name for Workload Identity

### LINKED ISSUE #735: Default metadata.clusterName for all cloud run resources on GKE
**Problem**
When we create GCP resources like PubSub Topics for objects in the cluster, it would be useful to use the cluster name as part of the GCP resource name. This helps with name uniqueness (technically Kubernetes object UIDs aren't unique across clusters) and helps users more easily find the resources associated with a cluster or object.

As @grac3gao points out in #715, the GKE metadata contains the cluster name and that's accessible from the pod. ObjectMeta also includes a `clusterName` field. We can use these features to automatically set cluster name for any object in a defaulting webhook, allowing controllers to use it for resource names. Both the webhook and the controller must consider this optional so they still work outside GKE. 

Contributor, User

**Exit Criteria**
A controller can create a GCP resource with a name or label containing the cluster name.

**Time Estimate (optional):** 
3

grac3gao:
Fixes #715 
Fixes #735 

## Proposed Changes

- Use an annotation `cluster-name` to include metadata.clusterName, and leave it as empty if the user is not running in GKE
- Rename the KSA for the Workload Identity
-

**Release Note**

```release-note

```

**Docs**


grac3gao:
ready to review

pkg/apis/duck/v1alpha1/annotations.go
```diff
@@ -161,3 +169,25 @@ func validateAnnotationNotExists(annotations map[string]string, annotation strin
...
+			return errs.Also(apis.ErrGeneric(
+				fmt.Sprintf(`not able to get cluster name with error: %s, please provide it by adding annotation "cluster-name=$CLUSTER_NAME"`, err.Error()),
+				fmt.Sprintf("metadata.annotations[%s]", ClusterNameAnnotation)))
+		}
+	}
+	return errs
+}
+
+func SetClusterNameAnnotation(obj *metav1.ObjectMeta, client metadataClient.Client) {
+	if client.OnGCE() {
```
grantr, line 187: If the annotation is already present, this method does nothing. Similarly, in the Validate method above, if the annotation is present we always return success. I think this is the right behavior.

Instead of checking for OnGCE first, it seems like we should check for the annotation first. Then we can avoid calling the metadata service most of the time.

This is optional and can be ignored or done in a followup.

pkg/apis/duck/v1alpha1/annotations_test.go
```diff
@@ -208,3 +210,68 @@ func TestValidateAutoscalingAnnotations(t *testing.T) {
...
+				t.Errorf("Unexpected differences (-want +got): %v", diff)
+			}
+		})
+	}
+}
+
+func TestValidateClusterNameAnnotation(t *testing.T) {
+	testCases := map[string]struct {
+		objMeta *v1.ObjectMeta
+		error   bool
```
grantr, line 253: Is there any way to test an error case for this method? Does the metadata testing client support mocking errors?

pkg/apis/events/v1alpha1/cloudauditlogssource_defaults_test.go
```diff
@@ -31,8 +34,19 @@ func TestCloudAuditLogsSource_SetDefaults(t *testing.T) {
 		expected *CloudAuditLogsSource
 	}{
 		"missing defaults": {
-			orig: &CloudAuditLogsSource{},
+			orig: &CloudAuditLogsSource{
```
grantr, line 37: If I understand the intent of this test, this change should be reverted so that the original is missing the default annotation. This applies to all similar tests below.

pkg/gclient/metadata/testing/client.go
```diff
@@ -0,0 +1,51 @@
...
+var (
+	clusterNameAttr = "cluster-name"
+	FakeClusterName = "fake-cluster-name"
+	FakeProjectID   = "fake-project-id"
+)
+
+type testMetadataClient struct {
+}
+
+func NewTestClient() metadata.Client {
```
grantr, line 32: Ideally this function could take some error mocking parameter(s) like https://github.com/google/knative-gcp/blob/f780beb7298acadbe75785fa2b49795a4f4066df/pkg/gclient/scheduler/testing/client.go#L49-L56
Then it would be possible to test error cases in `annotations_test.go`. That can be a followup if you prefer.

pkg/reconciler/identity/resources/service_account.go
```diff
@@ -17,29 +17,30 @@ limitations under the License.
...
 const (
-	workloadIdentityKey = "iam.gke.io/gcp-service-account"
+	WorkloadIdentityKey = "iam.gke.io/gcp-service-account"
 )
 
 // GenerateServiceAccountName generates a k8s ServiceAccount name according to GCP ServiceAccount
-func GenerateServiceAccountName(gServiceAccount string) string {
-	return strings.Split(gServiceAccount, "@")[0]
+func GenerateServiceAccountName(gServiceAccount, clusterName string) string {
+	return fmt.Sprintf("%s-%s", strings.Split(gServiceAccount, "@")[0], clusterName)
```
grantr, line 33: What happens to existing resources when this ServiceAccount name changes?

pkg/reconciler/messaging/channel/channel.go
```diff
@@ -168,7 +170,7 @@ func (r *Reconciler) syncSubscribers(ctx context.Context, channel *v1alpha1.Chan
 			ServiceAccount: channel.Spec.GoogleServiceAccount,
 			Secret:         channel.Spec.Secret,
 			Labels:         resources.GetPullSubscriptionLabels(controllerAgentName, channel.Name, genName, string(channel.UID)),
-			Annotations:    resources.GetPullSubscriptionAnnotations(channel.Name),
+			Annotations:    resources.GetPullSubscriptionAnnotations(channel.Name, clusterName),
```
grantr, line 173: Why is it necessary for the channel to set the cluster name? Wouldn't we expect this to happen in the webhook? Is it so that if the channel has a `cluster-name` value different from the default, its topics and pull subscriptions get the same value?

If so, seems like other resources that also create PullSubscriptions like CloudPubSubSource should have the same behavior. That can be added in a follow-up PR.

pkg/utils/metadata.go
```diff
@@ -17,19 +17,36 @@ limitations under the License.
 package utils
 
 import (
-	"cloud.google.com/go/compute/metadata"
+	metadataClient "github.com/google/knative-gcp/pkg/gclient/metadata"
+)
+
+var (
+	clusterNameAttr = "cluster-name"
```
grantr, line 24: Can this be a const instead of a var?

pkg/utils/metadata_test.go
```diff
@@ -0,0 +1,75 @@
...
+
+import (
+	"testing"
+
+	"github.com/google/go-cmp/cmp"
+
+	testingMetadataClient "github.com/google/knative-gcp/pkg/gclient/metadata/testing"
+)
+
+func TestProjectID(t *testing.T) {
```
grantr, line 27: These tests could also benefit from a way to mock errors from the metadata client. Currently there's no coverage of the error cases.

pkg/apis/duck/v1alpha1/annotations.go
```diff
@@ -161,3 +169,25 @@ func validateAnnotationNotExists(annotations map[string]string, annotation strin
 	}
 	return errs
 }
+
+// ValidateClusterNameAnnotation validates the cluster-name annotation.
+func ValidateClusterNameAnnotation(annotations map[string]string, errs *apis.FieldError, client metadataClient.Client) *apis.FieldError {
+	if client.OnGCE() {
```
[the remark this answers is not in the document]
grac3gao: I expected the KSA name for WI would include the cluster name, so it can avoid potential security issue like multiple clusters are using the same KSA name and Namespace name. (as all KSAs that share a name, Namespace name, and Workload Identity pool resolve to the same member name, and therefore share access to GSAs.)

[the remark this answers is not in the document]
grac3gao: Besides that, I don't find other bad things...

pkg/apis/events/v1alpha1/cloudauditlogssource_defaults_test.go
grac3gao: Yes, I also think the intent of the test is this. However, we directly use SetDefaults() here, I didn't find a way to put a testing client for metadate client in it. Thus, if I didn't add annotation here, in my local, it will show no cluster name annotation, but in prow, it will show cluster name annotation for the prow job project...

pkg/reconciler/identity/resources/service_account.go
grac3gao: Existing resources without the clusterName annotation will have error in Identity Reconciler:
```
 if clusterName == "" {
		err := fmt.Errorf(`not able to get cluster name, please provide it by adding annotation "cluster-name=$CLUSTER_NAME" to source`)
		status.MarkWorkloadIdentityFailed(identifiable.ConditionSet(), workloadIdentityFailed, err.Error())
		return nil, fmt.Errorf("failed to get cluster name: %w", err)
	}
```
but thinking together with a previous comment about leaving the clusterName annotation blank if user disable metadata, maybe I can allow the defaulting webhook to leave the annotation blank even if it is onGKE, and use the util.getClusterName to get clusterName again in Identity Reconciler, so that existing resources will get they clusterName even the annotation is not populated by the defaulting webhook.

grac3gao: And those users who disabled metadata access still need to provide the cluster name ....

pkg/reconciler/messaging/channel/channel.go
grac3gao: Yes, for WI, source and channel call Identity Reconciler function in there own reconciler, and create the KSA. (topic and pullsubscription reconciler only call Identity Reconciler function if the topic/pullsubscription obj doesn't have owner, which means they are created along without any up-level resources), and we pass the cluster annotation all the way down to publisher/receive-adaptor pod to make the KSA name to put into pod's serviceAccountName. This annotations setting here ensures that the KSA which reconcilers make has the same name as the topic/pullsubscription make.

grac3gao: For sources, they use the different pullsubscription/topic resources from channel. And pullsubscription resource for source already include this annotation arg.(I guess it is used for the autoscaling annotation before.)

pkg/apis/events/v1alpha1/cloudpubsubsource_validation.go
```diff
@@ -40,6 +42,7 @@ const (
 
 func (current *CloudPubSubSource) Validate(ctx context.Context) *apis.FieldError {
 	errs := current.Spec.Validate(ctx).ViaField("spec")
+	errs = duckv1alpha1.ValidateClusterNameAnnotation(current.Annotations, errs, metadataClient.NewDefaultMetadataClient())
```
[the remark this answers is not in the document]
grac3gao: They will be updated by the user.

Together with comment https://github.com/google/knative-gcp/pull/950#discussion_r421149115
It looks like that I can remove clusterName validation, (which means allow the clusterName to be empty in defaulting webhook, even if it is on GKE). I feel that not every GCP resource names need to include clusterName like WI, blocking this from the defaulting webhook may not that user-friendly. Let the Identity reconciler check clusterName annotation again when we want it to generate KSA name(we already do that in the Identity reconciler), then print out error if user need to update annotation.

In the meanwhile, adding immutability check, once clusterName annotation is set, it could not be changed.

Do you think that is proper ?

pkg/reconciler/identity/resources/service_account.go
grac3gao: We set the spec.googleServiceAccount is immutable, so before this change, the ServiceAccount name will not change. Yes, we need a immutable check also for the clusterName annotation.

pkg/apis/duck/v1alpha1/annotations.go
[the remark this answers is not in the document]
grantr: Another option is to avoid deleting the binding if no cluster name is present. This eliminates the problem of other clusters suddenly losing their credentials. However, it opens a security hole: a service account in a different cluster with a matching namespace and name will get the permissions of the bound GSA, even if no identity in the cluster ever had permissions to create that binding.

pkg/apis/events/v1alpha1/cloudpubsubsource_validation.go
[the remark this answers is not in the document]
grantr: We talked about this over video. It seems like a better user experience to remove the validation and replace it with an error message about the missing annotation in the object status. This makes the upgrade easier since missing cluster names can be corrected gradually.

grac3gao:
ready to review again

pkg/reconciler/broker/broker_test.go
```diff
@@ -160,7 +161,7 @@ func TestAllCases(t *testing.T) {
 	table.Test(t, MakeFactory(func(ctx context.Context, listers *Listers, cmw configmap.Watcher, testData map[string]interface{}) controller.Reconciler {
 		// Insert pubsub client for PostConditions and create fixtures
 		psclient, close := TestPubsubClient(ctx, testProject)
-		t.Cleanup(close)
+		t.Clenup(close)
```
grantr, line 164: This should be `t.Cleanup`

pkg/reconciler/identity/reconciler_test.go
```diff
@@ -74,7 +76,13 @@ func TestCreates(t *testing.T) {
 		expectedServiceAccount *corev1.ServiceAccount
 		wantCreates            []runtime.Object
 		wantErrCode            codes.Code
+		wantErr                error
+		addClusterName         bool
 	}{{
+		name:           "k8s service account doesn't exist, failed to get cluster name",
+		addClusterName: false,
+		wantErr:        fmt.Errorf(`failed to get cluster name: unable to get cluster name, please provide it by adding annotation "%s=$CLUSTER_NAME" to source`, duckv1alpha1.ClusterNameAnnotation),
```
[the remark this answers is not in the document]
Harwayne: If this was not user facing, then I would probably do it this way. But because this is user facing, I can see the desire for it being painful to change it. I'm ok either way, but think I prefer hard coding to make it harder to accidentally change a user visible string.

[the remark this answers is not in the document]
grac3gao: @grantr I removed this test, since it will show different error messages when running locally/running in prow. It is the same issue as the defaults test. If we didn't add clusterName annotation here, setDefaults() will set that annotation, and it can't take the testmedataClient in it. Thus, in prow, it can get the cluster name here...

pkg/apis/events/v1alpha1/cloudauditlogssource_defaults_test.go
grac3gao: I changed the name of this test. From "missing defaults" to "missing defaults, except cluster name annotation", similar for every resources.

pkg/reconciler/identity/reconciler_test.go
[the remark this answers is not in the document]
grantr: Ok thanks @Harwayne. Seems like this is affected by the same limitation mentioned above. @grac3gao please mention it in the issue from https://github.com/google/knative-gcp/pull/950#discussion_r422408427.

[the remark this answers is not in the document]
grac3gao: Just added that. Issue #1037

grantr:
Fixes #735

PR #959, opened 2020-04-28
Channel v1beta1 Fixes

Harwayne:
Should have been included in #615.

## Proposed Changes

- Inline `spec.SubscribableSpec`, as required by the duck v1beta1 Subscribable contract.
  - Alter the OpenAPIV3Schema to match.
- Add the messaging.knative.dev/subscribable annotation to Channel. This allows the correct creation of Subscriptions on v1beta1 Channels.
- Add v1beta1 `Channel`s to the E2E tests.

**Release Note**

```release-note
v1beta1 Channels remove a layer of nesting:
`spec.subscribable.subscribers` -> `spec.subscribers`

`Subscription`s can be made against v1beta1 `Channel`s.
```


pkg/apis/messaging/internal/constants.go
```diff
@@ -0,0 +1,20 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package internal
+
+// StoredChannelVersion is the version of the Channel that is stored in the API server.
+const StoredChannelVersion = "v1alpha1"
```
grantr, line 20: Just to confirm, this is the stored version of the PubSub Channel CRD, right? Not the stored version of the Channel CRD from Knative Eventing?

How will we ensure that when the stored version changes, this constant is also updated? Do we need a comment in the CRD yaml?

Harwayne: Yes, this is the version of the PubSub channel stored in the API server. Added a comment to the YAML.

Harwayne:
/hold

Pulling at a thread caused a lot of changes. Holding until I have updated the title and description.

Harwayne:
/retest

I swear this is working on my local machine...

Harwayne:
/hold cancel

This is ready for review.

pkg/apis/messaging/v1beta1/channel_defaults.go
```diff
@@ -40,10 +42,26 @@ func defaultSecretSelector() *corev1.SecretKeySelector {
...
+	// We need to set this to the _stored_ version of the Channel. If we set to anything other than
+	// the stored version, then when reading the stored version, conversion won't be called so
+	// nothing will set it to the stored version.
+	// Note that if a user sends a bad version of this annotation (e.g. sets it to v1beta1), then we
+	// won't overwrite their bad input. This is because the webhook:
+	// 1. Reads the stored version.
+	// 2. Converts to the desired version.
+	// 3. Defaults the desired version.
+	// So we don't know if the user or the converter put the value here, therefore we are forced to
+	// assume it was the converter and shouldn't change it.
```
yolocs, line 54: Do you mean the conversion happens before this function is invoked so that we have to default to stored version?

Harwayne: Yes, conversion occurs before this function is invoked. 

This actually isn't needed in the v1beta1 code path. It is only required in the storage version path (v1alpha1). It is only needed for the storage version, because we normally set this value during conversion, but if the storage version is requested, then no conversion occurs, so nothing sets it. I chose to add it here too so that we would not forget it when switching the storage version.

Eventing has the same logic, although only in the non-storage path:

https://github.com/knative/eventing/blob/d76621b63ac486411fdac4b9da07fd3a7a8e45b1/pkg/apis/messaging/v1beta1/in_memory_channel_defaults.go#L26-L36

Harwayne:
/unhold

I think I answered the question. If you want the comment expanded on, I'll do it in a follow up.

pkg/apis/messaging/v1beta1/channel_defaults.go
yolocs: Thanks for explaining!

Issue #958, opened 2020-04-28
Flaky unit test in SyncPool


liu-cong:
**Describe the bug**
https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/948/pull-google-knative-gcp-unit-tests/1255167042848296962

```
--- FAIL: TestSyncPool (0.00s)
    --- PASS: TestSyncPool/StartSyncPool_returns_error (0.00s)
    --- FAIL: TestSyncPool/Work_done_with_StartSyncPool (0.00s)
        pool_test.go:59: SyncOnce was not called twice
```
 **Expected behavior**

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative-GCP release version**

 **Additional context**

It seems to be rare. I have only seen this once so far.

cc @yolocs

PR #961, opened 2020-04-28
Add a tiny delay in SyncPool test to avoid flakiness

yolocs:
Fixes https://github.com/google/knative-gcp/issues/958

## Proposed Changes

-
-
-

**Release Note**

```release-note
NONE
```

**Docs**

### DIFF
--- a/pkg/broker/handler/pool/pool_test.go
+++ b/pkg/broker/handler/pool/pool_test.go
@@ -19,7 +19,6 @@ package pool
 import (
 	"context"
 	"fmt"
-	"sync/atomic"
 	"testing"
 	"time"
 
@@ -33,7 +32,8 @@ func TestSyncPool(t *testing.T) {
 
 	t.Run("StartSyncPool returns error", func(t *testing.T) {
 		syncPool := &fakeSyncPool{
-			returnErr: true,
+			returnErr:  true,
+			syncCalled: make(chan struct{}, 1),
 		}
 		_, gotErr := StartSyncPool(ctx, syncPool, make(chan struct{}))
 		if gotErr == nil {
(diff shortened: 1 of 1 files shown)


pkg/broker/handler/pool/pool_test.go
```diff
@@ -55,6 +56,7 @@ func TestSyncPool(t *testing.T) {
 			t.Errorf("SyncOnce was called more than once")
 		}
 		ch <- struct{}{}
+		time.Sleep(500 * time.Millisecond)
```
liu-cong, line 59: I haven't looked into the root cause of this yet. Is it easy to make the test 100% deterministic?

yolocs: After receiving a signal, the fake pool will increase its count asynchronously. And we were checking right away previously, which caused the flakiness. I don't see an obvious way to make it deterministic other than longer waiting.

liu-cong: One way to do this is to have a channel in the fakePool, and send a signal to the channel each time `SyncOnce` is called. The test can wait for the signals in the channel.

Normally unit test should be deterministic. Otherwise it may bite us again.

yolocs: I agree in principal. But in this case it's also possible our code has a bug and `SyncOnce` was never called and we would wait infinitely. To help we have to add a timeout, which is similar to what it's doing now. That's why I think it's not necessary to introduce such a mechanism.

ian-mi: One solution could be to use a buffered channel instead of an atomic counter in fakeSyncPool so that the test could block waiting for each sync instead of sleeping. Yes ideally this would also introduce a test timeout but I think this would be preferable to sleeping.

liu-cong: Fine with me. We can implement that if we see this again.

yolocs: I made the change you suggested to save some time

pkg/broker/handler/pool/pool_test.go
```diff
@@ -46,35 +46,36 @@ func TestSyncPool(t *testing.T) {
...
+	returnErr  bool
+	syncCalled chan struct{}
 }
 
-func (p *fakeSyncPool) SyncOnce(ctx context.Context) error {
-	if p.syncCount.Load() == nil {
-		p.syncCount.Store(0)
+func (p *fakeSyncPool) verifySyncOnceCalled(t *testing.T) {
+	select {
+	case <-time.After(500 * time.Millisecond):
```
ian-mi, line 70: Ah, I was thinking of adding a timeout to ctx but this works as well.

PR #965, opened 2020-04-28
Truncate pubsub names if greater than 255 chars

### LINKED ISSUE #831: #1 Broker reconciler doesn't validate length of pubsub resource names

grantr:
Fixes #831
Supersedes #942

Since we append the UID to the resource name, there's no need to use a hashed truncate to guarantee uniqueness of the truncated name as in `kmeta.ChildName`.

## Proposed Changes

- Truncate pubsub resource names deterministically if they would be greater than 255 characters

### DIFF
--- a/pkg/reconciler/broker/resources/names_test.go
+++ b/pkg/reconciler/broker/resources/names_test.go
@@ -18,6 +18,7 @@ package resources
 
 import (
 	"fmt"
+	"strings"
 	"testing"
 
 	"github.com/google/go-cmp/cmp"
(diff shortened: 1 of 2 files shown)


pkg/reconciler/broker/resources/names_test.go
```diff
@@ -46,12 +49,25 @@ func TestGenerateDecouplingTopicName(t *testing.T) {
 		n:    "more-dashes",
 		uid:  testUID,
 		want: fmt.Sprintf("cre-bkr_with-dashes_more-dashes_%s", testUID),
+	}, {
+		ns:   maxNamespace,
+		n:    maxName,
+		uid:  testUID,
+		want: fmt.Sprintf("cre-bkr_%s_%s_%s", maxNamespace, strings.Repeat("n", 146), testUID),
```
liu-cong, line 56: nit: define const maxNameLength=146 for better readablility.

grantr:
/hold cancel

@liu-cong RFAL

PR #972, opened 2020-04-29
Use wire to inject ingress Handler

### LINKED ISSUE #968: Simplify Ingress Handler Constructor and Options
**Describe the bug**
The ingress handler and multiTopicDecoupleSink constructors are unnecessarily complicated as they also support both constructing their eventing clients as well as using a client supplied as an option.

 **Expected behavior**
The handler and multiTopicDecoupleQueue should use an eventing client supplied as an argument.

ian-mi:
Fixes #968

## Proposed Changes

- Simplify handler and multiTopicDecoupleSink by removing dependency initialization logic.
- Simplify related opts to contain only the options needed to inject deps.
- Use wire to inject ingress Handler dependencies.


pkg/broker/ingress/options.go
```diff
@@ -20,61 +20,53 @@ import (
...
+	cev2 "github.com/cloudevents/sdk-go/v2"
+	cepubsub "github.com/cloudevents/sdk-go/v2/protocol/pubsub"
+	"github.com/google/wire"
 	"knative.dev/eventing/pkg/kncloudevents"
 )
 
-// HandlerOption is the option to configure ingress handler.
-type HandlerOption func(*handler) error
+// HandlerSet provides a handler with a real HTTPMessageReceiver and pubsub MultiTopicDecoupleSink.
+var HandlerSet wire.ProviderSet = wire.NewSet(
```
liu-cong, line 30: Let's move this to handler.go or create a provider.go?

pkg/broker/ingress/handler.go
```diff
@@ -67,34 +66,16 @@ type handler struct {
 }
 
 // NewHandler creates a new ingress handler.
-func NewHandler(ctx context.Context, options ...HandlerOption) (*handler, error) {
-	h := &handler{
-		logger: logging.FromContext(ctx),
+func NewHandler(ctx context.Context, httpReceiver HttpMessageReceiver, decouple DecoupleSink) *Handler {
```
liu-cong, line 69: With this simplification, I think we can use the `struct providers`: https://github.com/google/wire/blob/master/docs/guide.md#struct-providers

Not sure if it's recommended though.

ian-mi: I think in this case I'd prefer to use a constructor so that the Handler fields can remain hidden.

pkg/broker/ingress/options.go
ian-mi: Moved to handler.go

pkg/broker/ingress/args.go
```diff
@@ -0,0 +1,59 @@
...
+
+type Args struct {
+	IngressPort   int
+	ProjectID     string
+	PodName       string
+	ContainerName string
+}
+
+// NewHTTPMessageReceiver wraps kncloudevents.NewHttpMessageReceiver with type-safe options.
+func NewHTTPMessageReceiver(args Args) *kncloudevents.HttpMessageReceiver {
```
liu-cong, line 36: I prefer only passing relevant args(the port number). Same for the NewPubSubClient below.

pkg/broker/ingress/stats_reporter.go
```diff
@@ -103,10 +103,10 @@ func register() {
 }
 
 // NewStatsReporter creates a new StatsReporter.
-func NewStatsReporter(podName, containerName string) *StatsReporter {
+func NewStatsReporter(args Args) *StatsReporter {
```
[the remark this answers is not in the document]
ian-mi: I agree when the args have unique types but args cannot have the same type for dependency injection to work. It is also not a best practice for a constructor to take multiple arguments of the same type as it becomes too easy to accidentally swap arguments.
There are a couple possible solutions here. The one I've chosen is to create a common ingress args struct as ingress will always construct a handler from the same set of arguments. A more flexible approach might be to add type aliases to constructor arguments such as PodName, ContainerName, IngressPort etc. Either approach adds type safety so I am happy to switch to the latter if you find it preferable.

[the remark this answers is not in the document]
liu-cong: Didn't know the limitation of dependency injection, thanks for letting me know!

I didn't think of type safety when I made the comment but yes it's better to have type safety. 

My main concern was the fact that we expose params that the struct doesn't/shouldn't need. If the type alias solves both the type safety and unnecessary args exposure, I like that :)

Another approach I can think of is to define an `Args` type for each struct, e.g., `StatsReporterArgs`.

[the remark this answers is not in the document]
ian-mi: Sounds good, I actually had used separate structs in an earlier version but I felt it was becoming excessive. I think the ideal approach would use a mix of structs and typedefs on a case by case basis.

Issue #924, opened 2020-04-23
Failed to run e2e-tests.sh with E2E tests related to CloudSchedulerSource


capri-xiyue:
**Describe the bug**
When I ran `E2E_PROJECT_ID=$PROJECT_ID test/e2e-tests.sh`,
All E2E tests related to CloudSchedulerSource failed.
- --- FAIL: TestCloudSchedulerSourceBrokerWithPubSubChannel 
- --- FAIL: TestSmokeCloudSchedulerSourceSetup
- --- FAIL: TestCloudSchedulerSourceWithTargetTestImpl

It looks like these E2E tests failed because of the lack of app engine. We didn't set up the creation of app engine with correct time zone in our https://github.com/google/knative-gcp/blob/master/test/e2e-common.sh.
We need to fix it. Also need to figure out why E2E tests passes in prow.

 **Expected behavior**
All E2E tests pass when running command`E2E_PROJECT_ID=$PROJECT_ID test/e2e-tests.sh` 

 **To Reproduce**
Run `E2E_PROJECT_ID=$PROJECT_ID test/e2e-tests.sh` 
 **Knative-GCP release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

capri-xiyue:
This issue can be combined with https://github.com/google/knative-gcp/issues/766 and https://github.com/google/knative-gcp/issues/738

grantr:
This only happens when running tests on an existing cluster. Prow projects already have app engine setup.

capri-xiyue:
> This only happens when running tests on an existing cluster. Prow projects already have app engine setup.

To clarify,  this happens when we fail to create app engine and use `go test` on an existing cluster or when we run e2e with `test/e2e-tests.sh` directly which will creates a new cluster

capri-xiyue:
Closed this because it got fixed in https://github.com/google/knative-gcp/pull/973

Issue #766, opened 2020-04-03
Modify init_scripts of knative-gcp so that they can run multiple times(Idempotence)


capri-xiyue:
**Problem**
Currently, we have several init_scripts for example (https://github.com/google/knative-gcp/blob/master/hack/init_control_plane_gke.sh and https://github.com/google/knative-gcp/blob/master/hack/init_control_plane.sh). However, usera can only run it one time.  
I just found it's quite inconvenient if something goes wrong in the script but it doesn't support running for multiple rounds.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
The init scripts can run multiple times(Idempotence) 
**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

ericlem:
I'll see if i can get to any of this during upgrade work.  It would be useful there.
/assign

capri-xiyue:
Once we make init_control_plane.sh,init_control_plane_gke.sh, init_data_plane.sh idempotent so they can be run as part of e2e test setup process.

capri-xiyue:
This is related to https://github.com/google/knative-gcp/issues/738

capri-xiyue:
Closed this because it got fixed in https://github.com/google/knative-gcp/pull/973

Issue #738, opened 2020-03-30
Simplify e2e tests setup scripts


chizhg:
**Problem**
In [e2e-common.sh](https://github.com/google/knative-gcp/blob/master/test/e2e-common.sh), there are a lot of `gcloud` commands to configure the authentication mechanism. These are only used when users run the scripts on their local, and there is no way to keep them up-to-date.

**Exit Criteria**
Since it's a one-time setting for running it locally, we should direct users to run [./hack/init_control_plane.sh](https://github.com/google/knative-gcp/blob/master/hack/init_control_plane.sh) and [./hack/init_control_plane_gke.sh](https://github.com/google/knative-gcp/blob/master/hack/init_control_plane.sh) instead of creating a new service account every time they run the scripts. Then all these `gcloud` commands in the scripts can be removed.

/assign

capri-xiyue:
I'm working on https://github.com/google/knative-gcp/issues/766. These two issues are related to each other. @chizhg  I will just assign this issue to myself. Let me know if you already make some progress related to this issue.

capri-xiyue:
Closed this because it got fixed in https://github.com/google/knative-gcp/pull/973

Issue #896, opened 2020-04-20
Refactor control plane init scripts to avoid error message


liu-cong:
**Problem**
`./hack/init_control_plane_gke.sh` or `./hack/init_control_plane.sh` will print an error message if the script is trying to create the `cloud-run-events` GSA if it already exists. Though in our installation docs, we have a note explaining this is a safe error, it can still be scary especially for a first time user.

This is also brought up in PR review: https://github.com/google/knative-gcp/pull/894#discussion_r411508201

Ideally, the script should:
* Check if the GSA exists, if so, skip creating it

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

capri-xiyue:
Close this because it got fixed in https://github.com/google/knative-gcp/pull/973

PR #973, opened 2020-04-29
added e2e for gcp broker which sends and forwards events, refactored …

### LINKED ISSUE #837: E2E test new broker and receives/forwards an event
**Problem**

Add a very basic E2E tests to start up a new gcp broker, send an event, and make sure it is received as a first test at integrating the new broker into E2E tests.

This test should essentially be a copy of test/e2e/BrokerWithPubSubChannelTestImpl with a new
helper function equivalent to createBrokerWithPubSubChannel that can be used to create the
new style broker.  

This will also have to add a step to install yaml files to configure and deploy the control plane and shared data plane components of the new broker.

**Exit Criteria**
A new test that performs the same actions as BrokerWithPubSubChannelTestImpl on
the new gcp broker.

**Time Estimate (optional):** 
5

This will rely on a set of yaml's being produced that can be installed to configure the control and data planes of the new broker.

capri-xiyue:
Fixes #837 ,#924, #766 #738 

## Proposed Changes

- added e2e for gcp broker which sends and forwards events,
- Simplified and refactored e2e tests setup scripts
- Modify init_scripts of knative-gcp so that to achieve idempotence
## Desp:

- ./test/e2e-common.sh are the shared code between ./test/e2e-secret-tests.sh and ./test/e2e-wi-tests.sh
- ./hack/lib.sh are shared among other init_xxx scripts under ./hack repo AND ./test/e2e-common.sh AND./test/e2e-secret-tests.sh AND ./test/e2e-wi-tests.sh
- I deleted most of tear-down code in e2e-xxx-tests.sh because most of the set up code are now idempotent

**Release Note**


capri-xiyue:
There are some duplicate code of E2E tests of GCP Broker and Broker with PubSub Channel. I will create another issue to refactor it.

test/e2e/test_gcp_broker.go
```diff
@@ -0,0 +1,134 @@
...
+	}
+	// Check if resp CloudEvent hits the target Service.
+	if done := jobDone(client, targetName, t); !done {
+		t.Error("resp event didn't hit the target pod")
+		t.Failed()
+	}
+}
+
+func createGCPBroke(t *testing.T, client *lib.Client, targetName string) url.URL {
+	brokerName := helpers.AppendRandomString("gcp")
```
capri-xiyue, line 89: There are some duplicate code of E2E tests of GCP Broker and Broker with PubSub Channel. I will create another issue to refactor it.

capri-xiyue:
@chizhg It's ready for review

liu-cong:
I see this PR fixes many issues. Can we break it down to multiple smaller PRs?

capri-xiyue:
> I see this PR fixes many issues. Can we break it down to multiple smaller PRs?
All the issues are related to each other. That's why I combine them in one PR. Though the PR looks kind of large. But most of them are just script refactoring and variable renaming.

liu-cong:
How about having script refactor as one PR and broker e2e as a second PR?

capri-xiyue:
> How about having script refactor as one PR and broker e2e as a second PR?

The broker e2e also need the script refactor. We need some gcp broker set up script in the e2e set up script.

hack/lib.sh
```diff
@@ -0,0 +1,163 @@
...
+  echo "parameter cluster_location_type used when when enabling workload identity is'${cluster_location_type}'"
+  echo "cluster_location_option used when when enabling workload identity is'${cluster_location_option}'"
+  # Enable API
+  gcloud services enable iamcredentials.googleapis.com
+  # Enable workload identity.
+  echo "Enable Workload Identity"
+  local cluster_name="$(cut -d'_' -f4 <<<"$(kubectl config current-context)")"
+  local cluster_location="$(cut -d'_' -f3 <<<"$(kubectl config current-context)")"
+  gcloud container clusters update ${cluster_name} \
+    --${cluster_location_option}=${cluster_location} \
```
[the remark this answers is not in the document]
capri-xiyue: I believes the https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#enable_on_existing_cluster assumes that user set the correct `computer/zone` and `run/cluster_location` in gcloud config. If user doesn't set it or set it in the wrong cluster_ location or zone and user doesn't specify the `zone` or `region` when they run the `gcloud container clusters update`  command, it will fail.  I encountered some error when I run e2e tests when E2E set up script of test-infra create `regional` cluster instead of `zonal` cluster.   Here is the docs of the command FYI: https://cloud.google.com/sdk/gcloud/reference/container/clusters/update

hack/init_cloud_storage_source.sh
```diff
@@ -0,0 +1,30 @@
...
+# limitations under the License.
+
+# Usage: ./init_cloud_storage_source.sh
+# The current project set in gcloud MUST be the same as where the cluster is running.
+source $(dirname $0)/lib.sh
+
+PUBSUB_SERVICE_ACCOUNT_KEY_TEMP="cre-pubsub.json"
+
+# Download a JSON key for the service account.
+gcloud iam service-accounts keys create ${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP} \
```
grantr, line 24: Does this fail if the file already exists?

hack/init_control_plane_gke.sh
```diff
@@ -16,50 +16,27 @@
...
-
-# Grant permissions to the service account for the control plane to manage native GCP resources.
-gcloud projects add-iam-policy-binding ${PROJECT_ID} --member=serviceAccount:${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com --role roles/pubsub.admin
-gcloud projects add-iam-policy-binding ${PROJECT_ID} --member=serviceAccount:${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com --role roles/storage.admin
-gcloud projects add-iam-policy-binding ${PROJECT_ID} --member=serviceAccount:${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com --role roles/cloudscheduler.admin
-gcloud projects add-iam-policy-binding ${PROJECT_ID} --member=serviceAccount:${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com --role roles/logging.configWriter
-gcloud projects add-iam-policy-binding ${PROJECT_ID} --member=serviceAccount:${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com --role roles/logging.privateLogViewer
-gcloud projects add-iam-policy-binding ${PROJECT_ID} --member=serviceAccount:${GSA_NAME}@${PROJECT_ID}.iam.gserviceaccount.com --role roles/iam.serviceAccountAdmin
+source $(dirname $0)/lib.sh
+readonly CLUSTER_LOCATION_TYPE="zonal"
```
grantr, line 20: Is `CLUSTER_LOCATION_TYPE` really readonly? Line 25 sets it again.

hack/init_data_plane.sh
```diff
@@ -20,32 +20,27 @@
...
-else
-  NAMESPACE="$1"
-  echo "NAMESPACE provided, using ${NAMESPACE}"
-  kubectl create namespace $NAMESPACE
-fi
+source $(dirname $0)/lib.sh
 
-# Create the service account for the data plane
-gcloud iam service-accounts create ${SERVICE_ACCOUNT}
+readonly NAMESPACE="default"
```
grantr, line 25: Is `NAMESPACE` really readonly? Line 31 sets it again.

hack/lib.sh
```diff
@@ -0,0 +1,163 @@
...
+  gcloud projects add-iam-policy-binding ${project_id} \
+    --member=serviceAccount:${control_plane_service_account}@${project_id}.iam.gserviceaccount.com \
+    --role roles/logging.privateLogViewer
+
+}
+
+function init_pubsub_service_account() {
+  local project_id=${1}
+  local pubsub_service_account=${2}
+  echo "parameter project_id used when initiating data plane service account is'${project_id}'"
```
grantr, line 79: nit: `when initiating pubsub service account`

hack/lib.sh
```diff
@@ -0,0 +1,163 @@
...
+
+  echo "parameter project_id used when enabling workload identity is'${project_id}'"
+  echo "parameter control_plane_service_account used when when enabling workload identity is'${control_plane_service_account}'"
+  local cluster_location_option
+  if [ ${cluster_location_type} == ${ZONAL_CLUSTER_LOCATION_TYPE} ]; then
+    cluster_location_option=zone
+  elif [ ${cluster_location_type} == ${REGIONAL_CLUSTER_LOCATION_TYPE} ];then
+    cluster_location_option=region
+  else
+    echo >&2 "Fatal error: cluster_location_type used when when enabling workload identity has to '${ZONAL_CLUSTER_LOCATION_TYPE}' or '${REGIONAL_CLUSTER_LOCATION_TYPE}'"
```
grantr, line 115: nit: `has to` => `must be`

hack/lib.sh
```diff
@@ -0,0 +1,163 @@
...
+    exit 1
+  fi
+  echo "parameter cluster_location_type used when when enabling workload identity is'${cluster_location_type}'"
+  echo "cluster_location_option used when when enabling workload identity is'${cluster_location_option}'"
+  # Enable API
+  gcloud services enable iamcredentials.googleapis.com
+  # Enable workload identity.
+  echo "Enable Workload Identity"
+  local cluster_name="$(cut -d'_' -f4 <<<"$(kubectl config current-context)")"
+  local cluster_location="$(cut -d'_' -f3 <<<"$(kubectl config current-context)")"
```
grantr, line 125: These lines seem to rely on the user setting the kubectl context correctly before running the init script. That should be mentioned in the usage comment alongside `The current project set in gcloud MUST be the same as where the cluster is running.`

hack/lib.sh
```diff
@@ -0,0 +1,163 @@
...
+  while read -r pool_name
+  do
+  gcloud container node-pools update "${pool_name}" \
+    --cluster=${cluster_name} \
+    --${cluster_location_option}=${cluster_location} \
+    --workload-metadata=GKE_METADATA
+  done <<<"${pools}"
+
+  gcloud projects add-iam-policy-binding ${project_id} --member=serviceAccount:${control_plane_service_account}@${project_id}.iam.gserviceaccount.com --role roles/iam.serviceAccountAdmin
+  }
```
grantr, line 142: nit: This line is indented too far

hack/init_cloud_storage_source.sh
```diff
@@ -0,0 +1,30 @@
...
+  --iam-account=${PUBSUB_SERVICE_ACCOUNT}@${PROJECT_ID}.iam.gserviceaccount.com
```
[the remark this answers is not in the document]
capri-xiyue: Yep. PROJECT_ID should also be in hack/lib.sh. I think somehow I just deleted it. I will add it in hack/lib.sh.

hack/init_cloud_storage_source.sh
capri-xiyue: Yes. It will fail. But every time when run the script, we will delete the file in the end with `
rm ${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}`. Therefore, we can run this script multiple times.  Unless user has one file with the same name created by himself,  this failure won't happen. Or do you think we should allow user to pass the  `PUBSUB_SERVICE_ACCOUNT_KEY_TEMP` that they want to use?

hack/init_control_plane_gke.sh
capri-xiyue: This is a bug. I will fix it. This variable should not be readonly

hack/init_data_plane.sh
capri-xiyue: This is a bug. I will fix it. This variable should not be readonly.

hack/init_cloud_storage_source.sh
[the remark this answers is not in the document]
capri-xiyue: I will re-verify all these scripts under ./hack repo to see whether it works when I update the related docs.

hack/lib.sh
```diff
@@ -0,0 +1,163 @@
...
+
+  # Enable APIs.
+  gcloud services enable pubsub.googleapis.com
+  gcloud services enable storage-component.googleapis.com
+  gcloud services enable storage-api.googleapis.com
+  gcloud services enable cloudscheduler.googleapis.com
+  gcloud services enable cloudbuild.googleapis.com
+  gcloud services enable logging.googleapis.com
+  gcloud services enable stackdriver.googleapis.com
+  # Create the service account for the control plane if it doesn't exist
```
liu-cong, line 46: nice! Please also remove the "TODO(#896)" in https://github.com/google/knative-gcp/blob/master/docs/install/install-knative-gcp.md.

hack/lib.sh
```diff
@@ -0,0 +1,163 @@
...
+  local project_id=${1}
+  local control_plane_service_account=${2}
+  local cluster_location_type=${3}
+
+
+  echo "parameter project_id used when enabling workload identity is'${project_id}'"
+  echo "parameter control_plane_service_account used when when enabling workload identity is'${control_plane_service_account}'"
+  local cluster_location_option
+  if [ ${cluster_location_type} == ${ZONAL_CLUSTER_LOCATION_TYPE} ]; then
+    cluster_location_option=zone
```
liu-cong, line 111: why do we need `cluster_location_option `? why not use cluster_location_type directly?

hack/lib.sh
```diff
@@ -0,0 +1,163 @@
...
+    echo >&2 "Fatal error: cluster_location_type used when when enabling workload identity has to '${ZONAL_CLUSTER_LOCATION_TYPE}' or '${REGIONAL_CLUSTER_LOCATION_TYPE}'"
+    exit 1
+  fi
+  echo "parameter cluster_location_type used when when enabling workload identity is'${cluster_location_type}'"
+  echo "cluster_location_option used when when enabling workload identity is'${cluster_location_option}'"
+  # Enable API
+  gcloud services enable iamcredentials.googleapis.com
+  # Enable workload identity.
+  echo "Enable Workload Identity"
+  local cluster_name="$(cut -d'_' -f4 <<<"$(kubectl config current-context)")"
```
liu-cong, line 124: hmm, I don't think we can reliably parse the context to get cluster name and location. I believe user can change the context name.

hack/init_cloud_storage_source.sh
ericlem: It feels like it might be nicer to make the key name a little more obscure (so we don't have to worry about unrelated keys matching it), but also unilaterally delete it before this line.  I'm not 100% fond of deleting things i'm not sure we created, but as is if a previous run exits this script without finishing it (ctrl-c etc) the next run will fail in an obscure way due to the existence of this file.  alternately, may just check if this file exists prior to this call and tell the user to delete it if needed?

ericlem: That said, this isn't a regression as far as i can tell, so i could live with it as-is.

capri-xiyue:
-  ./test/e2e-common.sh are the shared code between ./test/e2e-secret-tests.sh and ./test/e2e-wi-tests.sh
-  ./hack/lib.sh are shared among other init_xxx scripts under ./hack repo AND ./test/e2e-common.sh AND./test/e2e-secret-tests.sh AND ./test/e2e-wi-tests.sh

In addition, I deleted the most tear-down code in e2e-xxx-tests.sh because most of the set up code are Idempotent

hack/lib.sh
capri-xiyue: cluster_location_type is `zonal` or `regional` based on what pantheon showes. But in the gcloud command, it should be `zone` or `region`.

hack/lib.sh
capri-xiyue: We need to add some comments like `The current project set in gcloud MUST be the same as where the cluster is running`

hack/init_cloud_storage_source.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
grantr: Clarification: scripts that are not executable, like `hack/lib.sh`, don't need these set directives.

hack/init_cloud_storage_source.sh
capri-xiyue: Currently the key name matches what the manual guide is using https://github.com/google/knative-gcp/blob/master/docs/install/authentication-mechanisms-gcp.md#option-2-kubernetes-secrets and https://github.com/google/knative-gcp/blob/master/docs/install/pubsub-service-account.md#option-2-export-service-account-keys-and-store-them-as-kubernetes-secrets.  Just use the same name for consistency. I thought we use such key name for simplicity.  Do you think we should use more obscure name in both init_xxx scripts and manual guide?As long as we keep key name consistent, I think it will be fine.

hack/lib.sh
capri-xiyue: Or do you think we should ask user for the project_id cluster name, cluster_location?

ericlem:
Have you run this in a fresh project with only the documented kubernetes/gke environment variables set?  We need to do that before checking it in to make sure it's not reliant on any pre-existing project settings.

capri-xiyue:
> Have you run this in a fresh project with only the documented kubernetes/gke environment variables set? We need to do that before checking it in to make sure it's not reliant on any pre-existing project settings.

I ran the e2e-xxx.tests.sh in a fresh project and it works. 
But I have not ran init_xxx.sh in a fresh project, I will try it when I modify the code based on code review.

hack/init_cloud_storage_source.sh
grantr: I suggest using `mktemp` to create temporary directories for each run. Then there's no state to consider on start or exit.

hack/lib.sh
grantr: I don't think we should ask the user. The script should be non-interactive.

We can use env vars for cluster name and location here, and only get values from kubectl context if those env vars are empty. I also recommend printing the chosen name and location so the user can see if the guess was correct.

capri-xiyue: What does your `env vars` mean? Like this one **https://cloud.google.com/run/docs/configuring/environment-variables#command-line? Could you help provide an exmaple?

hack/init_cloud_storage_source.sh
ericlem: I'm good putting it all in mktemp if that's doable.  Makes it pretty obvious they are disposable directories.

hack/lib.sh
capri-xiyue: Do you mean something like this 
`export E2E_PROJECT_ID="$(gcloud config get-value project)`
`export CLUSTER_NAME="$(gcloud config get-value run/cluster)`
`export CLUSTER_LOCATION="$(gcloud config get-value run/cluster_location)`

grantr: Looks good. You can also use bash default values to allow the user to override:

```sh
PROJECT_ID=${PROJECT_ID:-$(gcloud config get-value project)}
CLUSTER_NAME=${CLUSTER_NAME:-$(gcloud config get-value run/cluster)}
CLUSTER_LOCATION=${CLUSTER_LOCATION:-$(gcloud config get-value run/cluster_location)}
```
```

File: hack/lib.sh (hunk 12 above)
Comment (liu-cong) [reply hack/lib.sh:111]: I think you can tell if a cluster is zonal or regional by looking at its location format. But this can be a stretch goal. Feel free to dismiss.

File: test/e2e/test_gcp_broker.go [hunk 15]
```diff
@@ -0,0 +1,134 @@
...
+		t.Failed()
+	}
+	// Check if resp CloudEvent hits the target Service.
+	if done := jobDone(client, targetName, t); !done {
+		t.Error("resp event didn't hit the target pod")
+		t.Failed()
+	}
+}
+
+func createGCPBroke(t *testing.T, client *lib.Client, targetName string) url.URL {
```
liu-cong, line 88: this function does a lot more than creating a broker. Maybe call it something like setupBrokerAndTriggers, or divide it into multiple helper functions.

test/e2e/test_gcp_broker.go
```diff
@@ -0,0 +1,134 @@
...
+	defer lib.TearDown(client)
+
+	// Create a target Job to receive the events.
+	makeTargetJobOrDie(client, targetName)
+
+	u := createGCPBroke(t, client, targetName)
+
+	// Just to make sure all resources are ready.
+	// Sleep for 60 seconds because of the delay of mounted ConfigMaps of GCP Broker. We will decrease the sleep time later after we update the annotation of GCP Broker which triggers an immediate refresh of mounted ConfigMaps
+	time.Sleep(60 * time.Second)
```
liu-cong, line 67: This is implemented. We can try a much lower wait time now.

test/e2e/test_gcp_broker.go
capri-xiyue: Currently I just follow the code here https://github.com/google/knative-gcp/blob/8efed14a6bce20ec830a99b457938f3096f46691/test/e2e/test_broker_pubsub.go#L305. There are some duplicate code and I agree this function does a lot more than creating a broker. I will do the refactor in a follow up issue otherwise this PR will be much more larger considering the refactor of GCP Broker and Broker with PubSub Channel.

hack/init_cloud_storage_source.sh
capri-xiyue: `gcloud iam service-accounts keys create ${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP} \` this command won't fail when the file already exists. It will override the file. I will put it all in mktemp for both manual steps and init_xxx scripts.

hack/lib.sh
capri-xiyue: I will remove it when I work on updating the docs

chizhg:
/lgtm
/hold
In case other reviewers have further comments.

hack/init_control_plane_gke.sh
```diff
@@ -14,29 +14,47 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Usage: ./init_control_plane_gke.sh
-# The current project set in gcloud MUST be the same as where the cluster is running.
+# Usage: ./init_control_plane_gke.sh [CLUSTER_NAME] [CLUSTER_LOCATION] [CLUSTER_LOCATION_TYPE] [PROJECT_ID]
+#  where [CLUSTER_NAME] is an optional parameter to specify the cluster to use.
```
liu-cong, line 18:
```suggestion
# [CLUSTER_NAME] is an optional parameter to specify the cluster to use, default to gcloud(gcloud config get-value run/cluster_location).
```

hack/init_control_plane_gke.sh
```diff
@@ -14,29 +14,47 @@
...
+#  If CLUSTER_NAME not specified, we use the cluster set in gcloud(gcloud config get-value run/cluster_location).
```
liu-cong, line 19:
```suggestion
```

hack/init_control_plane_gke.sh
liu-cong: Same comments for other parameters.

test/e2e-wi-tests.sh
```diff
@@ -13,72 +13,67 @@
...
-      --role roles/logging.privateLogViewer
-    gcloud projects add-iam-policy-binding ${E2E_PROJECT_ID} \
-      --member=serviceAccount:${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
-      --role roles/cloudscheduler.admin
-        # Give iam.serviceAccountAdmin role to the Google service account.
-    gcloud projects add-iam-policy-binding ${E2E_PROJECT_ID} \
-      --member=serviceAccount:${CONTROL_PLANE_SERVICE_ACCOUNT}@${E2E_PROJECT_ID}.iam.gserviceaccount.com \
-      --role roles/iam.serviceAccountAdmin
+    init_control_plane_service_account ${E2E_PROJECT_ID} ${CONTROL_PLANE_SERVICE_ACCOUNT_NON_PROW}
+    local cluster_name="$(cut -d'_' -f4 <<<"$(kubectl config current-context)")"
```
liu-cong, line 67: The context can be **Renamed** so we cannot rely on the name of the context.  Can use what @grantr suggested:
```
CLUSTER_NAME=${CLUSTER_NAME:-$(gcloud config get-value run/cluster)}
CLUSTER_LOCATION=${CLUSTER_LOCATION:-$(gcloud config get-value run/cluster_location)}
```

As far as I can tell this is not a regression, so we can either fix it here or create an issue to follow up.

capri-xiyue: This is for the e2e-wi-tests. For e2e tests, it will override the kubectl context but it won't change the gcloud config. And the cluster won't get created until the e2e set up scripts get started.

liu-cong: I see. But the script also works with existing clusters, in which case the kubectl context is not overwritten, correct?

capri-xiyue: Yep. Just wanna clarify, for e2e tests with existing cluster, we don't run this e2e set up scripts. We will run the init_xxx.scripst under ./hack repo.

liu-cong: Thanks for the clarificaiton!

liu-cong:
/lgtm

It seems to me all comments have been resolved so I am going to unhold.
/unhold

Issue #905, opened 2020-04-21
Switch API Groups - pubsub.cloud.google.com -> internal.events.cloud.google.com


Harwayne:
**Problem**
Change the API Groups of existing resources.

* `Topic` and `PullSubscription` should be moved to `internal.events.cloud.google.com`.

**Additional context (optional)**
Should we make a breaking change and just drop the existing resources? Or should we move things over safely?

Harwayne:
Some ideas about how we can do this change.

Note: This assumes that all our controllers atomically swap. Either they are all on the old release or all on the new. Because we only have one replica and can use leader election, I think this is reasonable. But it does mean we also need to think about what happens during downgrades.

### Ignoring the old

Idea: Pretend the old one doesn't exist at all.

1. Move the CRD to the new Group.
1. User's upgrade.
1. The old CRD and COs continue to exist. The user has to manually delete them. The user **also** has to delete the now orphaned resources in GCP.

**Pros**
* Easy for us, no technical debt.
* Events are less likely to be lost, as everything will continue to work until a reconcile is required.

**Cons**
* Hard to clean up for the user.
* COs look normal. They don't tell users that they are effectively ghosts. Users would need to read the release notes.

### Deprecate, Delete, Remove

Idea: Make user do the work, but give them a release to get it done.

1. Copy the CRDs to the new Group. Adjust all code to use the new groups exclusively.
1. Keep running the old CRD controllers. Have them mark all the resources with deprecation notices.
1. Wait a release, let's call it 0.x+1.
1. Have the Webhook deny creation of any new COs of the old CRDs.
1. Have the old CRD controllers delete the underlying resources.
  - Possibly via a secondary controller which simply marks them for deletion.
1. Wait a release to 0.x+2.
1. Delete the old CRD controllers.
1. Tell the user to delete the old CRDs.
  - An upgrade script could do this too.

**Pros**
* Users see what is going on via introspecting objects (e.g. the CO will have the deprecation warning).
* All resources are cleaned up.
  - Well, the clean up is _attempted_. It may fail continuously while release 0.x+1 is installed.

**Cons**
* Takes a minimum of three releases.
* Will likely lead to duplicate events for most of release 0.x+1 because both old and new PullSubscriptions exist.

### Doppelgänger

Idea: Have the new PullSubscription controller determine if it is a Doppelgänger and if so, copy the other's GCP Subscription name.

1. Copy the CRDs to the new Group. Adjust all code to use the new groups exclusively.
[...]

PR #980, opened 2020-04-30
Add Deprecation Warnings to Topic and PullSubscription

Harwayne:
Helps with #905.

## Proposed Changes

- Add deprecation warnings to every `Topic` and `PullSubscription` in the `pubsub.cloud.google.com` API group.
    ```yaml
    - lastTransitionTime: "2020-05-01T22:56:05Z"
      message: This Kind is deprecated and the CRD has been made 'internal'. This
        means, end-users should not create objects of this CRD directly. If you need
        a non-internal variant, then please let us know by commenting on https://github.com/google/knative-gcp/issues/905.
        Moreover, the object must be deleted before upgrading to 0.16 to avoid orphaning
        the Google Cloud Platform resources that were created on-behalf of this object,
        such as Pub/Sub Topics and Subscriptions.
      reason: willBeRemoved
      severity: Warning
      status: "True"
      type: Deprecated
    ```

**Release Note**

```release-note
All `Topic` and `PullSubscription` objects in the `pubsub.cloud.google.com` API group will have deprecation warnings in their status. Those objects will no longer exist in the 0.16 release. All `Topic` and `PullSubscription` objects created by other reconcilers within `knative-gcp` will automatically migrate to the replacements without user interaction. **Migration may cause back logged events to be lost.**
```


Harwayne:
/hold

Holding at allow @akashrv to confirm or alter the deprecation message.

pkg/apis/pubsub/internal/deprecated_condition.go
```diff
@@ -0,0 +1,65 @@
...
+
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+)
+
+const deprecationMessage = "This object is deprecated and should be deleted. There is no public " +
+	"replacement. If you want a public replacement, please comment on " +
+	"https://github.com/google/knative-gcp/issues/905. This object should be deleted before " +
```
akashrv, line 31: Content is correct. We could mellow down a bit. How about: "This object (or kind) is deprecated and the CRD is made 'internal'. This means, end-users should not create objects of this CRD directly. If you need a non-internal (or a public) variant, then please let us know by commenting on https://github.com/google/knative-gcp/issues/905. Moreover, the object must be deleted before upgrading to 0.16 to avoid leaking (or orphaning) the Google Cloud Platform resources that were created on-behalf of this object, such as Pub/Sub topics and subscriptions."

Harwayne: Sounds good, how about (almost identical to yours, but removed the parenthesis):

`This Kind is deprecated and the CRD has been made 'internal'. This means, end-users should not create objects of this CRD directly. If you need a non-internal variant, then please let us know by commenting on https://github.com/google/knative-gcp/issues/905. Moreover, the object must be deleted before upgrading to 0.16 to avoid orphaning the Google Cloud Platform resources that were created on-behalf of this object, such as Pub/Sub Topics and Subscriptions.`

akashrv: /LGTM

Harwayne:
/hold cancel

Cancelling hold now that the wording has been agreed upon.

Harwayne:
Ping.

pkg/apis/pubsub/internal/deprecated_condition.go
```diff
@@ -0,0 +1,66 @@
...
+	"knative.dev/pkg/apis"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+)
+
+const deprecationMessage = "This Kind is deprecated and the CRD has been made 'internal'. This " +
+	"means, end-users should not create objects of this CRD directly. If you need a non-internal " +
+	"variant, then please let us know by commenting on " +
+	"https://github.com/google/knative-gcp/issues/905. Moreover, the object must be deleted " +
+	"before upgrading to 0.16 to avoid orphaning the Google Cloud Platform resources that were " +
+	"created on-behalf of this object, such as Pub/Sub Topics and Subscriptions."
```
grantr, line 34: Couple of grammar nits:
`end-users` -> `end users`
`on-behalf` -> `on behalf`

pkg/apis/pubsub/internal/deprecated_condition.go
```diff
@@ -0,0 +1,66 @@
...
+
+// deprecatedCondition is the condition to add to types that will be removed in 0.16.
+// See https://github.com/google/knative-gcp/issues/905 for more context.
+var deprecatedCondition = apis.Condition{
+	Type:     "Deprecated",
+	Reason:   "willBeRemoved",
```
grantr, line 40: Seems like this should be `WillBeRemoved` like other condition reasons

Harwayne: Done.

pkg/apis/pubsub/internal/deprecated_condition.go
Harwayne: Done.

Issue #317, opened 2019-10-03
StackDriver metrics e2e test does not work in PROW 


nachocano:
- I can consistently execute it locally and the test passes. In PROW, it doesn't. Needs further investigation...

`GOOGLE_APPLICATION_CREDENTIALS=<path_to_json> E2E_PROJECT_ID=<id> go test --tags=e2e ./test/e2e/... -run ^TestStorageStackDriverMetrics$ -count 1`
`
ok  	github.com/google/knative-gcp/test/e2e	156.576s
?   	github.com/google/knative-gcp/test/e2e/metrics	[no test files]
`

nachocano:
Log from PROW

--- FAIL: TestStorageStackDriverMetrics (144.18s)
    lifecycle.go:54: namespace is : "test-storage-stack-driver-metrics-nglsxtiv"
    test_storage.go:278: Last term message => {"success":true}
    test_storage.go:303: sleeping 1m0s to make sure metrics were pushed to stackdriver
    test_storage.go:336: Filter expression: metric.type="custom.googleapis.com/cloud.run/source/event_count" metric.label.event_source="//storage.googleapis.com/buckets/storage-e2e-test-knative-boskos-17" metric.label.namespace_name="test-storage-stack-driver-metrics-nglsxtiv" resource.type="global" metric.label.resource_group="storages.events.cloud.run" metric.label.event_type="google.storage.object.finalize" metric.label.name="storage-e2e-test-knative-boskos-17-storage" metric.label.response_code="500" metric.label.response_code_class="5xx"
    test_storage.go:346: failed to iterate over result: rpc error: code = NotFound desc = The metric referenced by the provided filter is unknown. Check the metric name and labels.
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1d23950]

chizhg:
Can [this](https://github.com/GoogleCloudPlatform/k8s-stackdriver/issues/198#issuecomment-415396464) be relevant?

chizhg:
/assign @chaodaiG 
Chao is oncall this week and he's willing to help :-)

chaodaiG:
This test failed the same way on my local desktop as in Prow, this is my first time running e2e tests from knative-gcp on my machine. The only difference is I invoked the test same way as in Prow, which is our standard way of testing:
[...]

PR #985, opened 2020-05-01
Update stackdriver e2e testing

bskaplan:
Updates for #317 

## Proposed Changes
- Updates the infrastructure for stackdriver tests to better work with the current implementation
- add a test that verifies stackdriver metrics for the broker. The test is currently disabled because the stackdriver implementation is currently broken due to an upstream config change.

### DIFF
--- a/test/e2e/lib/broker.go
+++ b/test/e2e/lib/broker.go
@@ -34,10 +34,10 @@ func AssertBrokerMetrics(t *testing.T, client *Client) {
 	// If we reach this point, the projectID should have been set.
 	projectID := os.Getenv(ProwProjectKey)
 	f := map[string]interface{}{
-		"metric.type":                 BrokerEventCountMetricType,
-		"resource.type":               BrokerMetricResourceType,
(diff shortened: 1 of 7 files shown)


test/e2e/lib/broker.go
```diff
@@ -0,0 +1,57 @@
...
+	"os"
+	"testing"
+	"time"
+
+	"github.com/google/knative-gcp/test/e2e/lib/metrics"
+	pkgmetrics "knative.dev/pkg/metrics"
+)
+
+func AssertBrokerMetrics(t *testing.T, client *Client) {
+	sleepTime := 4 * time.Minute
```
grantr, line 30: This is a long time! Where did this value come from?

bskaplan: Observations that data would show up about 2 minutes after the request was processed and wanting a buffer to make sure the tests won't be flaky. I don't like it, but it makes the test pass.

grantr: Thanks for figuring out how to make this pass!

PR #994, opened 2020-05-04
Reconcile gcp broker data plane deployments

### LINKED ISSUE #862: BrokerCell reconciler: static resources

liu-cong:
Fixes #862

## Proposed Changes

- Reconcile ingress/fanout/retry for brokercell.
-
-

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/helper.go
```diff
@@ -0,0 +1,61 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/equality"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/client-go/kubernetes"
+	appsv1listers "k8s.io/client-go/listers/apps/v1"
+	corev1listers "k8s.io/client-go/listers/core/v1"
+)
+
+// ReconcileDeployment reconciles the K8s Deployment 'd'.
+func ReconcileDeployment(kubeClient kubernetes.Interface, lister appsv1listers.DeploymentLister, d *v1.Deployment) (*v1.Deployment, error) {
```
grantr, line 14: This should take an event recorder or return an event to record. An event should be recorded whenever a deployment is created or updated.

pkg/reconciler/brokercell/brokercell.go
```diff
@@ -42,22 +66,85 @@ var _ bcreconciler.Finalizer = (*Reconciler)(nil)
...
+		return err
+	}
+	endpoints, err := reconciler.ReconcileService(r.kubeClientSet, r.serviceLister, r.endpointsLister, resources.MakeIngressService(ingressArgs))
+	if err != nil {
+		r.Logger.Errorf("Failed to reconcile ingress service for \"%s/%s\": %v", bc.Namespace, bc.Name, err)
+		bc.Status.MarkIngressFailed("IngressServiceFailed", "Failed to reconcile ingress service for \"%s/%s\": %v", bc.Namespace, bc.Name, err)
+		return err
+	}
+	bc.Status.PropagateIngressAvailability(endpoints)
+	bc.Status.IngressTemplate = "http://" + names.ServiceHostName(endpoints.GetName(), endpoints.GetNamespace())
```
grantr, line 83: The intent here is to signal to the broker controller how to generate the address for a broker, so it doesn't have to embed its knowledge of the ingress. It's kind of a synthetic and maybe useless separation, since the broker controller and brokercell controller are always in the same container.

Since this field isn't used by the broker controller yet, it should probably just be empty (or removed from the struct entirely, but that's a job for another PR).

pkg/reconciler/brokercell/brokercell.go
```diff
@@ -42,22 +66,85 @@ var _ bcreconciler.Finalizer = (*Reconciler)(nil)
 func (r *Reconciler) ReconcileKind(ctx context.Context, bc *intv1alpha1.BrokerCell) pkgreconciler.Event {
 	bc.Status.InitializeConditions()
 
+	// Reconcile ingress deployment and service
+	ingressArgs := r.makeIngressArgs(bc)
+	if _, err := reconciler.ReconcileDeployment(r.kubeClientSet, r.deploymentLister, resources.MakeIngressDeployment(ingressArgs)); err != nil {
+		r.Logger.Errorf("Failed to reconcile ingress deployment for \"%s/%s\": %v", bc.Namespace, bc.Name, err)
```
grantr, line 72: I think the namespace and name of the reconciled object is already a field on the log message, so no need to include them here (and elsewhere).

My preference is to use the non-sugared logger, so the variables are typed and checked at compile time. Using the sugared logger means these are only checked at runtime.

pkg/reconciler/brokercell/brokercell.go
```diff
@@ -42,22 +66,85 @@ var _ bcreconciler.Finalizer = (*Reconciler)(nil)
...
+		bc.Status.MarkIngressFailed("IngressDeploymentFailed", "Failed to reconcile ingress deployment for \"%s/%s\": %v", bc.Namespace, bc.Name, err)
```
grantr, line 73: The name and namespace is already present on the object. No need to add them to the status message.

pkg/reconciler/brokercell/brokercell.go
```diff
@@ -42,22 +66,85 @@ var _ bcreconciler.Finalizer = (*Reconciler)(nil)
...
 	return pkgreconciler.NewEvent(corev1.EventTypeNormal, "BrokerCellReconciled", "BrokerCell reconciled: \"%s/%s\"", bc.Namespace, bc.Name)
 }
 
 // FinalizeKind will be called when the resource is deleted.
 func (r *Reconciler) FinalizeKind(ctx context.Context, bc *intv1alpha1.BrokerCell) pkgreconciler.Event {
-	// TODO Finalize by:
-	// - Deleting the deployments. Wait until they're gone
-	// - Delete the configmap
-	// - When to delete the svc? Look at revision deletion to see if it specifies a deletion order
+	// Resources owned by this brokercell should automatically be garbage collected.
```
grantr, line 111: If so then just remove FinalizeKind entirely. No need for a finalizer.

pkg/reconciler/brokercell/brokercell_test.go
```diff
@@ -56,47 +63,415 @@ func init() {
...
+			Key:  "too/many/parts",
+		},
+		{
+			Name: "key not found",
+			Key:  testKey,
+		},
+		{
+			Name: "BrokerCell not found",
+			Key:  testKey,
+		},
```
grantr, line 78: This case can be removed. It's the same as the one above

pkg/reconciler/brokercell/controller.go
```diff
@@ -23,6 +23,12 @@ import (
...
 	v1alpha1brokercell "github.com/google/knative-gcp/pkg/client/injection/reconciler/intevents/v1alpha1/brokercell"
+	"github.com/kelseyhightower/envconfig"
+	"go.uber.org/zap"
+	kubeclient "knative.dev/pkg/client/injection/kube/client"
+	deploymentinformer "knative.dev/pkg/client/injection/kube/informers/apps/v1/deployment"
+	endpointsinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/endpoints"
+	serviceinformer "knative.dev/pkg/client/injection/kube/informers/core/v1/service"
 	configmap "knative.dev/pkg/configmap"
 	controller "knative.dev/pkg/controller"
 	logging "knative.dev/pkg/logging"
```
grantr, line 34: I recommend using [the logging package from eventing](https://github.com/knative/eventing/blob/master/pkg/logging/logging.go). It uses desugared loggers for better safety.

Also I wrote this code and it was wrong 😅

pkg/reconciler/brokercell/controller.go
```diff
@@ -41,11 +47,23 @@ func NewController(
 ) *controller.Impl {
 	logger := logging.FromContext(ctx)
 
-	brokercellInformer := brokercell.Get(ctx)
-
-	// TODO: setup additional informers here.
+	var env envConfig
+	if err := envconfig.Process("BROKER_CELL", &env); err != nil {
+		logger.Fatal("Failed to process env var", zap.Error(err))
```
grantr, line 52: This needs a desugared logger

pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -0,0 +1,148 @@
...
+	container := containerTemplate(args.Args)
+	return deploymentTemplate(args.Args, []corev1.Container{container})
+}
+
+// deploymentTemplate creates a template for data plane deployments.
+func deploymentTemplate(args Args, containers []corev1.Container) *appsv1.Deployment {
+	return &appsv1.Deployment{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:       args.BrokerCell.Namespace,
+			Name:            Name(args.BrokerCell.Name, args.ComponentName),
```
grantr, line 57: We should use kmeta.ChildName here to avoid name length issues.

pkg/reconciler/brokercell/brokercell.go
liu-cong: I like the separation though and prefer to keep this :) Although we can use static names for now, but in theory Broker controller doesn't know the ingress endpoint. BrokerCell controller should know since it creates the ingress. 

Is there any strong reason not to keep this?

grantr: :smiley: No strong reason. If we want to keep this, then it should be a URI template so the Broker controller doesn't have to know that the path is `/namespace/name`. I'll open an issue for this so it doesn't derail this PR.

pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -0,0 +1,148 @@
...
+	corev1 "k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/kmeta"
+	"knative.dev/pkg/system"
+)
+
+// MakeIngressDeployment creates the ingress Deployment object.
+func MakeIngressDeployment(args IngressArgs) *appsv1.Deployment {
```
yolocs, line 32: Does it complain test coverage if we don't have dedicated test?

pkg/reconciler/helper.go
```diff
@@ -0,0 +1,96 @@
...
+		if apierrs.IsAlreadyExists(err) {
+			return current, nil
+		}
+		if err == nil {
+			r.Recorder.Eventf(obj, corev1.EventTypeNormal, deploymentCreated, "Created deployment %s/%s", d.Namespace, d.Name)
+		}
+		return current, err
+	} else if err != nil {
+		return nil, err
+	} else if !equality.Semantic.DeepDerivative(d.Spec, current.Spec) {
```
yolocs, line 49: For consistency, you don't need to the `else` here

pkg/reconciler/brokercell/resources/deployments.go
liu-cong: Yes it does... But it should have good coverage in the controller test. Any workarounds?

yolocs: I know you can skip them in `.gitattributes`. But I'm not sure if you can still see the coverage of these files and our existing resources folder are all providing dedicated tests.

pkg/reconciler/brokercell/brokercell.go
liu-cong: Thanks opening the issue. I guess that issue is about having Broker controller to use the template? I updated this PR to return a URI template string. Let me know if it's good :)

pkg/reconciler/brokercell/brokercell.go
```diff
@@ -43,57 +44,85 @@ type envConfig struct {
...
 func (r *Reconciler) ReconcileKind(ctx context.Context, bc *intv1alpha1.BrokerCell) pkgreconciler.Event {
 	bc.Status.InitializeConditions()
 
 	// Reconcile ingress deployment and service
 	ingressArgs := r.makeIngressArgs(bc)
-	if _, err := reconciler.ReconcileDeployment(r.kubeClientSet, r.deploymentLister, resources.MakeIngressDeployment(ingressArgs)); err != nil {
-		r.Logger.Errorf("Failed to reconcile ingress deployment for \"%s/%s\": %v", bc.Namespace, bc.Name, err)
-		bc.Status.MarkIngressFailed("IngressDeploymentFailed", "Failed to reconcile ingress deployment for \"%s/%s\": %v", bc.Namespace, bc.Name, err)
+	if _, err := r.deploymentRec.ReconcileDeployment(bc, resources.MakeIngressDeployment(ingressArgs)); err != nil {
+		r.Logger.Desugar().Error("Failed to reconcile ingress deployment for \"%s/%s\": %v", zap.Any("namespace", bc.Namespace), zap.Any("name", bc.Name), zap.Error(err))
```
grantr, line 100: Sorry I misread this before and that made my earlier comment unclear. Instead of using `r.Logger`, get the logger from the context using `logging.FromContext(ctx)`. It contains additional structured fields specific to this reconcile call. You can use `knative.dev/eventing/pkg/logging` (which returns a desugared logger) to eliminate the `Desugar()` calls.

pkg/reconciler/brokercell/brokercell.go
grantr: Thanks, looks good. I updated the issue title to be more clear that it's about adding this functionality to broker controller.

PR #995, opened 2020-05-04
Make deployment controller listen to 'Create' and 'Delete' events of Secret

### LINKED ISSUE #971: Deployment controller needs to watch  'Create' secret
**Problem**
After #825 get merged, in the init state, if the user choose to use K8S Secret, once the real secret is in, the controller needs restart to take that secret in. Thus, we probably need to investigate for letting the controller also listens to "Create" that specific secret.

 @capri-xiyue mentioned that last time when she developed it, if the deployment controller listens to "Create" specific secret, it will result in infinite loop.

**Exit Criteria**
deployment controller will watch  'Create' secret, and force a restart for the controller pod.

**Time Estimate (optional):** 

grac3gao:
Fixes #971

## Proposed Changes

- deployment controller listens to 'Create'/'Delete' events of Secret, and forces a restart for the controller pod.
- controller pod will not run into infinite restarting loop
-

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/deployment/controller.go
```diff
@@ -63,21 +66,27 @@ func NewController(
...
+		AddFunc: func(obj interface{}) {
+			if _, ok := os.LookupEnv(envKey); !ok {
+				h(obj)
+			}
+		},
 		UpdateFunc: controller.PassNew(h),
-		DeleteFunc: doNothing,
+		// If secret is deleted, the controller pod will restart, in order to unset the envKey.
+		// This is needed when changing authentication configuration from k8s Secret to Workload Identity.
+		DeleteFunc: h,
```
capri-xiyue, line 90: what is the difference between `h` and `controller.PassNew(h)`.

grac3gao: The structure of `ResourceEventHandlerFuncs` is 
```
type ResourceEventHandlerFuncs struct {
	AddFunc    func(obj interface{})
	UpdateFunc func(oldObj, newObj interface{})
	DeleteFunc func(obj interface{})
}
```
For UpdateFunc, it is changing the old object (resource before update) to be the new object (resource after update). By using PassNew, only new object is forwarded. Checking the comments under PassNew Function, it is mentioned: PassNew makes it simple to create an UpdateFunc for use with cache.ResourceEventHandlerFuncs that can delegate the same methods as AddFunc/DeleteFunc but passing through only the second argument (which is the "new" object).

pkg/reconciler/deployment/controller.go
```diff
@@ -63,21 +66,27 @@ func NewController(
...
 	})
 	return impl
 }
 
-func handleUpdateOnly(h func(interface{})) cache.ResourceEventHandler {
+func handler(h func(interface{})) cache.ResourceEventHandler {
 	return cache.ResourceEventHandlerFuncs{
-		AddFunc:    doNothing,
+		// For AddFunc, only enqueue deployment key when envKey is not set.
+		// In such case, the controller pod haven't restarted before.
```
Harwayne, line 80:
```suggestion
		// In such case, the controller pod hasn't restarted before.
```

Issue #927, opened 2020-04-23
GCP Broker supports event "TTL"


liu-cong:
**Problem**
The channel based broker supports the “TTL” feature - # times an event can be sent to a broker by reply. We should implement similar logic.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
* Whenever an event is replied, increment its counter. Drop the event if the counter reaches the limit.
* individual unit tests in ingress, fanout and retry

**Additional context (optional)**
 * Fanout/retry increments the counter
* Ingress checks the counter and drops event if it reaches the limit

grantr:
Exit criteria should include only unit tests in specific components. Will create a separate issue for an E2E test of overall TTL behavior.

PR #996, opened 2020-05-04
Add event TTL (hops) to broker

yolocs:
Fixes https://github.com/google/knative-gcp/issues/927

## Proposed Changes

- add event TTL (hops) to broker
- Defaults to 255 hops which is the same as eventing core broker TTL

**Release Note**

```release-note
Add event TTL (hops) to broker
```

**Docs**


Harwayne:
FWIW, no one seemed to like me naming the property `TTL`. You might want to call it something else (e.g. `HopLimit`).

liu-cong:
> FWIW, no one seemed to like me naming the property `TTL`. You might want to call it something else (e.g. `HopLimit`).

There is an issue for renaming :) https://github.com/knative/eventing/issues/1018

Harwayne:
> > FWIW, no one seemed to like me naming the property `TTL`. You might want to call it something else (e.g. `HopLimit`).
> 
> There is an issue for renaming :) [knative/eventing#1018](https://github.com/knative/eventing/issues/1018)

Right. But because the property doesn't exist outside the Broker implementation itself (I don't think anyone actually added the configuration knob asked for in that issue), the name is entirely internal to the Broker. So now is a chance to change its name for this Broker implementation, if we want.

yolocs:
@liu-cong @Harwayne renamed to hops.

pkg/broker/eventutil/hops.go
```diff
@@ -0,0 +1,80 @@
...
+type Hops struct {
+	Logger *zap.Logger
+}
+
+// UpdateRemainingHops update an event with proper remaining hops.
+// 1. If the event doesn't have an existing hops value or have an invalid hops value,
+// it puts the preemptive hops in the event.
+// 2. If the event has an existing valid hops value,
+// it decrements it by 1.
+func (t *Hops) UpdateRemainingHops(event *event.Event, preemptiveHops int32) {
```
liu-cong, line 39:
```suggestion
func (h *Hops) UpdateRemainingHops(event *event.Event, preemptiveHops int32) {
```

pkg/broker/ingress/handler.go
```diff
@@ -134,6 +140,16 @@ func (h *handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
 		return
 	}
 
+	h.eventHops.UpdateRemainingHops(event, defaultEventHopsLimit)
```
liu-cong, line 143: Looks like the hops is decremented twice: once in the deliver  processor and once here. 

I think we only need to `GetRemainingHops` in ingress. Or what if we let the delivery processor to drop events that exceeds the hop limit? That way ingress doesn't need to do anything.

yolocs: No, in the deliver processor it sets the original hops without decrementing. The `UpdateRemainingHops` function only decrements hops when the event already has a Hops value. But the event reply won't have that (we also attempt to remove it no matter if it presents).

`UpdateRemainingHops` in ingress is needed (with the current approach) in order to set the initial hops. Doing it only in delivery processor is an interesting idea. I actually didn't see a problem but maybe there are some caveats. @Harwayne thoughts?

pkg/broker/eventutil/hops.go
```diff
@@ -0,0 +1,80 @@
...
+package eventutil
+
+import (
+	cetypes "github.com/cloudevents/sdk-go/pkg/cloudevents/types"
+	"github.com/cloudevents/sdk-go/v2/event"
+	"go.uber.org/zap"
+)
+
+const (
+	hopsAttribute = "kngcpbrokerhops"
```
Harwayne, line 26: Given this name, I don't know if it is counting up or down. How about `kngcpbrokerhopsremaining`?

pkg/broker/ingress/handler.go
Harwayne: I had two reasons for putting it in the ingress in the initial Broker implementation:
1. The event is dropped before being duplicated N times (for N Triggers).
1. It was a quick way to help prevent people from accidentally sending events directly to the internal Channel.

1 is not an issue for this Broker design. 2 could be, depending on how you want to think about it. There is no Channel object, but people could still send events directly to the PubSub Topic. By mandating that this extension is present, it makes it harder to do so by mistake (although a malicious actor would have no problem). 

I also liked the fact that all the 'drop the event' logic was in a single place. As it turned out, we never did any other form of filtering, so this was a moot point.

If nothing else, by doing it in the Ingress, we prevent another hop through PubSub, right? Whether that is worth the extra complexity of touching the attribute in both components is up to you.

pkg/broker/eventutil/hops.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+// Hops manages event hops.
+type Hops struct {
+	Logger *zap.Logger
```
Harwayne, line 31: I recommend having `ctx` passed in to the methods and using `logging.GetFromContext(ctx)` rather than storing it in the struct. That way any additional information about the event being processed is retained.

pkg/broker/ingress/handler.go
yolocs: I'm thinking we would drop the replied event right before forwarding to the ingress. It's a step before the ingress and shouldn't incur additional pubsub hop. If you don't see any other problems, I think I will proceed with having the logic in a single place.

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -101,6 +114,21 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
...
 
+	if hops <= 0 {
+		logging.FromContext(ctx).Debug("dropping event based on remaining hops status",
+			zap.String("target", tk),
+			zap.Int32("hops", hops),
+			zap.String("event.id", resp.ID()))
+		return nil
+	}
+
+	// Clean up potential hops from the reply. It really shouldn't happen though.
```
liu-cong, line 125: if you separate `SetHops` and `DecHops`, you can just do a `SetHops` here.

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -81,14 +84,24 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
 		return nil
 	}
 
+	// Hops is a broker local counter so remove any hops value before forwarding.
+	// Do not modify the original event as we need to send the original
+	// event to retry queue on failure.
+	copy := event.Clone()
+	// This will decrement the remaining hops if there is an existing value.
+	eventutil.UpdateRemainingHops(ctx, &copy, defaultEventHopsLimit)
```
liu-cong, line 92: I think it's more readable if you separate `SetHops` and `DecHops`.

pkg/broker/eventutil/hops.go
grantr: Since this is an internal attribute, I think it should be as short as possible. We can save people some money :grin:

yolocs:
> Only comment is that the hops name should be shorter. I don't need the name of an internal field that costs money to document itself.

Done

grantr:
> Only comment is that the hops name should be shorter. I don't need the name of an internal field that costs money to document itself.

I should clarify that the "I" here is intended to be the customer, not me personally.

To add some concrete numbers to this: at 1000 events per second, each requiring one publish and one pull, 16 bytes in an attribute key is about 80GB per month costing $3.31 (at $40 per TiB). The minimum size of a publish request is 1000 bytes, so if we send a single message per publish request, it might be that the size of most messages is less than the minimum.

I spoke with @Harwayne and he still prefers the longer name but doesn't object to a shorter one.

yolocs:
> To add some concrete numbers to this: at 1000 events per second, each requiring one publish and one pull, 16 bytes in an attribute key is about 80GB per month costing $3.31 (at $40 per TiB). The minimum size of a publish request is 1000 bytes, so if we send a single message per publish request, it might be that the size of most messages is less than the minimum.
> 
> I spoke with @Harwayne and he still prefers the longer name but doesn't object to a shorter one.

@grantr Since the attribute is mostly internal, I'd say let's use the shorter name. Could you re-stamp the PR?

PR #1012, opened 2020-05-06
Add mainhelper for common main logic

liu-cong:
## Proposed Changes
- Add a helper function similar to https://github.com/knative/pkg/blob/master/injection/sharedmain/main.go#L135
  - sharedmain is tailored for running controllers. the helper function here is a general purpose helper function.
  - I have also added an option to use fake k8s clients and informers, which makes it possible to run tests locally. One use case of this could be running some benchmark tests locally for quick iteration.

**Release Note**

```release-note
NONE
```

**Docs**


pkg/utils/mainhelper/boot_test.go
```diff
@@ -0,0 +1,128 @@
...
+
+func TestMain(m *testing.M) {
+	setup()
+	code := m.Run()
+	teardown()
+	os.Exit(code)
+}
+
+func TestBoot(t *testing.T) {
+	commonVerification := func(res *BootRes) {
```
ian-mi, line 51: This doesn't need to be anonymous.

pkg/utils/mainhelper/boot.go
```diff
@@ -0,0 +1,112 @@
...
+	"knative.dev/pkg/controller"
+	"knative.dev/pkg/injection/sharedmain"
+	"knative.dev/pkg/logging"
+	"knative.dev/pkg/profiling"
+)
+
+// BootRes holds a collection of objects after booting for convenient access by
+// other custom logic in the main function.
+type BootRes struct {
+	Ctx        context.Context
```
ian-mi, line 42: context should always be threaded through directly. I would make this a separate return result and also make context the first argument to boot rather than an option.

liu-cong: https://github.com/knative/pkg/blob/master/injection/sharedmain/main.go#L129

pkg/observability/observability.go
```diff
@@ -7,16 +7,19 @@ import (
...
 	tracingconfig "knative.dev/pkg/tracing/config"
 )
 
-// SetupDynamicConfig sets up logging, metrics, and tracing by watching observability
+// SetupDynamicConfigOrDie sets up logging, metrics, and tracing by watching observability
 // configmaps. Returns an updated context with logging and function to flush telemetry which should
 // be called before exit.
-func SetupDynamicConfig(ctx context.Context, componentName string) (context.Context, func(), error) {
+// The input context should have KubeClient injected.
+func SetupDynamicConfigOrDie(ctx context.Context, componentName string) (context.Context, *configmap.InformedWatcher, *profiling.Handler, func()) {
```
yolocs, line 22: nit: I feel a more idiomatic name could be `MustSetupDynamicConfig`.

pkg/utils/mainhelper/boot.go
```diff
@@ -0,0 +1,110 @@
...
+	KubeClient kubernetes.Interface
+	CMPWatcher configmap.Watcher
+	Cleanup    func()
+}
+
+// Similar to sharedmain.Main, Boot runs common logic in starting a main function,
+// it returns a result object that contains useful artifacts for later use.
+// Unlike sharedmain.Main, Boot is meant to be run as a helper function in any main
+// functions, while sharedmain.Main runs controllers with predefined method signatures.
+func Boot(component string, opts ...BootOption) (context.Context, *BootRes) {
```
yolocs, line 52: Just providing some naming idea: how about `Init()` like we have `google.Init()`

pkg/utils/mainhelper/boot.go
ian-mi: I prefer having two function definitions so that ctx can be passed directly as an argument rather than through an option. This is mostly a nit, however, so feel free to merge as is if you prefer.

pkg/observability/observability.go
liu-cong: I think I will keep `SetupDynamicConfigOrDie`, because this is consistent with other helper functions we call in sharedmain.

pkg/utils/mainhelper/boot.go
liu-cong: Sure, changed to `Init`

pkg/utils/mainhelper/boot.go
liu-cong: I don't see a use case to pass a context. The sharedmain.MainWithContext is not called by anything either as far as I can tell. I can either remove it or keep it as an option. Just leaving it as an option for now.

PR #1014, opened 2020-05-06
Register stats in NewStatsReporter instead of init

ian-mi:
Puts the latency measure inside of the stats reporter struct instead of a global variable and registers stats in NewStatsReporter instead of in init().


liu-cong:
With this change, you won't be able to get another StatsReporter instance due to duplicate view registration, right? But I don't think we expect more than one StatsReporter so this shouldn't be an issue.

/lgtm

/hold feel free to unhold.

ian-mi:
Yes, exactly. Since the stats reporter is provided by wire it would be easy to provide a single shared instance to multiple components if needed, but we expect that only the ingress handler should inject it.

PR #1015, opened 2020-05-06
Use wire to initialize fanout and retry SyncPools

### LINKED ISSUE #967: Simplify SyncPool Constructors and Options
**Describe the bug**
The broker pkg/broker/handler/pool/options.go includes a number of options for creating underlying pubsub and eventing clients:

```
// Options holds all the options for create handler pool.
type Options struct {
	// ProjectID is the project for pubsub.
	ProjectID string
...
	// PubsubClient is the pubsub client used to receive pubsub messages.
	PubsubClient *pubsub.Client
	// PubsubReceiveSettings is the pubsub receive settings.
	PubsubReceiveSettings pubsub.ReceiveSettings
	// CeClientOptions is the options used to create cloudevents client.
	CeClientOptions []ceclient.Option
}
```
The pool constructor constructs its eventing clients using these options, optionally overriding the pubsub client with the optional PubsubClient. This could be simplified by requiring these eventing clients to be provided by the caller.

 **Expected behavior**
NewSyncPool does not construct its dependencies and its options does not include the options of its dependencies.

ian-mi:
Simplifies option handling by removing client-related options from pool.Options leaving only the
pool-specific options.

Changes the sync pool implementations to always reuse a single pubsub client regardless of how the pool is initialized. Reusing clients is recommended by the pubsub documentation: https://pkg.go.dev/cloud.google.com/go/pubsub?tab=doc#Client.

Fixes #967


pkg/broker/handler/pool/providers.go
```diff
@@ -0,0 +1,69 @@
...
+	)
+)
+
+type (
+	ProjectID     string
+	DeliverClient ceclient.Client
+	RetryClient   ceclient.Client
+)
+
+func NewDeliverClient(hp *cehttp.Protocol, opts ...ceclient.Option) (DeliverClient, error) {
```
yolocs, line 53: Comment

pkg/broker/handler/pool/providers.go
```diff
@@ -0,0 +1,69 @@
...
+	return ceclient.NewObserved(hp, opts...)
+}
+
+// NewPubsubClient provides a pubsub client for the supplied project ID.
+func NewPubsubClient(ctx context.Context, projectID ProjectID) (*pubsub.Client, error) {
+	return pubsub.NewClient(ctx, string(projectID))
+}
+
+func NewRetryClient(ctx context.Context, client *pubsub.Client, opts ...ceclient.Option) (RetryClient, error) {
```
yolocs, line 62: Comment

cmd/broker/fanout/wire.go
```diff
@@ -0,0 +1,35 @@
...
+import (
+	"context"
+
+	"github.com/google/knative-gcp/pkg/broker/config/volume"
+	"github.com/google/knative-gcp/pkg/broker/handler/pool"
+	"github.com/google/wire"
+)
+
+func InitializeSyncPool(ctx context.Context, projectID pool.ProjectID, targetsVolumeOpts []volume.Option, opts ...pool.Option) (*pool.FanoutPool, error) {
+	panic(wire.Build(
```
yolocs, line 30: We may eventually all get familiar with wire but it seems it's too early to assume that (at least I'm not). So before that, I'd prefer document what it does wherever you use wire functions with some reference link.

cmd/broker/fanout/wire_gen.go
```diff
@@ -0,0 +1,50 @@
+// Code generated by Wire. DO NOT EDIT.
+
+//go:generate wire
+//+build !wireinject
```
yolocs, line 4: Also document how to add new dependency with wire.

pkg/broker/handler/pool/providers.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// Provider set providing the default fanout and retry clients.
+	ProviderSet = wire.NewSet(
+		cehttp.New,
+		NewDeliverClient,
+		NewPubsubClient,
+		NewRetryClient,
+		wire.Value([]cehttp.Option(nil)),
+		wire.Value(DefaultCEClientOpts),
+	)
```
yolocs, line 44: Similarly, more detailed comments would be appreciated.

pkg/broker/handler/pool/fanout.go
```diff
@@ -14,39 +14,40 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package fanout
+package pool
```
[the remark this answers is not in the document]
ian-mi: It depends on what you mean by necessary. The retry and fanout sync pools implement the same interface and share the same options and many of the same dependencies. For this reason it feels natural for the retry and fanout pools to live in the same package and it also allows shared code to gain test coverage from the pool tests.

PR #1018, opened 2020-05-07
Add StatsReporter for Broker fanout metrics

zargarpur:
A follow up PR will use this StatsReporter to emit the metrics defined in
here.

These metrics are the same one's emitted for the Trigger metrics in OSS
knative. Which are also listed in https://cloud.google.com/monitoring/api/metrics_knative
(when you search for "internal/eventing")


pkg/broker/fanout/stats_reporter.go
```diff
@@ -0,0 +1,161 @@
+package fanout
```
yolocs, line 1: I wonder if it makes sense to have all stats_reporters in sub-packages of metrics. This looks more like a "delivery stats reporter". We can put this under `metrics/delivery/stats_reporter`. Similarly for ingress. Thoughts?

ian-mi: How about pkg/observability/delivery_stats.go or pkg/metrics/delivery_stats.go? I think a package per reporter might be too much.

yolocs: I don't have strong preference. I usually do different package names with the same struct name (like I did for the sync pool), but I'm fine with this approach too.

zargarpur: I'll do pkg/metrics/delivery_stats.go for now, mainly because we can create subdirectories later if there ends up being a lot of files in pkg/metrics in the future

zargarpur: I ended up doing @yolocs suggestion since the both the ingress's and delivery's StatsReporter had the same function and metric names which caused a conflict being in the same package. Ending up with:

`pkg/metrics/delivery/stats_reporter.go`
     And going to create a follow up PR to move the ingress metrics to
`pkg/metrics/ingress/stats_reporter.go`

liu-cong: You can rename the variables or make them a field in the struct.  In fact, @ian-mi just did this for ingress recently, if you sync with master. What about `metrics/broker_stats_reporter` and `metrics/trigger_stats_reporter`?

zargarpur: Sync'd with @liu-cong offline and ended up renaming and moving some fields into the struct. Now everything's in `pkg/metrics` like @ian-mi suggested

pkg/metrics/delivery_reporter.go
```diff
@@ -0,0 +1,167 @@
...
+import (
+	"context"
+	"fmt"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+	"knative.dev/pkg/metrics"
+	"strconv"
+	"time"
```
liu-cong, line 28: nit: move standard libraries at the top and separate them from others

pkg/metrics/delivery_reporter.go
```diff
@@ -0,0 +1,167 @@
...
+
+type DeliveryReporter struct {
+	podName               PodName
+	containerName         ContainerName
+	dispatchTimeInMsecM   *stats.Float64Measure
+	processingTimeInMsecM *stats.Float64Measure
+}
+
+type DeliveryReportArgs struct {
+	namespace  string
```
liu-cong, line 39: I think these should be public as they will be used in a different pkg

pkg/metrics/delivery_reporter.go
```diff
@@ -0,0 +1,167 @@
...
+		// and the delivery to the Trigger subscriber.
+		processingTimeInMsecM: stats.Float64(
+			"event_processing_latencies",
+			"The time spent processing an event before it is dispatched to a Trigger subscriber",
+			stats.UnitMilliseconds,
+		),
+	}
+
+	if err := r.register(); err != nil {
+		return nil, fmt.Errorf("failed to register ingress stats: %w", err)
```
liu-cong, line 116:
```suggestion
		return nil, fmt.Errorf("failed to register delivery stats: %w", err)
```

pkg/metrics/delivery_reporter_test.go
```diff
@@ -0,0 +1,129 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"context"
+	reportertest "github.com/google/knative-gcp/pkg/metrics/testing"
```
liu-cong, line 21: nit: move this to its own import group

pkg/metrics/delivery_reporter_test.go
```diff
@@ -0,0 +1,129 @@
...
+	reportertest.ExpectMetrics(t, func() error {
+		return r.ReportEventProcessingTime(context.Background(), args, 1100*time.Millisecond)
+	})
+	reportertest.ExpectMetrics(t, func() error {
+		return r.ReportEventProcessingTime(context.Background(), args, 9100*time.Millisecond)
+	})
+	metricstest.CheckDistributionData(t, "event_processing_latencies", wantTags, 2, 1100.0, 9100.0)
+}
+
+func TestMetricsWithEmptySourceAndTypeFilter(t *testing.T) {
```
liu-cong, line 99: Seem like you can consolidate all three  tests to a table test

pkg/metrics/delivery_reporter.go
```diff
@@ -0,0 +1,167 @@
...
+	broker     string
+	trigger    string
+	filterType string
+}
+
+func (r *DeliveryReporter) register() error {
+	return view.Register(
+		&view.View{
+			Name:        "event_count",
+			Description: "Number of events received by a Trigger",
```
yolocs, line 49: Number of events delivered to a Trigger subscriber?

pkg/metrics/delivery_reporter.go
```diff
@@ -0,0 +1,167 @@
...
+			"The time spent dispatching an event to a Trigger subscriber",
+			stats.UnitMilliseconds,
+		),
+		// processingTimeInMsecM records the time spent between arrival at the Broker
+		// and the delivery to the Trigger subscriber.
+		processingTimeInMsecM: stats.Float64(
+			"event_processing_latencies",
+			"The time spent processing an event before it is dispatched to a Trigger subscriber",
+			stats.UnitMilliseconds,
+		),
```
yolocs, line 112: How would these be calculated? Via some custom attribute on the event?

pkg/metrics/delivery_reporter.go
zargarpur: Done

pkg/metrics/delivery_reporter.go
zargarpur: Good catch, updated

pkg/metrics/delivery_reporter_test.go
zargarpur: Done

pkg/metrics/delivery_reporter.go
zargarpur: Done

pkg/metrics/delivery_reporter_test.go
zargarpur: I'll leave it for now in this PR, but will create a follow one to look into it since I haven't done that before in golang

pkg/metrics/delivery_reporter.go
zargarpur: Updated

pkg/metrics/delivery_reporter.go
zargarpur: I have not looked to closely into integrating it yet, I'm going to do that as soon as I submit this PR but what do you mean by "some custom attribute on the event?"

liu-cong: The ingress adds an event arrival time attribute to the event. I guess that's what Chen is asking?

pkg/metrics/delivery_reporter.go
```diff
@@ -0,0 +1,167 @@
...
+				TriggerNameKey,
+				TriggerFilterTypeKey,
+				ResponseCodeKey,
+				ResponseCodeClassKey,
+				PodNameKey,
+				ContainerNameKey,
+			},
+		},
+		&view.View{
+			Description: r.dispatchTimeInMsecM.Description(),
```
liu-cong, line 64: nit: The view will by default inherit Name and Description from the Measure, let's either explicitly add `Name:r.dispatchTimeInMsecM.Name()` or remove the Description. I personally prefer explicitly add them here.

liu-cong:
/lgtm
Will leave for @ian-mi to approve.

pkg/metrics/delivery_reporter.go
zargarpur: Added the names explicitly

pkg/metrics/constants.go
```diff
@@ -0,0 +1,45 @@
...
+	EventTypeKey         = tag.MustNewKey(metricskey.LabelEventType)
+	TriggerNameKey       = tag.MustNewKey(metricskey.LabelTriggerName)
+	TriggerFilterTypeKey = tag.MustNewKey(metricskey.LabelFilterType)
+
+	ResponseCodeKey      = tag.MustNewKey(metricskey.LabelResponseCode)
+	ResponseCodeClassKey = tag.MustNewKey(metricskey.LabelResponseCodeClass)
+
+	PodNameKey       = tag.MustNewKey(metricskey.PodName)
+	ContainerNameKey = tag.MustNewKey(metricskey.ContainerName)
+)
```
ian-mi, line 45: Maybe name this tags.go?

zargarpur: @ian-mi I'm not sure why `e2e-secret-tests.sh` is showing up in my PR, I sync'd with master and I guess it was caught in the middle for some reason but I made no changes and the permissions are in line with https://github.com/google/knative-gcp/commit/12200f7088cd728b7e1a683fe5b6db80b7a36468#diff-098f6bcd4621d373cade4e832627b4f6

zargarpur: Renamed to `tags.go`

PR #1024, opened 2020-05-07
Receive messages directly from subscription in the broker handler

### LINKED ISSUE #963: The Broker Should Pull Messages Directly from the Pubsub Client
**Describe the bug**
The broker currently uses the pubsub CE bindings protocol to pull messages. The bindings package, however, is designed to allow direct control of the underlying protocol and it is not necessary to use CE protocol when not making use of the CE client. Directly pulling messages using the pubsub client has the following advantages:

1. the pubsub client manages goroutines and concurrency limits so it is not necessary for the broker handler to do so.
2. the handler will have access to the pubsub request context.
3. removes unnecessary abstraction and allows the handler to access the underlying pubsub message if necessary.

ian-mi:
This removes unnecessary indirection as well as serialization through the CE transport channel and allows the pubsub message context to be preserved during processing. Also removes the handler Concurrency option as concurrency is controlled using the subscription's receive settings. The handler will now no longer create idle worker goroutines, goroutines will instead be created by the pubsub client as needed.

Fixes #963

### DIFF
--- a/pkg/broker/handler/handler.go
+++ b/pkg/broker/handler/handler.go
@@ -18,12 +18,12 @@ package handler
 
 import (
 	"context"
-	"io"
 	"sync/atomic"
 	"time"
 
+	"cloud.google.com/go/pubsub"
 	"github.com/cloudevents/sdk-go/v2/binding"
-	"github.com/cloudevents/sdk-go/v2/protocol/pubsub"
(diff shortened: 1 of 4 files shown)


ian-mi:
Thanks! We should be able to attach the trace context to the pubsub request context, I will add that in my next PR.

PR #1026, opened 2020-05-08
Broker returns 202 instead of 200 when events are accepted

liu-cong:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

cc @grantr

### DIFF
--- a/pkg/broker/ingress/handler.go
+++ b/pkg/broker/ingress/handler.go
@@ -40,9 +40,6 @@ const (
 	// TODO(liu-cong) configurable timeout
 	decoupleSinkTimeout = 30 * time.Second
 
-	// defaultPort is the defaultPort number for the ingress HTTP receiver.
-	defaultPort = 8080
-
 	// EventArrivalTime is used to access the metadata stored on a
 	// CloudEvent to measure the time difference between when an event is
 	// received on a broker and before it is dispatched to the trigger function.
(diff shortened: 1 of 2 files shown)


pkg/broker/ingress/handler.go
```diff
@@ -140,10 +135,13 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
...
 		}
-		response.WriteHeader(statusCode)
-		response.Write([]byte(msg))
+		nethttp.Error(response, msg, statusCode)
 		return
 	}
+
+	// According to the data plane spec (https://github.com/knative/eventing/blob/master/docs/spec/data-plane.md), a
+	// non-callable SINK (which broker is) MUST respond with 202 Accepted if the request is accepted.
+	response.WriteHeader(nethttp.StatusAccepted)
```
bskaplan, line 144: Is it safe to hardcode StatusAccepted here or should this be set to statusCode?

pkg/broker/ingress/handler.go
```diff
@@ -116,15 +113,13 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
...
 		return
 	}
 	ns, broker := pieces[1], pieces[2]
 	event, msg, statusCode := h.toEvent(request)
 	if event == nil {
-		response.WriteHeader(statusCode)
-		response.Write([]byte(msg))
+		nethttp.Error(response, msg, statusCode)
 		return
 	}
```
yolocs, line 124: nit: it seems the `toEvent` func will only return bad request status code. Maybe just check err and return bad request here?

pkg/broker/ingress/handler.go
liu-cong: It's safer to return hardcode 202 here since this is the end of the request with no error. Although statusCode should also be 202 here. Let me know if you have other concerns.

pkg/broker/ingress/handler.go
liu-cong: `toEvent` can return different bad status codes, so we need it to return statusCode instead of just an error.

liu-cong: Actually I am wrong :) Let me refactor it

pkg/broker/ingress/handler.go
liu-cong: I refactored it a little bit. Let me know if this is better.

Issue #702, opened 2020-03-25
Enable verification of LICENSE check


nachocano:
**Describe the bug**
While trying to get this PR https://github.com/google/knative-gcp/pull/701 in, I kept on getting a build-test failure. All other PROW jobs passed. It seems that there is a problem with the LICENSE check. The third_party/VENDOR_LICENSE was updated, I ran update-codegen multiple times, but still PROW complaints that is old and that I should run it again...  

 **Expected behavior**
Build tests job pass, with the license check.

 **Knative-GCP release version**
HEAD

 **Additional context**
https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/701/pull-google-knative-gcp-build-tests/1242615382368325632

PR #1027, opened 2020-05-08
Run shellcheck on all shell scripts

chizhg:
## Proposed Changes
- Run `shellcheck` on all shell scripts
`shellcheck -f diff test/*.sh | git apply`
`shellcheck -f diff hack/*.sh | git apply`
- The new version of `ko` has a nice support for Go modules, then there is no need to `sed` when uploading eventing test images
- Fixes https://github.com/google/knative-gcp/issues/702

**Release Note**

```release-note
NONE
```


hack/verify-codegen.sh
```diff
@@ -52,9 +52,8 @@ diff -Naupr --no-dereference \
 diff -Naupr --no-dereference \
   "${REPO_ROOT_DIR}/pkg" "${TMP_DIFFROOT}/pkg" || ret=1
 
-# TODO uncomment this once we fix https://github.com/google/knative-gcp/issues/702
-# diff -Naupr --no-dereference \
-#  "${REPO_ROOT_DIR}/third_party" "${TMP_DIFFROOT}/third_party" || ret=1
+diff -Naupr --no-dereference \
```
liu-cong, line 55: I see #702 is still open. Is it fixed?

hack/lib.sh
```diff
@@ -153,18 +153,21 @@ function storage_admin_set_up() {
   local project_id=${1}
   local pubsub_service_account=${2}
   local pubsub_service_account_key_temp=${3}
```
liu-cong, line 155: Looks like this param is not needed any more?

hack/verify-codegen.sh
liu-cong: Sorry I didn't notice this PR actually fixes it

hack/lib.sh
chizhg: Good catch, removed.

hack/verify-codegen.sh
chizhg: Yeah it was fixed a while ago.

hack/init_cloud_storage_source.sh
```diff
@@ -21,18 +21,9 @@ set -o errexit
...
 
-# Download a JSON key for the service account.
-gcloud iam service-accounts keys create "${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}" \
-  --iam-account="${PUBSUB_SERVICE_ACCOUNT}"@"${PROJECT_ID}".iam.gserviceaccount.com
-
-storage_admin_set_up "${PROJECT_ID}" "${PUBSUB_SERVICE_ACCOUNT}" "${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}"
-
-# Remove the tmp file.
-rm "${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}"
+storage_admin_set_up "${PROJECT_ID}" "${PUBSUB_SERVICE_ACCOUNT}"
```
ian-mi, line 29: Why these changes?

chizhg: [This line](https://github.com/google/knative-gcp/blob/master/hack/lib.sh#L166) is too long and some logic is unnecessary to just get the GCS service account name, so I simplified it a bit. And Cong pointed out that one arg in this function is needed anymore after my change, so I also removed it from here.

May need @capri-xiyue to take a second look.

hack/lib.sh
```diff
@@ -128,43 +128,44 @@ function enable_workload_identity(){
...
     --role roles/iam.serviceAccountAdmin
   }
 
 function storage_admin_set_up() {
   echo "Update ServiceAccount for Storage Admin"
   local project_id=${1}
   local pubsub_service_account=${2}
-  local pubsub_service_account_key_temp=${3}
+  local project_number="$(gcloud projects describe "${project_id}" --format="value(projectNumber)")"
+  local access_token="$(gcloud auth application-default print-access-token)"
```
capri-xiyue, line 156: Which credential does this command use? The credential for the control plane or the credential for the data plane?
https://cloud.google.com/sdk/gcloud/reference/auth/application-default/print-access-token. 
In the docs, it looks like it may not always be the credential for the data plane.

hack/lib.sh
```diff
@@ -128,43 +128,44 @@ function enable_workload_identity(){
...
-  gcloud projects add-iam-policy-binding ${project_id} \
-    --member=serviceAccount:${pubsub_service_account}@${project_id}.iam.gserviceaccount.com \
+  gcloud projects add-iam-policy-binding "${project_id}" \
+    --member=serviceAccount:"${pubsub_service_account}"@"${project_id}".iam.gserviceaccount.com \
     --role roles/storage.admin
-  export GCS_SERVICE_ACCOUNT=`curl -s -X GET -H "Authorization: Bearer \`GOOGLE_APPLICATION_CREDENTIALS=${pubsub_service_account_key_temp} gcloud auth application-default print-access-token\`" "https://www.googleapis.com/storage/v1/projects/${project_id}/serviceAccount" | grep email_address | cut -d '"' -f 4`
-  gcloud projects add-iam-policy-binding ${project_id} \
-    --member=serviceAccount:"${GCS_SERVICE_ACCOUNT}" \
+
+  curl -s -X GET -H "Authorization: Bearer ${access_token}" "https://www.googleapis.com/storage/v1/projects/${project_id}/serviceAccount"
```
capri-xiyue, line 167: Why do we still need this command here?

hack/lib.sh
```diff
@@ -128,43 +128,44 @@ function enable_workload_identity(){
...
+  gcloud projects add-iam-policy-binding "${project_id}" \
```
capri-xiyue, line 168: We need to update the docs accordingly https://github.com/google/knative-gcp/blob/master/docs/examples/cloudstoragesource/README.md

hack/init_cloud_storage_source.sh
capri-xiyue: Have some questions regarding the removal of ${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}. Already added some comments in lib.sh

hack/lib.sh
chizhg: This command is needed to activate the GCS service account.

hack/lib.sh
chizhg: The doc does not need to be updated. Here it's using Option 1 to get the GCS service account, except instead of visiting the pantheon UI, it's using the curl command to activate the service account.

hack/lib.sh
chizhg: You are right, I have added the service account key back and used it to get the `access_token`.

liu-cong:
/lgtm

I will defer to @capri-xiyue to approve

Issue #842, opened 2020-04-15
Flaky E2E tests related to Cloud Audit Logs


capri-xiyue:
**Describe the bug**
`TestCloudAuditLogsSource` is kind of flaky. It is related to the SLO of CloudAuditLogsSource.
CAL source misses the CloudAuditLogs which gets created shortly after CAL source becomes ready( https://github.com/ian-mi/knative-gcp/pull/2#pullrequestreview-333515256).
The error logs of prow looks as follows: https://gubernator.knative.dev/build/knative-prow/pr-logs/pull/google_knative-gcp/784/pull-google-knative-gcp-integration-tests/1248021891256422401/

 **Expected behavior**
The test should consistently pass.

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative-GCP release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

grantr:
This is less flaky than it used to be, but still a bit flaky.

capri-xiyue:
@ian-mi All E2E tests related to CAL is still flaky. I investigated the logs, the root cause of that CAL fails to emit a event even though it shows it's ready. Can you help investigate it to see how we can fix it?

nlopezgi:
Closing as this is old. Please open a new bug if seen again

Issue #759, opened 2020-04-02
Workload Identity has delay in receive-adapter


grac3gao:
**Describe the bug**
After Workload Identity reconciled for a source in the data plane (create a k8s Service Account, set iam policy binding as WorkloadIdentityUser to the GoogleServiceAccount with the K8s ServiceAccount, annotate the k8s ServiceAccount), the source will mark ready after topic, pullsubscription and other condition are all marked ready. 

But the receive-adapter will continue with error state for about 1min, error message is `Unable to generate token; IAM returned 403 Forbidden: The caller does not have permission`, which looks like it didn't get that policy binding.

There is a delay for the workload identity functionality working in the receive-adapter pod. During this 1min delay, user can publish message (the publisher doesn't have this issue, the publisher permission is given by the controller?), but they need to wait for 1 min to receive that message. When the receive-adapter is in error state, if you check the GoogleServiceAcount's iam policy, you'll find the role is already bound, and the k8s service account annotation is also there. Thus, it might be the cold start time for the workload identity to put a new member into this project's workload identity pool. If this GSA is already used for other sources in the same namespace, then the newly applied source will not have this problem, as they are using same ksa and bindings.

I was wondering if the source's ready status condition should also check the receive-adapter's state in this case, so that user will not sent message when receive-adapter is in error state? Would like to hear about opinions for how to deal with this.

Also, this makes me think about the initial condition when enabling control plane's Workload Identity. It also needs to wait for 1 min at the beginning, I'll check that.

grac3gao:
https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#deployment-status

Deployments have this condition that we can check to see if minimum replicas are ready:
- lastTransitionTime: 2016-10-04T12:25:42Z
    lastUpdateTime: 2016-10-04T12:25:42Z
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"

grac3gao:
For the control plane, because we have that 1 min delay, once the control plane is enabled WI, user need to wait for 1 min to apply resources. Otherwise, resources will still fail for permission issue. 
If the user apply resources before configuring any credential in control plane, after configuring WI, restarting controller immediately will not lead to reconcile resource successfully. It will still fail for the first 1 min, and because of exponential back off, usually need to wait for 5 min reSync time to make such resource ready. (Restarting controller help resource become ready immediately when using secret)

yolocs:
[...]

PR #1028, opened 2020-05-08
added e2e test of gcp broker with four sources and fixed typo of docs

### LINKED ISSUE #639: E2E tests for GCP Broker with Sources
The existing E2E tests for the pubsub channel-based broker should be copied to use the new broker.

/area broker

capri-xiyue:
Fixes #639 

## Proposed Changes

- added e2e test of gcp broker with four sources
- refactored the duplicate code between gcp broker and broker with pubsub channel
- fixed typo of docs


capri-xiyue:
Considering the size of the code review, I will separate https://github.com/google/knative-gcp/issues/639 with two PRs. One for GCP Broker with four sources and the other for GCP broker with [Broker flow to multiple triggers
](https://github.com/knative/eventing/blob/master/test/e2e/broker_channel_flow_test.go)

test/e2e/e2e_test.go
```diff
@@ -313,3 +313,31 @@ func TestGCPBroker(t *testing.T) {
 	defer cancel()
 	GCPBrokerTestImpl(t, authConfig)
 }
+
+// TestCloudPubSubSourceWithGCPBroker tests we can knock a Knative Service from a GCPBroker from a CloudPubSubSource.
+func TestCloudPubSubSourceWithGCPBroker(t *testing.T) {
+	cancel := logstream.Start(t)
+	defer cancel()
```
ian-mi, line 320: Just `defer logstream.Start(t)`?

test/e2e/test_auditlogs.go
```diff
@@ -48,15 +43,15 @@ func CloudAuditLogsSourceWithTestImpl(t *testing.T, authConfig lib.AuthConfig) {
 	defer lib.TearDown(client)
 
 	// Create a target Job to receive the events.
-	lib.MakeAuditLogsJobOrDie(client, methodName, project, resourceName, serviceName, targetName)
+	lib.MakeAuditLogsJobOrDie(client, lib.MethodName, project, resourceName, lib.ServiceName, targetName)
```
ian-mi, line 46: Should these still be arguments now that that constants have moved?

ian-mi:
I would be reluctant to add an additional audit log test until #842 is fixed.

test/e2e/test_auditlogs.go
capri-xiyue: I think it will be better if we still make these to be arguments so that callers can customize the AuditLogsJobs.

capri-xiyue:
> I would be reluctant to add an additional audit log test until #842 is fixed.

I think we can add this test first and the later fix  [#842](https://github.com/google/knative-gcp/issues/842).  Considering we already enabled E2E for CloudAuditLogs and CloudAduditLogs for Broker with PubSub Channel. I don't think adding this one will increase much more flakiness.  
If we think CloudAuditLogs is really flaky, I think we should increase the priority of [#842](https://github.com/google/knative-gcp/issues/842)

test/e2e/e2e_test.go
capri-xiyue: Does 
`cancel := logstream.Start(t)
 defer cancel()`
equal to 
`defer logstream.Start(t)`?

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
```diff
@@ -0,0 +1,331 @@
...
+	senderJob := resources.SenderJob(senderName, []v1.EnvVar{{
+		Name:  "BROKER_URL",
+		Value: brokerURL.String(),
+	}})
+	client.CreateJobOrFail(senderJob)
+
+	// Check if dummy CloudEvent is sent out.
+	if done := jobDone(client, senderName, t); !done {
+		t.Error("dummy event wasn't sent to broker")
+		t.Failed()
```
liu-cong, line 50: Why call `t.Failed()`? I think it only returns true of false??

test/e2e/e2e_test.go
```diff
@@ -313,3 +313,31 @@ func TestGCPBroker(t *testing.T) {
...
+	AuditLogsSourceBrokerWithGCPBrokerTestImpl(t, authConfig)
+}
+
+// TestCloudSchedulerSourceWithGCPBroker tests we can knock a Knative Service from a GCPBroker from a CloudSchedulerSource.
+func TestCloudSchedulerSourceWithGCPBroker(t *testing.T) {
+	cancel := logstream.Start(t)
+	defer cancel()
+	SchedulerSourceWithGCPBrokerTestImpl(t, authConfig)
+}
```
chizhg, line 343: Add one new line at the end of this file?

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
```diff
@@ -0,0 +1,331 @@
...
+		dummyTriggerName,
+		eventingtestresources.WithBroker(brokerName),
+		eventingtestresources.WithAttributesTriggerFilter(
+			eventingv1alpha1.TriggerAnyFilter, eventingv1alpha1.TriggerAnyFilter,
+			map[string]interface{}{"type": "e2e-testing-dummy"}),
+		eventingtestresources.WithSubscriberServiceRefForTrigger(kserviceName),
+	)
+}
+
+func createTriggerWithTargetServiceSubscriber(client *lib.Client, brokerName, targetName string) {
```
liu-cong, line 271: I don't see the difference between `createTriggerWithTargetServiceSubscriber` and `createTriggerWithKServiceSubscriber`, except for the trigger name

liu-cong: How about `func reateTriggers(client *lib.Client, brokerName string, targets ...string)

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
```diff
@@ -0,0 +1,331 @@
...
+		eventingtestresources.WithAttributesTriggerFilter(
+			eventingv1alpha1.TriggerAnyFilter, eventingv1alpha1.TriggerAnyFilter,
+			map[string]interface{}{"type": "e2e-testing-resp"}),
+		eventingtestresources.WithSubscriberServiceRefForTrigger(targetName),
+	)
+}
+
+func createTriggersAndKService(client *lib.Client, brokerName, targetName string) {
+	// Create the Knative Service.
+	kserviceName := CreateKService(client)
```
liu-cong, line 285: Let's move this out for consistency - the targetName is created outside of this function. Let's make this function only create triggers, not target services

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
```diff
@@ -0,0 +1,331 @@
...
+	// Create the Knative Service.
+	kservice := resources.ReceiverKService(
+		kserviceName, client.Namespace)
+	client.CreateUnstructuredObjOrFail(kservice)
+	return kserviceName
+
+}
+
+func createTriggerWithKServiceSubscriber(client *lib.Client, brokerName, kserviceName string) {
+	dummyTriggerName := "dummy-broker-" + brokerName
```
liu-cong, line 260: This trigger is not "dummy", maybe just call it kserviceTrigger

test/e2e/lib/auditlogs.go
```diff
@@ -25,6 +25,11 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
+const (
```
chizhg, line 28: Are the two constant only used by the auditlogs test? If yes maybe it would be better to move them out of this package, or rename them to be less confusing?

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
```diff
@@ -0,0 +1,331 @@
...
+	var url apis.URL = apis.URL(brokerURL)
+	// Just to make sure all resources are ready.
+	time.Sleep(5 * time.Second)
+
+	// Create the CloudSchedulerSource.
+	lib.MakeSchedulerOrDie(client, sName, data, targetName, authConfig.PubsubServiceAccount,
+		kngcptesting.WithCloudSchedulerSourceSinkURI(&url),
+	)
+
+	msg, err := client.WaitUntilJobDone(client.Namespace, targetName)
```
liu-cong, line 218: use `jobDone`?

test/e2e/lib/auditlogs.go
capri-xiyue: These two constants are used both in CloudAuditLogsSourceWithTestImpl of test/e2e/test_auditlogs.go and BrokerEventTransformationTestWithAuditLogsSourceHelper of test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go. I moved it here to avoid cycle import. I will rename them to be less confusing.

test/e2e/test_auditlogs.go
```diff
@@ -31,11 +31,6 @@ import (
 	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
 )
 
-const (
```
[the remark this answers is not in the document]
capri-xiyue: These two constants are used both in CloudAuditLogsSourceWithTestImpl of test/e2e/test_auditlogs.go and BrokerEventTransformationTestWithAuditLogsSourceHelper of test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go.  I move these two constants to libs because clients use the same args.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
```diff
@@ -0,0 +1,331 @@
+package helpers
+
+import (
```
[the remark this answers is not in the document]
capri-xiyue: I followed the naming convention in eventing. For example: https://github.com/knative/eventing/blob/master/test/e2e/helpers/broker_event_transformation_test_helper.go

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
```diff
@@ -0,0 +1,331 @@
...
+func jobDone(client *lib.Client, podName string, t *testing.T) bool {
+	msg, err := client.WaitUntilJobDone(client.Namespace, podName)
+	if err != nil {
+		t.Error(err)
+		return false
+	}
+	if msg == "" {
+		t.Error("No terminating message from the pod")
+		return false
+	} else {
```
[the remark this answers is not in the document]
capri-xiyue: Why we don't need the `else` here? I'm kind of confused.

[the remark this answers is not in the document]
liu-cong: there is a `return` in `if`

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
capri-xiyue: I refactored the code. This is what we used to use https://github.com/google/knative-gcp/blob/bb7cfe32ebeb07932f0143a1a7a4092dd218195a/test/e2e/test_broker_pubsub.go#L88. I thought when the tests fail, we will use t.Failed to mark `Fail` for this test case.

liu-cong: t.Failed() does not mark the test to fail, it only return the status.

t.Error already marks the test fail.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
capri-xiyue: The extensions are also different. One use map[string]interface{}{"type": "e2e-testing-dummy"}) and the other use map[string]interface{}{"type": "e2e-testing-resp"}). 
Most of the code are already refactored into the function 
`func (client *Client) CreateTriggerOrFail(name string, options ...resources.TriggerOption) *v1alpha1.Trigger `

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
capri-xiyue: This function is used for all GCP Broker with different sources. That's why we move it here.
We already have specific function for creating triggers.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
capri-xiyue: I thought we call it dummy because it doesn't filter anything and it's used for the dummy event. That's the naming convention for both knative-eventing and knative-gcp

docs/examples/cloudschedulersource/README.md
```diff
@@ -17,7 +17,7 @@ scheduled events from
    [`CloudSchedulerSource`](cloudschedulersource.yaml)
 
    ```shell
-   export APP_ENGINE_LOCATION=us-central1
+   export APP_ENGINE_LOCATION=us-central
    gcloud app create --region=$APP_ENGINE_LOCATION
```
bskaplan, line 21: Does app engine accept us-central1 as the region? App Engine calls it us-central.

test/e2e/test_gcp_broker.go
```diff
@@ -51,83 +48,60 @@ Note: the number denotes the sequence of the event that flows in this test case.
 */
 
 func GCPBrokerTestImpl(t *testing.T, authConfig lib.AuthConfig) {
```
bskaplan, line 50: FYI, this may have a small conflict with #1022 which adds the same metrics-checking logic to this test impl as is in the broker_pubsub test.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
liu-cong: we do not need to follow those. I would either rename it or add a comment, the name is really misleading

test/e2e/e2e_test.go
ian-mi: Ah, I misread this.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
liu-cong: Then I don't think we need this function. I would just inline CreateKService and CreateTrigger

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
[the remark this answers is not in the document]
liu-cong: let at least get rid of the `helper` in the file, it's already in `helper` package

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
liu-cong: You can save some duplicated code by doing`func reateTrigger(client *lib.Client, brokerName, triggerName, extensions)`, or just inline client.Core.CreateTriggerOrFail. But up to you.

test/e2e/test_gcp_broker.go
capri-xiyue: Yep. There will be a conflict since this PR did a refactor of the duplicate code between GCP Broker and Broker with PubSub Channel. But I think that should be quite fast to resolve the conflicts.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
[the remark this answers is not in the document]
capri-xiyue: ok

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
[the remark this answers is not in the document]
capri-xiyue: Thanks for the clarification. I missed the `return` above.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
capri-xiyue: `t.Error()` use `c.Fail()` but `t.Failed()` call `c.Failed()`. I think previous we use t.Failed to mark failed reports and stops the execution.
// Fail marks the function as having failed but continues execution.
func (c *common) Fail() {
	if c.parent != nil {
		c.parent.Fail()
	}
	c.mu.Lock()
	defer c.mu.Unlock()
	// c.done needs to be locked to synchronize checks to c.done in parent tests.
	if c.done {
		panic("Fail in goroutine after " + c.name + " has completed")
	}
	c.failed = true
}

// Failed reports whether the function has failed.
func (c *common) Failed() bool {
	c.mu.RLock()
	failed := c.failed
	c.mu.RUnlock()
	return failed || c.raceErrors+race.Errors() > 0
}

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
capri-xiyue: If we don't have this function, all test cases will need to inline these 5 codes. That's why I change it to one function. For the above comments, I did't create anotherfunc createTrigger(client *lib.Client, brokerName, triggerName, extensions) is because this function will end up calling just one line 
`client.Core.CreateTriggerOrFail(
                 triggerName,
		eventingtestresources.WithBroker(brokerName),
		eventingtestresources.WithAttributesTriggerFilter(
			eventingv1alpha1.TriggerAnyFilter, eventingv1alpha1.TriggerAnyFilter,
			map[string]interface{}{"type": "e2e-testing-resp"}),
		eventingtestresources.WithSubscriberServiceRefForTrigger(targetName),
	)`

docs/examples/cloudschedulersource/README.md
capri-xiyue: No, app engine doesn't accept us-central1 as the region.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
capri-xiyue: I will rename it.

test/e2e/test_gcp_broker.go
capri-xiyue: I solved the merge conflicts. But you may need to add more metrics-checking for the new E2E tests added in this PR once this PR got merged.

test/e2e/lib/helpers/broker_event_transformation_test_with_source_helper.go
liu-cong: My main point about this one is it's unclear to me what it does but reading its name. I feel better if the KService is created outside of this function, and this is changed to `createTriggersForKserviceAndTargetService`...

But I would'd say this is a blocker.

capri-xiyue: I will change the naming  to `createTriggersForKserviceAndTargetService`

capri-xiyue:
The wi tests are quite flaky. I tested in my local with single test case, it is not that flaky.  It looks like running all E2E tests together increase the flakiness . But when it fails because of the error of cre-pull instead of anything related to gcp-broker:
"error":"unable to create sunscription \"cre-pull-99b0ac99-9d81-44a9-b340-236b4c917490\", rpc error: code = Unauthenticated desc = transport: compute: Received 403 \nUnable to generate token; IAM returned 403 Forbidden: The caller does not have permission\n\nThis error could be caused by a missing IAM policy binding on the target IAM service account.\n\nYou can create the necessary policy binding with:\n\n gcloud iam service-accounts add-iam-policy-binding \\\n --role=roles/iam.workloadIdentityUser \\\n --member=\"serviceAccount:xiyue-opensource-project-wi.svc.id.goog[test-cloud-storage-source-with-g-c-p-broker-s7xxf/cre-pubsub]\" \\\n cre-pubsub@xiyue-opensource-project-wi.iam.gserviceaccount.com\n\nFor more information, refer to the Workload Identity documentation:\n\n https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity\n\n","stacktrace":"main.main\n\tgithub.com/google/knative-gcp/cmd/pubsub/receive_adapter/main.go:93\nruntime.main\n\truntime/proc.go:203"}

capri-xiyue:
This pull-google-knative-gcp-integration-tests is also flaky. It's flaky mainly because of CloudAuditLogsSource and CloudStorageSource.

grac3gao:
Checked with failure log, looks like integration test and wi test are all becoming more flaky. We have flaky tests: AuditLogSource, AuditLogSourceWithPubSubChannel, StorageSourceWithPubSubChannel. As GCP broker shares the same function as PubSubChannel Broker, we have two more flaky tests: AuditLogSourceWithGCPBroker, StorageSourceWithGCPBroker for each e2e tests now, this generally increases both e2e tests flakiness. 

A known issue for Workload Identity is WI has delay in receive-adapter: https://github.com/google/knative-gcp/issues/759. According to the error message, it seems that the receive-adapter is still in that delay recovering period, and the test is timeout. This further increases the flakiness in wi tests.

PR #1045, opened 2020-05-11
[master] golang format tools

mattmoor:
Produced via:
  `gofmt -s -w $(find -path './vendor' -prune -o -path './third_party' -prune -o -type f -name '*.go' -print)`
  `goimports -w $(find -name '*.go' | grep -v vendor | grep -v third_party)`
/assign grantr nachocano
/cc grantr nachocano


grantr:
I wonder if a tool exists that can filter files by the directives in .gitattributes?

ian-mi:
I think the easiest solution would be to modify matt's script to prune generated go files.

mattmoor:
Generally I exclude directories, but I haven’t had to for any other code generators.  Have you considered running gofmt on the generator output?  Pretty sure there K8s code generators all do this.

FYI I generally ignore the github notifications for these, so if you want a change just DM me in slack (I caught this by chance).

ian-mi:
wire does gofmt itself but does not call goimports. It's not easily possible to add a call to goimports since the logic is hidden in golang.org/x/tools/internal/imports. I think it'd be preferable to prune wire files from goimports rather than reformat them so that the generated code doesn't diverge from the output of go generate.

ian-mi:
I've revised the goimports command to skip wire_gen.go files:
```bash
find -name '*.go' \! -name wire_gen.go -exec goimports -w {} \+ -o \( -path ./vendor -o -path ./third_party \) -prune
```

Harwayne:
> Generally I exclude directories, but I haven’t had to for any other code generators. Have you considered running gofmt on the generator output? Pretty sure there K8s code generators all do this.
> 
> FYI I generally ignore the github notifications for these, so if you want a change just DM me in slack (I caught this by chance).

#1120 runs the wire outputs through `goimports`.

ian-mi:
The build issue appears to be due to the v1 package name in the pkg/reconciler/helper.go. I've created #1122 to workaround this, but I've noticed that my version of goimports does not get confused the by the v1 package. Maybe the goimports version used to generate this needs to be updated?

Harwayne:
Should we just close this and see what the bot creates tomorrow morning?

nachocano:
Yes, closing

/close

PR #1062, opened 2020-05-13
Upgrade Knative Dependency Versions

ian-mi:
Pulls in new broker tracing helpers.


pkg/broker/ingress/handler.go
```diff
@@ -24,16 +24,16 @@ import (
 	"strings"
 	"time"
 
-	"go.uber.org/zap"
+	_ "knative.dev/pkg/metrics/testing"
```
liu-cong, line 27: Why do we need this?

ian-mi: This is necessary to set a test metrics domain which is now required by metrics.Record: https://github.com/google/knative-gcp/pull/1062/files#diff-e0bb1c99fd51561af061d576f06f39b0R157-R169

liu-cong: I see. Maybe just add a comment here as having a "testing" package seems a bit odd.

ian-mi: Unfortunately this is required everywhere metrics are used, not just this test. I'd prefer not to add comments since the _ already indicates that it's for initialization purposes and the tests print a helpful error message if this is omitted.

PR #1064, opened 2020-05-13
Broker tracing conventions

### LINKED ISSUE #1063: GCP Broker Follows Eventing Conventions for Spans and Attributes.
**Problem**
https://github.com/knative/eventing/pull/2983 defines tracing conventions for brokers that the GCP broker should implement.

Event Consumer

**Exit Criteria**
Manual verification that conforming spans are emitted.

**Time Estimate (optional):** 
3

ian-mi:
Fixes #1063

## Proposed Changes

- Emit Broker Ingress and Trigger Spans.
- Add the messaging.system, messaging.protocol, messaging.destination, and messaging.id attributes as specified in https://github.com/knative/eventing/blob/master/docs/spec/broker.md.

**Release Note**

```release-note
- 🎁 The GCP Broker now emits broker and trigger trace spans named broker:<name>.<namespace> and trigger:<name>.<namespace> respectively.
```

### DIFF
--- a/pkg/broker/handler/processors/filter/processor.go
+++ b/pkg/broker/handler/processors/filter/processor.go
@@ -61,12 +61,7 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
 		Namespace: target.Namespace,
 		Name:      target.Name,
 	}
-	var span *trace.Span
-	if dt, ok := extensions.GetDistributedTracingExtension(*event); ok {
-		ctx, span = dt.StartChildSpan(ctx, kntracing.TriggerMessagingDestination(trigger))
-	} else {
-		ctx, span = trace.StartSpan(ctx, kntracing.TriggerMessagingDestination(trigger))
-	}
+	ctx, span := startSpan(ctx, trigger, event)
 	defer span.End()
 	if span.IsRecordingEvents() {
 		span.AddAttributes(
@@ -86,6 +81,14 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
 	}
 	logging.FromContext(ctx).Debug("event does not pass filter for target", zap.Any("target", target))
 	return nil
+}
+
+func startSpan(ctx context.Context, trigger types.NamespacedName, event *event.Event) (context.Context, *trace.Span) {
+	if dt, ok := extensions.GetDistributedTracingExtension(*event); ok {
+		return dt.StartChildSpan(ctx, kntracing.TriggerMessagingDestination(trigger))
+	} else {
+		return trace.StartSpan(ctx, kntracing.TriggerMessagingDestination(trigger))
+	}
 }
 
 func (p *Processor) passFilter(ctx context.Context, attrs map[string]string, event *event.Event) bool {
--- a/pkg/broker/handler/processors/filter/processor.go
+++ b/pkg/broker/handler/processors/filter/processor.go
@@ -18,17 +18,19 @@ package filter
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/cloudevents/sdk-go/v2/event"
 	"github.com/cloudevents/sdk-go/v2/extensions"
 	"go.opencensus.io/trace"
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/eventing/pkg/logging"
+	kntracing "knative.dev/eventing/pkg/tracing"
 
 	"github.com/google/knative-gcp/pkg/broker/config"
 	handlerctx "github.com/google/knative-gcp/pkg/broker/handler/context"
 	"github.com/google/knative-gcp/pkg/broker/handler/processors"
+	"github.com/google/knative-gcp/pkg/tracing"
 )
 
 // Processor is the processor to filter events based on trigger filters.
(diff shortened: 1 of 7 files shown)


ian-mi:
Manually verified to produce the expected spans in Stackdriver.

pkg/broker/ingress/handler.go
```diff
@@ -24,18 +24,22 @@ import (
...
+	"github.com/google/knative-gcp/pkg/tracing"
 	"github.com/google/wire"
+	"go.opencensus.io/trace"
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/eventing/pkg/kncloudevents"
 	"knative.dev/eventing/pkg/logging"
+	kntracing "knative.dev/eventing/pkg/tracing"
+
+	_ "knative.dev/pkg/metrics/testing"
```
Harwayne, line 42: Why is a testing package being used in production code?

I realize you just moved this line, but can it be removed instead?

pkg/broker/ingress/handler.go
```diff
@@ -127,15 +136,26 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
 
 	event.SetExtension(EventArrivalTime, cev2.Timestamp{Time: time.Now()})
 
+	ctx, span := trace.StartSpan(ctx, kntracing.BrokerMessagingDestination(broker))
```
Harwayne, line 139: Should this check for an existing distributed trace? For example, if something has been replied back into the Broker. Or if the source that sent this already had started a trace.

pkg/broker/ingress/handler.go
```diff
@@ -127,15 +136,26 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
...
+	defer span.End()
+	if span.IsRecordingEvents() {
+		span.AddAttributes(
+			kntracing.MessagingSystemAttribute,
+			tracing.PubSubProtocolAttribute,
+			kntracing.BrokerMessagingDestinationAttribute(broker),
```
Harwayne, line 145: What does this attribute represent?

For example, if this attribute is on a span, how can I interpret that span?

I ask because it looks like in the ingress we are annotating the span that sends the request to GCP PubSub with the broker as the destination. While in the filtering code I think the annotation states the trigger name while sending to the trigger's destination.

pkg/broker/handler/processors/filter/processor.go
```diff
@@ -55,10 +57,24 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
...
+		ctx, span = dt.StartChildSpan(ctx, kntracing.TriggerMessagingDestination(trigger))
+	} else {
+		ctx, span = trace.StartSpan(ctx, kntracing.TriggerMessagingDestination(trigger))
+	}
+	defer span.End()
+	if span.IsRecordingEvents() {
+		span.AddAttributes(
+			kntracing.MessagingSystemAttribute,
+			tracing.PubSubProtocolAttribute,
+			kntracing.TriggerMessagingDestinationAttribute(trigger),
```
Harwayne, line 75: Should this include any form of type information? As-is, I can't tell the difference between different objects that have the same name and namespace (e.g. K8s service vs Broker vs KafkaChannel).

pkg/broker/ingress/handler.go
ian-mi: The existing trace context should always be set in the HTTP tracecontext so there is no need to check for the extension.

pkg/broker/ingress/handler.go
ian-mi: This attribute considers each broker or trigger to be an event destination. On ingress a single broker span will be created and on fanout multiple trigger spans will be created identifiable by the messaging.destination attribute as well as the span name.

pkg/broker/handler/processors/filter/processor.go
ian-mi: messaging.destination is prefixed by either broker or trigger and will contain the namespaced name.

pkg/broker/handler/processors/filter/processor.go
```diff
@@ -55,10 +57,24 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
 		return nil
 	}
 
+	trigger := types.NamespacedName{
+		Namespace: target.Namespace,
+		Name:      target.Name,
+	}
+	var span *trace.Span
```
yolocs, line 64: nit: might as well wrap this into a local func so that here it will look like `ctx, span := startSpan(ctx, target, event); defer span.End()`

pkg/broker/handler/processors/filter/processor.go
```diff
@@ -55,10 +57,24 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
 		return nil
 	}
 
+	trigger := types.NamespacedName{
```
yolocs, line 60: Do we need tracing on forwarding replied events?

ian-mi: Event replies should propagate trace context over HTTP to the broker ingress, creating new broker and trigger spans.

pkg/broker/handler/processors/filter/processor.go
```diff
@@ -55,10 +57,19 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
...
+	ctx, span := startSpan(ctx, trigger, event)
+	defer span.End()
+	if span.IsRecordingEvents() {
+		span.AddAttributes(
+			kntracing.MessagingSystemAttribute,
+			tracing.PubSubProtocolAttribute,
+			kntracing.TriggerMessagingDestinationAttribute(trigger),
+			kntracing.MessagingMessageIDAttribute(event.ID()),
+		)
 	}
```
yolocs, line 73: nit: can this be wrapped into `startSpan` func?

PR #1066, opened 2020-05-13
Reconcilers delete old Topics and PullSubscriptions

### LINKED ISSUE #905: Switch API Groups - pubsub.cloud.google.com -> internal.events.cloud.google.com

Harwayne:
Fixes #905.

## Proposed Changes

- Reconcilers delete owned and controlled `Topic` and `PullSubscription` in `pubsub.cloud.google.com`.
  - Topic
      - First have their PropagationPolicy updated to `NoDelete`, so that the Topic is not deleted in GCP.
      - Deleted _before_ the creation of the new Topic. This is because the old and new Topics both use the same `Service` name, so we will never have both be `Ready=true` at the same time.
          - This will lead to potential lost events for Channels during upgrade while the new Topic's reconciler tries to create the Service.
  - PullSubscription
    - Deleted only after the new PullSubscription is `Ready=true`.
        - This means there will likely be double event delivery for a period of time.
        - The old PullSubscription is not drained, so anything in it that was not delivered will be lost.

**Release Note**

```release-note
`Source`s and `Channel`s will delete deprecated `Topic`s and `PullSubscription`s. `Channel`s may become unavailable during the upgrade, but should quickly reconcile back to being healthy. `Source`s will continue sending events and may for a short period of time send the same events multiple times. Any old events in the `PullSubscription` will be lost (e.g. an event from four hours ago that has not yet been acknowledge by the Source's sink).
```

**Docs**


Harwayne:
This is ready for review.

pkg/reconciler/intevents/reconciler.go
```diff
@@ -175,15 +189,101 @@ func (psb *PubSubBase) ReconcilePullSubscription(ctx context.Context, pubsubable
 		}
 	}
 
+	// TODO Compare the actual PullSubscription we found with the desired one and update as needed.
```
nachocano, line 192: is this TODO still relevant?

pkg/reconciler/intevents/reconciler.go
```diff
@@ -175,15 +189,101 @@ func (psb *PubSubBase) ReconcilePullSubscription(ctx context.Context, pubsubable
...
+		err = psb.deleteOldPubSubPullSubscriptionCO(ctx, pubsubable, ps)
+		if err != nil {
+			return ps, err
+		}
+	}
+
 	return ps, nil
 }
 
+func (psb *PubSubBase) deleteOldPubSubTopicCO(_ context.Context, pubsubable duck.PubSubable, t *inteventsv1alpha1.Topic) pkgreconciler.Event {
```
nachocano, line 214: nit: need the CO at the end? same for the pullSubscription deletion

nachocano: something that has been useful for me in this deprecation stuff has been to add TODOs saying remove after 0.x is cut, or something like that... not strong opinion, but might be easier for others to know what to remove

pkg/duck/v1alpha1/identifiable.go
```diff
@@ -17,13 +17,17 @@ limitations under the License.
...
+	"k8s.io/apimachinery/pkg/runtime"
 	"knative.dev/pkg/apis"
 	"knative.dev/pkg/kmeta"
 
 	duckv1alpha1 "github.com/google/knative-gcp/pkg/apis/duck/v1alpha1"
 )
 
 type Identifiable interface {
+	// runtime.Object can be removed once the old Topic and PullSubscription are removed.
+	runtime.Object
```
nachocano, line 29: why is this one needed?

Harwayne: The event recorder requires a `runtime.Object`. In the `deleteOldPubSub*CO` methods, I write an event if the old CO is not owned by the pubsubable.

pkg/reconciler/intevents/reconciler.go
Harwayne: Nope :)

Bad merge with your change that just added this.

pkg/reconciler/intevents/reconciler.go
Harwayne: Done.

Harwayne:
/unhold

Comments addressed.

Issue #1039, opened 2020-05-09
Enable tracing when running E2E tests


capri-xiyue:
**Problem**
Currently it's quite difficult to debug with flaky E2E tests like TestCloudAuditLogs and TestCloudStorageSourceWithPubSubChannel.

Usually, the error msg is just the resp event didn't hit the target. But we don't know in which procedure the event gets stuck which results in the test failure.

If we can enable tracing when running E2E tests, it will improve the development experience with E2E tests.

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

capri-xiyue:
@ian-mi  FYI.

PR #1069, opened 2020-05-14
Print metrics from all Pods in the namespace if the test fails.

Harwayne:
Easier, not as effective, alternative to #1039.

## Proposed Changes

- Print metrics from all Pods in the namespace if the test fails.

**Release Note**

```release-note
NONE
```

---

### How this is useful

To use this, grep through the build log for `Metrics logs`. The sources tend to have very good representation of all the events that went through them. The log line includes the 'root owner' of the Pod. So rather than say `Metrics logs for Pod xyz`, it will say `Metrics logs for root "CloudAuditLogsSource/xyz`. This is useful for things like sources because we normally only care about the root owner, not all the intermediaries (e.g. for most Sources: `Pod -> ReplicaSet -> Deployment -> PullSubscription -> Source`). This tends to not be useful for Broker pieces as all those Pods are rooted by the Broker itself, so it is more work to figure out if this is for the ingress, filter, etc.

A few examples from this PR:

#### Example 1

> pull-google-knative-gcp-integration-tests	[659cd07](https://github.com/google/knative-gcp/commit/659cd07f37963ff9af43bc7be6a47252bc5046b6)	[link](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/1069/pull-google-knative-gcp-integration-tests/1261086850609582082)

This failure printed the metrics. In particular:

```
TestCloudAuditLogsSourceBrokerWithPubSubChannel: metrics.go:130: Metrics logs for root "CloudAuditLogsSource/auditlogs-e2e-test-fvanszag": <Empty Metrics>
```

The metrics were empty, so the CloudAuditLogsSource never received an event.

#### Example 2

> pull-google-knative-gcp-integration-tests	[231229b](https://github.com/google/knative-gcp/commit/231229bf1d6031bf3711018362d32ae8da8264b8)	[link](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/1069/pull-google-knative-gcp-integration-tests/1261361030118772736)	`/test pull-google-knative-gcp-integration-tests`

From the logs:
```
TestCloudStorageSourceBrokerWithPubSubChannel: metrics.go:135: Metrics logs for root "CloudStorageSource/storage-e2e-test-knative-boskos-08-storage-prslonku": # HELP source_event_count Number of events sent
        # TYPE source_event_count counter
        source_event_count{event_source="//storage.googleapis.com/buckets/storage-e2e-test-knative-boskos-08",event_type="com.google.cloud.storage.object.delete",name="storage-e2e-test-knative-boskos-08-storage-prslonku",namespace_name="test-cloud-storage-source-broker-with-pub-sub-channel-v5zld",resource_group="cloudstoragesources.events.cloud.google.com",response_code="500",response_code_class="5xx"} 2
        source_event_count{event_source="//storage.googleapis.com/buckets/storage-e2e-test-knative-boskos-08",event_type="com.google.cloud.storage.object.finalize",name="storage-e2e-test-knative-boskos-08-storage-prslonku",namespace_name="test-cloud-storage-source-broker-with-pub-sub-channel-v5zld",resource_group="cloudstoragesources.events.cloud.google.com",response_code="500",response_code_class="5xx"} 2
```

So the downstream sink rejected the event from the source. The source tried to send the event, but it got a 500 response. With this information, we can easily look at the logs for the Source pod for more information.

#### Example 3

If we see the following in the logs:

```
    broker_event_transformation_test_with_source.go:451: Metrics logs for root "CloudStorageSource/storage-e2e-test-harwayne2-storage-nqdmdrnl": # HELP source_event_count Number of events sent
        # TYPE source_event_count counter
        source_event_count{event_source="//storage.googleapis.com/buckets/storage-e2e-test-harwayne2",event_type="com.google.cloud.storage.object.delete",name="storage-e2e-test-harwayne2-storage-nqdmdrnl",namespace_name="test-cloud-storage-source-broker-with-pub-sub-channel-95f6f",resource_group="cloudstoragesources.events.cloud.google.com",response_code="202",response_code_class="2xx"} 1
        source_event_count{event_source="//storage.googleapis.com/buckets/storage-e2e-test-harwayne2",event_type="com.google.cloud.storage.object.finalize",name="storage-e2e-test-harwayne2-storage-nqdmdrnl",namespace_name="test-cloud-storage-source-broker-with-pub-sub-channel-95f6f",resource_group="cloudstoragesources.events.cloud.google.com",response_code="202",response_code_class="2xx"} 1
```

Then we know that the source sent two events, one of each type, to the downstream sink and that that sink accepted the event with a 202. Therefore any problems that caused the test to fail must be after the source.


Harwayne:
> pull-google-knative-gcp-integration-tests	[659cd07](https://github.com/google/knative-gcp/commit/659cd07f37963ff9af43bc7be6a47252bc5046b6)	[link](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/1069/pull-google-knative-gcp-integration-tests/1261086850609582082)

This failure printed the metrics. In particular:

```
TestCloudAuditLogsSourceBrokerWithPubSubChannel: metrics.go:130: Metrics logs for root "CloudAuditLogsSource/auditlogs-e2e-test-fvanszag": 
```

The metrics were empty, so the CloudAuditLogsSource never received an event.

In contrast, this is what the Storage source shows when it saw two events:

```
source_event_count{event_source="//storage.googleapis.com/buckets/storage-e2e-test-harwayne2",event_type="com.google.cloud.storage.object.delete",name="storage-e2e-test-harwayne2-storage-nqdmdrnl",namespace_name="test-cloud-storage-source-broker-with-pub-sub-channel-95f6f",resource_group="cloudstoragesources.events.cloud.google.com",response_code="202",response_code_class="2xx"} 1
source_event_count{event_source="//storage.googleapis.com/buckets/storage-e2e-test-harwayne2",event_type="com.google.cloud.storage.object.finalize",name="storage-e2e-test-harwayne2-storage-nqdmdrnl",namespace_name="test-cloud-storage-source-broker-with-pub-sub-channel-95f6f",resource_group="cloudstoragesources.events.cloud.google.com",response_code="202",response_code_class="2xx"} 1
```

Harwayne:
> pull-google-knative-gcp-integration-tests	[231229b](https://github.com/google/knative-gcp/commit/231229bf1d6031bf3711018362d32ae8da8264b8)	[link](https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/1069/pull-google-knative-gcp-integration-tests/1261361030118772736)	`/test pull-google-knative-gcp-integration-tests`

This one had a more interesting failure:

```
TestCloudStorageSourceBrokerWithPubSubChannel: metrics.go:135: Metrics logs for root "CloudStorageSource/storage-e2e-test-knative-boskos-08-storage-prslonku": # HELP source_event_count Number of events sent
        # TYPE source_event_count counter
        source_event_count{event_source="//storage.googleapis.com/buckets/storage-e2e-test-knative-boskos-08",event_type="com.google.cloud.storage.object.delete",name="storage-e2e-test-knative-boskos-08-storage-prslonku",namespace_name="test-cloud-storage-source-broker-with-pub-sub-channel-v5zld",resource_group="cloudstoragesources.events.cloud.google.com",response_code="500",response_code_class="5xx"} 2
        source_event_count{event_source="//storage.googleapis.com/buckets/storage-e2e-test-knative-boskos-08",event_type="com.google.cloud.storage.object.finalize",name="storage-e2e-test-knative-boskos-08-storage-prslonku",namespace_name="test-cloud-storage-source-broker-with-pub-sub-channel-v5zld",resource_group="cloudstoragesources.events.cloud.google.com",response_code="500",response_code_class="5xx"} 2
```

So the downstream channel rejected the event from the source.

capri-xiyue:
Thanks @Harwayne for giving me a demo of how this feature works. I think this feature will be very helpful for us to figure out the flakiness of the pipeline.
/lgtm
/hold just in case @chi has any other comments

test/e2e/lib/metrics.go
```diff
@@ -0,0 +1,201 @@
...
+	for _, c := range pod.Spec.Containers {
+		for _, p := range c.Ports {
+			if p.Name == "metrics" {
+				metricsPort = int(p.ContainerPort)
+				break
+			}
+		}
+	}
+	if metricsPort < 0 {
+		client.T.Logf("Pod '%v' does not have a metrics port", podName)
```
chizhg, line 75: `%q` for string instead of `'%v'`, the same for other places as well.

test/e2e/lib/metrics.go
```diff
@@ -0,0 +1,201 @@
...
+		Items: []corev1.Pod{
+			pod,
+		},
+	}
+	localPort, err := findAvailablePort()
+	if err != nil {
+		client.T.Logf("Unable to find an avaiable port for Pod '%v': %v", podName, err)
+		return
+	}
+	// There is almost certainly a better way to do this, but for now, just use kubectl to port
```
chizhg, line 95: What would be a better way to do this? Can we add an issue for further tracking? I would be interested in learning more :)

test/e2e/lib/metrics.go
```diff
@@ -0,0 +1,201 @@
...
+	if resp.ContentLength == 0 {
+		client.T.Logf("Pod had no metrics reported '%v' (root %q)", podName, root)
+		return
+	}
+	b, err := ioutil.ReadAll(resp.Body)
+	if err != nil {
+		client.T.Logf("Unable to read HTTP response body for root %q: %v", root, err)
+		return
+	}
+	client.T.Logf("Metrics logs for root %q: %s", root, string(b))
```
chizhg, line 135: I feel it not quite easy to find these logs as `build-log.txt` is usually a big file.
You can add a new regex in Prow [Spyglass config](https://github.com/knative/test-infra/blob/7a28f47cdbcb11b92359528689b57030fafc127d/config/prod/prow/core/config.yaml#L62-L84) and use it to highlight the metric logs here. But it's up to you.

chizhg:
/lgtm

Feel free to unhold if you want to leave my comments to followup PRs.

test/e2e/lib/metrics.go
Harwayne: I don't know either :P 

It just seems like there _should_ be an easier way to do this. If nothing else, a programmatic call to a library that does this, rather than spawning a new process would be preferred. As I was forcing this to fail locally to make sure things worked, I accidentally orphaned a few of these port forward processes. If it was an in-process library call that wouldn't be possible. It also would allow me to remove the polling loop waiting for the port forwarding to actually start working. And I would know if something went wrong (as opposed to today where I never read the output of the executed kubectl call).

Here is a blog post that has a promising title (haven't actually read it yet): https://gianarb.it/blog/programmatically-kube-port-forward-in-go

Filed https://github.com/google/knative-gcp/issues/1083. This should also be done in knative/pkg, should I make an issue there too?

test/e2e/lib/metrics.go
Harwayne: Done.

test/e2e/lib/metrics.go
Harwayne: https://github.com/knative/test-infra/pull/2082

I'll leave it as a draft PR until this one merges.

Harwayne:
/unhold

All comments addressed.

PR #1070, opened 2020-05-14
Add some missing pieces in docs and some improvement

liu-cong:
Fixes #

## Proposed Changes

- Fix format issue in Broker yaml
- Add `GOOGLE_APPLICATION_CREDENTIALS` for running tests locally. Without this, some tests are not authorized to call pubsub.
- Removed duplicated commands for running with `WI`

**Release Note**

```release-note
NONE
```

**Docs**


test/e2e/README.md
```diff
@@ -30,26 +30,24 @@ knative-gcp should be added under [knative-gcp e2e test lib](lib).
...
-1.  [Pub/Sub Enabled Service Account](../../docs/install/pubsub-service-account.md)
-    installed.
-1.  [GCP Broker Deployment](../../docs/install/install-gcp-broker.md#deployment)
-    and
-    [GCP Broker Authentication Setup](../../docs/install/install-gcp-broker.md#authentication-setup-for-gcp-broker).
+1.  Create a [Pub/Sub Enabled Service Account](../../docs/install/pubsub-service-account.md).
+    Download a credential file and set `GOOGLE_APPLICATION_CREDENTIALS` env var. This is used by
+    some tests(e.g., `TestSmokePullSubscription`) to authorize the Google SDK clients.
+    ```
+    gcloud iam service-accounts keys create cre-pubsub.json \
```
capri-xiyue, line 46: Do we have to set the `GOOGLE_APPLICATION_CREDENTIALS` after we create a Pun/Sub Enabled Service Account?

test/e2e/README.md
```diff
@@ -30,26 +30,24 @@ knative-gcp should be added under [knative-gcp e2e test lib](lib).
...
+    --iam-account=cre-pubsub@$PROJECT_ID.iam.gserviceaccount.com
+    export GOOGLE_APPLICATION_CREDENTIALS=./cre-pubsub.json
+    ```
+1.  [Install GCP Broker](../../docs/install/install-gcp-broker.md#deployment).
```
capri-xiyue, line 50: Is it a little bit confusing here if we don't say it clearly that only deployment and authentication set up are needed?
People may forget to do the authentication set up since you only link to the `deployment` session or people may think they have to create a GCP Broker by themselves.

test/e2e/README.md
liu-cong: Yes, `create a Pun/Sub Enabled Service Account` only asks you to create the GSA, but not use it in local test (by setting the env var)

test/e2e/README.md
liu-cong: As part of the deployment, we ask user to set up authentication. It's now the first step in the deployment guide.

test/e2e/README.md
capri-xiyue: I never set `GOOGLE_APPLICATION_CREDENTIALS`.
I just create the Pub/Sub Enabled Service Account but it works for me.

test/e2e/README.md
capri-xiyue: In this case, why the link is ../../docs/install/install-gcp-broker.md#deployment instead of https://github.com/liu-cong/knative-gcp/blob/docs-new/docs/install/install-gcp-broker.md?

liu-cong: oops, updated link

test/e2e/README.md
liu-cong: Can you check you `GOOGLE_APPLICATION_CREDENTIALS ` and run `go test --tags=e2e ./test/e2e/... -run TestSmokePullSubscription -v`

capri-xiyue: When I ran `echo $GOOGLE_APPLICATION_CREDENTIAL`, it shows empty. But when I ran `go test --tags=e2e ./test/e2e/... -run TestSmokePullSubscription -v`, it passes.

capri-xiyue:
/lgtm.
/approve
Though we have some open questions regarding whether we need to set up `GOOGLE_APPLICATION_CREDENTIALS `. But it looks like it depends on whether we run it in Mac or Linux. That's not a blocker for this PR. We can figure that out later.

liu-cong:
/hold

Let's figure out why @capri-xiyue can run without the `GOOGLE_APPLICATION_CREDENTIALS` env var. Or maybe have someone else to try if they can run without the env var.

test/e2e/README.md
capri-xiyue: After I fetched the latest code of master branch, I need to provide  `GOOGLE_APPLICATION_CREDENTIALS`, otherwise it will fail

test/e2e/README.md
```diff
@@ -30,26 +30,24 @@ knative-gcp should be added under [knative-gcp e2e test lib](lib).
...
-1.  [GCP Broker Deployment](../../docs/install/install-gcp-broker.md#deployment)
-    and
-    [GCP Broker Authentication Setup](../../docs/install/install-gcp-broker.md#authentication-setup-for-gcp-broker).
+1.  Create a [Pub/Sub Enabled Service Account](../../docs/install/pubsub-service-account.md).
+    Download a credential file and set `GOOGLE_APPLICATION_CREDENTIALS` env var. This is used by
+    some tests(e.g., `TestSmokePullSubscription`) to authorize the Google SDK clients.
+    ```
+    gcloud iam service-accounts keys create cre-pubsub.json \
+    --iam-account=cre-pubsub@$PROJECT_ID.iam.gserviceaccount.com
+    export GOOGLE_APPLICATION_CREDENTIALS=./cre-pubsub.json
```
capri-xiyue, line 48: It looks like it should be the absolute path instead of relative path

test/e2e/README.md
liu-cong: Updated to absolute path

PR #1079, opened 2020-05-15
Add Delivery Processor Benchmarks

### LINKED ISSUE #1078: Add Delivery Processor Benchmark
**Problem**
The fanout processor should have a microbenchmark measuring delivery performance.

ian-mi:
Fixes #1078

### DIFF
--- a/pkg/broker/handler/processors/deliver/processor_test.go
+++ b/pkg/broker/handler/processors/deliver/processor_test.go
@@ -59,13 +59,7 @@ func TestInvalidContext(t *testing.T) {
 }
 
 func TestDeliverSuccess(t *testing.T) {
-	sampleEvent := event.New()
-	sampleEvent.SetID("id")
-	sampleEvent.SetSource("source")
-	sampleEvent.SetSubject("subject")
-	sampleEvent.SetType("type")
-	sampleEvent.SetTime(time.Now())
-
+	sampleEvent := newSampleEvent()
 	sampleReply := sampleEvent.Clone()
 	sampleReply.SetID("reply")
 
@@ -77,8 +71,8 @@ func TestDeliverSuccess(t *testing.T) {
 		wantReply  *event.Event
 	}{{
 		name:       "success",
-		origin:     &sampleEvent,
-		wantOrigin: &sampleEvent,
+		origin:     sampleEvent,
+		wantOrigin: sampleEvent,
 		reply:      &sampleReply,
 		wantReply: func() *event.Event {
 			copy := sampleReply.Clone()
@@ -92,7 +86,7 @@ func TestDeliverSuccess(t *testing.T) {
 			eventutil.UpdateRemainingHops(context.Background(), &copy, 1)
 			return &copy
 		}(),
-		wantOrigin: &sampleEvent,
+		wantOrigin: sampleEvent,
 		reply:      &sampleReply,
 	}}
 
@@ -280,12 +274,7 @@ func TestDeliverFailure(t *testing.T) {
 				DeliverTimeout:     500 * time.Millisecond,
 			}
 
-			origin := event.New()
-			origin.SetID("id")
-			origin.SetSource("source")
-			origin.SetSubject("subject")
-			origin.SetType("type")
-			origin.SetTime(time.Now())
+			origin := newSampleEvent()
 
 			rctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 			defer cancel()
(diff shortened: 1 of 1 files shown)


pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -321,6 +322,109 @@ func TestDeliverFailure(t *testing.T) {
...
+	sampleEvent.SetType("type")
+	sampleEvent.SetTime(time.Now())
+
+	deliverClient, err := ceclient.NewDefault()
+	if err != nil {
+		b.Fatalf("failed to create requester cloudevents client: %v", err)
+	}
+	targetSvr := httptest.NewServer(NoReplyHandler(struct{}{}))
+	defer targetSvr.Close()
+	ingressSvr := httptest.NewServer(NoReplyHandler(struct{}{}))
```
liu-cong, line 348: nit: probably don't need this since there is no reply?

pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -321,6 +322,109 @@ func TestDeliverFailure(t *testing.T) {
...
+
+func (NoReplyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	w.WriteHeader(http.StatusAccepted)
+	// Read body to allow the connection to be reused.
+	io.Copy(ioutil.Discard, req.Body)
+	req.Body.Close()
+}
+
+func BenchmarkDeliveryNoReply(b *testing.B) {
+	sampleEvent := event.New()
```
liu-cong, line 335: nit: add a `sampleEvent` helper function that can be reused in tests.

pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -321,6 +322,109 @@ func TestDeliverFailure(t *testing.T) {
...
+		b.Fatalf("failed to create requester cloudevents client: %v", err)
+	}
+	targetSvr := httptest.NewServer(NoReplyHandler(struct{}{}))
+	defer targetSvr.Close()
+	ingressSvr := httptest.NewServer(NoReplyHandler(struct{}{}))
+	defer ingressSvr.Close()
+
+	broker := &config.Broker{Namespace: "ns", Name: "broker"}
+	target := &config.Target{Namespace: "ns", Name: "target", Broker: "broker", Address: targetSvr.URL}
+	testTargets := memory.NewEmptyTargets()
```
liu-cong, line 353: nit: can directly user memory.NewTargets(target)

ian-mi:
I get the following results on my machine:
```
goos: linux
goarch: amd64
pkg: github.com/google/knative-gcp/pkg/broker/handler/processors/deliver
BenchmarkDeliveryNoReply-16      	   10000	    121653 ns/op
BenchmarkDeliveryWithReply-16    	    4434	    247350 ns/op
```

Soon Prow will be able to run these benchmarks on each PR allowing us to detect relative changes in benchmarking results.

pkg/broker/handler/processors/deliver/processor_test.go
ian-mi: This would require constructing the TargetsConfig first.

pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -321,6 +311,94 @@ func TestDeliverFailure(t *testing.T) {
 	}
 }
 
+type NoReplyHandler struct{}
+
+func (NoReplyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
+	w.WriteHeader(http.StatusAccepted)
+	// Read body to allow the connection to be reused.
```
yolocs, line 318: May I get a reference link? Just curious how it works.

ian-mi: Sure, this is documented in the [HTTP Response](https://golang.org/pkg/net/http/#Response) Body field:

> The http Client and Transport guarantee that Body is always
> non-nil, even on responses without a body or responses with
> a zero-length body. It is the caller's responsibility to
> close Body. The default HTTP client's Transport may not
> reuse HTTP/1.x "keep-alive" TCP connections if the Body is
 > not read to completion and closed.

I am currently working on converting HTTP client code to use CE bindings which will give us direct control over this behavior in the deliver code as well as I'm not sure that pubsub transport always handles this properly.

ian-mi: On second look this only applies to response bodies, not request bodies, so this is not actually necessary (but shouldn't hurt).

PR #1080, opened 2020-05-15
Unpin Cloudevents SDK and Upgrade to v2.0.0-rc4

ian-mi:
Requires skipping a Fanout E2E test case due to a race condition. @yolocs could you help take a look at this test issue and determine if it's due to a faulty test or a regression in the SDK?


pkg/broker/handler/pool/fanout_test.go
```diff
@@ -200,6 +200,7 @@ func TestFanoutSyncPoolE2E(t *testing.T) {
 	})
 
 	t.Run("event with delivery timeout was sent to retry queue", func(t *testing.T) {
+		t.SkipNow()
```
Harwayne, line 203: Add a comment linking to an issue.

ian-mi: After some more digging I was able to discover the issue. I added context cancellation to pubsub receive in the SDK (https://github.com/cloudevents/sdk-go/pull/469) which caused goroutines in this test which previously ignored the context timeout and blocked forever to correctly return with an error. This SDK issue was covering up a bug which would have otherwise been detected by this test.

#1087 should fix this issue, I will wait to merge this PR until after the fix is merged.

ian-mi:
/unhold
Ready to review now that #1087 has been merged.

PR #1090, opened 2020-05-19
Give the receiver time to shutdown before stopping the knockdown process.

Harwayne:
## Proposed Changes

- Give the receiver time to shutdown before stopping the knockdown process.
    - Before this change the knockdown job would succeed, but just close its connection to the caller without responding, leading to misleading error logs in tests.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/test/test_images/auditlogs_target/main.go
+++ b/test/test_images/auditlogs_target/main.go
@@ -40,6 +40,8 @@ const (
 	resourceName = "resourceName"
 )
 
+var shutDown = make(chan struct{}, 1)
+
 func main() {
 	client, err := cloudevents.NewDefaultClient()
 	if err != nil {
@@ -63,16 +65,22 @@ func main() {
 	timer := time.NewTimer(time.Second * time.Duration(duration))
 	defer timer.Stop()
 	go func() {
-		<-timer.C
-		// Write the termination message if time out
-		fmt.Printf("time out to wait for event with type %q source %q subject %q service_name %q method_name %q resource_name %q .\n",
-			r.Type, r.Source, r.Subject, r.ServiceName, r.MethodName, r.ResourceName)
-		if err := r.writeTerminationMessage(map[string]interface{}{
-			"success": false,
-		}); err != nil {
-			fmt.Printf("failed to write termination message, %s.\n", err.Error())
+		select {
+		case <-shutDown:
+			// Give the receiver a little time to finish responding.
+			time.Sleep(time.Second)
+			os.Exit(0)
+		case <-timer.C:
+			// Write the termination message if time out
+			fmt.Printf("time out to wait for event with type %q source %q subject %q service_name %q method_name %q resource_name %q .\n",
+				r.Type, r.Source, r.Subject, r.ServiceName, r.MethodName, r.ResourceName)
+			if err := r.writeTerminationMessage(map[string]interface{}{
+				"success": false,
+			}); err != nil {
+				fmt.Printf("failed to write termination message, %s.\n", err.Error())
+			}
+			os.Exit(0)
 		}
-		os.Exit(0)
 	}()
 
 	if err := client.StartReceiver(context.Background(), r.Receive); err != nil {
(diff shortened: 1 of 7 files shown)


test/test_images/auditlogs_target/main.go
```diff
@@ -63,16 +65,22 @@ func main() {
...
-		if err := r.writeTerminationMessage(map[string]interface{}{
-			"success": false,
-		}); err != nil {
-			fmt.Printf("failed to write termination message, %s.\n", err.Error())
+		select {
+		case <-shutDown:
+			// Give the receiver a little time to finish responding.
+			time.Sleep(time.Second)
+			os.Exit(0)
+		case <-timer.C:
```
ian-mi, line 73: Rather than use a timer and shutdown channel, why not just attach a timeout to the receiver context and communicate the result back to the main thread with a channel? The main thread can then cleanup and exit normally without needing to call os.Exit from a goroutine.

Harwayne:
> These seem to be doing very similar things. Should we combine these into a single shared target image?

I had thought about it, but because these are all main packages, they can't import each other and I didn't see a great place to put the test lib. I'll take a quick look and see if I can find a reasonable place.

ian-mi:
Is it feasible to combine these into a single image and configure the behavior through env vars?

test/test_images/auditlogs_target/main.go
Harwayne: Done. 

But this assumes that the server shuts down gracefully when the context is closed. I looked at the code and it seems to today, but I don't think that is guaranteed in the future.

Harwayne:
> Is it feasible to combine these into a single image and configure the behavior through env vars?

A future improvement would be to have the common lib validate attributes. That would cover two of the tests, IIRC. Validating the data is harder, I would want to think about it some more before making the generic framework.

test/test_images/internal/knockdown/knockdown.go
```diff
@@ -0,0 +1,130 @@
...
+//     FooProperty string `envconfig:...`
+// }
+type Config struct {
+	Time time.Duration `envconfig:"TIME" required:"true"`
+}
+
+// cancel will cancel the main receiver's context. Calling this will cause Main() to become
+// unblocked and call os.Exit(0). In general, write{Successful,Failed}TerminationMessage should be
+// called first.
+var cancel func()
```
ian-mi, line 46: might as well add this to receiver?

test/test_images/internal/knockdown/knockdown.go
```diff
@@ -0,0 +1,130 @@
...
+
+	fmt.Printf("Waiting to receive event (timeout in %s)...", config.Time.String())
+
+	// Note, we assume that closing the context will gracefully shutdown the receiver. Having looked
+	// at the current implementation, this is true. But nothing guarantees it in the future.
+	if err := client.StartReceiver(ctx, r.Receive); err != nil {
+		log.Fatal(err)
+	}
+
+	os.Exit(0)
```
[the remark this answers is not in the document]
Harwayne: I agree it is likely not needed. But because the correctness of the tests depends on it happening, I would prefer to keep it here.

test/test_images/internal/knockdown/knockdown.go
Harwayne: Done.

test/test_images/internal/knockdown/knockdown.go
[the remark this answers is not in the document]
ian-mi: My concern with calling os.Exit is that it will skip defers. It is very easy for someone to add a defer to this function and not realize that it won't get called. It's also possible that the caller will want to do some additional cleanup or perform additional checks after Main exits.

ian-mi:
/lgtm
/hold
I've added an explanation of why I don't like calling os.Exit, but it is probably fine to use it this way in test images. Feel free to merge you prefer keeping it.

test/test_images/internal/knockdown/knockdown.go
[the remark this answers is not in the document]
Harwayne: Switched. Although I don't think it is useful in this instance, because this is for fairly simple test code.

Harwayne:
/unhold

Comment addressed.

PR #1094, opened 2020-05-19
Add trigger controller

### LINKED ISSUE #1053: Allow triggers to be reconciled without a broker
**Problem**
While we are still investigating #828, we can go ahead an implement a Trigger controller to reconciler triggers even without a Broker. 

Reasons for doing this:
* It's really a bad UX not being able to delete triggers. We should fix this.
* It seems that regardless what we choose to do for #828, we need the ability to reconcile Triggers without a Broker.
* While we are still investigating #828, a few people think it's expected to lose events if a broker is deleted, which is the same behavior as we have today. 

Developer

**Exit Criteria**
Triggers can be reconciled regardless of existence of their Brokers.
Manual test
E2e Test?

**Time Estimate (optional):** 

* Create a new Trigger controller (can reuse most of the existing Trigger reconciliation logic in Broker controller)
* We don't have to delete the retry topic and pull subscription if a Trigger's Broker doesn't exist. They don't cost money.

liu-cong:
Fixes #1053

I tried to reuse existing code as much as I can to reduce the amount of changes. I will have follow up PRs to do some cleanup and refactoring.

## Proposed Changes

- Add a separate trigger controller that reconciles the retry topic/pull subscription for Triggers
- The controller reconciles/finalizes triggers if their brokers have the `googlecloud` brokerclass, or they have the `googlecloud` finalizer.
- Changes to `targets-config` reconciliation logic
  - The controller deletes broker entry if a broker is deleted, previously the entry still exists, which creates garbage
  - Previously the broker entry and each trigger entry is individually updated, now every time a broker is reconciled, the entire broker entry is reconstructed. This has the following benefits:
     - The trigger controller doesn't need to manipulate trigger entries in the configmap
     - It's more reliable to always reconstruct compared to incremental edits

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/broker/broker.go
```diff
@@ -193,6 +173,29 @@ func (r *Reconciler) reconcileBroker(ctx context.Context, b *brokerv1beta1.Broke
...
+	r.flagTargetsForUpdate()
+	return nil
+}
+
+// reconcileConfig reconstructs the data entry for the given broker in targets-config.
+func (r *Reconciler) reconcileConfig(ctx context.Context, b *brokerv1beta1.Broker, triggers []*brokerv1beta1.Trigger) {
+	// TODO Maybe get rid of BrokerMutation and add Delete() and Upsert(broker) methods to TargetsConfig. Now we always
+	//  delete or update the entire broker entry and we don't need partial updates per trigger.
+	// The code can be simplified to r.targetsConfig.Upsert(brokerConfigEntry)
+	// First delete the broker entry.
```
ian-mi, line 196: I had written a targets config reconciler which supports the interface you want here and allows directly constructing the broker config proto rather than working with the mutation interface: https://github.com/google/knative-gcp/pull/914. Directly constructing the proto is a bit nicer since we are not constrained to using the mutation interface and individual mutations do not need to be threadsafe. I had closed that since the brokercell reconciler was meant to replace this but maybe it is worth reviving after this PR is merged since the brokercell reconciler is not yet ready.

pkg/reconciler/trigger/controller.go
```diff
@@ -0,0 +1,137 @@
...
+	controllerAgentName = "trigger-controller"
+	// finalizerName is the name of the finalizer that this controller adds to the Triggers that it reconciles.
+	finalizerName = "googlecloud"
+)
+
+// filterBroker is the function to filter brokers with proper brokerclass.
+var filterBroker = pkgreconciler.AnnotationFilterFunc(eventingv1beta1.BrokerClassAnnotationKey, brokerv1beta1.BrokerClass, false /*allowUnset*/)
+
+func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl {
+	// TODO initialize project ID here via env var or metadata.
```
ian-mi, line 60: Do we want to support reconciling across multiple projects?

liu-cong: I would say no for now unless there are some compelling use cases. Do you have any use case in mind ?

pkg/reconciler/broker/broker.go
liu-cong: Once we do 1054, we can start moving the target-config reconciliation to brokercell reconciler (assuming a static brokercell).

yolocs: I disagree that updating proto is better. Reconciler shouldn't worry about the internals of targets config. Again, I'm likely biased but just giving my voice.

pkg/reconciler/broker/broker.go
```diff
@@ -206,12 +209,36 @@ func (r *Reconciler) reconcileBroker(ctx context.Context, b *brokerv1beta1.Broke
 			m.SetState(config.State_UNKNOWN)
 		}
 	})
+	// Insert each Trigger to the config.
+	for _, t := range triggers {
```
yolocs, line 213: This loop should be done inside the broker mutation function. Your current approach will be more expensive on memory.

pkg/reconciler/broker/broker.go
```diff
@@ -193,6 +173,29 @@ func (r *Reconciler) reconcileBroker(ctx context.Context, b *brokerv1beta1.Broke
...
+
+// reconcileConfig reconstructs the data entry for the given broker in targets-config.
+func (r *Reconciler) reconcileConfig(ctx context.Context, b *brokerv1beta1.Broker, triggers []*brokerv1beta1.Trigger) {
+	// TODO Maybe get rid of BrokerMutation and add Delete() and Upsert(broker) methods to TargetsConfig. Now we always
+	//  delete or update the entire broker entry and we don't need partial updates per trigger.
+	// The code can be simplified to r.targetsConfig.Upsert(brokerConfigEntry)
+	// First delete the broker entry.
+	r.targetsConfig.MutateBroker(b.Namespace, b.Name, func(m config.BrokerMutation) { m.Delete() })
+	// Then reconstruct the broker entry and insert it
 	r.targetsConfig.MutateBroker(b.Namespace, b.Name, func(m config.BrokerMutation) {
```
yolocs, line 199: This should be done in the same mutation as the triggers below.

pkg/reconciler/broker/broker.go
```diff
@@ -193,6 +173,29 @@ func (r *Reconciler) reconcileBroker(ctx context.Context, b *brokerv1beta1.Broke
...
+	// Update config map
+	r.flagTargetsForUpdate()
+	return nil
+}
+
+// reconcileConfig reconstructs the data entry for the given broker in targets-config.
+func (r *Reconciler) reconcileConfig(ctx context.Context, b *brokerv1beta1.Broker, triggers []*brokerv1beta1.Trigger) {
+	// TODO Maybe get rid of BrokerMutation and add Delete() and Upsert(broker) methods to TargetsConfig. Now we always
+	//  delete or update the entire broker entry and we don't need partial updates per trigger.
+	// The code can be simplified to r.targetsConfig.Upsert(brokerConfigEntry)
```
yolocs, line 195: If you apply my suggestions below, the updates will (hopefully) be less verbose. I hope we don't have to leak the targets config internals in the reconciler (which will be leaked if we allow upserting broker). So one alternative is to have `m.Reset()` to reset an existing broker (or do nothing if not exist)

pkg/reconciler/broker/broker.go
liu-cong: I am not sure I follow what do you mean by "internals" of targets config. Currently the `UpsertTargets` interface already takes in a `config.Target` object. What I am suggesting here is to add an interface to take in `config.Broker` because now we don't upsert each trigger one by one. If we think the `config.Broker` proto is "internal", we can always add another abstraction but I don't think it's needed.

pkg/reconciler/trigger/controller.go
grantr: I haven't heard of a need for reconciling across multiple projects.

Notes for a future PR: Using an env var here makes tests more complicated. One option for setting project ID in this and other controllers is to add it to the context created in cmd/controller/main.go. Then controllers have a uniform way of getting the global project ID if they need it, and there's only one piece of code responsible for retrieving the project id from metadata or env var.

pkg/reconciler/trigger/trigger.go
```diff
@@ -61,43 +68,51 @@ type TriggerReconciler struct {
...
-		//return fmt.Errorf("Couldn't fetch Broker from context")
+var _ triggerreconciler.Interface = (*Reconciler)(nil)
+var _ triggerreconciler.Finalizer = (*Reconciler)(nil)
+
+// ReconcileKind implements Interface.ReconcileKind.
+func (r *Reconciler) ReconcileKind(ctx context.Context, t *brokerv1beta1.Trigger) pkgreconciler.Event {
+	b, err := r.brokerLister.Brokers(t.Namespace).Get(t.Spec.Broker)
+
+	// If there is an error getting the Broker or the Broker is deleted, we mark the Trigger status
+	// regardless of whether the Trigger belongs to this controller or not. It doesn't hurt.
```
grantr, line 87: If multiple trigger controllers are doing this and they disagree on the status message, then they'll fight over the message forever. This should probably check for the condition name and status (BrokerReady, Unknown) before attempting to set. Ultimately we'll want a shared trigger controller in eventing doing this.

pkg/reconciler/trigger/trigger.go
```diff
@@ -61,43 +68,51 @@ type TriggerReconciler struct {
...
-	} else {
-		t.Status.PropagateBrokerStatus(&b.Status)
+		return r.FinalizeKind(ctx, t)
 	}
 
+	if !filterBroker(b) {
+		// Call Finalizer anyway in case the Trigger still holds GCP Broker related resources.
+		// If a Trigger used to point to a GCP Broker but now has a Broker with a different brokerclass,
+		// we should clean up resources related to GCP Broker.
+		return r.FinalizeKind(ctx, t)
```
grantr, line 105: If a trigger is created for a broker that doesn't have this broker class, will the `googlecloud` finalizer be added anyway? I think it's automatically added in genreconciler when FinalizeKind is present, before Reconcile is even called. So most likely every Trigger will have the `googlecloud` finalizer even if they use another broker class.

I think that means that every trigger with a different broker class will be finalized on every update. The unit test of a trigger with a not found broker seems to confirm this (because it finalizes that trigger). Might want to confirm with a real-world test.

If that behavior is confirmed then we should disable the automatic finalizer logic in genreconciler (by renaming FinalizeKind). Then we can add the finalizer only to Triggers that point to a googlecloud broker.

pkg/reconciler/broker/broker.go
yolocs: `config.Broker` is bit different from `config.Target` IMO. The "internal" I'm referring to is the map in the `config.Broker`. To construct a `config.Broker`, you will have to "manually" create the map rather than having the lib handle that. On the other side, `config.Target` is pure data.

pkg/reconciler/broker/broker.go
yolocs: Actually "delete" cannot be done in the same mutation (in the existing implementation) as it will ignore all other operations. This is because the mutation stages all operations without order (my bad) before "committing". If a "delete" operation appears in the mutation, it will treat it as "you want to change broker data and delete it in the same transaction" which essentially equals to "delete the broker". In my other comment, I proposed to add a "reset" function to the mutation.

pkg/reconciler/broker/broker.go
```diff
@@ -193,7 +173,31 @@ func (r *Reconciler) reconcileBroker(ctx context.Context, b *brokerv1beta1.Broke
...
 	r.targetsConfig.MutateBroker(b.Namespace, b.Name, func(m config.BrokerMutation) {
+		// First delete the broker entry.
+		m.Delete()
```
yolocs, line 198: See my other comment. I will add a reset func for now before we decide how we want to change the interface.

yolocs: For the purpose of this PR. I think you can move the delete into a separate mutation and I will modify it along with my reset func.

yolocs, line 198: See my other comment. Let me fix the `delete` func. Either wait for my change or go ahead and having delete in the separate mutation in the PR (and I will fix that later).

pkg/reconciler/trigger/trigger.go
liu-cong: Your understanding is correct. I forgot to add a TODO here to explain the finalizer scenario and future improvements.

> I think that means that every trigger with a different broker class will be finalized on every update.

Correct. This is unnecessary and it will also cause read operation on the topic and pull subscription.

>  we should disable the automatic finalizer logic in genreconciler (by renaming FinalizeKind).

Unfortunately if we do this, the FinalizeKind logic won't be called when the trigger object is deleted. The genreconciler will do nothing because it thinks there is no finalizer to call if the reconciler doesn't implement FinalizeKind.

I have thought about this and came up with a few options:
1. The current behavior. Pros: least effort; Cons: unnecessary FinalizeKind calls for non-GCP broker. 
1. Use a custom controller instead of genreconciler. This essentially means we need to copy most of the  genreconciler boil plate code and make some modifications.
1. Upstream genreconciler changes to make it more flexible, specifically, add a `FilterKind` interface to let the genreconciler skip a trigger if it doesn't point to a GCP broker. I have proposed this interface before https://github.com/knative/pkg/issues/1149

pkg/reconciler/broker/broker.go
yolocs: https://github.com/google/knative-gcp/pull/1142

pkg/reconciler/broker/broker.go
yolocs: Made a fix: https://github.com/google/knative-gcp/pull/1142

pkg/reconciler/trigger/trigger.go
liu-cong: I personally prefer Option 3. If this sounds good to you, I will add a TODO for that.

pkg/reconciler/broker/broker.go
liu-cong: It seems to me making `Delete` a separate Mutation is good for now. We can revisit the interface as a follow up.
I am not sure a `reset` is needed. I believe a single `Upsert` is enough. With regards to the map in the config.Broker, I don't think that's a big deal. If we really want to hide that, we can add an interface like so: `Upsert(broker, triggers)` and let the config handle the mapping.

yolocs: Yea, I didn't add a "reset" in the fix (I fixed "delete" instead). Do notice that you have to stage triggers in a var before you upsert. The actual amount of code would likely look similar to what you have today.

liu-cong: Yeah the amount of code is likely similar. But that's not the point. I was trying to achieve: 1, simplified interface as the interaction is simplified; 2. avoid maintaining states and concurrency in the config lib since we don't need that with brokercell.

Anyways, we don't need to figure out the exact change now. We can do that when we reconcile the config in the brokercell.

pkg/reconciler/trigger/controller.go
```diff
@@ -0,0 +1,137 @@
...
+	}
+
+	impl := triggerreconciler.NewImpl(ctx, r, withAgentAndFinalizer)
+	r.kresourceTracker = duck.NewListableTracker(ctx, conditions.Get, impl.EnqueueKey, controller.GetTrackerLease(ctx))
+	r.addressableTracker = duck.NewListableTracker(ctx, addressable.Get, impl.EnqueueKey, controller.GetTrackerLease(ctx))
+	r.uriResolver = resolver.NewURIResolver(ctx, impl.EnqueueKey)
+
+	r.Logger.Info("Setting up event handlers")
+
+	triggerInformer.Informer().AddEventHandlerWithResyncPeriod(controller.HandleAll(impl.Enqueue), reconciler.DefaultResyncPeriod)
```
yolocs, line 94: Do we need to filter on trigger label `eventing.BrokerLabelKey`?

pkg/reconciler/trigger/controller_test.go
```diff
@@ -0,0 +1,76 @@
...
+	_ "knative.dev/pkg/client/injection/ducks/duck/v1/addressable/fake"
+	_ "knative.dev/pkg/client/injection/ducks/duck/v1/conditions/fake"
+	_ "knative.dev/pkg/client/injection/kube/informers/apps/v1/deployment/fake"
+	_ "knative.dev/pkg/client/injection/kube/informers/core/v1/configmap/fake"
+	_ "knative.dev/pkg/client/injection/kube/informers/core/v1/endpoints/fake"
+	_ "knative.dev/pkg/client/injection/kube/informers/core/v1/pod/fake"
+	_ "knative.dev/pkg/injection/clients/dynamicclient/fake"
+
+	// Why is serving needed here?
+	_ "knative.dev/serving/pkg/client/injection/informers/serving/v1/service/fake"
```
yolocs, line 43: What if you remove it?

pkg/reconciler/trigger/trigger.go
yolocs: Dumb question: why would we reconcile such a trigger (trigger not pointing a gcp broker) in the first place? Broker in trigger spec is immutable.

Also for option 2, there might not be that much boilerplate code from *just* the FinalizeKind. We can still use the majority of the genreconciler.

pkg/reconciler/broker/broker.go
```diff
@@ -205,13 +209,33 @@ func (r *Reconciler) reconcileBroker(ctx context.Context, b *brokerv1beta1.Broke
...
+				target := &config.Target{
+					Id:        string(t.UID),
+					Name:      t.Name,
+					Namespace: t.Namespace,
+					Broker:    b.Name,
+					Address:   t.Status.SubscriberURI.String(),
+					RetryQueue: &config.Queue{
+						Topic:        resources.GenerateRetryTopicName(t),
+						Subscription: resources.GenerateRetrySubscriptionName(t),
+					},
```
yolocs, line 225: At this stage, are we sure the trigger retry topic/sub exist? Unlike when we have broker/trigger reconcilers together, this might no longer hold true, right? Is there a way to tell from the trigger status (I remember Grant added custom trigger status for gcp broker triggers).

yolocs: I care because the retry service will keep trying to connect to the retry sub.

pkg/reconciler/trigger/controller_test.go
grantr: This comment is cargo-culted from similar tests in this repo. When the serving import is missing the test fails, but I don't remember what the error message is.

liu-cong: Interesting, the error is gone if I delete it... I am pretty sure there was an error when I tried this last week.

I believe the error used to be 
```
--- FAIL: TestNew (0.00s)
    logger.go:130: 2020-05-28T10:29:15.343-0700 PANIC   service/service.go:48   Unable to fetch knative.dev/serving/pkg/client/informers/externalversions/serving/v1.ServiceInformer from context.
panic: Unable to fetch knative.dev/serving/pkg/client/informers/externalversions/serving/v1.ServiceInformer from context. [recovered]
        panic: Unable to fetch knative.dev/serving/pkg/client/informers/externalversions/serving/v1.ServiceInformer from context.
```

Going to delete this import wherever applicable. It seems that pkg/reconciler/intevents/topics still needs this.

pkg/reconciler/broker/broker.go
liu-cong: Yeah ideally we should get the pubsub resources from broker/trigger status. Unfortunately the custom status doesn't work today due to webhook issues. This is yet another reason to prioritize custom trigger status.

I believe the retry will eventually work once the pubsub resources exist, but let me verify that.  Given that this should happen rarely in practice, we can work on it as a follow up.

@yolocs also pointed out that retry can check `target.State` field and should not establish pubsub connection if trigger is not ready.

pkg/reconciler/trigger/controller.go
liu-cong: No because we need to be able to reconcile orphaned triggers without a broker.

pkg/reconciler/trigger/trigger.go
liu-cong: > why would we reconcile such a trigger (trigger not pointing a gcp broker) in the first place? Broker in trigger spec is immutable.

Consider this scenario. A trigger was pointed to a gcp broker. Then the broker was updated to use a different brokerclass. Now we want to tear down trigger pubsub resources. We don't know if users will actually do this. But in theory they can.

We could do option 2 but I really don't like the trigger controller to be "special". 
I don't think it's very high  priority because: 1. users are likely to only use gcp broker. 2. even if they use a different broker, this results in a couple of unnecessary calls to pubsub. not great but doesn't really hurt too much. 
For the above reasons I prefer to have a clean solution.

pkg/reconciler/broker/broker.go
liu-cong: I tested retry and it will eventually recover once the pubsub topic/sub exist. #1154 to track the improvements that @yolocs proposed.

pkg/reconciler/trigger/trigger.go
yolocs: I don't think you can change the broker class. https://github.com/knative/eventing/blob/master/pkg/apis/eventing/v1beta1/broker_validation.go#L58-L59

pkg/reconciler/trigger/controller.go
yolocs: When would that happen? It seems trigger spec broker and broker class are both immutable.

pkg/reconciler/trigger/trigger.go
grantr: A broker could be deleted and replaced with a broker with a different class. From the trigger controller's perspective, this will be the same as a broker changing class.

pkg/reconciler/trigger/controller.go
liu-cong: Sorry let me clarify. Do mean filter on the **existence** of the label or actually check the broker is GCP broker?

Let me assume you meant the latter . If you delete a broker, then its triggers become orphaned. We'd like to be able to still delete the triggers (in this case reconcile is a noop).

pkg/reconciler/trigger/trigger.go
liu-cong: Right it's immutable. But you can still delete a broker and recreate it with a different brokerclass.

grantr: It's fine to defer this for later. I'd love to end up at option 3, but realistically I'd probably attempt option 2 first, then move the new code upstream later. It will be easier to justify as an upstream option once we have a demonstrated solution.

pkg/reconciler/broker/broker.go
grantr: The custom status fields would only indicate the resource names, not the readiness. They are not necessary here because the reconcilers updating the configmap and creating the resources will always be in the same container, so they can use the same name generation function. Not perfect encapsulation, but good enough for now.

pkg/reconciler/trigger/trigger.go
liu-cong: SG. Created an issue to follow up https://github.com/google/knative-gcp/issues/1164

liu-cong:
Thanks @yolocs and @grantr for the review. There are many tricky bits and thanks for your comments.

pkg/reconciler/trigger/trigger.go
yolocs: Nothing blocking. Just an idea: if we can (or already?) label the trigger with the initial broker class, then even if the broker got recreated with a different class we can still handle that deterministically, right?

PR #1095, opened 2020-05-19
Remove TestBrokerTracing.

### LINKED ISSUE #818: TestBrokerTracing - Flaky
**Describe the bug**
`TestBrokerTracing` has been very flaky recently in prow (anecdotally, failing >75% of the time). 

 **Expected behavior**
The test should consistently pass.

**Exit Criteria**
Running the test via Prow succeeds 20 times in a row.

 **To Reproduce**
Remove the ignore for the test and run it through prow repeatedly.

Harwayne:
Fixes #818.

## Proposed Changes

- Remove the `TestBrokerTracing` test. 
    - It was redundant because the Channel was already tested directly, so we do not need to test the Channel based Broker on top of it.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/test/e2e/e2e_test.go
+++ b/test/e2e/e2e_test.go
@@ -137,22 +137,6 @@ func TestBrokerDeadLetterSink(t *testing.T) {
 	e2ehelpers.BrokerDeadLetterSinkTestHelper(t, "ChannelBasedBroker" /*brokerClass*/, channelTestRunner, lib.DuplicatePubSubSecret)
 }
 
-func TestBrokerTracing(t *testing.T) {
-	t.Skip("Skipping tracing tests due to flakiness. See https://github.com/google/knative-gcp/issues/818.")
-	if authConfig.WorkloadIdentity {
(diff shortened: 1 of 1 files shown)


grantr:
Isn't this test still technically relevant because it ensures the broker is passing tracing through to the channel correctly?

Harwayne:
> Isn't this test still technically relevant because it ensures the broker is passing tracing through to the channel correctly?

I would consider that a test of the Broker itself, rather than the Channel. That test exists in the knative/eventing repo. I don't think we need to duplicate it here. Especially as this will not be the recommended Broker for people to use.

nachocano:
> > Isn't this test still technically relevant because it ensures the broker is passing tracing through to the channel correctly?
> 
> I would consider that a test of the Broker itself, rather than the Channel. That test exists in the knative/eventing repo. I don't think we need to duplicate it here. Especially as this will not be the recommended Broker for people to use.

I'm not sure if in the internal repo the eventing tests are running. They were disabled AFAIK. Until those are enabled, I think it might be worth leaving this one here.
Only this one flakes and not the channel one?

PR #1099, opened 2020-05-20
Support delivering to a sink in another namespace

### LINKED ISSUE #720: CloudPubSubSource Sink Namespace not honored
**Describe the bug**
According to the spec, you can set the namespace of the sink Service but it is not honored and you get the following error.

```
▶ kubectl describe cloudpubsubsource.events.cloud.google.com/odr-develop -n knative-eventing                                                                                                                                                       
Name:         odr-develop
Namespace:    knative-eventing
API Version:  events.cloud.google.com/v1alpha1
Kind:         CloudPubSubSource
Metadata:
  Creation Timestamp:  2020-03-27T01:57:04Z
  Generation:          1
Spec:
  Ack Deadline:        30s
  Project:             summa-develop
  Retention Duration:  168h0m0s
  Secret:
    Key:   key.json
    Name:  google-cloud-key
  Sink:
    Ref:
      API Version:  serving.knative.dev/v1
      Kind:         Service
      Name:         go-report-engine
      Namespace:    odr-feature-gre
  Topic:            odr-develop
Status:
  Conditions:
    Last Transition Time:  2020-03-27T01:57:04Z
    Message:               failed to get ref &ObjectReference{Kind:Service,Namespace:knative-eventing,Name:go-report-engine,UID:,APIVersion:serving.knative.dev/v1,ResourceVersion:,FieldPath:,}: services.serving.knative.dev "go-report-engine" not found
    Reason:                InvalidSink
    Status:                False
    Type:                  PullSubscriptionReady
    Last Transition Time:  2020-03-27T01:57:04Z
    Message:               failed to get ref &ObjectReference{Kind:Service,Namespace:knative-eventing,Name:go-report-engine,UID:,APIVersion:serving.knative.dev/v1,ResourceVersion:,FieldPath:,}: services.serving.knative.dev "go-report-engine" not found
    Reason:                InvalidSink
    Status:                False
    Type:                  Ready
  Observed Generation:     1
Events:
  Type    Reason   Age                    From                                       Message
  ----    ------   ----                   ----                                       -------
  Normal  Updated  2m37s (x2 over 2m37s)  cloud-run-events-pubsub-source-controller  Updated CloudPubSubSource "odr-develop"

▶ kubectl get cloudpubsubsource.events.cloud.google.com/odr-develop -n knative-eventing -o yaml
apiVersion: events.cloud.google.com/v1alpha1
kind: CloudPubSubSource
metadata:
  creationTimestamp: "2020-03-27T01:57:04Z"
  name: odr-develop
  namespace: knative-eventing
spec:
  ackDeadline: 30s
  project: summa-develop
  retentionDuration: 168h0m0s
  secret:
    key: key.json
    name: google-cloud-key
  sink:
    ref:
      apiVersion: serving.knative.dev/v1
      kind: Service
      name: go-report-engine
      namespace: odr-feature-gre
  topic: odr-develop
status:
  conditions:
  - lastTransitionTime: "2020-03-27T01:57:04Z"
    message: 'failed to get ref &ObjectReference{Kind:Service,Namespace:knative-eventing,Name:go-report-engine,UID:,APIVersion:serving.knative.dev/v1,ResourceVersion:,FieldPath:,}:
      services.serving.knative.dev "go-report-engine" not found'
    reason: InvalidSink
    status: "False"
    type: PullSubscriptionReady
  - lastTransitionTime: "2020-03-27T01:57:04Z"
    message: 'failed to get ref &ObjectReference{Kind:Service,Namespace:knative-eventing,Name:go-report-engine,UID:,APIVersion:serving.knative.dev/v1,ResourceVersion:,FieldPath:,}:
      services.serving.knative.dev "go-report-engine" not found'
    reason: InvalidSink
    status: "False"
    type: Ready
  observedGeneration: 1
```

 **Expected behavior**
You can have a knative Service that is in another namespace from the cloudpubsubsource

 **To Reproduce**
Create KNative Service and CloudPubSubSource in another with reference to the Service as a sink.

v0.13.0

 **Additional context**

nachocano:
Fixes #720

## Proposed Changes

-  🐛 Allowing delivering events to a sink in another namespace.

**Release Note**

```release-note
Your sources can now send events to a sink in another namespace using sink.ref, instead of just sink.uri
```


nachocano:
> The code looks fine, but I'm still not sure this is a good idea. What this effectively means is that any user that can create a pull subscription (directly or indirectly via a source) can target any addressable in the cluster. If nothing else, it gives them a style of read permission throughout the cluster.
> 
> For example, I can make PS in namespace 'a'. I have no read permissions in namespace 'b'. But I know that another team uses eventing in namespace 'b'. I guess at the name of the pubsub topic they created, by creating PS in namespace "a" with different sinks until it resolves. Now I can send events to that Topic, even though I wouldn't normally be able to see it.

It's a valid concern. Note that if you use the sink.uri instead of the sink.ref you are able to do that as well... See https://github.com/google/knative-gcp/issues/720#issuecomment-624738906. Slightly more work, but same thing, you can send to a sink in another namespace...

capri-xiyue:
The code looks fine for me.
Same concerns with @Harwayne , I'm not sure whether it's a good idea.
In addition, should we add some unit tests or e2e tests to verify it works?

nachocano:
> > The code looks fine, but I'm still not sure this is a good idea. What this effectively means is that any user that can create a pull subscription (directly or indirectly via a source) can target any addressable in the cluster. If nothing else, it gives them a style of read permission throughout the cluster.
> > For example, I can make PS in namespace 'a'. I have no read permissions in namespace 'b'. But I know that another team uses eventing in namespace 'b'. I guess at the name of the pubsub topic they created, by creating PS in namespace "a" with different sinks until it resolves. Now I can send events to that Topic, even though I wouldn't normally be able to see it.
> 
> It's a valid concern. Note that if you use the sink.uri instead of the sink.ref you are able to do that as well... See [#720 (comment)](https://github.com/google/knative-gcp/issues/720#issuecomment-624738906). Slightly more work, but same thing, you can send to a sink in another namespace...

@grantr @akashrv what do you think on this one. We briefly covered it in the grooming, and the feeling was that we should support it. @Harwayne raises a valid concern here, although with sink.uri we would still be able to target any addressable in the cluster as well. Is that something that security with istio can address? @yolocs

nachocano:
> The code looks fine for me.
> Same concerns with @Harwayne , I'm not sure whether it's a good idea.
> In addition, should we add some unit tests or e2e tests to verify it works?

Makes sense. I will create a follow up issue for UTs. There are none for this reconciler. We need to add those. Same with E2E test, I'll create an issue and put it in our backlog. BTW, I tried in my cluster and it works.

nachocano:
> > The code looks fine for me.
> > Same concerns with @Harwayne , I'm not sure whether it's a good idea.
> > In addition, should we add some unit tests or e2e tests to verify it works?
> 
> Makes sense. I will create a follow up issue for UTs. There are none for this reconciler. We need to add those. Same with E2E test, I'll create an issue and put it in our backlog. BTW, I tried in my cluster and it works.

I take back the UTs. There are there in the "subclasses". Will add some... Thanks for the catch

akashrv:
> > > The code looks fine, but I'm still not sure this is a good idea. What this effectively means is that any user that can create a pull subscription (directly or indirectly via a source) can target any addressable in the cluster. If nothing else, it gives them a style of read permission throughout the cluster.
> > > For example, I can make PS in namespace 'a'. I have no read permissions in namespace 'b'. But I know that another team uses eventing in namespace 'b'. I guess at the name of the pubsub topic they created, by creating PS in namespace "a" with different sinks until it resolves. Now I can send events to that Topic, even though I wouldn't normally be able to see it.
> > 
> > 
> > It's a valid concern. Note that if you use the sink.uri instead of the sink.ref you are able to do that as well... See [#720 (comment)](https://github.com/google/knative-gcp/issues/720#issuecomment-624738906). Slightly more work, but same thing, you can send to a sink in another namespace...
> 
> @grantr @akashrv what do you think on this one. We briefly covered it in the grooming, and the feeling was that we should support it. @Harwayne raises a valid concern here, although with sink.uri we would still be able to target any addressable in the cluster as well. Is that something that security with istio can address? @yolocs

Rephrasing the exact case you mentioned to make sure I understood it correctly? - User A has create a source in nsA and route events to an addressable in nsB. And that security is a concern.
If security is a concern, then Is there anything today that stops them from doing the same just by creating a pod in nsA that routes to the addressable in nsB? Or if they can directly use the url to the addressable in nsB? I believe that security is a concern that there will be network policies on the namespace that will stop this traffic irrespective of using addressable or direct uri or forwarding through a pod.

Does that make sense? Please let me know if I understood it correctly and/or if I am missing something.

Could we create an issue in Knative/Eventing and get OSS perspective on this?

nachocano:
/hold for UTs...
and until we decide if we want this

Harwayne:
> Rephrasing the exact case you mentioned to make sure I understood it correctly? - User A has create a source in nsA and route events to an addressable in nsB. And that security is a concern.
> If security is a concern, then Is there anything today that stops them from doing the same just by creating a pod in nsA that routes to the addressable in nsB? Or if they can directly use the url to the addressable in nsB? I believe that security is a concern that there will be network policies on the namespace that will stop this traffic irrespective of using addressable or direct uri or forwarding through a pod.

There are two ways that a PS can send information:
- URI
- Addressable

We don't stop anything from hitting a URI. So if user A can craft the URI, then they can make a Pod in nsA that will send traffic to it.

However user A _cannot_ make a Pod in nsA that can lookup the address of addressable in nsB, unless the KSA that Pod runs under has adequate RBAC to GET the addressable resource.

If nothing else, this change will give user A an oracle to confirm which addressables exist (and are Ready=true) in nsB. I don't know how hard of a security boundary this is, but I'm wary of creating oracles in general.

I remember thinking about the general cross-namespace communication in the past (with @grantr I think) and my conclusion at the time was that to make this secure, we would need to have users create the two sides of 'bridge', a sender in nsA and a receiver in nsB. That way we guarantee people with adequate RBAC wanted this to occur.

liu-cong:
I once wrote a doc about cross namespace communication in Eventing: https://docs.google.com/document/d/1I4nkj3tRmgUxRBu9Xhi65wQkrViTkG6kKsQjjos5ddg/edit#

If we want to allow this, we also need to toggle a flag in the admission controller because by default different namespace in a KReference is not allowed: https://github.com/knative/eventing/blob/master/vendor/knative.dev/pkg/apis/duck/v1/knative_reference.go#L68

akashrv:
Interesting (the flag in knative_reference). Please do not block this one
on me as I may not get time to go into details. If you could connect and
come up with a decision I am fine with it.

Regards,
Akash

nachocano:
> I once wrote a doc about cross namespace communication in Eventing: https://docs.google.com/document/d/1I4nkj3tRmgUxRBu9Xhi65wQkrViTkG6kKsQjjos5ddg/edit#
> 
> If we want to allow this, we also need to toggle a flag in the admission controller because by default different namespace in a KReference is not allowed: https://github.com/knative/eventing/blob/master/vendor/knative.dev/pkg/apis/duck/v1/knative_reference.go#L68

Thanks for the pointers @liu-cong . I didn't know you already worked on this. Will sync with you tomorrow.

nachocano:
if we agree to have this, I'd like to have it before the 0.15 cut

config/core/resources/old_pullsubscription.yaml
```diff
@@ -103,15 +103,11 @@ spec:
                     kind:
                       type: string
                       minLength: 1
+                    namespace:
+                      type: string
```
yolocs, line 107: Not sure about the purpose of leaving this old yaml. Could you explain?

nachocano: after the 0.15 cut I think we can remove it. It was due to a migration of the api group @Harwayne did in order to follow the proper deprecation policy.

nachocano:
/hold cancel

Met with folks and agreed we will let this in. Once we tighten up security, we will revisit, and maybe provide a knob for operators to enable/disable this feature. Worth looking at the serving feature flag, or a SubjectAccessReview with custom verbs proposed by @grantr and @liu-cong in a document above.

liu-cong:
I think we should also bring this up in Knative Eventing. I can cleanup the doc a little bit and discuss it with the community

nachocano:
> I think we should also bring this up in Knative Eventing. I can cleanup the doc a little bit and discuss it with the community

Yes, agree. That would be great.

PR #1103, opened 2020-05-20
Pubsub converts between Pubsubs, not intevents.

Harwayne:
## Proposed Changes

- Pubsub converts between Pubsubs, not intevents.

**Release Note**

```release-note
NONE
```


ericlem:
ah yes, that will help...
/lgtm

PR #1117, opened 2020-05-21
Pin knative/pkg and knative/test-infra to release-0.15

Harwayne:
## Proposed Changes

-  Pin `knative/pkg` and `knative/test-infra` to `release-0.15` in anticipation of the 0.15 release.

This was created by:
1. Manually altering `hack/update-deps.sh`.
1. Running `hack/update-deps.sh --upgrade`.

**Release Note**

```release-note
NONE
```

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,9 +32,8 @@ require (
 	k8s.io/apimachinery v0.18.1
 	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
 	knative.dev/eventing v0.14.1-0.20200520223157-166fc6cd13be
-	knative.dev/pkg v0.0.0-20200520073958-94316e20e860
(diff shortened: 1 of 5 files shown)


Harwayne:
@ian-mi is this the correct way to pin dependencies with `go mod`?

ian-mi:
Yes, but you should just change the VERSION var instead.

Harwayne:
> Yes, but you should just change the VERSION var instead.

I thought about that, but didn't because the four places that user `$VERSION` do not all share the same version. Specifically, neither eventing nor serving have the release-0.15 branch yet. S we would need a second variable for them.

This also came up on the equivalent PR to serving, where they decided to hard code it this way: https://github.com/knative/serving/pull/7996#discussion_r427456195.

ian-mi:
In that case this should work. Ideally knative would tag releases with a version tag, the current method essentially pins the dependency to the HEAD commit of the release-0.15 branch.

PR #1120, opened 2020-05-21
Automatically run the wire generated code through goimports.

Harwayne:
## Proposed Changes

- Automatically run the wire generated code through `goimports`.
  - See the discussion on #1045 for why this is useful.

**Release Note**

```release-note
NONE
```


Harwayne:
> /home/prow/go/src/github.com/google/knative-gcp/hack/update-codegen.sh: line 78: goimports: command not found

Looks like prow doesn't have `goimports`...

@chizhg Can I just install it myself? `go get golang.org/x/tools/cmd/goimports` Can it be added to prow's image?

chizhg:
> > /home/prow/go/src/github.com/google/knative-gcp/hack/update-codegen.sh: line 78: goimports: command not found
> 
> Looks like prow doesn't have `goimports`...
> 
> @chizhg Can I just install it myself? `go get golang.org/x/tools/cmd/goimports` Can it be added to prow's image?

It seems we don't have other use cases that need `goimports`. So I would prefer to have here if it does not take a long time to install.
If it's indeed taking a long time, or some other repos also need it, we can certainly add it to the prow-tests image, which is defined at https://github.com/knative/test-infra/blob/master/images/prow-tests/Dockerfile

ian-mi:
The goimports task has already been modified to exclude wire generated files. I'd prefer not to do extra formatting of generated code as this can result in the formatter fighting the code generator.

Harwayne:
> The goimports task has already been modified to exclude wire generated files. I'd prefer not to do extra formatting of generated code as this can result in the formatter fighting the code generator.

I didn't realize that.

/close

Issue #876, opened 2020-04-16
Determine resource request/limit for broker data plane components


liu-cong:
**Problem**
We need to run some benchmark tests to determine a reasonable resource requests/limits for broker data plane.

- [ ] Understand how serving determined these limits and "reasonable" QPS
- [ ] https://github.com/google/knative-gcp/issues/981
- [ ] ingress
- [ ] fanout
- [ ] retry

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?
Operator

**Exit Criteria**
A recommended resource request/limit and why.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**

grantr:
This is P1 for the requests, but not P1 for the limits value or repeatable benchmarks.

grantr:
Clarification on scope: request is needed to determine minimum cluster footprint. Optimal request size for performance is not necessary for R1.

yolocs:
### Setup

- n1-standard-4 3 nodes cluster
- [Workloads](https://github.com/yolocs/ce-test-actor/tree/master/gen/profile)
    - namespace-1: 500x healthy subscribers; seeding one event every 5s
    - namespace-2: 150x slow subscribers (delay 15m); seeding one event every 5m
    - namespace-3: 200x err subscribers (always return 500); seeding one event every 2m
    - namespace-4: 10x echo subscribers (echo back events); seeding one event every 2m

Accumulating long connections and retries in retry service. Every 2m there is a surge of requests because of the echos.

### Results (all single pod)

```bash
k top pods                                                                                                                                                                  130 ↵
NAME                             CPU(cores)   MEMORY(bytes)
broker-fanout-76867ffc8-rdpzn    1670m        166Mi
broker-ingress-746f69945-hvvfz   610m         64Mi
broker-retry-7cc4559df5-l72bh    570m         239Mi
```

**tl;dr:** Fanout is pretty CPU intensive while retry is slightly more memory intensive when there are a lot of events to be retried.

### Results (more replicas)

Fanout - 3 replicas

Retry - 3 replicas

```bash
k top pods
NAME                             CPU(cores)   MEMORY(bytes)
broker-fanout-76867ffc8-jfbrq    1047m        291Mi
broker-fanout-76867ffc8-rdpzn    984m         166Mi
broker-fanout-76867ffc8-rqwzn    722m         277Mi
[...]

Issue #1107, opened 2020-05-20
Add HPA for broker components


yolocs:
**Problem**
Related: https://github.com/google/knative-gcp/issues/1102 and https://github.com/google/knative-gcp/issues/876
Broker fanout/retry resource usage scales with num of triggers. To ensure proper delivery rate, it would make sense to install HPA for these components based on the resource limit they have.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

yolocs:
We also need incorporate HPA in brokercell.

Issue #1102, opened 2020-05-20
Retry pod memory usage grows without bound


grantr:
**Describe the bug**
As mentioned in #876, when the retry service has consumers with long timeouts, its memory usage grows until evicted.

 **Expected behavior**
Retry service memory usage remains stable as load increases.

 **To Reproduce**
https://github.com/google/knative-gcp/issues/876#issuecomment-630950580

 **Additional context**
This is likely caused by max outstanding bytes being set too high. With hundreds of triggers each consuming messages slower than they're coming in, eventually the retry service will use max outstanding bytes * number of triggers. We can reduce the max outstanding bytes/messages as suggested in #876.

yolocs:
Resolved from the PR

PR #1121, opened 2020-05-21
Update broker resource limits and add HPA

yolocs:
Fixes https://github.com/google/knative-gcp/issues/876, https://github.com/google/knative-gcp/issues/1102, https://github.com/google/knative-gcp/issues/1107

## Proposed Changes

- Use resource request/limit proposed in #876 
- Limit max outstanding messages to 100 in retry service (which limits outbound connections for each subscriber)
- Limit overall outstanding bytes for each subscriber to 3M. So with 1500M limit we can *roughly* support 500 subscribers in each pod.
- Add HPA

With the load I'm running to profile broker, here is what HPA looks like (initially all have 1 replica):

```
k get hpa
NAME                 REFERENCE                   TARGETS                        MINPODS   MAXPODS   REPLICAS   AGE
broker-fanout-hpa    Deployment/broker-fanout    241225045333m/900Mi, 74%/95%   1         20        6          107m
broker-ingress-hpa   Deployment/broker-ingress   41536170666m/900Mi, 77%/95%    1         10        3          107m
broker-retry-hpa     Deployment/broker-retry     1453053952/1400Mi, 42%/95%     1         20        2          107m
```

This is a snapshot after a refresh deployment. When it gets stable, there will be even less fanout replicas.

Caveats:

In case where the overall retry load is low except one subscriber. It won't trigger retry HPA to scale up. But because we limit outstanding messages/bytes for each subscriber, the retry delivery rate will be locked down to that. Going to create an issue for that.

**Release Note**

```release-note
Update broker resource limits; limit outstanding messages/bytes for each subscriber in retry service; add HPA
```

**Docs**


yolocs:
When it smoothed out

```
k get hpa
NAME                 REFERENCE                   TARGETS                     MINPODS   MAXPODS   REPLICAS   AGE
broker-fanout-hpa    Deployment/broker-fanout    234119168/900Mi, 87%/95%    1         20        3          4d21h
broker-ingress-hpa   Deployment/broker-ingress   38553600/900Mi, 66%/95%     1         10        2          4d21h
broker-retry-hpa     Deployment/broker-retry     796178432/1400Mi, 40%/95%   1         20        2          4d21h
```

config/broker/deployments/hpa.yaml
```diff
@@ -0,0 +1,77 @@
...
+      name: cpu
+      target:
+        type: Utilization
+        averageUtilization: 95
+  - type: Resource
+    resource:
+      name: memory
+      target:
+        type: AverageValue
+        averageValue: 900Mi
```
liu-cong, line 25: This should be smaller than 500 (the limit)?

PR #1123, opened 2020-05-22
Add resync period to brokercell controller

liu-cong:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/reconciler/brokercell/controller.go
+++ b/pkg/reconciler/brokercell/controller.go
@@ -62,7 +62,7 @@ func NewController(
 
 	logger.Info("Setting up event handlers.")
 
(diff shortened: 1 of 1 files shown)


capri-xiyue:
The code looks good to me. I think I missed some context. Just curious why do we need resync here? Is there any change outside of the cluster that we need to periodically check?

liu-cong:
> The code looks good to me. I think I missed some context. Just curious why do we need resync here? Is there any change outside of the cluster that we need to periodically check?

It's just a general practice to make the system more resilient, in case certain events are lost, or not fully handled. We hope with resync the system will eventually drive the state to the desired.

PR #1131, opened 2020-05-26
Add Delivery Processor MicroBenchmarks Using Fake HTTP Clients

ian-mi:
Adds fake HTTP microbenchmarks alongside the existing real HTTP microbenchmarks. The fake HTTP benchmarks are useful because they better isolate the performance of the processor code and can provide more useful profiling data of this code. The real HTTP benchmarks remain useful as they help measure the performance of the HTTP client itself and can uncover performance issues that only manifest when using real HTTP connections such as improper HTTP connection reuse.  Both sets of benchmarks have been refactored to share common benchmark helpers.

Benchmark Results:
```
goos: linux
goarch: amd64
pkg: github.com/google/knative-gcp/pkg/broker/handler/processors/deliver
BenchmarkDeliveryNoReply-16                	  106274	     10956 ns/op	    8522 B/op	     116 allocs/op
BenchmarkDeliveryWithReply-16              	   54400	     24015 ns/op	   17085 B/op	     251 allocs/op
BenchmarkDeliveryNoReplyFakeClient-16      	  482514	      2515 ns/op	    4347 B/op	      62 allocs/op
BenchmarkDeliveryWithReplyFakeClient-16    	  294411	      3772 ns/op	    7322 B/op	     109 allocs/op
PASS
ok  	github.com/google/knative-gcp/pkg/broker/handler/processors/deliver	5.267s
```

### DIFF
--- a/pkg/broker/handler/processors/deliver/processor_test.go
+++ b/pkg/broker/handler/processors/deliver/processor_test.go
@@ -367,6 +367,9 @@ func (r fakeRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
 	return nil, fmt.Errorf("URL not found: %q", req.URL)
 }
 
+// BenchmarkDeliveryNoReplyFakeClient benchmarks delivery using a fake HTTP client. Use of the
+// fake client helps to better isolate the performance of the processor itself and can provide
+// better profiling data.
 func BenchmarkDeliveryNoReplyFakeClient(b *testing.B) {
 	targetAddress := "target"
 	httpClient := http.Client{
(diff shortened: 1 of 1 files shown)


pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -318,12 +325,107 @@ func (NoReplyHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
...
+			},
+		}),
+	}
+	targetAddress = fakeTargetAddress
+	return
+}
+
+func BenchmarkDeliveryWithReplyFakeClient(b *testing.B) {
+	benchmarkWithReply(b, fakeIngressAddress, makeFakeTargetWithReply)
+}
```
yolocs, line 424: Add a comment similar to what you've put in the PR description. This helps us later know that the one with the fake client could be a baseline.

Issue #867, opened 2020-04-16
BrokerCell creation controller


grantr:
**Problem**
The first time a Broker is created, we should create a BrokerCell to service it. This can be a separate controller from the normal Broker and BrokerCell controller for simplicity.

If this BrokerCell is modified, the controller shouldn't attempt to reconcile away the differences. If the BrokerCell is deleted, the controller should create a new one.

This controller is optional but recommended since it reduces the footprint of eventing when no eventing objects have been created.

**Time Estimate (optional):** 
2

**Additional context (optional)**
If we later add support for multiple BrokerCells (see #866) this controller should create the BrokerCell a Broker is assigned to if it doesn't exist.

grantr:
This is now P1 based on outcome of #1077

PR #1132, opened 2020-05-26
Create brokercell on demand

### LINKED ISSUE #1145: Broker controller watch BrokerCell readiness and update Ready status on change
**Problem**
Broker controller should update Broker readiness when BrokerCell status changes via watch on BrokerCell.

**Exit Criteria**
BrokerCell readiness changes and Broker readiness is immediately updated.

liu-cong:
Ref #867
Fixes #1145 

## Proposed Changes

- The broker controller also creates brokercell if it doesn't exist
- Since the ingress will be managed by brokercell, broker status will look at brokercell status instead of ingress
-

**Release Note**

```release-note

```

**Docs**


liu-cong:
cc @yolocs

pkg/reconciler/broker/brokercell.go
```diff
@@ -0,0 +1,76 @@
...
+		b.Status.MarkBrokerCelllUnknown("NotReady", "Brokercell %s/%s is not ready", bc.Namespace, bc.Name)
+	}
+
+	//TODO(#1019) Use the IngressTemplate of brokercell.
+	ingressServiceName := brokercellresources.Name(bc.Name, brokercellresources.IngressName)
+	b.Status.SetAddress(&apis.URL{
+		Scheme: "http",
+		Host:   names.ServiceHostName(ingressServiceName, bc.Namespace),
+		Path:   fmt.Sprintf("/%s/%s", b.Namespace, b.Name),
+	})
```
yolocs, line 72: Don't we have this already?

pkg/reconciler/broker/brokercell.go
```diff
@@ -0,0 +1,76 @@
...
+	if apierrs.IsNotFound(err) {
+		want := resources.CreateBrokerCell(b)
+		bc, err = r.RunClientSet.InternalV1alpha1().BrokerCells(system.Namespace()).Create(want)
+		if err != nil {
+			logging.FromContext(ctx).Error("Error creating brokercell", zap.String("namespace", b.Namespace), zap.String("broker", b.Name), zap.Error(err))
+			b.Status.MarkBrokerCelllFailed("CreationFailed", "Failed to create %s/%s", want.Namespace, want.Name)
+			return err
+		}
+		r.Recorder.Eventf(b, corev1.EventTypeNormal, brokerCellCreated, "Created brokercell %s/%s", bc.Namespace, bc.Name)
+	}
```
yolocs, line 58: Usually we reconcile with both creation and update. Is there a specific reason here to only do creation?

liu-cong: The purpose here is to ensure a brokercell exists (and create a default one if it doesn't). Once it's created, we allow the brokercell to be updated (although there is nothing to update currently). 

This is also documented in #867.

pkg/reconciler/broker/brokercell.go
liu-cong: This part is updated: previously we get ingress address assuming a static ingress deployment, this is to dynamically update it based on the brokercell.

pkg/reconciler/broker/resources/brokercell.go
```diff
@@ -0,0 +1,40 @@
...
+
+import (
+	"github.com/google/knative-gcp/pkg/apis/broker/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/system"
+
+	inteventsv1alpha1 "github.com/google/knative-gcp/pkg/apis/intevents/v1alpha1"
+)
+
+// Right now we onlyb support one brokercell in the system namespace in the cluster.
```
yolocs, line 27: Typo: onlyb

pkg/reconciler/broker/broker.go
```diff
@@ -149,30 +153,16 @@ func (r *Reconciler) reconcileBroker(ctx context.Context, b *brokerv1beta1.Broke
 	b.Status.InitializeConditions()
 	b.Status.ObservedGeneration = b.Generation
 
+	if err := r.ensureBrokerCellExists(ctx, b); err != nil {
+		return fmt.Errorf("brokercell reconcile failed: %v", err)
+	}
```
yolocs, line 158: Will the brokercell controller fight with the broker controller? It seems that they will both reconcile data plane deployments?

liu-cong: No. The broker controller only creates a `brokercell` resource if needed. The actual reconciliation of data plane is done by the brokercell controller.

pkg/reconciler/broker/broker.go
```diff
@@ -287,7 +277,7 @@ func (r *Reconciler) reconcileDecouplingTopicAndSubscription(ctx context.Context
 		topic, err = client.CreateTopicWithConfig(ctx, topicID, topicConfig)
 		if err != nil {
 			logger.Error("Failed to create Pub/Sub topic", zap.Error(err))
-			b.Status.MarkTopicFailed("CreationFailed", "Topic creation failed: %w", err)
+			b.Status.MarkTopicFailed("TopicCreationFailed", "Topic creation failed: %w", err)
```
liu-cong, line 280: This is the `REASON` field when you do `kubectl get broker my-broker`.

pkg/reconciler/broker/create_brokercell.go
```diff
@@ -0,0 +1,89 @@
...
+			// There can be a race condition where the informer is not updated. In this case we directly
+			// read from the API server.
+			bc, err = r.RunClientSet.InternalV1alpha1().BrokerCells(want.Namespace).Get(want.Name, metav1.GetOptions{})
+			if err != nil {
+				logging.FromContext(ctx).Error("Failed to get the brokercell from the API server", zap.String("namespace", b.Namespace), zap.String("broker", b.Name), zap.Error(err))
+				return err
+			}
+		}
+		if err == nil {
+			r.Recorder.Eventf(b, corev1.EventTypeNormal, brokerCellCreated, "Created brokercell %s/%s", bc.Namespace, bc.Name)
```
yolocs, line 70: nit: created => ensured. Could be confusing if there is an existing brokercell.

liu-cong: Actually I think `create` is accurate here: Line 53 tries to create a brokercell, and this line only records the event when there is no error.

PR #1135, opened 2020-05-27
Use HTTP Client + CE SDK v2 Bindings in Deliver Processor

ian-mi:
Directly use the HTTP client in the deliver processor using the cloudevents SDK v2 bindings. This significantly improves the performance of fanout delivery:

```
name                            old time/op    new time/op    delta
DeliveryNoReply-16                15.0µs ± 1%    11.8µs ± 1%  -21.28%  (p=0.000 n=19+19)
DeliveryWithReply-16              25.8µs ± 1%    23.4µs ± 1%   -9.39%  (p=0.000 n=20+20)
DeliveryNoReplyFakeClient-16      3.03µs ± 1%    1.96µs ± 2%  -35.20%  (p=0.000 n=20+20)
DeliveryWithReplyFakeClient-16    4.88µs ± 2%    3.04µs ± 2%  -37.78%  (p=0.000 n=19+16)

name                            old alloc/op   new alloc/op   delta
DeliveryNoReply-16                8.64kB ± 0%    7.40kB ± 0%  -14.33%  (p=0.000 n=19+18)
DeliveryWithReply-16              17.2kB ± 0%    14.7kB ± 0%  -14.81%  (p=0.000 n=20+20)
DeliveryNoReplyFakeClient-16      4.35kB ± 0%    3.23kB ± 0%  -25.82%  (p=0.000 n=20+20)
DeliveryWithReplyFakeClient-16    7.33kB ± 0%    4.84kB ± 0%  -34.00%  (p=0.000 n=20+16)

name                            old allocs/op  new allocs/op  delta
DeliveryNoReply-16                   116 ± 0%        94 ± 0%  -18.97%  (p=0.000 n=19+20)
DeliveryWithReply-16                 251 ± 0%       210 ± 0%  -16.33%  (p=0.000 n=20+20)
DeliveryNoReplyFakeClient-16        62.0 ± 0%      41.0 ± 0%  -33.87%  (p=0.000 n=20+20)
DeliveryWithReplyFakeClient-16       109 ± 0%        69 ± 0%  -36.70%  (p=0.000 n=20+20)
```

### DIFF
--- a/pkg/broker/handler/processors/deliver/processor.go
+++ b/pkg/broker/handler/processors/deliver/processor.go
@@ -19,12 +19,15 @@ package deliver
 import (
 	"context"
 	"fmt"
+	"net/http"
 	"time"
 
+	"github.com/cloudevents/sdk-go/v2/binding"
+	"github.com/cloudevents/sdk-go/v2/binding/spec"
 	ceclient "github.com/cloudevents/sdk-go/v2/client"
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
 	"github.com/cloudevents/sdk-go/v2/event"
-	"github.com/cloudevents/sdk-go/v2/protocol"
+	cehttp "github.com/cloudevents/sdk-go/v2/protocol/http"
 	"go.uber.org/zap"
 	"knative.dev/pkg/logging"
 
@@ -41,7 +44,7 @@ type Processor struct {
 	processors.BaseProcessor
 
 	// DeliverClient is the cloudevents client to send events.
-	DeliverClient ceclient.Client
+	DeliverClient *http.Client
 
 	// Targets is the targets from config.
 	Targets config.ReadonlyTargets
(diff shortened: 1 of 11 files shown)


pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -101,41 +104,69 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
...
+	if err := replyResp.Body.Close(); err != nil {
+		logging.FromContext(ctx).Warn("failed to close reply response body", zap.Error(err))
 	}
+	return nil
+}
 
-	// For post-delivery processing.
-	return p.Next().Process(ctx, event)
+func (p *Processor) sendMsg(ctx context.Context, address string, msg binding.Message, transformers ...binding.Transformer) (*http.Response, error) {
+	req, err := http.NewRequestWithContext(ctx, "POST", address, nil)
```
yolocs, line 162: nit: use `http.MethodPost`?

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -101,41 +104,69 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
...
-	resp, res := p.DeliverClient.Request(cecontext.WithTarget(dctx, target.Address), copy)
-	if !protocol.IsACK(res) {
+	if err := p.deliver(dctx, target, broker, (*binding.EventMessage)(&copy), hops); err != nil {
 		if !p.RetryOnFailure {
-			return fmt.Errorf("target delivery failed: %v", res.Error())
+			return err
 		}
 
-		logging.FromContext(ctx).Warn("target delivery failed", zap.String("target", tk), zap.String("error", res.Error()))
+		logging.FromContext(ctx).Warn("target delivery failed", zap.String("target", tk), zap.Error(err))
```
Harwayne, line 112: Desugar or switch to `knative.dev/eventing/pkg/logging`.

This mistake is repeated a few times in the file, so I suggest switching the import.

ian-mi: I'm not sure that the logging throughput is high enough here for desugaring to be useful. Profiling data shows very little time is spent in the logger and desugaring may actually make things slower because we recreate a desugared logger on every call instead of storing the logger.

ian-mi:
I'm not sure exactly why the CE client is so slow. The main performance benefit of the binding code is supposed be removing the intermediate serialization to the event type. I have not yet removed use of the event type from the fanout code, however, and still see significant improvements so I suspect there is some unnecessary overhead in the CE client causing the performance gap.

pkg/broker/handler/processors/deliver/processor.go
ian-mi: Resolved offline that desugared logger is still worth using for safety reasons.

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -101,41 +104,69 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
...
 		return p.sendToRetryTopic(ctx, target, event)
```
grantr, line 113: This is a comment on the [above strategy](https://github.com/google/knative-gcp/pull/1135/files#diff-b746f19389275f3eb9488200919af700R93) of cloning the event to remove the hops attribute and keeping the original around for possible retry. I'm curious why this cloning is necessary. Could we remove the hops attribute from `event` and add it back if needed? Is there an issue with concurrent access?

I know this isn't really part of this PR so feel free to ignore.

ian-mi: Yes, this event may be fanned out to multiple processors and it would not be safe to mutate the event here. My next idea for optimization was to write a message transformer to mutate hops when writing the outgoing http message rather than cloning the event. I will start work on that optimization once this PR is merged.

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -101,41 +104,69 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
...
+
+	respMsg := cehttp.NewMessageFromHttpResponse(resp)
+	if respMsg.ReadEncoding() == binding.EncodingUnknown {
+		// No reply
 		return nil
 	}
 
 	if hops <= 0 {
+		_, id := respMsg.GetAttribute(spec.ID)
 		logging.FromContext(ctx).Debug("dropping event based on remaining hops status",
```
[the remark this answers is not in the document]
ian-mi: Sounds good, I will create a follow-up PR to address this.

n3wscott:
> I wonder what's in CE SDk that is causing the overhead? Is the http protocol not implemented efficiently?

conversion to and from the Event type is a lot of memory copying. Staying with Message and using the Bindings should speed things up for middleware

ian-mi:
What makes this unexpected is that we haven't yet removed the unnecessary conversion to event and still see a significant performance benefit. I haven't looked too deep into profiling data but I suspect some of this overhead was due to using the client with metrics enabled.

Issue #1141, opened 2020-05-27
Targets config broker mutation should handle delete properly when it's not the last operation


yolocs:
**Describe the bug**
Currently `mutation.Delete()` once called will delete the broker no matter what.

 **Expected behavior**
If I do:

```
mutation.Delete()
mutation.SetID("...")
```

It really should delete the broker and then create a new one for me.

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative-GCP release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

PR #1142, opened 2020-05-27
Handle broker mutation delete properly

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1141

## Proposed Changes

- calling `mutation.Delete()` resets the broker instantly

**Release Note**

```release-note
Calling `mutation.Delete()` in targets.Config resets the broker instantly
```

**Docs**

### DIFF
--- a/pkg/broker/config/memory/memory.go
+++ b/pkg/broker/config/memory/memory.go
@@ -29,26 +29,31 @@ type brokerMutation struct {
 }
 
 func (m *brokerMutation) SetID(id string) config.BrokerMutation {
+	m.delete = false
 	m.b.Id = id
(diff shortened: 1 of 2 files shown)


liu-cong:
The code LGTM. But I think it's getting a bit tricky in maintaining the states. I think calling `Delete` separately is probably better for now.

How about saving the effort and revisit this once we start to reconcile the config in brokercell?

yolocs:
@liu-cong  I think as long as there is no bug in this code we should get this merged because it's the right behavior to have. Whether maintaining this state is tricky I believe is something we can revisit. We can even change the interface if needed. Thoughts?

PR #1152, opened 2020-05-28
Renaming internal test package to int

nachocano:
Renaming test images internal package to int, otherwise it was causing problems in other repos.
Error I was getting:

`Seeing vendor/github.com/google/knative-gcp/test/test_images/pubsub_target/main.go:23:2: use of internal package github.com/google/knative-gcp/test/test_images/internal/knockdown not allowed`

According to https://golang.org/doc/go1.4#internalpackages, it should work, but we might be messing up something somewhere.

**Release Note**

```release-note
NONE
```

**Docs**


nachocano:
/close 

Thanks @Harwayne for taking care of it! I'm still a bit puzzled

PR #1153, opened 2020-05-28
Add HPA to brokercell controller

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1107

## Proposed Changes

- Reconcile HPA for broker components

**Release Note**

```release-note
Reconcile HPA for broker components in brokercell 
```

**Docs**


pkg/reconciler/brokercell/brokercell.go
```diff
@@ -95,13 +94,22 @@ var _ bcreconciler.Interface = (*Reconciler)(nil)
 func (r *Reconciler) ReconcileKind(ctx context.Context, bc *intv1alpha1.BrokerCell) pkgreconciler.Event {
 	bc.Status.InitializeConditions()
 
-	// Reconcile ingress deployment and service
+	// Reconcile ingress deployment, HPA and service.
 	ingressArgs := r.makeIngressArgs(bc)
-	if _, err := r.deploymentRec.ReconcileDeployment(bc, resources.MakeIngressDeployment(ingressArgs)); err != nil {
+	ind, err := r.deploymentRec.ReconcileDeployment(bc, resources.MakeIngressDeployment(ingressArgs))
+	if err != nil {
 		logging.FromContext(ctx).Error("Failed to reconcile ingress deployment for \"%s/%s\": %v", zap.Any("namespace", bc.Namespace), zap.Any("name", bc.Name), zap.Error(err))
```
liu-cong, line 101:
```suggestion
		logging.FromContext(ctx).Error("Failed to reconcile ingress deployment", zap.Any("namespace", bc.Namespace), zap.Any("name", bc.Name), zap.Error(err))
```

pkg/reconciler/brokercell/resources/hpa.go
```diff
@@ -0,0 +1,65 @@
...
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"knative.dev/pkg/kmeta"
+)
+
+// MakeHorizontalPodAutoscaler makes an HPA for the given arguments.
+func MakeHorizontalPodAutoscaler(deployment *appsv1.Deployment, args AutoscalingArgs) *hpav2beta1.HorizontalPodAutoscaler {
+	var one int32 = 1
+	q := resource.MustParse(args.AvgMemoryUsage)
```
liu-cong, line 31: nit: q->memory? Not sure what q means here.

pkg/reconciler/brokercell/brokercell_test.go
```diff
@@ -140,12 +150,62 @@ func TestAllCases(t *testing.T) {
 			},
 			WantErr: true,
 		},
+		{
+			Name: "Ingress HorizontalPodAutoscaler.Create error",
```
liu-cong, line 154: Looks like we can add a couple of new test cases:
 - HPA exists with same spec and no updates needed
- HPA exists with different spec and update succeeds

pkg/reconciler/brokercell/brokercell.go
```diff
@@ -95,13 +94,22 @@ var _ bcreconciler.Interface = (*Reconciler)(nil)
...
 		bc.Status.MarkIngressFailed("IngressDeploymentFailed", "Failed to reconcile ingress deployment: %v", err)
 		return err
 	}
+
+	ingressHPA := resources.MakeHorizontalPodAutoscaler(ind, r.makeIngressHPA(bc))
+	if err := r.reconcileAutoscaling(ctx, bc, ingressHPA); err != nil {
+		logging.FromContext(ctx).Error("Failed to reconcile ingress HPA for \"%s/%s\": %v", zap.Any("namespace", bc.Namespace), zap.Any("name", bc.Name), zap.Error(err))
+		bc.Status.MarkIngressFailed("HorizontalPodAutoscalerFailed", "Failed to reconcile ingress HorizontalPodAutoscaler: %v", err)
```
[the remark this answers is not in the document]
yolocs: I don't see a very compelling reason to not treat HPA as part of the component unit (e.g. ingress unit). If HPA fails, it could have a real impact on the ingress availability under some circumstances. That's why I prefer my current approach.

If we really want to treat HPA as a 2nd party citizen, we can probably add a *single* condition type for HPA and indicate which HPA has failed in the message.

pkg/reconciler/brokercell/resources/hpa.go
yolocs: quantity. It's the return type from the parsing. Will make it more clear.

pkg/reconciler/brokercell/brokercell.go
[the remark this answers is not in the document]
liu-cong: I am fine with the current approach. I was debating whether we should treat HPA as 2nd party citizen or even whether we should return an error if HPA fails since very likely the data plane still works without the scaling. But I guess it's too tricky and complicated if we that.

liu-cong:
Talked to @yolocs offline and I believe the integ test failures are due to missing hpa APIs in the cluster. @yolocs is going to try using v2beta2 instead of v2beta1.

Issue #1154, opened 2020-05-28
Don't start retry pull sub if the trigger is not ready


yolocs:
Once we have separated trigger controller from the broker controller, there is no guarantee that trigger's retry topic/sub will present before retry service get them.

liu-cong:
I tested that if the pubsub topic/sub don't exist, the retry pod will just keep retrying (every ~25s) until they show up. So it will eventually recover. For that reason, this is not super important.

PR #1159, opened 2020-05-28
Only start the pull handler when its resource is ready

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1154

## Proposed Changes

- fanout/retry pool will not start pull sub when the resource is not ready

**Release Note**

```release-note
Fanout/retry pool will not start pull sub when the resource is not ready
```

**Docs**

### DIFF
--- a/pkg/broker/handler/pool/fanout.go
+++ b/pkg/broker/handler/pool/fanout.go
@@ -136,6 +136,12 @@ func (p *FanoutPool) SyncOnce(ctx context.Context) error {
 			p.pool.Delete(b.Key())
 		}
 
+		// Don't start the handler if broker is not ready.
+		// The decouple topic/sub might not be ready at this point.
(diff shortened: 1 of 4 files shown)


pkg/broker/handler/pool/fanout.go
```diff
@@ -136,6 +136,12 @@ func (p *FanoutPool) SyncOnce(ctx context.Context) error {
 			p.pool.Delete(b.Key())
 		}
 
+		// Don't start the handler if broker is not ready.
+		// The decouple topic/sub might not be ready at this point.
+		if b.State != config.State_READY {
+			return true
```
grantr, line 142: This will be retried again after 15 seconds, correct?

yolocs: That is right.

PR #1161, opened 2020-05-28
added or modified smoke test of four sources and gcp broker

### LINKED ISSUE #909: Test deletion of sources also deletes GCP resources
**Problem**
We need to verify sources get created and get deleted correctly
For example, when we delete a cloudstoragesource, all the underlying pull subscriptions and topic etc. should also get deleted

**Exit Criteria**
Smoke Testing for knative-gcp of 4 sources(pubsub, storage, scheduler, auditlogs)

**Time Estimate (optional):** 

capri-xiyue:
Fixes #909 #1165  

## Proposed Changes

- modified or added smoke test of pubsub, storage, scheduler, auditlogs sources
- added smoke test of gcp broker
- use "github.com/google/knative-gcp/pkg/apis/broker/v1beta1" when creating gcp broker instead of "knative.dev/eventing/pkg/apis/eventing/v1beta1"

**Release Note**

```release-note

```

**Docs**


test/e2e/lib/creation.go
```diff
@@ -72,6 +72,16 @@ func (c *Client) CreatePubSubOrFail(pubsub *eventsv1alpha1.CloudPubSubSource) {
 	c.Tracker.AddObj(pubsub)
 }
 
+func (c *Client) CreatePubSubOrFailWithoutClean(pubsub *eventsv1alpha1.CloudPubSubSource) {
```
capri-xiyue, line 75: The only difference between `CreatePubSubOrFailWithoutClean` and `CreatePubSubOrFail` is this line https://github.com/capri-xiyue/knative-gcp/blob/ffdbf0872bef3a4e755b01ee3a812a41be4449d2/test/e2e/lib/creation.go#L72
For such smoke tests in this PR, we will delete the sources by ourselves, therefore we don't need the clean

test/e2e/lib/auditlogs.go
```diff
@@ -75,3 +81,24 @@ func MakeAuditLogsJobOrDie(client *Client, methodName, project, resourceName, se
...
+	if err != nil {
+		t.Fatalf("failed to create LogAdmin client, %s", err.Error())
+	}
+	defer client.Close()
+
+	_, err = client.Sink(ctx, sinkID)
+	if err != nil {
+		if status.Code(err) == codes.NotFound {
+			return false
+		} else {
```
chizhg, line 99: else is not needed.

capri-xiyue:
At first, I tried to avoid using methods like https://github.com/google/knative-gcp/blob/53a1bd6bce81008487a66fba80739ca0ef671a84/test/e2e/lib/creation.go#L75 since it has tracker that we don't need to use in smoke test happy path since we will deleete the source by ourselves https://github.com/google/knative-gcp/blob/53a1bd6bce81008487a66fba80739ca0ef671a84/test/e2e/lib/creation.go#L83.
However, if I don't use that method, it will make the code kind of complex since I need to handle unhappy paths that the test fail and but we still need to delete the sources no matter what.

Therefore, in the end, I still choose to use https://github.com/google/knative-gcp/blob/53a1bd6bce81008487a66fba80739ca0ef671a84/test/e2e/lib/creation.go#L75  with `	c.Tracker.AddObj(xxxx)`. The side effect is that when the test passes, if you run go test -v, it will show similar logs that `tracker can't delete source, not found'

capri-xiyue:
In addition, I found one bug of sources(source can still be ready even though the sink(k8s service) doesn't exist). Already filed another issue for that bug https://github.com/google/knative-gcp/issues/1171

chizhg:
lgtm from my side.
Defer to @nachocano and @ericlem  for applying the `/lgtm` label.

test/e2e/lib/deletion.go
```diff
@@ -0,0 +1,67 @@
...
+
+
+func (c *Client) DeleteGCPBrokerOrFail(name string) {
+	c.T.Helper()
+	brokers := c.KnativeGCP.EventingV1beta1().Brokers(c.Namespace)
+	err := brokers.Delete(name, &metav1.DeleteOptions{})
+	if err != nil {
+		c.T.Fatalf("Failed to delete gcp broker %s/%s: %v", c.Namespace, name, err)
+	}
+}
```
nachocano, line 67: add empty line at the end of the file

test/e2e/lib/get.go
```diff
@@ -0,0 +1,64 @@
...
+
+func (c *Client) GetAuditLogsOrFail(name string) *eventsv1alpha1.CloudAuditLogsSource{
+	c.T.Helper()
+	auditLogs:= c.KnativeGCP.EventsV1alpha1().CloudAuditLogsSources(c.Namespace)
+	existing, err := auditLogs.Get(name, metav1.GetOptions{})
+	if err != nil {
+		c.T.Fatalf("Failed to get auditlogs %s/%s: %v", c.Namespace, name, err)
+	}
+	return existing
+}
```
nachocano, line 64: same here

test/e2e/lib/resources/knative_gcp.go
```diff
@@ -0,0 +1,58 @@
...
+
+// WithBrokerClassForBrokerV1Beta1 returns a function that adds a brokerClass
+// annotation to the given Broker.
+func WithBrokerClassForBrokerV1Beta1(brokerClass string) BrokerV1Beta1Option {
+	return func(b *v1beta1.Broker) {
+		annotations := b.GetAnnotations()
+		if annotations == nil {
+			annotations = make(map[string]string, 1)
+		}
+		annotations["eventing.knative.dev/broker.class"] = brokerClass
```
nachocano, line 40: if there is a constant somewhere, rather use that?

test/e2e/test_auditlogs.go
```diff
@@ -32,6 +32,71 @@ import (
...
+		t.Errorf("Expected topic%q to exist", topicID)
+	}
+
+	createdSubExists := lib.SubscriptionExists(t, subID)
+	if !createdSubExists {
+		t.Errorf("Expected subscription %q to exist", subID)
+	}
+	client.DeleteAuditLogsOrFail(auditlogsName)
+	//Wait for 20 seconds for topic, subscription and notification to get deleted in gcp
+	time.Sleep(20*time.Second)
```
nachocano, line 81: might be worth adding this magic numbers to some constant. If we start seeing flakes, we may need to increase them? Hope not...

Something like WaitDeletionTime or something like that...

test/e2e/test_gcp_broker.go
```diff
@@ -94,10 +98,54 @@ func SchedulerSourceWithGCPBrokerTestImpl(t *testing.T, authConfig lib.AuthConfi
...
+	}
+
+	createdSubExists := lib.SubscriptionExists(t, subID)
+	if !createdSubExists {
+		t.Errorf("Expected subscription %q to exist", subID)
+	}
+
+	client.DeleteGCPBrokerOrFail(brokerName)
+	//Wait for 10 seconds for subscription to get deleted in gcp
+	time.Sleep(10 * time.Second)
```
nachocano, line 130: same here, if you can... Why was 20 the one above and this one 10?

test/e2e/test_pubsub.go
```diff
@@ -51,8 +52,21 @@ func SmokeCloudPubSubSourceTestImpl(t *testing.T, authConfig lib.AuthConfig) {
...
+	subID := createdPubSub.Status.SubscriptionID
 
+	createdSubExists := lib.SubscriptionExists(t, subID)
+	if !createdSubExists {
+		t.Errorf("Expected subscription %q to exist", subID)
+	}
+
+	client.DeletePubSubOrFail(psName)
+	//Wait for 10 seconds for subscription to get deleted in gcp
+	time.Sleep(10*time.Second)
```
nachocano, line 65: same if you can

test/e2e/lib/resources/knative_gcp.go
capri-xiyue: No. In knative eventing, we also use string there https://github.com/knative/eventing/blob/bdc48890f0e46a95e1bd4bb253b27b3af9ed5d05/test/lib/resources/eventing.go#L219.
I agree we should use constant, but we need to change that in eventing

test/e2e/test_scheduler.go
```diff
@@ -19,32 +19,65 @@ package e2e
...
 
-	client.CreateSchedulerOrFail(scheduler)
-	client.Core.WaitForResourceReadyOrFail(sName, lib.CloudSchedulerSourceTypeMeta)
+	createdSubExists := lib.SubscriptionExists(t, subID)
+	if !createdSubExists {
+		t.Errorf("Expected subscription %q to exist", subID)
+	}
+	client.DeleteSchedulerOrFail(schedulerName)
+	//Wait for 20 seconds for topic, subscription and job to get deleted in gcp
+	time.Sleep(20 * time.Second)
```
nachocano, line 65: same

test/e2e/test_storage.go
```diff
@@ -37,6 +37,64 @@ import (
...
+		t.Errorf("Expected topic%q to exist", topicID)
+	}
+
+	createdSubExists := lib.SubscriptionExists(t, subID)
+	if !createdSubExists {
+		t.Errorf("Expected subscription %q to exist", subID)
+	}
+	client.DeleteStorageOrFail(storageName)
+	//Wait for 20 seconds for topic, subscription and notification to get deleted in gcp
+	time.Sleep(20*time.Second)
```
nachocano, line 79: one more so that you don't forget

test/e2e/test_gcp_broker.go
capri-xiyue: StackdriverSink for CloudAuditLogsSource and Scheduler Job for CloudSchedulerSource takes more time to create. That's why I don't use a constant. The wait time depends on the resources needed.

nachocano:
/lgtm
/hold 
in case you want to do those changes

test/e2e/test_auditlogs.go
capri-xiyue: StackdriverSink for CloudAuditLogsSource and Scheduler Job for CloudSchedulerSource takes more time to create. That's why I don't use a constant. The wait time depends on the resources needed. That's why I don't use a consistent constant time to control it. I can also use a consistent constant time for all sources if you prefer.

test/e2e/test_gcp_broker.go
nachocano: I'm fine with it. Just do 2 * constant, and document why, so that other folks don't get confused

test/e2e/test_auditlogs.go
nachocano: 2 * GcpResourceWaitDeletionTime or something like that

test/e2e/lib/resources/constants.go
```diff
@@ -53,3 +55,9 @@ const (
 const (
 	KServiceKind string = "Service"
 )
+
+// WaitDeletionTime for deleting resources
+const (
+	WaitDeletionTime  = 20 * time.Second
```
nachocano, line 61: sounds good, will put this one to 10, and then in the few places you need 20, just multiply?

nachocano: I see you went with 20 everywhere. My only worry is that tests will last even longer... But I'm ok if you leave it as is

test/e2e/lib/resources/knative_gcp.go
nachocano: ack

PR #1162, opened 2020-05-28
Garbage collect brokercell if no brokers are using it

### LINKED ISSUE #1157: Brokercell garbage collection
**Problem**
Delete the brokercell if there are no brokers using it to reduce idle resource usage. This is the reverse process of #867

liu-cong:
Fixes #1157

## Proposed Changes

- For brokercells automatically created by GCP broker in #1132 (that has `internal.events.cloud.google.com/creator: googlecloud` annotation), if there are no brokers pointing to it, we garbage collect it.
- We don't gc brokercells that were not created by GCP broker to avoid surprises.
-

**Release Note**

```release-note
Garbage collect brokercell if no brokers are using it
```

**Docs**

cc @grantr


pkg/reconciler/brokercell/brokercell.go
```diff
@@ -93,6 +97,11 @@ var _ bcreconciler.Interface = (*Reconciler)(nil)
 
 // ReconcileKind implements Interface.ReconcileKind.
 func (r *Reconciler) ReconcileKind(ctx context.Context, bc *intv1alpha1.BrokerCell) pkgreconciler.Event {
+	if r.shouldGC(ctx, bc) {
```
yolocs, line 100: Would this controller be informed at all when brokers get deleted? Or it purely relies on resync period?

liu-cong: It doesn't watch brokers currently. I think using resync period is good enough, or maybe better since we'd like to create brokercell quickly but garbage collect slowly.

yolocs: It just occurred to me - why don't we do it in the broker controller? If it's the broker controller creates the brokercells, shouldn't it be the one deletes them?

liu-cong: Good question! It's simple and reliable to do it in the brokercell controller than the broker controller.

- Simple

It's tricky to handle concurrency in broker controller. Suppose you are deleting all brokers at the same time, hard to tell if the brokercell should be gc'ed.

- Reliable

If for some reason we didn't delete the brokercell in the broker controller (due to race, or due to missing event, bug, etc), and if all the brokers are deleted, then we don't have chance to retry. While in brokercell controller, as long as the object still exists, we will have chances to continue looking at it.

grantr: Seems like the creation and deletion of brokercells can be its own controller since it's independent of both the broker and brokercell controllers. That keeps the logic separate and maybe easier to follow.

Not intending to block this PR, just happened to see this comment.

liu-cong: Yes it's cleaner to have a separate controller. But meanwhile there are more code to write, so I chose the current approach for speed :)

liu-cong: Created an issue to follow up: https://github.com/google/knative-gcp/issues/1196

yolocs: @grantr you mean just deletion? I don't understand how creation can be split out from the brokercell controller.

@liu-cong Sounds like a fine plan for short term.

yolocs: Maybe add a TODO with a brief explanation. Otherwise lgtm

liu-cong: For the creation, the new controller will just watch for brokers and create a brokercell if needed. The existing broker controller only looks at brokercell status and updates broker status accordingly.

Not sure if this is better than the current approach though.

liu-cong: @yolocs 

Explained the choice as a comment and added TODO.

PR #1170, opened 2020-05-29
Remove static GCP broker data plane  yamls

### LINKED ISSUE #867: BrokerCell creation controller

liu-cong:
Fixes #867

## Proposed Changes

- Remove static yamls for GCP broker data plane as now they are created dynamically by brokercell (https://github.com/google/knative-gcp/pull/1132)
- Copy the SA, role and rolebinding in the 200-rbac.yaml to existing rbac files
-

**Release Note**

```release-note
Remove static GCP broker data plane yamls
```

**Docs**


liu-cong:
This PR removes the `config/broker` folder as the broker data plane will be dynamically created with brokercell.
@ericlem Let me know if there is anything requried for go mod (I remember you talked about the dummy go file).

cc @nachocano as you are looking at the release yamls

nachocano:
> This PR removes the `config/broker` folder as the broker data plane will be dynamically created with brokercell.
> @ericlem Let me know if there is anything requried for go mod (I remember you talked about the dummy go file).
> 
> cc @nachocano as you are looking at the release yamls

Are you planning on cherry-picking this and other broker stuff to 0.15?

ericlem:
When we merge it into cloudrun-eventing we'll have to remove the entries pointing at the config/broker directories from hack/tools.go (the version in cloudrun-eventing) and the copy_yaml command for it from hack/update-deps.sh (the version in cloudrun-eventing)

yolocs:
It just occurred to me - would it be useful to keep these yamls somewhere for easier testing?

liu-cong:
> It just occurred to me - would it be useful to keep these yamls somewhere for easier testing?

For consistency I would say no since we don't keep yamls for other dynamically created resources. Is there any specific use case you have in mind?

yolocs:
@liu-cong I was thinking the case where I want to create a broker data plane manually. But never mind, once we move to brokercell the existing yamls won't work anyway.

liu-cong:
> I was thinking the case where I want to create a broker data plane manually. But never mind, once we move to brokercell the existing yamls won't work anyway.

Yeah I can see sometimes you want to manually deploy the yamls for troubleshooting for example. We could create a `example yaml` folder to put these yamls. But I am afraid we will forget to keep them up to date... So better just remove them if we won't maintain them.

yolocs:
/hold
It just occurred to me. Will this PR break e2e test? https://github.com/google/knative-gcp/blob/848d632c199e189e6aedd0594c2e392c529d527a/test/lib.sh#L38

liu-cong:
> Will this PR break e2e test?

Indeed it will. And this explains the test failures :)

Also removed the yaml from e2e tests. Let's see if the tests will pass.

Issue #882, opened 2020-04-17
Support liveness probes in broker ingress


liu-cong:
**Problem**
- Related to #785. 
- serve health and readiness endpoints
- Also add probes in the deployment yamls (or the BrokerCell controller if it replaces the yaml)

Ref: https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot/

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**

grantr:
How to test:
- Unit tests that verify endpoints are being served and accurate
- E2E test that verifies when ingress pods change state the broker readiness changes. How do we force a change in the broker readiness?

Let's first investigate how to test these probes in E2E tests.

PR #1179, opened 2020-06-01
Add broker ingress liveness probe

yolocs:
Fixes https://github.com/google/knative-gcp/issues/882

**Release Note**

```release-note
Add broker ingress liveness probe
```

**Docs**


pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -34,6 +35,20 @@ func MakeIngressDeployment(args IngressArgs) *appsv1.Deployment {
...
 	container.Ports = append(container.Ports, corev1.ContainerPort{Name: "http", ContainerPort: int32(args.Port)})
+	container.LivenessProbe = &corev1.Probe{
+		Handler: corev1.Handler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path:   "/healthz",
+				Port:   intstr.FromInt(args.Port),
+				Scheme: corev1.URISchemeHTTP,
+			},
+		},
+		FailureThreshold:    3,
```
grantr, line 46: Where did these numbers come from?

yolocs: Used the same from Eventing core broker.

nachocano: I recall I had some problems with these numbers on the old implementation... https://github.com/knative/eventing/issues/2060

yolocs: Ack. The fact that Eventing core hasn't fixed the issue might be an indicator that this is not a common problem? We can always adjust these numbers if we find problem.

Issue #1059, opened 2020-05-13
Add  liveness probes to Broker fanout/retry


liu-cong:
**Problem**
Similar to #882 but probably more important as ingress error are more easily surfaced to the user.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

grantr:
Readiness might only make sense if a service is targeting.

PR #1185, opened 2020-06-01
Add liveness probes for fanout and retry

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1059

## Proposed Changes

- Add a health check server along with the handler sync pool; it keeps the last sync time and report "unhealthy" if there has been no sync for over a threshold.
- Add liveness probe to brokercell deployment template

It's hard to say if this is effective (enough). Open to suggestions :)

**Release Note**

```release-note
Add liveness probes for fanout and retry
```

**Docs**


pkg/broker/handler/pool/pool.go
```diff
@@ -27,25 +30,86 @@ type SyncPool interface {
...
+func (c *healthChecker) start(ctx context.Context) {
+	c.reportHealth()
+	// TODO: allow changing port? It's pure internal though.
+	srv := &http.Server{
+		Addr:    ":8080",
+		Handler: c,
+	}
+
+	go func() {
+		logging.FromContext(ctx).Info("Starting the sync pool health checker...")
```
grantr, line 60: Maybe better as Debug?

cmd/broker/retry/main.go
```diff
@@ -50,6 +50,9 @@ type envConfig struct {
 	// 3Mi. We also want to limit the memory usage from each subscription.
 	OutstandingBytesPerSub int `envconfig:"OUTSTANDING_BYTES_PER_SUB" default:"3000000"`
 
+	// MaxStaleDuration is the max duration of the handler pool without being synced.
+	MaxStaleDuration time.Duration `envconfig:"MAX_STALE_DURATION" default:"1m"`
```
grantr, line 54: What's the reasoning behind this value or is it arbitrary?

pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -58,6 +59,20 @@ func MakeFanoutDeployment(args FanoutArgs) *appsv1.Deployment {
 			corev1.ResourceCPU:    resource.MustParse("1500m"),
 		},
 	}
+	container.LivenessProbe = &corev1.Probe{
+		Handler: corev1.Handler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path:   "/healthz",
+				Port:   intstr.FromInt(8080),
```
grantr, line 66: We should probably add 8080 to the container's port list.

pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -73,6 +88,20 @@ func MakeRetryDeployment(args RetryArgs) *appsv1.Deployment {
 			corev1.ResourceCPU:    resource.MustParse("1000m"),
 		},
 	}
+	container.LivenessProbe = &corev1.Probe{
+		Handler: corev1.Handler{
+			HTTPGet: &corev1.HTTPGetAction{
+				Path:   "/healthz",
+				Port:   intstr.FromInt(8080),
```
grantr, line 95: We should probably add 8080 to the container's port list.

pkg/broker/handler/pool/pool.go
liu-cong: I think this is only called once so an Info is OK.

cmd/broker/fanout/main.go
```diff
@@ -43,6 +43,9 @@ type envConfig struct {
 	HandlerConcurrency     int    `envconfig:"HANDLER_CONCURRENCY"`
 	MaxConcurrencyPerEvent int    `envconfig:"MAX_CONCURRENCY_PER_EVENT"`
 
+	// MaxStaleDuration is the max duration of the handler pool without being synced.
+	MaxStaleDuration time.Duration `envconfig:"MAX_STALE_DURATION" default:"1m"`
```
liu-cong, line 47: We should validate that this value is greater than 15s (the sync period).

pkg/broker/handler/pool/pool.go
```diff
@@ -27,25 +30,86 @@ type SyncPool interface {
...
+
+func (c *healthChecker) lastTime() time.Time {
+	c.mux.RLock()
+	defer c.mux.RUnlock()
+	return c.lastReportTime
+}
+
+func (c *healthChecker) start(ctx context.Context) {
+	c.reportHealth()
+	// TODO: allow changing port? It's pure internal though.
```
liu-cong, line 53: Using a fixed port can lead to flaky tests if the port is not available during tests.

cmd/broker/retry/main.go
liu-cong: There is a maximum sync period of 15s. So I think the value should be greater than that. `1m` means we fail the probe if at least 4 sync loops have failed.

pkg/broker/handler/pool/pool.go
```diff
@@ -27,25 +30,86 @@ type SyncPool interface {
...
 }
 
-func watch(ctx context.Context, syncPool SyncPool, syncSignal <-chan struct{}) {
+func watch(ctx context.Context, syncPool SyncPool, syncSignal <-chan struct{}, c *healthChecker) {
 	for {
 		select {
 		case <-ctx.Done():
 			return
 		case <-syncSignal:
 			if err := syncPool.SyncOnce(ctx); err != nil {
```
liu-cong, line 108: Is an error equal to not live? The question is, do we think restarting the pod (due to liveness probe failure) is the right thing to do if we the sync fails? Or do we think it's better for the pool to keep trying?

On a side note, it seems that the SyncOnce method never returns an error... The `errs` counter is always 0 I believe.

pkg/broker/handler/pool/pool.go
yolocs: Yes, it will be once.

cmd/broker/retry/main.go
yolocs: That's right. I will add more comment.

pkg/broker/handler/pool/pool.go
yolocs: You mean in unit test? Unless we assign random port in the test you're referring, it won't make a difference right?

pkg/broker/handler/pool/pool.go
yolocs: We should write `SyncOnce` to be robust enough to not easily return error. As you noticed, we don't really return error today (let me also remove the useless `errs`). The error should only be returned on unrecoverable failures in which case we would want a pod restart. Make sense?

pkg/broker/handler/pool/pool.go
liu-cong: Yeah in unit tests. And it's pretty easy to run into that.

You can make the http server a dependency and use httptest.NewServer in tests while still using a fixed port in real code.

yolocs: I still think it makes sense to have the sync pool own the lifecycle of the health check server. Using httptest.Server makes it difficult to do that (it's not an interface). So instead I allow it to change health check port.

pkg/broker/handler/pool/pool_test.go
```diff
@@ -81,3 +123,18 @@ func (p *fakeSyncPool) SyncOnce(ctx context.Context) error {
...
+	addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
+	if err != nil {
+		return 0, err
+	}
+
+	l, err := net.ListenTCP("tcp", addr)
+	if err != nil {
+		return 0, err
+	}
+	defer l.Close()
```
liu-cong, line 138: Should let the caller close it, right?

yolocs: This is a trick to get a random port. If we don't close here, the http server won't be able to use this port.

liu-cong: Oh I see, you close the listener so the port can be used by the test. I guess in theory there could be a race that something else claims the port as soon as it's released here but it may never happen in reality.

pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -73,6 +73,26 @@ func MakeFanoutDeployment(args FanoutArgs) *appsv1.Deployment {
 			corev1.ResourceCPU:    resource.MustParse("1500m"),
 		},
 	}
+	container.Ports = append(container.Ports,
+		corev1.ContainerPort{
+			Name:          "http-health",
+			ContainerPort: 8080,
```
liu-cong, line 79: nit: Use the `DefaultHealthCheckPort` constant.

pkg/broker/handler/pool/pool.go
grantr: If I'm interpreting the probe settings correctly, the container will only be restarted if this fails 3 times in a row.

PR #1188, opened 2020-06-02
Brokercell controller watches data plane components

### LINKED ISSUE #912: BrokerCell watch ingress endpoints and update READY status immediately
**Problem**
Currently broker controller relies on the 5min global resync to update broker status if ingress endpoint status changes. 

**Exit Criteria**
* Broker status changes immediately as its ingress endpoint status changes.
* Is unit test/integ test applicable?

liu-cong:
Fixes #912

I manually verified the change. Not sure how to test this automatically.

## Proposed Changes

- Brokercell watches the components it creates to update status imnmediately
-
-

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/brokercell/controller.go
```diff
@@ -67,7 +71,27 @@ func NewController(
...
+		if mo, ok := obj.(metav1.Object); ok {
+			bcName, exist := mo.GetLabels()[resources.BrokerCellLabelKey]
+			if !exist {
+				// This object is not created by brokercell
+				return
+			}
+			impl.EnqueueKey(types.NamespacedName{Namespace: mo.GetNamespace(), Name: bcName})
+		}
+	})
+}
```
yolocs, line 97: Can you use something like this? Since they're "children" of the brokercell.

```
cache.FilteringResourceEventHandler{
  FilterFunc: controller.FilterGroupKind(appsv1.Kind("Deployment")),
  Handler:    controller.HandleAll(impl.EnqueueControllerOf),
}
```

liu-cong: Unfortunately this doesn't work for endpoints (the endpoints are indirectly created by the ingress service and therefore doesn't have the owner reference). I could do this for other resources if you prefer, but the current approach is generic for all resource created by brokercell.

yolocs: But the annotations will be passed to the endpoint? Interesting. Maybe add a comment explaining why of doing this.

liu-cong: In fact, I found an existing helper function that does similar to what I do. Changed to that :)

I suppose we don't need a `FilterFunc` here. The filter func is applied to the object in the informer, for instance, in the hpaInformer. I think the objects we get from the hpsInformer should be an HPA object.

yolocs: Nice. Good to know.

Issue #962, opened 2020-04-28
GCP Broker can't show reason when it's not ready


capri-xiyue:
**Describe the bug**
GCP Broker can't show reason when it's not ready. I have to run `kubectl describer broker XXX` to check the root cause why it's not ready.
<img width="478" alt="Screen Shot 2020-04-28 at 12 21 32 PM" src="https://user-images.githubusercontent.com/52932582/80537295-a402dd80-8958-11ea-8fd7-6ffa04608325.png">
<img width="1012" alt="Screen Shot 2020-04-28 at 12 22 14 PM" src="https://user-images.githubusercontent.com/52932582/80537300-a6653780-8958-11ea-933b-6f3beb8f1f1c.png">

 **Expected behavior**
GCP Broker will be able to show reason why it's not ready.

 **To Reproduce**
Create a GCP broker without control plane set up.(Missing secret `google-cloud-key` in namespace `cloud-run-events` and disable workload identity)

 **Knative-GCP release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

capri-xiyue:
https://docs.google.com/document/d/1-I26hHg1KxAuOW9CxWNTBmHz-qAlb5CZInDhR0OdQ20/edit?usp=sharing

liu-cong:
It seems that Triggers have the same issue: The `REASON` and `SUBSCRIBER_URI` columns are empty. Maybe they can be fixed together. If not, let's create a new issue for Triggers.

```
k get triggers.eventing.knative.dev --all-namespaces
NAMESPACE                  NAME              READY   REASON   BROKER        SUBSCRIBER_URI   AGE
cloud-run-events-example   goodbye-display   True             test-broker                    30m
cloud-run-events-example   hello-display     True             test-broker                    30m
```

capri-xiyue:
> It seems that Triggers have the same issue: The `REASON` and `SUBSCRIBER_URI` columns are empty. Maybe they can be fixed together. If not, let's create a new issue for Triggers.
> 
> ```
> k get triggers.eventing.knative.dev --all-namespaces
> NAMESPACE                  NAME              READY   REASON   BROKER        SUBSCRIBER_URI   AGE
> cloud-run-events-example   goodbye-display   True             test-broker                    30m
> cloud-run-events-example   hello-display     True             test-broker                    30m
> ```

Should it show reason when the `Ready` status is True?
I remember last time when I checked Trigger, it will show `Reason` when the `Ready` is `False`

[...]

PR #1189, opened 2020-06-02
Mark sub-status when GCP Broker is not ready

grac3gao:
related #962

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/broker/broker.go
```diff
@@ -287,7 +288,7 @@ func (r *Reconciler) reconcileDecouplingTopicAndSubscription(ctx context.Context
 		topic, err = client.CreateTopicWithConfig(ctx, topicID, topicConfig)
 		if err != nil {
 			logger.Error("Failed to create Pub/Sub topic", zap.Error(err))
-			b.Status.MarkTopicFailed("CreationFailed", "Topic creation failed: %w", err)
+			b.Status.MarkTopicFailed("TopicCreationFailed", "Topic creation failed: %w", err)
```
yolocs, line 291: Surprising to me that we don't need to change unit test accordingly?

liu-cong: I don't think we test this case currently. Maybe it's not easy to inject an error in pstest?

Issue #930, opened 2020-04-23
GCP Broker Component Level Benchmarks


liu-cong:
**Problem**
Add micro benchmarks for each component
 * These tests help identify bottlenecks easily
 * Ideally these tests can run locally so developers can use them to find performance improvements/regression
 * Nice-to-haves
   * Enable a command such as /benchmark-broker-ingress in github to trigger a test against a PR. this is a convenient tool to use for any PRs that may affect performance.
   * Run a continuous job and integrate with Mako so we have a history

**Components**
- [x] Delivery Processor #1078
- [x] Ingress Handler #1108 
- [ ] fanout processor
- [ ] filter processor
- [ ] fanout pool
- [ ] retry pool

Note: May be benchmarking the pools will also include the processors. In that case individual benchmarks for the processor can be skipped.

yolocs:
It would be helpful to define "performance". Historically we have latency test in Mako which is only measured by one client seeding events and one subscriber. But that doesn't seem to provide good simulation of the system running under heavy load or with various subscribers (e.g. always returning error, slow response, etc.). I would also imagine a proper simulation would incur quite some cost especially when we run it continuously.

PR #1195, opened 2020-06-02
adding test for broker-gcp with steady rps rate

nlopezgi:
Moves forward #876 and #930 

## Proposed Changes

- Adds a new benchmark for broker-gcp that runs at a steady pace for 5 minutes. Steady pace is set at 500 rps which is below the max rps which increasing pace test shows can be safely handled. (this is very much open to discussion, tuning)
- Increases the max rps for the existing broker-gcp tests (increased to 800 max, I want to see it hit the pole)

Reasoning: 
- See [here](http://highscalability.com/blog/2015/10/5/your-load-generator-is-probably-lying-to-you-take-the-red-pi.html) for a conceptual talk about doing performance testing.
- See [here](https://github.com/slinkydeveloper/eventing-bench-results/blob/master/11-14-2019/README.md#results) for a recent performance analysis of knative eventing components.

Key takeaways from these two references (as related to this PR):
- We should not be measuring latencies with a performance test that has an increasing rps pace
- The tests we have (increasing rps pace) are useful, but for a different purpose (to find the breaking point of the SUT). However, it seems to me the current pace does not achieve that goal.

Will update PR with some sample runs (on dev env) once they start coming through on Mako.

### DIFF
--- a/test/performance/benchmarks/broker-gcp-steady/continuous/100-broker-gcp-setup.yaml
+++ b/test/performance/benchmarks/broker-gcp-steady/continuous/100-broker-gcp-setup.yaml
@@ -1,0 +1,90 @@
+# Copyright 2019 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: eventing.knative.dev/v1beta1
+kind: Broker
+metadata:
+  name: gcp
+  namespace: default
+  annotations:
+    eventing.knative.dev/broker.class: googlecloud
+spec: {}
+
+---
+
+apiVersion: eventing.knative.dev/v1beta1
+kind: Trigger
+metadata:
+  name: broker-gcp
+  namespace: default
+spec:
+  broker: gcp
+  subscriber:
+    ref:
+      apiVersion: v1
+      kind: Service
+      name: broker-gcp-receiver
+
+---
+
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: perf-gcpbroker
+  namespace: default
+
+---
+
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRoleBinding
+metadata:
+  name: perf-gcpbroker
+subjects:
+  - kind: ServiceAccount
+    name: perf-gcpbroker
+    namespace: default
+roleRef:
+  kind: ClusterRole
+  name: cluster-admin
+  apiGroup: rbac.authorization.k8s.io
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: broker-gcp-receiver
+spec:
+  selector:
+    role: broker-gcp-receiver
+  ports:
+    - name: http
+      port: 80
+      targetPort: cloudevents
+      protocol: TCP
+
+---
+
+apiVersion: v1
+kind: Service
+metadata:
+  name: broker-gcp-aggregator
+spec:
+  selector:
+    role: broker-gcp-aggregator
+  ports:
+    - name: grpc
+      port: 10000
+      targetPort: grpc
+      protocol: TCP
--- a/test/performance/benchmarks/broker-gcp-steady/continuous/200-broker-gcp.yaml
+++ b/test/performance/benchmarks/broker-gcp-steady/continuous/200-broker-gcp.yaml
@@ -45,7 +45,7 @@ spec:
               - "--roles=sender,receiver"
               - "--sink=http://broker-ingress.cloud-run-events/default/gcp"
               - "--aggregator=broker-gcp-aggregator:10000"
-              - "--pace=100:10,200:20,400:30,500:60,600:60,700:60,800:60"
+              - "--pace=500:300"
             env:
               - name: POD_NAME
                 valueFrom:
--- a/test/performance/benchmarks/broker-gcp-steady/dev.config
+++ b/test/performance/benchmarks/broker-gcp-steady/dev.config
@@ -1,0 +1,76 @@
+# Create this benchmark with the mako tool: mako create_benchmark dev.config
+# Update this benchmark with the mako tool: mako update_benchmark dev.config
+# Learn more about the mako tool at
+# https://github.com/google/mako/blob/master/docs/CLI.md.
+
+project_name: "Knative"
+benchmark_name: "Development - GCP Broker Latency & Throughput at Steady Pace"
+description: "Measure latency and throughput of the Google Cloud Broker at Steady Pace."
+
+# Human owners for manual benchmark adjustments.
+owner_list: "bkaplan@google.com"
+owner_list: "chizhg@google.com"
+owner_list: "conliu@google.com"
+owner_list: "cshou@google.com"
+owner_list: "danyinggu@google.com"
+owner_list: "elemar@google.com"
+owner_list: "gracegao@google.com"
+owner_list: "grantrodgers@google.com"
+owner_list: "harwayne@google.com"
+owner_list: "ianmi@google.com"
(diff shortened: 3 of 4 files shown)


nlopezgi:
A nice run with steady rps pace and a latency spike that would be interesting to explore more: https://mako.dev/run?run_key=5206983072284672&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1

A run pushing the rps to 800 that shows the broker can still be pushed even further: https://mako.dev/run?run_key=6483677569613824&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1

nlopezgi:
broker-gcp is still holding at 1.2k rps (https://mako.dev/run?run_key=4787605419851776&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1) Will push further and update PR once I found a limit

ian-mi:
Awesome, thanks for moving this forward! Have you investigate whether it would be possible for mako to automatically increase load until the maximum sustainable throughput is found? Or do you propose using the old tests for this purpose?

nlopezgi:
> Awesome, thanks for moving this forward! Have you investigate whether it would be possible for mako to automatically increase load until the maximum sustainable throughput is found? Or do you propose using the old tests for this purpose?

I have not seen a way to do this automatically. It seems we will need to maintain at least two benchmarks for each SUT, one to find the max throughput and one to test latency at close to that max.

yolocs:
> broker-gcp is still holding at 1.2k rps ([mako.dev/run?run_key=4787605419851776&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1](https://mako.dev/run?run_key=4787605419851776&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1)) Will push further and update PR once I found a limit

I expect on happy path, it will go beyond the limit of the aggregator. Also I think these we have HPA for broker.

nlopezgi:
> > broker-gcp is still holding at 1.2k rps ([mako.dev/run?run_key=4787605419851776&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1](https://mako.dev/run?run_key=4787605419851776&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1)) Will push further and update PR once I found a limit
> 
> I expect on happy path, it will go beyond the limit of the aggregator. Also I think these days we have HPA for broker.

I'll keep pushing, if we find the aggregator is the limit, then this benchmark (the one with increasing pace) is pretty useless to have running, imo

nlopezgi:
So at 2.5k rps things do get really messy (https://mako.dev/run?run_key=4688571392851968&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1), and imo this would be what I would expect to see if the SUT was hitting its limits (as opposed to the aggregator), wdyt @yolocs ?

yolocs:
> So at 2.5k rps things do get really messy ([mako.dev/run?run_key=4688571392851968&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1](https://mako.dev/run?run_key=4688571392851968&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1)), and imo this would be what I would expect to see if the SUT was hitting its limits (as opposed to the aggregator), wdyt @yolocs ?

By just looking at the chart I can't really tell what's causing the noise. I wonder how you tell if it's aggregator's problem or not? Do you by any chance have access to the cluster? It seems to me SUT can still handle the load. Maybe give it more load and see what will happen?

grantr:
A  previous test measured the test infrastructure's capacity as 10k rps, though that test doesn't run regularly so I'm not sure what its capacity is today. https://github.com/knative/eventing/tree/master/test/performance/direct

ian-mi:
To obtain consistent results it would be ideal to deploy broker components with autoscaling disabled and with hard resource limits. If we are hitting test infrastructure limits we can then reduce resource limits until the broker component under test becomes the bottleneck.

grantr:
@ian-mi beat me to it, but: does it make sense to limit the broker HPAs to 1 or N replicas for these tests? Or maybe that should be a different test?

Practically we may not be able to do this without changes to BrokerCell controller because it reconciles HPA maxreplicas.

nlopezgi:
@yolocs 

> By just looking at the chart I can't really tell what's causing the noise. I wonder how you tell if it's aggregator's problem or not? Do you by any chance have access to the cluster? It seems to me SUT can still handle the load. Maybe give it more load and see what will happen?

I do have access to the cluster as this is running in my dev project. The aggregator logs look ok and show no errors occurred (I'm comparing this to how I saw the aggregator behave when I created more than 10 receivers and a single sender and it seemed it was getting swamped by the receiver messages). A recent log (https://mako.dev/run?run_key=5688279556947968&~pl=1&~pe=1&~st=1&~dl=1&~de=1&~dt=1&~pet=1&~det=1) shows how the latency spikes to over 5s at around 2.1k rps. I dont know enough about this system to know how its expected to scale, but to me this seems like the SUT is not able to keep up with that load and latency is going way higher than it should be expected. But I agree this might be some other part of the test infra that is not scaling (the sender or receiver, though those logs also look fine) other than the SUT.
 
@grantr 
> A previous test measured the test infrastructure's capacity as 10k rps, though that test doesn't run regularly so I'm not sure what its capacity is today. https://github.com/knative/eventing/tree/master/test/performance/direct

what do you mean here by "the test infrastructure" and how was the setup there different than here? For different setups you will probably get vastly different results. With this setup it seems I'm hitting some limit at 2.1-2.5k rps. I'm also surprised you mention numbers that high as none of the benchmarks in knative/eventing also come close to trying that load (imc channel goes up to 1.5k - https://github.com/knative/eventing/blob/master/test/performance/benchmarks/channel-imc/continuous/200-channel-imc.yaml).  
Could you point me to the 10k test results it so I can try to replicate them?
So far I was looking at the best benchmark/analysis I have seen, which is Francesco's (https://github.com/slinkydeveloper/eventing-bench-results/blob/master/11-14-2019/README.md#results). I'd be happy to try something different, but once we settle on a setup, and tools for testing (so far Mako is all I see as a possibility for sustainably running tests and collecting their data, which is a premise for anything to be useful for anything else other than a snapshot of time), I plan on trying to roll it out consistently for all knative/eventing and knative-gcp benchmarks, otherwise all these benchmarks are pretty pointless imo (or I might be missing something altogether about their purpose).

@ian-mi 
> To obtain consistent results it would be ideal to deploy broker components with autoscaling disabled and with hard resource limits. If we are hitting test infrastructure limits we can then reduce resource limits until the broker component under test becomes the bottleneck.

I guess I was assuming these tests were better setup than they really are. I'd be happy to modify them to disable autoscaling (not sure where to start looking for the knobs for that, if you could send me a pointer, that would be great). 

> @ian-mi beat me to it, but: does it make sense to limit the broker HPAs to 1 or N replicas for these tests? Or maybe that should be a different test?
> 
> Practically we may not be able to do this without changes to BrokerCell controller because it reconciles HPA maxreplicas.

If we cant really control the SUT in a way that allows us to strictly control resources then, imo, having these tests at all is pretty pointless.

Overall I think this PR is highlighting we have some deep gaps in our current benchmarks. I don't think I understand what value the existing ones provide nor what value we want them to actually provide. I'll set up a brief sync with everyone involved to try to get some clarification.

grantr:
> what do you mean here by "the test infrastructure" and how was the setup there different than here?

The test infrastructure I'm referring to is a single pod sending events to itself in a loopback configuration. There is no SUT, or more accurately, the SUT is the event sender and event receiver. Details are in https://github.com/knative/eventing/blob/master/test/performance/direct/200-direct-perf.yaml.

I don't have a record of test results, so I retract my 10k number :smiley: but if this loopback test seems valuable we could easily turn it into a real Mako benchmark.

grantr:
The autoscaling of broker components is quite new so it's not surprising to me that it needs a bit of work to align well with perf testing. It's probably a good idea to make the maximum scale of each Broker deployment configurable. However, even without this improvement we can use [quotas](https://kubernetes.io/docs/concepts/policy/resource-quotas/#compute-resource-quota) to limit the resources available to the broker deployments.

nlopezgi:
Closing, this PR was meant to elicit conversation first. Will come back later with an updated one.

PR #1206, opened 2020-06-03
Use "trigger" prefix for fanout & retry metrics

zargarpur:
Currently, the metric names that attempt to be emitted for fanout &
retry use their component names, so they end up being:
- knative.dev/internal/eventing/broker_fanout
- knative.dev/internal/eventing/broker_retry

But we expect the metric names to start with:
- knative.dev/internal/eventing/trigger


pkg/utils/mainhelper/init.go
```diff
@@ -49,7 +49,7 @@ type InitRes struct {
 // it returns a result object that contains useful artifacts for later use.
 // Unlike sharedmain.Main, Init is meant to be run as a helper function in any main
 // functions, while sharedmain.Main runs controllers with predefined method signatures.
-func Init(component string, opts ...InitOption) (context.Context, *InitRes) {
```
liu-cong, line 52: How about making metricNamespace an option, and by default it's the same as `component`? e.g.m Init(component, WithMetricNamespace("trigger"))

zargarpur: Updated, great suggestion!

pkg/utils/mainhelper/options.go
```diff
@@ -54,6 +55,9 @@ func newInitArgs(component string, opts ...InitOption) initArgs {
 		// Set up signals so we handle the first shutdown signal gracefully.
 		args.ctx = signals.NewContext()
 	}
+	if args.metricNamespace == "" {
```
liu-cong, line 58: Can you add a unit test for the new option? Otherwise lgtm

cmd/broker/ingress/main.go
```diff
@@ -63,7 +62,7 @@ func main() {
 		ingress.Port(env.Port),
 		ingress.ProjectID(projectID),
 		metrics.PodName(env.PodName),
-		metrics.ContainerName(containerName),
+		metrics.ContainerName(component),
```
nachocano, line 65: I'd add a comment of why we are doing this...

pkg/utils/mainhelper/options.go
zargarpur: Done

cmd/broker/ingress/main.go
zargarpur: I just made this consistent with what the fanout & retry files do, I think ContainerName was a way to include "ingress" as the container name while keeping the metrics working with the "broker" component, but now we can specify the metric name via this PR

PR #1212, opened 2020-06-04
Drop non-event messages

### LINKED ISSUE #1178: Non-event message in pubsub causes fanout/retry to panic (SDK bug)
**Describe the bug**
NOTE: This is a CE SDK bug but let's track it here. 

I manually published some random messages (not a CloudEvent) in the retry topic, retry encountered panic and the pod run into CrashLoop due to container keeps failing

```
$ k get pod

default-brokercell-fanout-6499f4c977-dwwh7    0/1     CrashLoopBackOff   5          4m19s

$ k describe pod broker-fanout

Warning  BackOff    2m43s (x10 over 4m35s)  kubelet, gke-dev-default-pool-ffb946b5-b54z  Back-off restarting failed container
```

Stacktrace:
```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0x9ff4ff]

goroutine 148 [running]:
github.com/cloudevents/sdk-go/v2/binding.(*messageToEventBuilder).SetStructuredEvent(0xc000ea6880, 0x1d7c780, 0xc000058018, 0x0, 0x0, 0x1d34800, 0xc000eae960, 0x1990180, 0xbde)
        github.com/cloudevents/sdk-go/v2@v2.0.0-RC4/binding/to_event.go:70 +0xaf
github.com/cloudevents/sdk-go/v2/protocol/pubsub.(*Message).ReadStructured(0xc000eae930, 0x1d7c780, 0xc000058018, 0x1d34f40, 0xc000ea6880, 0x19e22d6, 0x3)
        github.com/cloudevents/sdk-go/v2@v2.0.0-RC4/protocol/pubsub/message.go:71 +0xf4
github.com/cloudevents/sdk-go/v2/binding.DirectWrite(0x1d7c780, 0xc000058018, 0x1d64c40, 0xc000eae930, 0x1d34f40, 0xc000ea6880, 0x1d84dc0, 0xc000ea6880, 0x0, 0x0, ...)
        github.com/cloudevents/sdk-go/v2@v2.0.0-RC4/binding/write.go:35 +0x24e
github.com/cloudevents/sdk-go/v2/binding.ToEvent(0x1d7c740, 0xc00031d5c0, 0x1d64c40, 0xc000eae930, 0x0, 0x0, 0x0, 0x17f3ce0, 0x60dc54, 0xc000c30140)
        github.com/cloudevents/sdk-go/v2@v2.0.0-RC4/binding/to_event.go:47 +0x2b7
github.com/google/knative-gcp/pkg/broker/handler.(*Handler).receive(0xc00031d500, 0x1d7c740, 0xc00031d5c0, 0xc000c4e280)
        github.com/google/knative-gcp/pkg/broker/handler/handler.go:74 +0x9c
cloud.google.com/go/pubsub.(*Subscription).Receive.func2.2(0x17ce1e0, 0xc000c4e280)
        cloud.google.com/go/pubsub@v1.3.2-0.20200506222144-2c46308f8465/subscription.go:793 +0x80
cloud.google.com/go/pubsub/internal/scheduler.(*ReceiveScheduler).Add.func1(0xc000eae900, 0x17ce1e0, 0xc000c4e280, 0xc0008ba0e0)
        cloud.google.com/go/pubsub@v1.3.2-0.20200506222144-2c46308f8465/internal/scheduler/receive_scheduler.go:82 +0x3a
created by cloud.google.com/go/pubsub/internal/scheduler.(*ReceiveScheduler).Add
        cloud.google.com/go/pubsub@v1.3.2-0.20200506222144-2c46308f8465/internal/scheduler/receive_scheduler.go:80 +0x367
```

 **To Reproduce**
* Create a GCP broker
* Find the topic in pubsub, publish a message such as "Hello World"

 **Additional context**

liu-cong:
Fixes #1178

## Proposed Changes

- If a message is not an event, we drop it, otherwise we will keep retrying indefinitely.
- For now we just log and drop it, we could consider moving it to a DLQ or some other persistence storage.
-

**Release Note**

```release-note
Drop non-event messages.
```

**Docs**


pkg/broker/handler/handler.go
```diff
@@ -74,8 +75,17 @@ func (h *Handler) IsAlive() bool {
 func (h *Handler) receive(ctx context.Context, msg *pubsub.Message) {
 	ctx = metrics.StartEventProcessing(ctx)
 	event, err := binding.ToEvent(ctx, cepubsub.NewMessage(msg))
+	// The following errors can be returned by ToEvent and are not retryable.
+	// TODO Should ToEvent consolidate them and return the genric ErrCannotConvertToEvent?
+	if errors.Is(err, binding.ErrCannotConvertToEvent) || errors.Is(err, binding.ErrNotStructured) || errors.Is(err, binding.ErrUnknownEncoding) || errors.Is(err, binding.ErrNotBinary) {
+		logging.FromContext(ctx).Error("failed to convert received message to an event, check the msg format", zap.Any("message", msg), zap.Error(err))
```
grantr, line 81: Max message data size is 10MB. Worst case, if all messages in the topic are the wrong format and very large we use a lot of resources logging them. Maybe print the message id and first N bytes of data?

pkg/broker/handler/handler.go
```diff
@@ -74,8 +75,17 @@ func (h *Handler) IsAlive() bool {
...
+		// Ack the message so it won't be retried.
+		// TODO Should this go to the DLQ once DLQ is implemented?
+		//msg.Ack()
+		//return
+	}
 	if err != nil {
-		logging.FromContext(ctx).Error("failed to convert received message to an event", zap.Any("message", msg), zap.Error(err))
+		logging.FromContext(ctx).Error("unknown error when converting the received message to an event", zap.Any("message", msg), zap.Error(err))
```
grantr, line 88: Same logging comment as above

pkg/broker/handler/handler.go
```diff
@@ -74,8 +75,17 @@ func (h *Handler) IsAlive() bool {
 func (h *Handler) receive(ctx context.Context, msg *pubsub.Message) {
 	ctx = metrics.StartEventProcessing(ctx)
 	event, err := binding.ToEvent(ctx, cepubsub.NewMessage(msg))
+	// The following errors can be returned by ToEvent and are not retryable.
+	// TODO Should ToEvent consolidate them and return the genric ErrCannotConvertToEvent?
+	if errors.Is(err, binding.ErrCannotConvertToEvent) || errors.Is(err, binding.ErrNotStructured) || errors.Is(err, binding.ErrUnknownEncoding) || errors.Is(err, binding.ErrNotBinary) {
+		logging.FromContext(ctx).Error("failed to convert received message to an event, check the msg format", zap.Any("message", msg), zap.Error(err))
+		// Ack the message so it won't be retried.
+		// TODO Should this go to the DLQ once DLQ is implemented?
+		//msg.Ack()
```
grantr, line 84: Was this intended to be commented out?

pkg/broker/handler/handler_test.go
```diff
@@ -122,12 +122,26 @@ func TestHandler(t *testing.T) {
...
+			t.Fatalf("failed to seed event to pubsub: %v", err)
+		}
+		gotEvent := nextEventWithTimeout(eventCh)
+		// The message should be Acked and should not reach the processor
+		if gotEvent != nil {
+			t.Errorf("processor should receive 0 events but got: %+v", gotEvent)
+		}
+		ctx, _ := context.WithTimeout(context.Background(), time.Millisecond)
+		msg, err := p.Receive(ctx)
+		t.Errorf("msg: %v, err: %v", msg, err)
```
grantr, line 136: Looks like this will change once the sdk fix is in place

pkg/broker/handler/handler.go
```diff
@@ -74,8 +75,17 @@ func (h *Handler) IsAlive() bool {
 func (h *Handler) receive(ctx context.Context, msg *pubsub.Message) {
 	ctx = metrics.StartEventProcessing(ctx)
 	event, err := binding.ToEvent(ctx, cepubsub.NewMessage(msg))
+	// The following errors can be returned by ToEvent and are not retryable.
+	// TODO Should ToEvent consolidate them and return the genric ErrCannotConvertToEvent?
+	if errors.Is(err, binding.ErrCannotConvertToEvent) || errors.Is(err, binding.ErrNotStructured) || errors.Is(err, binding.ErrUnknownEncoding) || errors.Is(err, binding.ErrNotBinary) {
```
yolocs, line 80: nit: will be clearer if wrapped to a func

pkg/broker/handler/handler.go
yolocs: Maybe log both error and debug. Error has the shorter message and debug has the full message?

pkg/broker/handler/handler.go
yolocs: If we log the full message, then I think we should ack the message now. The reason being 1) retry such messages will continue to fail so it doesn't make sense to do so 2) since we log the message (with debug), developers can still check what's in the message that was causing the problem

liu-cong: oops, sorry forgot to uncomment, was doing some testing.

pkg/broker/handler/handler.go
liu-cong: I was debating whether to log the full msg. This should really not happen in normal conditions. Since we don't have a DLQ, without logging it's difficult for the user to debug.

Maybe picking a big N such as 5000 is good enough to cover most cases?

liu-cong: Log full msg in debug, and truncate the Data field to 2000 chars in error (other fields are either fixed size or a map of attributes)

pkg/broker/handler/handler.go
```diff
@@ -102,3 +100,22 @@ func (h *Handler) receive(ctx context.Context, msg *pubsub.Message) {
...
+func logEventConversionError(ctx context.Context, pm *pubsub.Message, err error, msg string) {
+	maxLen := 2000
+	truncated := pm
+	if len(pm.Data) > maxLen {
+		copy := *pm
+		copy.Data = copy.Data[:maxLen]
+		truncated = &copy
+	}
+	logging.FromContext(ctx).Debug(msg, zap.Any("message", pm), zap.Error(err))
+	logging.FromContext(ctx).Error(msg, zap.Any("message", truncated), zap.Error(err))
```
grantr, line 120: Optional, but maybe suggest in the field key that the message data is truncated by calling it `message-truncated`?

liu-cong: Good idea, done

Issue #1100, opened 2020-05-20
Broker retry retries sending failed events immediately


liu-cong:
**Problem**
A short explanation of the problem, including relevant restrictions.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A retried message delivery that fails is retried only after some static timeout.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

yolocs:
The go pubsub client we're using doesn't provide a way to directly write AckDeadline. Instead it [manages the AckDeadline](https://godoc.org/cloud.google.com/go/pubsub#hdr-Deadlines) for us. We can directly call RPC to set the AckDeadline (e.g. [like how it's done in the client](https://github.com/googleapis/google-cloud-go/blob/master/pubsub/iterator.go#L424-L427)) but it's hard to tell how well it can work with the client we're using (which also sets the AckDeadline dynamically).

If we stick with the existing client, the only way to alternate AckDeadline seems to be 1) explicitly holding on to the messages and call `nack()` after a period or 2) implicitly holding on to the messages by not calling either `ack()` or `nack()` and messages will eventually expire after `MaxExtension`. Both workarounds will create "congestion" for a pull client as we also set `MaxOutstandingBytes` and `MaxOutstandingMessages`. It could work well in the case where a subscriber consistently rejects events (e.g. service not avaialble). But it could potentially cause unfairness when the subscriber accepts certain events but not others. A batch of unacceptable events could block future acceptable events being retried *efficiently*. 

For release 1, I think it's ok to proceed with the option 1 as it could provide knobs on how fast we `nack()`.

grantr:
Unfortunately calling `nack()` in the high-level client always sets the deadline to zero: https://github.com/googleapis/google-cloud-go/blob/df3d655c5e3fc7d1e62fabf96358c3a52a874f4c/pubsub/iterator.go#L343-L347. It would be a useful feature for the higher-level client to be able to nack with a timeout, or to enable the same sort of backoff as described in https://godoc.org/cloud.google.com/go/pubsub#hdr-Deadlines.

[...]

PR #1218, opened 2020-06-04
Add client side backoff retries for pubsub messages

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1100

## Additional Changes

- added handler constructor to initialize new fields
- removed `pool` package as it feels a bit unnature to have a deeper package `pkg/broker/handler/pool` depend on a parent package `pkg/broker/handler` implementation; no functional change involved

**Release Note**

```release-note
Add client side backoff retries for pubsub messages
```

**Docs**


pkg/broker/handler/fanout_test.go
```diff
@@ -28,7 +28,7 @@ import (
 
 	"github.com/google/knative-gcp/pkg/broker/config"
 	"github.com/google/knative-gcp/pkg/broker/eventutil"
-	pooltesting "github.com/google/knative-gcp/pkg/broker/handler/pool/testing"
+	pooltesting "github.com/google/knative-gcp/pkg/broker/handler/testing"
```
liu-cong, line 31:
```suggestion
	handlertesting "github.com/google/knative-gcp/pkg/broker/handler/testing"
```

pkg/broker/handler/handler_test.go
```diff
@@ -141,6 +138,91 @@ func TestHandler(t *testing.T) {
...
+	if len(delays) != desiredErrCount {
+		t.Errorf("retry count got=%d, want=%d", len(delays), desiredErrCount)
+	}
+	if delays[0] != time.Millisecond {
+		t.Errorf("initial nack delay got=%v, want=%v", delays[0], time.Millisecond)
+	}
+	// We expect exponential backoff until MaxBackoff
+	for i := 1; i < len(delays); i++ {
+		wantDelay := 2 * delays[i-1]
+		if wantDelay > 16*time.Millisecond {
```
liu-cong, line 217: this seems unnecessary. Looks like it can only happen when len(delays)>desiredErrCount, but we already have an `Errorf`, we can even make it `Fatalf`

pkg/broker/handler/handler_test.go
```diff
@@ -141,6 +138,91 @@ func TestHandler(t *testing.T) {
...
+	testEvent.SetSubject("subject")
+	testEvent.SetType("type")
+
+	if err := p.Send(ctx, binding.ToMessage(&testEvent)); err != nil {
+		t.Errorf("failed to seed event to pubsub: %v", err)
+	}
+
+	// Wait until all desired errors were returned.
+	// Then stop the handler by cancel the context.
+	time.Sleep(time.Second)
```
liu-cong, line 205: Instead of doing this, I recommend adding a channel in the fake processor to store the event. It should make the test faster and deterministic.

pkg/broker/handler/handler_test.go
```diff
@@ -141,6 +138,91 @@ func TestHandler(t *testing.T) {
...
+	processors.BaseProcessor
+	desiredErrCount, currErrCount int
+}
+
+func (p *alwaysErrProc) Process(_ context.Context, _ *event.Event) error {
+	if p.currErrCount < p.desiredErrCount {
+		p.currErrCount++
+		return errors.New("always error")
+	}
+	return nil
```
liu-cong, line 151: The processor can store the event in a channel once succeeding and the test can wait for the event.

pkg/broker/handler/handler.go
```diff
@@ -86,9 +107,13 @@ func (h *Handler) receive(ctx context.Context, msg *pubsub.Message) {
 		defer cancel()
 	}
 	if err := h.Processor.Process(ctx, event); err != nil {
-		logging.FromContext(ctx).Error("failed to process event", zap.Any("event", event), zap.Error(err))
+		backoffPeriod := h.retryLimiter.When(msg.ID)
+		logging.FromContext(ctx).Error("failed to process event; backoff nack", zap.Any("event", event), zap.Float64("backoffPeriod", backoffPeriod.Seconds()), zap.Error(err))
```
ian-mi, line 111: Use zap.Duration

pkg/broker/handler/handler_test.go
yolocs: This is necessary because I intentionally had enough errors to reach max delay more than once. This is to ensure the code never delays more than the max delay.

cmd/broker/retry/main.go
```diff
@@ -58,6 +58,9 @@ type envConfig struct {
 
 	// Max to 10m.
 	TimeoutPerEvent time.Duration `envconfig:"TIMEOUT_PER_EVENT"`
+
+	MinRetryBackoff time.Duration `envconfig:"MIN_RETRY_BACKOFF" default:"1s"`
+	MaxRetryBackoff time.Duration `envconfig:"MAX_RETRY_BACKOFF" default:"30s"`
```
liu-cong, line 63: I was wondering if this should be a bit longer, like a few minutes?

yolocs: How long would you suggest? I don't have a preference. I simply don't want to over blocking the queue because of the limitation we have today.

yolocs: Changed to 1m now.

liu-cong: I just think if the backoff reaches 30s, it's very likely the issue is not transient, and some longer backoff is probably better.

But just a guess. We can always tune this later. 1min seems good to me.

Issue #1210, opened 2020-06-04
Avoid creating ksvcs in Sources


nachocano:
**Problem**
Most of our Sources create a Topic object, which underneath ends up creating the topic in GCP. Since its inception, the Topic construct has been an Addressable. It has a data plane component, which is a Knative Service called publisher, which receives HTTP and outputs to PubSub. This publisher makes sense for Channels but not much for Sources. 

We haven't payed much attention to it in Sources as the ksvc created scales down to zero. But even so, we shouldn't be creating those in the first place.
One potential solution would be to change the ready condition of Topic to make the Addressable readiness part mandatory only for Channels but not for Sources. This is already distinguishable from labels that we are setting when creating the Topic object.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
Sources shouldn't create unnecessary ksvcs.

PR #1226, opened 2020-06-05
Avoid creating publishers when creating Topics from Sources

nachocano:
Fixes https://github.com/google/knative-gcp/issues/1210

## Proposed Changes

- 🐛 Changing the condition set of Topic to only have TopicExists. Addressable and PublisherReady do not inform readiness, unless the Topic was created from an object different than a Source.
- This will also help in the `gcloud trigger create`, as we were seeing some Publisher reconciliation errors.

**Release Note**

```release-note
CloudStorageSource, CloudSchedulerSource and CloudAuditLogsSource do not create a Knative Service along with a Topic.
```

**Docs**


pkg/reconciler/intevents/topic/topic.go
```diff
@@ -107,6 +108,12 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, topic *v1alpha1.Topic) r
 	// Set the topic being used.
 	topic.Status.TopicID = topic.Spec.Topic
 
+	// If the Topic was created from a Source, then we don't need to create the Publisher.
+	if _, ok := topic.Labels[SourceLabelKey]; ok {
```
Harwayne, line 112: My preference is to use a spec field to do this, rather than introspect the labels. Something like `spec.publisher`, which defaults to `true`, but will be set to `false` when creating Topics for sources.

nachocano:
Makes sense and I like the name as well

capri-xiyue:
Maybe I missed some context. So for Sources, we don't need topic with publisher?
In addition, when will we need topic with publisher?

nachocano:
> Maybe I missed some context. So for Sources, we don't need topic with publisher?
> In addition, when will we need topic with publisher?

for source we don't. For the channel we do and if you just create a Topic itself, although we make it internal now...

Issue #1192, opened 2020-06-02
GCP Broker doesn't work outside GCP


grantr:
**Describe the bug**
https://knative.slack.com/archives/C93E33SN8/p1591044047127700 reports that GCP Broker doesn't work on EKS. Broker reconcile failed:

`broker-controller  failed to reconcile broker: decoupling topic reconcile failed: metadata: GCE metadata "project/project-id"  not defined`

The reporter tried to set the `PROJECT_ID` env var on the controller deployment but this didn't fix the issue.

 **Expected behavior**
GCP Broker runs correctly on non-GCP compute resources.

 **To Reproduce**
> We followed all steps in README and installed everything using `ko apply`  with `KO_DOCKER_REPO=<ECR(AWS's container registry>` .

 **Knative-GCP release version**
0.15.0

 **Additional context**
The root cause is likely an assumption that the GCE metadata server is available.

This issue is scoped to broker specifically, but it's likely that other resources in this repo also don't work outside of GCP for the same reason.

Issue #1190, opened 2020-06-02
Shall Finalizer mark resource sub-conditions when it failed?


grac3gao:
**Problem**
In FinalizeKind (sources and broker). When deleting resources, we only return error without marking corresponding status failed (showing reasons like SubscriptionDeleteFailed). Thus, when users are trying to delete a resource and got stuck, they can’t track the reason through resource (resource still show ready is true if the last reconciling loop for it is true). 

Note: by checking kubectl describe, it will show a related warning event for that

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

grantr:
It might not be possible to update the status of a finalized resource. If not, we could emit an event for this case.

@grac3gao do you have any specific examples of issues with blocked finalized objects?

grac3gao:
According to the discussion in Eventing WG and issue https://github.com/knative/eventing/issues/3305
it is worthwhile to mark conditions in Finalizer. This issue will track the broker/trigger finalizer. #1294 is used to track finalizer in source and channel

PR #1239, opened 2020-06-08
Set correct broker status if topic is deleted

### LINKED ISSUE #1233: Should Mark Broker/Trigger not READY if topic is deleted by the user
**Describe the bug**

For context see #997. This issue is to mark the correct status, not to recover.

 **Expected behavior**

 **To Reproduce**

 **Additional context**

### LINKED ISSUE #1235: Update Broker status when the decouple pull sub has a deleted topic
**Describe the bug**
Broken out from #997.

When the Broker's decouple pull subscription has a deleted topic, the Broker status should be unready with an error message to the user about what went wrong and how to recover. This may later be recovered automatically, but for now we'll need users to manually fix it by deleting the existing pull sub.

 **Expected behavior**
After a user deletes the topic, the reconciler behavior we're aiming for is:
- Create a new topic. New messages will flow into this topic.
- Notice that the existing pull sub is for a deleted topic. Leave the existing pull sub in place. 
- Update the Broker's status to unready and direct users to delete the old pull subscription.

After this first reconcile, previously acknowledged messages from the old topic will continue to be delivered by the old pull sub, but new messages from the new topic will not be delivered (because there is no pull sub for it yet).

After the user deletes the old pull sub, a new one will be created on the next reconcile and the Broker's status can flip back to ready.

liu-cong:
Fixes #1233
Fixes #1235

## Proposed Changes

- If the topic of a subscription is "_deleted-topic_", mark Broker/Trigger condition false
- Also did some cleanup/refactoring
  - Moved topic/sub reconciliation logic into `pkg/reconclier/utils/pubsub` so it can be shared
  - Force project ID initialization in controller creation.
  - If a trigger condition is not True due to broker, prefix the reason with `Broker-`, otherwise it can be confusing. Example: `Broker-DeletedTopic`
  

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/apis/broker/v1beta1/trigger_lifecycle.go
+++ b/pkg/apis/broker/v1beta1/trigger_lifecycle.go
@@ -65,11 +65,11 @@ func (ts *TriggerStatus) PropagateBrokerStatus(bs *BrokerStatus) {
 
 	switch {
 	case bc.Status == corev1.ConditionUnknown:
-		ts.MarkBrokerUnknown(bc.Reason, bc.Message)
+		ts.MarkBrokerUnknown("Broker-"+bc.Reason, bc.Message)
 	case bc.Status == corev1.ConditionTrue:
 		triggerCondSet.Manage(ts).MarkTrue(eventingv1beta1.TriggerConditionBroker)
 	case bc.Status == corev1.ConditionFalse:
-		ts.MarkBrokerFailed(bc.Reason, bc.Message)
+		ts.MarkBrokerFailed("Broker-"+bc.Reason, bc.Message)
 	default:
 		ts.MarkBrokerUnknown("BrokerUnknown", "The status of Broker is invalid: %v", bc.Status)
 	}
--- a/pkg/reconciler/broker/broker.go
+++ b/pkg/reconciler/broker/broker.go
@@ -30,7 +30,7 @@ import (
 	"go.uber.org/zap"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apierrs "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/labels"
 	appsv1listers "k8s.io/client-go/listers/apps/v1"
@@ -47,23 +47,17 @@ import (
 	brokerreconciler "github.com/google/knative-gcp/pkg/client/injection/reconciler/broker/v1beta1/broker"
 	brokerlisters "github.com/google/knative-gcp/pkg/client/listers/broker/v1beta1"
 	inteventslisters "github.com/google/knative-gcp/pkg/client/listers/intevents/v1alpha1"
-	metadataClient "github.com/google/knative-gcp/pkg/gclient/metadata"
 	"github.com/google/knative-gcp/pkg/reconciler"
 	"github.com/google/knative-gcp/pkg/reconciler/broker/resources"
 	brokercellresources "github.com/google/knative-gcp/pkg/reconciler/brokercell/resources"
-	"github.com/google/knative-gcp/pkg/utils"
+	reconcilerutilspubsub "github.com/google/knative-gcp/pkg/reconciler/utils/pubsub"
 )
 
 const (
 	// Name of the corev1.Events emitted from the Broker reconciliation process.
-	brokerReconcileError = "BrokerReconcileError"
-	brokerReconciled     = "BrokerReconciled"
-	brokerFinalized      = "BrokerFinalized"
-	topicCreated         = "TopicCreated"
-	brokerCellCreated    = "BrokerCellCreated"
-	subCreated           = "SubscriptionCreated"
-	topicDeleted         = "TopicDeleted"
-	subDeleted           = "SubscriptionDeleted"
+	brokerReconciled  = "BrokerReconciled"
+	brokerFinalized   = "BrokerFinalized"
+	brokerCellCreated = "BrokerCellCreated"
 
 	targetsCMName         = "broker-targets"
 	targetsCMKey          = "targets"
@@ -105,8 +99,6 @@ type Reconciler struct {
 	// between multiple controller workers.
 	targetsNeedsUpdate chan struct{}
 
-	// projectID is used as the GCP project ID when present, skipping the
-	// metadata server check. Used by tests.
 	projectID string
 
 	// pubsubClient is used as the Pubsub client when present.
@@ -233,20 +225,10 @@ func (r *Reconciler) reconcileConfig(ctx context.Context, b *brokerv1beta1.Broke
 func (r *Reconciler) reconcileDecouplingTopicAndSubscription(ctx context.Context, b *brokerv1beta1.Broker) error {
 	logger := logging.FromContext(ctx)
 	logger.Debug("Reconciling decoupling topic", zap.Any("broker", b))
-	// get ProjectID from metadata if projectID isn't set
-	projectID, err := utils.ProjectID(r.projectID, metadataClient.NewDefaultMetadataClient())
-	if err != nil {
-		logger.Error("Failed to find project id", zap.Error(err))
-		b.Status.MarkTopicUnknown("ProjectIdNotFound", "Failed to find project id: %w", err)
-		return err
-	}
-	// Set the projectID in the status.
-	//TODO uncomment when eventing webhook allows this
-	//b.Status.ProjectID = projectID
 
 	client := r.pubsubClient
 	if client == nil {
-		client, err := pubsub.NewClient(ctx, projectID)
+		client, err := pubsub.NewClient(ctx, r.projectID)
 		if err != nil {
 			logger.Error("Failed to create Pub/Sub client", zap.Error(err))
 			b.Status.MarkTopicUnknown("PubSubClientCreationFailed", "Failed to create Pub/Sub client: %w", err)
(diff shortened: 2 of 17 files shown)


pkg/apis/broker/v1beta1/trigger_lifecycle.go
```diff
@@ -65,11 +65,11 @@ func (ts *TriggerStatus) PropagateBrokerStatus(bs *BrokerStatus) {
 
 	switch {
 	case bc.Status == corev1.ConditionUnknown:
-		ts.MarkBrokerUnknown(bc.Reason, bc.Message)
+		ts.MarkBrokerUnknown("Broker-"+bc.Reason, bc.Message)
```
grantr, line 68: Optional nit: I think I'd prefer `/` as separator here. That seems to be the preferred condition name separator for Pod Lifecycle according to https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-readiness-gate.

grantr, line 68: Separate discussion :)

I'm not sure why the `Broker` prefix is necessary. Isn't the condition type already `Broker`? So the reason and message will already be scoped to the broker's status. Can you give an example that illustrates why this is confusing?

pkg/reconciler/broker/controller.go
```diff
@@ -66,7 +67,11 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 	// Attempt to create a pubsub client for all worker threads to use. If this
 	// fails, pass a nil value to the Reconciler. They will attempt to
 	// create a client on reconcile.
-	client, err := newPubsubClient(ctx, "")
+	projectID, err := utils.ProjectID(os.Getenv(utils.ProjectIDEnvKey), metadataClient.NewDefaultMetadataClient())
+	if err != nil {
+		logging.FromContext(ctx).Fatal("Failed to get project ID", zap.Error(err))
```
grantr, line 72: This will crash the controller container, stopping any controllers from running unless the project id is present. No objects will be reconciled or have statuses updated until the project id is correctly provided, even if those objects don't depend on the configured project id.

Why is that better than the current strategy?

(which is)
- Attempt to get the project id, and if successful create a single client for all worker threads to use
- If client creation fails, defer and attempt to create a client on each reconcile
- For each reconcile that fails to create a client, set an error status on the object being reconciled

liu-cong: I reverted the change with a small improvement: don't read projectID again if it's not empty in the Reconciler.

pkg/apis/broker/v1beta1/trigger_lifecycle.go
grantr: Makes sense; I hadn't considered the kubectl experience.

pkg/reconciler/broker/controller.go
grantr: I agree we could stand to do more investigation about what this failure really means, and if it's transient or permanent.

We want to eliminate the debug path for the user persona of "resource status isn't saying anything useful, find the controller pod and look at its status/logs". The user persona shouldn't need to be aware of where or how the controller is running. They only care about the resource.

For a concrete example, when was the last time you had to look at the k8s apiserver logs? Might be worth looking into how init failures are handled in storage plugins.

pkg/reconciler/testing/pstest.go
```diff
@@ -52,10 +52,10 @@ func SubscriptionWithTopic(id string, tid string) PubsubAction {
 	}
 }
 
-func TopicAndSub(id string) PubsubAction {
+func TopicAndSub(tid, sid string) PubsubAction {
```
grantr, line 55: Why did this change? Is there a need to use different topic and subscription names?

pkg/reconciler/trigger/controller.go
```diff
@@ -58,11 +58,11 @@ const (
...
-	projectID := ""
-
 	triggerInformer := triggerinformer.Get(ctx)
 
+	// If there is an error, the projectID will be empty. The reconciler will retry
+	// to get the projectID during reconciliation.
+	projectID, _ := utils.ProjectID(os.Getenv(utils.ProjectIDEnvKey), metadataClient.NewDefaultMetadataClient())
 	// Attempt to create a pubsub client for all worker threads to use. If this
 	// fails, pass a nil value to the Reconciler. They will attempt to
 	// create a client on reconcile.
```
grantr, line 68: Why is Trigger reconciler still using the old init style? Is this planned to be changed in a future PR?

pkg/reconciler/trigger/trigger.go
```diff
@@ -37,21 +37,17 @@ import (
...
-	triggerReadinessChanged   = "TriggerReadinessChanged"
-	triggerReconcileFailed    = "TriggerReconcileFailed"
-	triggerUpdateStatusFailed = "TriggerUpdateStatusFailed"
-	topicCreated              = "TopicCreated"
-	subCreated                = "SubscriptionCreated"
-	topicDeleted              = "TopicDeleted"
-	subDeleted                = "SubscriptionDeleted"
+	triggerReconciled = "TriggerReconciled"
+	triggerFinalized  = "TriggerFinalized"
+	topicDeleted      = "TopicDeleted"
```
grantr, line 49: Do we expect to switch to PubsubReconciler's delete functionality in a future PR?

pkg/reconciler/trigger/trigger.go
```diff
@@ -175,17 +169,20 @@ func hasGCPBrokerFinalizer(t *brokerv1beta1.Trigger) bool {
...
-		trig.Status.MarkTopicUnknown("ProjectIdNotFound", "Failed to find project id: %w", err)
-		return err
+
+	projectID := r.projectID
+	if projectID == "" {
+		var err error
+		projectID, err = utils.ProjectID("", metadataClient.NewDefaultMetadataClient())
+		if err != nil {
+			logger.Error("Failed to find project id", zap.Error(err))
+			trig.Status.MarkTopicUnknown("ProjectIdNotFound", "Failed to find project id: %w", err)
```
grantr, line 179: Also MarkSubscriptionUnknown?

pkg/reconciler/utils/pubsub/subscription.go
```diff
@@ -0,0 +1,100 @@
...
+	if subExists {
+		config, err := sub.Config(ctx)
+		if err != nil {
+			logger.Error("Failed to get Pub/Sub subscription Config", zap.Error(err))
+			updater.MarkSubscriptionUnknown("SubscriptionConfigUnknown", "Failed to get Pub/Sub subscription Config: %w", err)
+			return nil, err
+		}
+		if config.Topic != nil && config.Topic.String() == deletedTopic {
+			logger.Error("The topic is deleted")
+			updater.MarkSubscriptionFailed("DeletedTopic", "Pull subscriptions should be recreated to work with recreated topic")
```
grantr, line 60: Change the `should` to `must` to be clear that there's no other option.

pkg/reconciler/utils/pubsub/subscription.go
```diff
@@ -0,0 +1,100 @@
...
+}
+
+func (r *Reconciler) DeleteSubscription(ctx context.Context, id string, obj runtime.Object) error {
+	logger := logging.FromContext(ctx)
+	logger.Debug("Deleting decoupling sub")
+
+	sub := r.client.Subscription(id)
+	exists, err := sub.Exists(ctx)
+	if err != nil {
+		logger.Error("Failed to verify Pub/Sub subscription exists", zap.Error(err))
```
grantr, line 87: @grac3gao this is one place we could potentially update the Subscription's status while finalizing for #1190

pkg/reconciler/broker/broker.go
```diff
@@ -233,106 +227,67 @@ func (r *Reconciler) reconcileConfig(ctx context.Context, b *brokerv1beta1.Broke
...
-	if err != nil {
-		logger.Error("Failed to find project id", zap.Error(err))
-		b.Status.MarkTopicUnknown("ProjectIdNotFound", "Failed to find project id: %w", err)
-		return err
+
+	projectID := r.projectID
+	if projectID == "" {
+		var err error
+		// get ProjectID from metadata if projectID isn't set
+		projectID, err = utils.ProjectID("", metadataClient.NewDefaultMetadataClient())
```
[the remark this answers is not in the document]
liu-cong: right, forgot about that...

pkg/reconciler/broker/broker.go
```diff
@@ -344,61 +299,25 @@ func (r *Reconciler) deleteDecouplingTopicAndSubscription(ctx context.Context, b
...
-	projectID, err := utils.ProjectID(r.projectID, metadataClient.NewDefaultMetadataClient())
-	if err != nil {
-		logger.Error("Failed to find project id", zap.Error(err))
-		return err
-	}
-
 	client := r.pubsubClient
 	if client == nil {
-		client, err := pubsub.NewClient(ctx, projectID)
+		client, err := pubsub.NewClient(ctx, r.projectID)
```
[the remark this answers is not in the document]
liu-cong: This should be the same as `reconcileDecouplintTopicAndSubscription`, I missed this one.

pkg/reconciler/broker/controller.go
```diff
@@ -63,10 +63,13 @@ func NewController(ctx context.Context, cmw configmap.Watcher) *controller.Impl
 	podInformer := podinformer.Get(ctx)
 	bcInformer := brokercellinformer.Get(ctx)
 
+	// If there is an error, the projectID will be empty. The reconciler will retry
+	// to get the projectID during reconciliation.
+	projectID, _ := utils.ProjectID(os.Getenv(utils.ProjectIDEnvKey), metadataClient.NewDefaultMetadataClient())
```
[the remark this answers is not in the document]
liu-cong: Since this is in the initialization, we probably won't see any. But logging it anyways.

pkg/reconciler/trigger/trigger.go
liu-cong: I should have just done it :)

pkg/reconciler/testing/pstest.go
liu-cong: The new PubSubReconciler tests need this. And I think as a helper method, this should allow different names anyway.

pkg/reconciler/broker/controller_test.go
```diff
@@ -43,7 +45,7 @@ import (
 
 func TestNew(t *testing.T) {
 	ctx, _ := SetupFakeContext(t)
-
+	os.Setenv(utils.ProjectIDEnvKey, "test")
```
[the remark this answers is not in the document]
liu-cong: In fact this is not going to make any difference for the test case since we ignore the  `ProjectID` failure. Removed it since it's not necessary.

pkg/reconciler/trigger/controller.go
liu-cong: Which old init style exactly? I think both broker and trigger controllers are using the same init style.

grantr:
I believe this also fixes #1192 because it adds support for the `PROJECT_ID` env var.

pkg/reconciler/trigger/controller.go
grantr: Disregard this, I thought `newPubsubClient` was not used in broker/controller.go anymore, but it is back now (or maybe never left).

pkg/reconciler/utils/pubsub/subscription.go
```diff
@@ -0,0 +1,100 @@
...
+
+	// Check if the topic of the subscription is "_deleted-topic_"
+	if subExists {
+		config, err := sub.Config(ctx)
+		if err != nil {
+			logger.Error("Failed to get Pub/Sub subscription Config", zap.Error(err))
+			updater.MarkSubscriptionUnknown("SubscriptionConfigUnknown", "Failed to get Pub/Sub subscription Config: %w", err)
+			return nil, err
+		}
+		if config.Topic != nil && config.Topic.String() == deletedTopic {
```
nachocano, line 58: @capri-xiyue this part we need in our code... Sync with @liu-cong 
I'll assign you the issue. Thanks!!!

pkg/reconciler/utils/pubsub/subscription.go
grac3gao: @grantr Thanks for pointing out this!

PR #1241, opened 2020-06-08
Use wire to inject the IAMPolicyManager and gcpauth.Store.

Harwayne:
## Proposed Changes

- Use wire to inject the IAMPolicyManager and gcpauth.Store.
- Based heavily on #944.

**Release Note**

```release-note
NONE
```


cmd/controller/wire.go
```diff
@@ -0,0 +1,52 @@
...
+	"github.com/google/wire"
+	"knative.dev/pkg/injection"
+)
+
+func InitializeControllers(ctx context.Context) ([]injection.ControllerConstructor, error) {
+	panic(wire.Build(
+		Controllers,
+		ClientOptions,
+		iam.PolicyManagerSet,
+		wire.Struct(new(gcpauth.StoreSingleton)),
```
Harwayne, line 41: Is this the right way to do a singleton? What I want is to create a single gcpauth.Store. Because I only seem to be able to do that while the various controllers are being created (I am using the passed in configmap.Watcher, I suppose I could just make my own here instead), I made a wrapping singleton struct that uses sync.Once to make sure we only make one of them.

ian-mi: I think this is a good way to do this. Of course it would be ideal if the cmw could be injected by wire as well but that will require significant changes to controller injection.

cmd/controller/main.go
```diff
@@ -34,24 +38,49 @@ import (
...
-		auditlogs.NewController,
-		storage.NewController,
-		scheduler.NewController,
-		pubsub.NewController,
-		build.NewController,
-		staticpullsubscription.NewController,
-		kedapullsubscription.NewController,
-		topic.NewController,
-		channel.NewController,
+	ctx := signals.NewContext()
```
Harwayne, line 48: I can't call this. It can only be called once per process, the second call immediately panics. `sharedmain.Main` calls it internally.

So, i think our choices are to either create singletons for the IAM client and IAM policy manager. And have them lazily initialize, similar to how gcpauth.StoreSingleton is being initialized. But at that point, is it better to use wire than just using a static singleton variable, like `iam.PolicyManagerSingleton`?

The other thing we could do is start altering the pkg code to either allow passing in constructor constructors (`func(context.Context, configmap.Watcher) injection.ControllerConstructor`) or a way to create the context (and configmap.Watcher) before creating the constuctor list for sharedmain.Main. My preference is the latter, but I don't know how knative will feel about that change. Thoughts?

ian-mi: Can't this be fixed by called sharedmain.MainWithContext instead?

I had been working on a branch to generate wire provider sets to perform informer injection along with a refactored controller struct which would allow the entire controller be injected by wire. I haven't spent the time to finish this work up however, as it is a fairly significant divergence from the ctx based injection that is baked into much of pkg. If you think this sounds worthwhile I can dedicate some time to getting that PR out.

Harwayne: Yes it can 😄

PR #1243, opened 2020-06-08
pin eventing to 0603 nightly

capri-xiyue:
## Proposed Changes

- pin eventing to 0603 nightly


nachocano:
06/04 should work? Or it doesn't?

nachocano:
I think this is the PR that removed v1alpha1 https://github.com/knative/eventing/pull/3275. 
Was pushed on the 5th. There is no nightly on 06/04 apparently, so we go with 06/03...

PR #1249, opened 2020-06-09
Remove spec.googleServiceAccount in v1alpha1

### LINKED ISSUE #1205: Remove spec.googleServiceAccount in v1alpha1
**Problem**
`spec.googleServiceAccount` in v1alpha1 is deprecated. We will remove it in 0.16 release. 
`spec.serviceAccountName` replaces it in v1beta1.

**Exit Criteria**
Removing spec.googleServiceAccount, together, deleting any controller logic related on googleServiceAccount

**Time Estimate (optional):** 

grac3gao:
Fixes #1205

## Proposed Changes

- Removing spec.googleServiceAccount, together, deleting any controller logic related on googleServiceAccount
-
-

**Release Note**

```release-note

Field spec.googleServiceAccount in v1alpha1 is removed, together with related controller logic
```

**Docs**

### DIFF
--- a/pkg/apis/duck/v1alpha1/credentials.go
+++ b/pkg/apis/duck/v1alpha1/credentials.go
@@ -30,20 +30,13 @@ const (
 )
 
 var (
-	// The format of gServiceAccountName is service-account-name@project-id.iam.gserviceaccount.com
-	// Service account name must be between 6 and 30 characters (inclusive),
-	// must begin with a lowercase letter, and consist of lowercase alphanumeric characters that can be separated by hyphens.
-	// Project IDs must start with a lowercase letter and can have lowercase ASCII letters, digits or hyphens,
-	// must be between 6 and 30 characters.
-	gsaValidationRegex = regexp.MustCompile(`^[a-z][a-z0-9-]{5,29}@[a-z][a-z0-9-]{5,29}.iam.gserviceaccount.com$`)
-
 	// The name of a k8s ServiceAccount object must be a valid DNS subdomain name.
 	// https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names
 	ksaValidationRegex = regexp.MustCompile(`^[A-Za-z0-9](?:[A-Za-z0-9\-]{0,61}[A-Za-z0-9])?$`)
 )
 
 // ValidateCredential checks secret and service account.
-func ValidateCredential(secret *corev1.SecretKeySelector, gServiceAccountName string, kServiceAccountName string) *apis.FieldError {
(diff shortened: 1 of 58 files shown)


pkg/apis/duck/v1alpha1/credentials.go
```diff
@@ -30,37 +30,24 @@ const (
...
-	} else {
-		var errs *apis.FieldError
-		if kServiceAccountName != "" {
-			errs = errs.Also(validateK8sServiceAccount(kServiceAccountName))
-		}
-		if gServiceAccountName != "" {
-			errs = errs.Also(validateGCPServiceAccount(gServiceAccountName))
-		}
-		return errs
+	} else if kServiceAccountName != "" {
```
capri-xiyue, line 47: Is it possible that both secret and kServiceAccountName is nil or "", do we allow that?

grac3gao: We allow that case, defaulting webhook will add default credential to it. (which is set up in configmap config-gcp-auth).

pkg/reconciler/events/scheduler/scheduler.go
```diff
@@ -187,7 +187,7 @@ func (r *Reconciler) deleteJob(ctx context.Context, scheduler *v1alpha1.CloudSch
 func (r *Reconciler) FinalizeKind(ctx context.Context, scheduler *v1alpha1.CloudSchedulerSource) reconciler.Event {
 	// If k8s ServiceAccount exists and it only has one ownerReference, remove the corresponding GCP ServiceAccount iam policy binding.
```
nachocano, line 188: this comment still applies? Don't recall if we agreed on removing the GSA, did we?
we might be removing a default one in some cases?

grac3gao: This is referring to the default one. In the identity finalizer, it will do nothing if the KSA not pair to any GSA in the config-gcp-auth. I'll updating a more clear comment.

nachocano: ok, and by removal you mean removing the iam role binding, or whatever is called. Not the GSA itself, right?

grac3gao: Yes, just remove the iam policy binding, not the GSA itself.

Harwayne:
/hold

Please add a release notes section saying that the field has been removed. The release-notes section is what we use to create the release notes and this is definitely a user-visible change.

grac3gao:
@Harwayne Done. Sorry, I forget to add release note....

PR #1268, opened 2020-06-11
Recreate subscription if topic has been deleted

### LINKED ISSUE #1259: Broker ingress incorrectly responds success after topic is deleted and recreated
The broker controller will create a new topic when it notices the original topic is deleted. This new topic will be picked up by the broker ingress via the targets config and event publish will start succeeding again, since as far as the ingress is concerned everything is working. But those events will actually be dropped by pubsub because there are no subscriptions on the topic yet.

So this is a situation in which the Broker's control plane behavior is correct, but the data plane behavior is incorrect. I think the correct behavior from the data plane, given the Broker controller's current lack of auto-healing, is to reject any events until a subscription is in place.

I'm hesitant to recommend that the ingress wait until the Broker status is ready to accept events. This might be an irrational opinion, but it feels like that allows the Broker controller to be lazy and produce an invalid, incomplete targets configuration that the data plane has to ignore. I'd rather the configuration always be valid and complete. But in the short term, I'd be fine with the ingress rejecting events on Brokers with unready status (and maybe it's the right thing for the long-term too, I'm not sure).

An alternative: the Broker controller could refuse to create a new topic if the "deleted topic" subscription exists. I think I prefer this solution since it avoids the ingress needing to check broker status.

Broken out of #1235

liu-cong:
Fixes #1259

## Proposed Changes

- Recreate the topic and subscription if topic is manually deleted.
-
-

**Release Note**

```release-note
If topic of a broker/trigger is manually deleted, the topic and subscription will be recreated. This may cause unacknowledged events to be lost.
```

**Docs**

### DIFF
--- a/pkg/reconciler/utils/pubsub/subscription.go
+++ b/pkg/reconciler/utils/pubsub/subscription.go
@@ -18,7 +18,7 @@ package pubsub
 
 import (
 	"context"
-	"errors"
+	"fmt"
 
 	"cloud.google.com/go/pubsub"
 	"go.uber.org/zap"
@@ -34,8 +34,6 @@ const (
 	subCreated   = "SubscriptionCreated"
 	subDeleted   = "SubscriptionDeleted"
 )
-
-var deletedTopicErr = errors.New("topic of the subscription has been deleted")
 
 func (r *Reconciler) ReconcileSubscription(ctx context.Context, id string, subConfig pubsub.SubscriptionConfig, obj runtime.Object, updater StatusUpdater) (*pubsub.Subscription, error) {
 	logger := logging.FromContext(ctx)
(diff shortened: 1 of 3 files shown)


pkg/reconciler/utils/pubsub/topic_test.go
```diff
@@ -143,10 +142,6 @@ func newTestRunner(t *testing.T, tc testCase) (*testRunner, func()) {
 }
 
 func (r *testRunner) verify(t *testing.T, tc testCase, su *utilspubsubtesting.StatusUpdater, err error) {
-	if tc.wantErr != err {
-		t.Fatalf("Expect error, got: %v, want: %v", err, tc.wantErr)
```
grantr, line 147: Should this be Errorf? Why does it need to be Fatalf?

pkg/reconciler/utils/pubsub/subscription.go
```diff
@@ -56,26 +54,20 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, id string, subCo
 			return nil, err
 		}
 		if config.Topic != nil && config.Topic.String() == deletedTopic {
-			logger.Error("The topic is deleted")
-			updater.MarkSubscriptionFailed("DeletedTopic", "Pull subscriptions must be recreated to work with recreated topic")
-			return nil, deletedTopicErr
+			logger.Warn("Detected deleted topic. Going to recreate the pull subscription.")
```
grantr, line 57: I think this should still be error level. The error message should warn that data loss may occur if the subscription has unacknowledged messages.

grantr, line 57: Can we also record an Event (type Warn) saying that a deleted topic was detected? That way this situation is more clear from describing the resource. As written, the only indication from `kubectl describe` that something was amiss is a pair of Deleted Sub and Created Sub events (type Normal). A type Warn Event about the deleted topic is more clear that something exceptional has occurred.

pkg/reconciler/utils/pubsub/topic_test.go
liu-cong: This is removed :)

pkg/reconciler/utils/pubsub/subscription.go
liu-cong: Good idea!

liu-cong:
Thanks for updating the release note.

Issue #1265, opened 2020-06-11
Stabilize broker components


yolocs:
**Describe the bug**
Under certain setup and scale, the broker components become unstable. Here are a couple of problems I have observed.

Setup: 1 broker, 100 triggers (subscribing to the same event), 100 ingress qps.

Problem 1: kube-dns stopped responding. Errors from fanout pods are like `{"level":"warn","ts":"2020-06-10T18:54:20.552Z","logger":"broker-fanout","caller":"deliver/processor.go:118","msg":"target delivery failed","commit":"a12cfb0","target":"loadtest/testbroker/trigger-actor-57","error":"Post \"http://actor-57.loadtest.svc.cluster.local/\": dial tcp: lookup actor-57.loadtest.svc.cluster.local on 10.4.0.10:53: dial udp 10.4.0.10:53: operation was canceled"}`

Problem 2: unstable memory usage causing pods being killed for OOM. When I increased ingress qps from 10 to 100, there will be a surge of memory usage. With our current memory limit, existing fanout pods got killed instantly (before HPA could kick in) which ended up with crash lookback. However, after I removed the memory limit and repeated the same process, the fanout pods eventually consumed less memory than our limit after the initial memory usage spike.

Problem 3: Number of triggers to fanout to (which we can't control) dominates memory usage. Ideally we want a higher pubsub pull concurrency for better throughput. However, when there are a little bit more triggers, the cost on memory will be at least x (# of triggers). With a memory limit, this can easily lead to OOM and causing problem2. But if we limit pull concurrency, then we run the risk of low throughput even when there are not many triggers.

 **Proposed solutions**

- [x] I believe user can modify the scale of kube-dns in kube-system namespace
- [x] Increase `MaxIdleConns` of delivery HTTP client to increase connection reuse (so that less load will be put on dns)
- [ ] Remove all memory limits or set those limits really high (e.g. 5Gi)
  - The GKE guideline now suggests setting memory limit = request (we could potentially make an exception)
  - This could be controversial if we end up setting request as high as the limit.

liu-cong:
cc

grantr:
Thank you for doing this research @yolocs! Great to have this clear breakdown of the issues.

My initial thoughts:

Problem 1: This seems like a bug. We should be reusing connections and only talking to DNS when a connection is made. Hopefully a fix is possible.

[...]

PR #1269, opened 2020-06-11
Tune broker components

yolocs:
Some mitigation for https://github.com/google/knative-gcp/issues/1265

## Proposed Changes

- Adjusted broker components memory resource
  - Increased ingress memory limit to 1000Mi
  - Set memory limits much high than requested
- Adjusted HPA
  - The avg memory usage is set to half of the limit for fanout/retry (this hopefully helps to mitigate some problem from surging memory usage)
  - Lower the max replicas for fanout/retry as I've seen a higher error rate of dns when reaching more than 10 replicas
- Relaxed liveness probe timeout (would help reduce some noise)
- Override `MaxIdleConns` for delivery HTTP client (would help reusing connections and reduce the chance of dns error)

**Release Note**

```release-note
- Adjusted broker components memory resource
  - Increased ingress memory limit to 1000Mi
  - Set memory limits much high than requested
- Adjusted HPA
  - The avg memory usage is set to half of the limit for fanout/retry (this hopefully helps to mitigate some problem from surging memory usage)
  - Lower the max replicas for fanout/retry as I've seen a higher error rate of dns when reaching more than 10 replicas
- Relaxed liveness probe timeout (would help reduce some noise)
- Override `MaxIdleConns` for delivery HTTP client (would help reusing connections and reduce the chance of dns error)
```

**Docs**


pkg/broker/handler/providers.go
```diff
@@ -37,6 +38,11 @@ var (
 
 	DefaultHTTPClient = &http.Client{
 		Transport: &ochttp.Transport{
+			Base: &http.Transport{
+				MaxIdleConns:        1000,
+				MaxIdleConnsPerHost: 1000,
```
liu-cong, line 43: Since generally we should have more than 1 targets, I think this value should be smaller than `MaxIdleConns`. 


I suggest the following numbers. @grantr any thoughts on this?

```
MaxIdelConns: 1000,
// Assume an average of 20 triggers, just a guess. This makes sure idle conns are more fairly distributed across targets.
MaxIdleConnsPerHost: 50,
// Also set max conns per target to limit the resource usage per target. This should also help limit the memory usage.
MaxConnsPerHost:100,
```

yolocs: Open to suggestion. Some of my thinking:

1. We don't have to guarantee fairness here. When you have 100 triggers but only one is actively receiving events, I see no reason to not have it take more idle connections. It _should_ naturally evolve to "active triggers have more idle conns and less active ones have less". Although it might worth setting `MaxIdleConnsPerHost` a little bit less than `MaxIdleConns`. Under high concurrency I occasionally saw errors of failure to put new idle conn for a host because it's reaching max idle conns.

2. Setting `MaxConnsPerHost` should be able to effectively limit resource usage. But there is a balance between this and maxing throughput. Especially when we don't have "queue length based" auto scaling.

liu-cong: RE 
#1. We need to find a good balance point. A high `MaxIdleConnsPerHost` deals with "small number of active triggers" scenario well but not so well with many active triggers. So I took a guess of on average 20 active triggers. That might be too big. Ideally we can make this configurable but we have to pick a guess for now. How about setting `MaxIdleConnsPerHost` to 100 or 200?

#2 Agree it's a tradeoff. Since we can run into OOM, setting this will help. Maybe something like 200 is a good balance point?

yolocs: I'm thinking a bigger number like 500 for both `MaxIdleConnsPerHost` and `MaxConnsPerHost`. Since now I set high memory limit with HPA on 50%, it should help with the OOM problem (not always but should help some). Thoughts?

liu-cong: Is it better to have MaxIdleConnsPerHost< MaxConnsPerHost? This gives more room for other triggers to maintain some idle connection. The tradeoff is that it's less efficient for handling surges. 

 I don't think we can come up with perfect numbers We may need to tune it later anyway. I will lgtm with whatever you come up with :)

yolocs: Make sense. Set them to 500 now.

Issue #1261, opened 2020-06-10
Broker accepts events only when the broker state is ready


yolocs:
When topic/sub creation fails, we should mark broker not ready in the config. And ingress shouldn't accept events in that case

PR #1272, opened 2020-06-11
Have broker ingress reject events when the broker under requesting is not ready

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1261

## Proposed Changes

-
-
-

**Release Note**

```release-note
Have broker ingress reject events when the broker under requesting is not ready
```

**Docs**


pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -71,6 +71,10 @@ func (m *multiTopicDecoupleSink) getTopicForBroker(ns, broker string) (string, e
 		m.logger.Warn("config is not found for", zap.Any("ns", ns), zap.Any("broker", broker))
 		return "", fmt.Errorf("%q/%q: %w", ns, broker, ErrNotFound)
 	}
+	if brokerConfig.State != config.State_READY {
+		m.logger.Debug("broker is not ready", zap.Any("ns", ns), zap.Any("broker", broker))
+		return "", fmt.Errorf("%q/%q is not ready: %w", ns, broker, ErrNotFound)
```
liu-cong, line 76: nit: create a new error `ErrNotReady`

yolocs: Sure. Do you expect a different http status code in that case?

liu-cong: Probably, how about `503 Service Unavailable`? This implies that the unavailability is temporary: https://www.w3.org/Protocols/rfc2616/rfc2616-sec10.html

yolocs: By 404 I meant it to mean something like "ready broker xxx is not found". But now I think 503 is better.

PR #1273, opened 2020-06-11
Run ingress handler benchmark with multiple event payload sizes

ian-mi:
Reduce maximum event payload size to 100000 bytes to avoid overflowing the pubsub client buffer.

Benchstat results:
```
name                            time/op
IngressHandler/0_bytes-16       10.9µs ± 2%
IngressHandler/1000_bytes-16    11.1µs ± 2%
IngressHandler/100000_bytes-16   122µs ± 3%
```

### DIFF
--- a/pkg/testing/event.go
+++ b/pkg/testing/event.go
@@ -1,0 +1,48 @@
+/*
+Copyright 2020 Google LLC.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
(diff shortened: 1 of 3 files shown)


pkg/testing/event.go
```diff
@@ -0,0 +1,48 @@
...
+	event.SetType("type")
+	event.SetTime(time.Now())
+	if size > 0 {
+		data := make([]byte, size)
+		_, err := rand.Read(data)
+		if err != nil {
+			t.Fatal(err)
+		}
+		event.DataEncoded = data
+		return &event
```
liu-cong, line 45: nit: no need to return here.

Issue #1118, opened 2020-05-21
Unneeded Topic Connections are Not Cleaned Up By Ingress


ian-mi:
**Describe the bug**
The ingress pubsub CE client will retain topics forever. If a broker is deleted the topic connection will die but will not be removed from the topic map.

 **Expected behavior**
When a broker config is removed from targets the corresponding topic connection should be cleanly stopped and removed.

grantr:
@ian-mi to clarify, this is a memory leak in the container, not a GCP resource leak?

ian-mi:
Yes, this will leak a small amount of memory and potentially some amount of additional overhead but is not a resource leak.

liu-cong:
If we continue to use CE pubsub protocol, looks like the `Close` method should be implemented: https://github.com/cloudevents/sdk-go/blob/master/protocol/pubsub/v2/protocol.go#L243

ian-mi:
Implementing close would be useful for closing the entire protocol but would not help for cleaning up individual topic connections.

PR #1278, opened 2020-06-12
Use pubsub bindings in broker ingress

ian-mi:
## Proposed Changes

- Directly use pubsub client and bindings in ingress handler. Removes unnecessary overhead and allows for better control over the metrics and trace spans that get emitted. Results in a ~10% performance improvement for small events. For very large events the reduction in overhead is negligible.

## Benchmark Results
```
name                            old time/op  new time/op  delta
IngressHandler/0_bytes-16       11.0µs ± 3%   9.7µs ± 2%  -11.63%  (p=0.000 n=9+10)
IngressHandler/1000_bytes-16    11.2µs ± 1%  10.1µs ± 2%   -9.54%  (p=0.000 n=10+10)
IngressHandler/100000_bytes-16   122µs ± 5%   123µs ± 4%     ~     (p=0.798 n=8+8)
```

### DIFF
--- a/pkg/broker/ingress/multi_topic_decouple_sink.go
+++ b/pkg/broker/ingress/multi_topic_decouple_sink.go
@@ -43,7 +43,8 @@ func NewMultiTopicDecoupleSink(ctx context.Context, brokerConfig config.Readonly
 		logger:       logging.FromContext(ctx),
 		pubsub:       client,
 		brokerConfig: brokerConfig,
-		topics:       make(map[types.NamespacedName]*pubsub.Topic),
+		// TODO(#1118): remove Topic when broker config is removed
+		topics: make(map[types.NamespacedName]*pubsub.Topic),
 	}
 }
 
--- a/pkg/broker/ingress/multi_topic_decouple_sink.go
+++ b/pkg/broker/ingress/multi_topic_decouple_sink.go
@@ -109,10 +109,9 @@ func (m *multiTopicDecoupleSink) updateTopicForBroker(broker types.NamespacedNam
 		if topic.ID() == topicID {
 			// Topic already updated.
 			return topic, nil
-		} else {
-			// Stop old topic.
-			m.topics[broker].Stop()
 		}
+		// Stop old topic.
+		m.topics[broker].Stop()
 	}
 	topic := m.pubsub.Topic(topicID)
 	m.topics[broker] = topic
--- a/pkg/broker/ingress/multi_topic_decouple_sink.go
+++ b/pkg/broker/ingress/multi_topic_decouple_sink.go
@@ -19,11 +19,17 @@ package ingress
 import (
 	"context"
 	"fmt"
+	"sync"
 
+	"cloud.google.com/go/pubsub"
+	"go.opencensus.io/trace"
 	"go.uber.org/zap"
+	"k8s.io/apimachinery/pkg/types"
 
+	cepubsub "github.com/cloudevents/sdk-go/protocol/pubsub/v2"
 	cev2 "github.com/cloudevents/sdk-go/v2"
-	cecontext "github.com/cloudevents/sdk-go/v2/context"
+	"github.com/cloudevents/sdk-go/v2/binding"
+	"github.com/cloudevents/sdk-go/v2/extensions"
 	"github.com/cloudevents/sdk-go/v2/protocol"
 	"github.com/google/knative-gcp/pkg/broker/config"
 	"knative.dev/eventing/pkg/logging"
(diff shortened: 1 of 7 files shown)


pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -53,27 +63,81 @@ type multiTopicDecoupleSink struct {
...
-		return "", fmt.Errorf("decouple queue of %q/%q: %w", ns, broker, ErrIncomplete)
+		return "", fmt.Errorf("decouple queue of %q: %w", broker, ErrIncomplete)
 	}
 	return brokerConfig.DecoupleQueue.Topic, nil
 }
+
+func (m *multiTopicDecoupleSink) getExistingTopic(broker types.NamespacedName) (*pubsub.Topic, bool) {
+	m.topicsMut.RLock()
+	defer m.topicsMut.RUnlock()
+	topic, ok := m.topics[broker]
```
liu-cong, line 141: nit: `return m.topics[broker]`

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -53,27 +63,81 @@ type multiTopicDecoupleSink struct {
...
+	topicID, err := m.getTopicIDForBroker(broker)
+	if err != nil {
+		return nil, err
+	}
+
+	if topic, ok := m.topics[broker]; ok {
+		if topic.ID() == topicID {
+			// Topic already updated.
+			return topic, nil
+		} else {
```
liu-cong, line 112: nit: `else` is not neede

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -32,19 +38,23 @@ import (
...
 
 // multiTopicDecoupleSink implements DecoupleSink and routes events to pubsub topics corresponding
 // to the broker to which the events are sent.
 type multiTopicDecoupleSink struct {
-	// client talks to pubsub.
-	client cev2.Client
+	// pubsub talks to pubsub.
+	pubsub *pubsub.Client
+	// map from brokers to topics
+	topics    map[types.NamespacedName]*pubsub.Topic
```
liu-cong, line 56: It seems that if a broker is deleted, its entry is not removed from this map. To do the removal, I think we need to watch for the configmap file.

pkg/broker/ingress/multi_topic_decouple_sink.go
ian-mi: This does not work since this is a special case in go where the result is overloaded as both a 1 and 2 value expression and go will not perform type inference.

pkg/broker/ingress/multi_topic_decouple_sink.go
ian-mi: Yes, good catch, I've already created #1118 to track this. Note this is also an issue when using the pubsub transport as we do currently.

liu-cong: Right, I believe the existing CE SDK client doesn't provide a way for the clean up. Looks like it should be implemented here: https://github.com/cloudevents/sdk-go/blob/master/protocol/pubsub/v2/protocol.go#L243

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -53,31 +63,84 @@ type multiTopicDecoupleSink struct {
...
+	topic, err := m.getTopicForBroker(types.NamespacedName{Namespace: ns, Name: broker})
 	if err != nil {
 		return err
 	}
-	ctx = cecontext.WithTopic(ctx, topic)
-	return m.client.Send(ctx, event)
+
+	dt := extensions.FromSpanContext(trace.FromContext(ctx).SpanContext())
+	msg := new(pubsub.Message)
+	if err := cepubsub.WritePubSubMessage(ctx, binding.ToMessage(&event), msg, dt.WriteTransformer()); err != nil {
```
liu-cong, line 73: It seems like we need to work with `message` instead of `event` for tracing support? In `handler.go`, we transform http request to message and then to event. Can we add the span context there to avoid another transformation?

pkg/broker/ingress/multi_topic_decouple_sink.go
liu-cong: Can you add a TODO here and link the issue?

pkg/broker/ingress/multi_topic_decouple_sink.go
ian-mi: This is a bit of refactoring that will be needed before the event conversion can be elided as the SDK requires that all attribute and extension access be done through transformers.

ian-mi: I will do this refactoring in a separate PR

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -53,31 +64,84 @@ type multiTopicDecoupleSink struct {
...
+		return nil, err
+	}
+
+	if topic, ok := m.getExistingTopic(broker); ok {
+		// Check that the broker's topic ID hasn't changed.
+		if topic.ID() == topicID {
+			return topic, nil
+		}
+	}
+
```
liu-cong, line 95: `updateTopicForBroker` does these checks too. It seems to me you can just directly call `updateTopicForBroker`. (maybe rename it to `getOrCreateTopicForBroker`

liu-cong: will lgtm once this is addressed.

ian-mi: This would require obtaining the mutex on every request. Most requests should only require an RLock.

liu-cong: right. In that case is it necessary for `updateTopicForBroker` to do the same checks? How about 
```
func (m *multiTopicDecoupleSink) updateTopicForBroker(broker types.NamespacedName, topicID string) (*pubsub.Topic, error) {
	m.topicsMut.Lock()
	defer m.topicsMut.Unlock()

	if topic, ok := m.topics[broker]; ok {
		// Stop old topic.
		m.topics[broker].Stop()
	}
	topic := m.pubsub.Topic(topicID)
	m.topics[broker] = topic
	return topic, nil
}
```

liu-cong: I guess there could be a race condition where the given `topicID` is not up to date.

ian-mi: Yes, to avoid potentially creating unnecessary topic connections it's necessary to recheck state after obtaining the mutex. This path will only be hit once per topic so the overall impact should be negligible.

ian-mi:
Thanks for the review!

PR #1282, opened 2020-06-15
Increase the time out for metrics assertions in an attempt to deflake…

Harwayne:
## Proposed Changes

- Increase the time out for metrics assertions in an attempt to deflake the test.

We see the following style flakes in roughly equal proportion:

1. 
```
    TestGCPBrokerMetrics: test_gcp_broker.go:68: unexpected metric "knative.dev/eventing/trigger/event_count" count (-want, +got) =   map[string]int64{
        - 	"resp-broker-gcp-vumhovdj":    1,
        - 	"trigger-broker-gcp-vumhovdj": 1,
          }
```
2. 
```
    TestGCPBrokerMetrics: test_gcp_broker.go:68: unexpected broker metric count (-want, +got) =   map[string]int64{
        - 	"e2e-dummy-event-type":              1,
        + 	"e2e-dummy-event-type":              0,
        - 	"e2e-testing-resp-event-type-dummy": 1,
        + 	"e2e-testing-resp-event-type-dummy": 0,
          }
```

We hope that this change helps resolve the first. 

**Release Note**

```release-note
NONE
```

### DIFF
--- a/test/e2e/lib/metrics/stackdriver.go
+++ b/test/e2e/lib/metrics/stackdriver.go
@@ -103,7 +103,7 @@ func CheckAssertions(t *testing.T, assertions ...Assertion) {
 		t.Error(err)
 		return
 	}
-	timeout := time.After(5 * time.Minute)
+	timeout := time.After(8 * time.Minute)
 	for {
 		errors := make([]error, 0, len(assertions))
 		for _, assertion := range assertions {
@@ -121,7 +121,8 @@ func CheckAssertions(t *testing.T, assertions ...Assertion) {
 				t.Error(err)
 			}
 			return
-		case <-time.After(5 * time.Second):
+		default:
+			time.Sleep(5 * time.Second)
 		}
 	}
 }


test/e2e/lib/metrics/stackdriver.go
```diff
@@ -121,7 +121,8 @@ func CheckAssertions(t *testing.T, assertions ...Assertion) {
 				t.Error(err)
 			}
 			return
-		case <-time.After(5 * time.Second):
+		default:
+			time.Sleep(5 * time.Second)
```
ian-mi, line 125: This change seems unnecessary?

Harwayne: The only reason I did that is because I had to reread the original two or three times to convince myself what was going on. I find this clearer, but if you prefer the original, I can switch back.

ian-mi: I see, lets go ahead and submit this while the tests are passing since the difference is minimal.

Issue #969, opened 2020-04-28
Integration test verifying Broker Retry metrics to Stackdriver


Harwayne:
**Exit Criteria**
- All metrics defined in the Broker Retry component are exported to Stackdriver. 
- Integration test verifies the metric is in Stackdriver.

grantr:
This issue now only tracks the integration test piece

PR #1284, opened 2020-06-16
Verify broker retry metrics in e2e test

yolocs:
Fixes https://github.com/google/knative-gcp/issues/969

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


yolocs:
The broker metrics test actually passed in the latest WI tests.

/retest

test/e2e/lib/helpers/broker_event_transformation_test_with_source.go
```diff
@@ -165,14 +166,43 @@ func BrokerEventTransformationMetricsTestHelper(client *lib.Client, projectID st
...
 		},
-		lib.TriggerMetricAssertion{
-			ProjectID:       projectID,
-			BrokerName:      brokerName,
-			BrokerNamespace: client.Namespace,
-			StartTime:       start,
-			CountPerTrigger: map[string]int64{
-				trigger.Name:     1,
-				respTrigger.Name: 1,
+		lib.TriggerMetricWithRespCodeAssertion{
```
ian-mi, line 169: The downside to structuring the assertion in this way is that it will not be able to detect if metrics are emitted with some other response code whereas previously this would have failed the test. An alternative approach would be to add the response code to the count map key.

yolocs: Good point.

test/e2e/lib/helpers/broker_event_transformation_test_with_source.go
```diff
@@ -164,43 +164,21 @@ func BrokerEventTransformationMetricsTestHelper(client *lib.Client, projectID st
...
-				CountPerTrigger: map[string]int64{
-					trigger.Name: 2,
-				},
+		lib.TriggerMetricAssertion{
+			ProjectID:       projectID,
+			BrokerName:      brokerName,
+			BrokerNamespace: client.Namespace,
+			StartTime:       start,
+			CountPerTriggerWithRespCode: map[string]int64{
+				lib.TriggerNameWithRespCode(trigger.Name, http.StatusAccepted):     1,
```
ian-mi, line 173: nit: please make this a struct instead

ian-mi: If this would be too difficult feel free to merge as is.

yolocs: You meant make trigger + resp code a struct?

ian-mi: Right, my idea was to make the trigger + code key a struct instead of string.

test/e2e/lib/helpers/broker_event_transformation_test_with_source.go
```diff
@@ -169,16 +169,16 @@ func BrokerEventTransformationMetricsTestHelper(client *lib.Client, projectID st
...
-				lib.TriggerNameWithRespCode(trigger.Name, http.StatusBadRequest): 2,
-			},
-			CountPerTriggerNoRespCode: map[string]int64{
+				{Name: trigger.Name, RespCode: http.StatusBadRequest}: 2,
+
 				// For metrics without response code, we expect 3 trigger deliveries (first 2 from delivery failures).
-				trigger.Name:     3,
-				respTrigger.Name: 1,
+				{Name: trigger.Name}:     3,
+				{Name: respTrigger.Name}: 1,
```
ian-mi, line 181: I'm not sure this will work with the maps combined unless the keys are filtered somehow. One possibility would be derive a no resp code map from the resp code map to use in the processing metrics. Another possibility would be to go back to having two separate maps as before.

ian-mi: Another possibility might be to use some fancy cmp opts to compare only the relevant keys but that might be too fiddly.

yolocs: Ah you're right

ian-mi:
/lgtm
Great work!

PR #1295, opened 2020-06-17
Mark condition in broker/trigger finalizer when the deletion failed

### LINKED ISSUE #1190: Shall Finalizer mark resource sub-conditions when it failed?

grac3gao:
Fixes #1190

## Proposed Changes

- Mark condition in broker/trigger finalizer when the deletion failed
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/reconciler/broker/broker.go
+++ b/pkg/reconciler/broker/broker.go
@@ -300,8 +300,8 @@ func (r *Reconciler) deleteDecouplingTopicAndSubscription(ctx context.Context, b
 	projectID, err := utils.ProjectID(r.projectID, metadataClient.NewDefaultMetadataClient())
 	if err != nil {
 		logger.Error("Failed to find project id", zap.Error(err))
-		b.Status.MarkTopicUnknown("FinalizeTopicUnknown", "Failed to find project id: %w", err)
-		b.Status.MarkSubscriptionUnknown("FinalizeSubscriptionUnknown", "Failed to find project id: %w", err)
+		b.Status.MarkTopicUnknown("FinalizeTopicProjectIdNotFound", "Failed to find project id: %w", err)
+		b.Status.MarkSubscriptionUnknown("FinalizeSubscriptionProjectIdNotFound", "Failed to find project id: %w", err)
 		return err
 	}
 
@@ -310,8 +310,8 @@ func (r *Reconciler) deleteDecouplingTopicAndSubscription(ctx context.Context, b
 		client, err := pubsub.NewClient(ctx, projectID)
 		if err != nil {
 			logger.Error("Failed to create Pub/Sub client", zap.Error(err))
-			b.Status.MarkTopicUnknown("FinalizeTopicUnknown", "Failed to create Pub/Sub client: %w", err)
-			b.Status.MarkSubscriptionUnknown("FinalizeSubscriptionUnknown", "Failed to create Pub/Sub client: %w", err)
+			b.Status.MarkTopicUnknown("FinalizeTopicPubSubClientCreationFailed", "Failed to create Pub/Sub client: %w", err)
+			b.Status.MarkSubscriptionUnknown("FinalizeSubscriptionPubSubClientCreationFailed", "Failed to create Pub/Sub client: %w", err)
 			return err
 		}
 		defer client.Close()
--- a/pkg/reconciler/broker/broker.go
+++ b/pkg/reconciler/broker/broker.go
@@ -246,6 +246,7 @@ func (r *Reconciler) reconcileDecouplingTopicAndSubscription(ctx context.Context
 		if err != nil {
 			logger.Error("Failed to create Pub/Sub client", zap.Error(err))
 			b.Status.MarkTopicUnknown("PubSubClientCreationFailed", "Failed to create Pub/Sub client: %w", err)
+			b.Status.MarkSubscriptionUnknown("PubSubClientCreationFailed", "Failed to create Pub/Sub client: %w", err)
 			return err
 		}
 		defer client.Close()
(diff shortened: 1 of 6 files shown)


pkg/reconciler/utils/pubsub/topic.go
```diff
@@ -71,11 +71,13 @@ func (r *Reconciler) DeleteTopic(ctx context.Context, id string, obj runtime.Obj
 	exists, err := topic.Exists(ctx)
 	if err != nil {
 		logger.Error("Failed to verify Pub/Sub topic exists", zap.Error(err))
+		updater.MarkTopicUnknown("FinalizeTopicUnknown", "failed to verify Pub/Sub topic exists: %w", err)
 		return err
 	}
 	if exists {
 		if err := topic.Delete(ctx); err != nil {
 			logger.Error("Failed to delete Pub/Sub topic", zap.Error(err))
+			updater.MarkTopicFailed("FinalizeTopicFailed", "failed to delete Pub/Sub topic: %w", err)
```
liu-cong, line 80: This is tricky. If we fail to delete the topic, depending on how you look at it, the Broker can be treated as "READY" because it can function properly.

Same comment for the pull subscription.

pkg/reconciler/utils/pubsub/subscription.go
```diff
@@ -71,24 +71,27 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, id string, subCo
...
 	exists, err := sub.Exists(ctx)
 	if err != nil {
 		logger.Error("Failed to verify Pub/Sub subscription exists", zap.Error(err))
+		updater.MarkSubscriptionUnknown("FinalizeSubscriptionUnknown", "failed to verify Pub/Sub subscription exists: %w", err)
 		return err
 	}
 	if exists {
-		return r.deleteSubscription(ctx, sub, obj)
+		if err = r.deleteSubscription(ctx, sub, obj, updater); err != nil {
+			updater.MarkSubscriptionFailed("FinalizeSubscriptionFailed", "failed to delete Pub/Sub subscription: %w", err)
```
liu-cong, line 87: In my opinion, we shouldn't mark Broker as not ready because the pull subscription still exists and the Broker works properly. Instead, once we successfully delete the pull subscription, we should MarkSubscriptionFailed.

I know this depends on how do you look at it. If you treat "BrokerBeingDeleted" in combination with "FinalizeSubscriptionFailed" as a condition, then it makes sense. However, if I am a user, I probably just look at the overall status.

pkg/reconciler/utils/pubsub/topic.go
grac3gao: Sync offline, we think marking unknown would be more reasonable than failed. If the topic deletion failed, you can't tell whether the remaining topic is still working or not. Probably the topic is still working after deletion failed, then, it is not proper to mark failed.

pkg/reconciler/broker/broker.go
```diff
@@ -307,6 +310,8 @@ func (r *Reconciler) deleteDecouplingTopicAndSubscription(ctx context.Context, b
 		client, err := pubsub.NewClient(ctx, projectID)
 		if err != nil {
 			logger.Error("Failed to create Pub/Sub client", zap.Error(err))
+			b.Status.MarkTopicUnknown("FinalizeTopicUnknown", "Failed to create Pub/Sub client: %w", err)
```
liu-cong, line 313: The reason  can be "FinalizeTopicPubSubClientCreationFailed"

pkg/reconciler/utils/pubsub/subscription.go
```diff
@@ -71,24 +71,27 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, id string, subCo
...
+			updater.MarkSubscriptionUnknown("FinalizeSubscriptionDeletionFailed", "failed to delete Pub/Sub subscription: %w", err)
+			return err
+		}
 	}
-
 	return nil
 }
 
-func (r *Reconciler) deleteSubscription(ctx context.Context, sub *pubsub.Subscription, obj runtime.Object) error {
+func (r *Reconciler) deleteSubscription(ctx context.Context, sub *pubsub.Subscription, obj runtime.Object, updater StatusUpdater) error {
```
liu-cong, line 94: It seems that "updater" is not used.

pkg/reconciler/trigger/trigger.go
```diff
@@ -240,6 +241,8 @@ func (r *Reconciler) deleteRetryTopicAndSubscription(ctx context.Context, trig *
 	projectID, err := utils.ProjectID(r.projectID, metadataClient.NewDefaultMetadataClient())
 	if err != nil {
 		logger.Error("Failed to find project id", zap.Error(err))
+		trig.Status.MarkTopicUnknown("FinalizeTopicUnknown", "Failed to find project id: %w", err)
```
liu-cong, line 244: nit: same as above, the reason can be more specific

pkg/reconciler/trigger/trigger.go
```diff
@@ -248,6 +251,8 @@ func (r *Reconciler) deleteRetryTopicAndSubscription(ctx context.Context, trig *
 		client, err := pubsub.NewClient(ctx, projectID)
 		if err != nil {
 			logger.Error("Failed to create Pub/Sub client", zap.Error(err))
+			trig.Status.MarkTopicUnknown("FinalizeTopicUnknown", "Failed to create Pub/Sub client: %w", err)
```
liu-cong, line 254: same nit

liu-cong:
/lgtm
/approve

Thanks @grac3gao

Issue #1209, opened 2020-06-04
Rename CE types and change other attributes such as dataschema


nachocano:
**Problem**
There has been recent discussions on renaming the CE types attributes of the different sources, to include things like versioning and to make them more consistent. 
Similarly, the schemas will be hosted by Google instead of here in Github.
Creating this issue so that we keep track of it

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
CE attributes agreed upon and changed.  Potentially, a post-install job to rename all trigger filtering on previous values

nachocano:
fyi @yolocs @akashrv

nachocano:
@yolocs when doing this one, if you can put the renamed strings in `register.go` as it's typically done in eventing, instead of the v1beta1 types, that would be great? Otherwise we are duplicating a bunch of constants between v1alpha1 and v1beta1. 
This will help with https://github.com/google/knative-gcp/issues/1274

PR #1296, opened 2020-06-17
Moving to cev2 - publisher, receive adapter - Implementing KRShaped

### LINKED ISSUE #1166: Implement KRShaped for Sources
**Problem**
Most resources are now implementing [KRShaped](https://github.com/knative/pkg/blob/08a86da47d28b270dfa1bddf1f73a8944eef883e/apis/duck/v1/kresource_type.go#L31). We should do so in sources. 
Example PR here: https://github.com/knative/eventing/pull/3226

Contributor

**Exit Criteria**
Sources implement KRShaped duck

### LINKED ISSUE #1167: Migrate publisher and receiveAdapter to ce v2
**Problem**
We should migrate the publisher and receive adapter to start using the CloudEvents SDK v2 bindings.

Contributor

**Exit Criteria**
Using the CE SDK v2

https://github.com/knative/eventing/pull/3206
https://github.com/knative/eventing/pull/3205
and sources in eventing-contrib

nachocano:
Fixes #1167
Fixes #1166
Helps with https://github.com/google/knative-gcp/issues/1209

## Proposed Changes

- 🧽  Moving publisher and receive adapter to cev2 and using wire.
- 🧽  Updating some of the CE formats due to ongoing conversations. There were undocumented extensions we were adding that are not needed. The release notes of @yolocs PR should include all the breaking changes we are planning to  make.
- 🧽 Implementing KRShaped in all resources otherwise it does not compile

- TODOs
  - Fix adapter UTs to be done in a follow up as it might take sometime and don't want to block @capri-xiyue nor @yolocs  https://github.com/google/knative-gcp/issues/1333
 - Move e2e tests to cev2. https://github.com/google/knative-gcp/issues/1336

**Release Note**

```release-note
We are using CE SDK v2 for Sources and Channels, with all its performance improvements
```

**Docs**


nachocano:
/assign @Harwayne 
/assign @ian-mi 
/assign @yolocs 

Guys this is a WIP PR, but it is kind of big so I want to get your input asap. Especially on some ugly part of our old receive adapter that IIRC we use for channels. There is a tracing part that I'm not sure what we are actually doing...

@yolocs this will help out with the changes to events schemas. The strings hasn't been changed yet, but I refactored things so that it would be easy to do...

Either way, I'm having seconds thoughts about including this as it's big and kind of last minute. We can always do the changes of the events without moving to cev2, and then do the move...
The upside of doing this now is that I've heard from folks that the old sdk was not very performant. So this might improve the performance of our sources data plane, which we have been neglecting so far.

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,202 @@ package adapter
...
+		// TODO check if this is OK
 		// Update the tracing information to use the span returned by the transformer.
-		ctx = trace.NewContext(ctx, trace.FromContext(transformedCTX))
-	}
+		// ctx = trace.NewContext(ctx, trace.FromContext(transformedCTX))
+		if span := trace.FromContext(ctx); span.IsRecordingEvents() {
+			span.Annotate(ceclient.EventTraceAttributes(e),
+				"Event reply",
+			)
+		}
```
nachocano, line 182: @ian-mi @Harwayne 
here is the part of tracing that I have no clue about. Is this somewhat equivalent to the old:
`trace.NewContext(ctx, trace.FromContext(transformedCTX))` ?

nachocano: OK, seems that the e2e tests failing are due to tracing problems. I'm surely doing something wrong...

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,202 @@ package adapter
...
+	// Done channel has been closed, we need to gracefully shutdown. The cancel() method will start its
+	// shutdown, if it hasn't finished in a reasonable amount of time, just return an error.
+	cancel()
+	select {
+	case err := <-errCh:
+		return err
+	case <-time.After(30 * time.Second):
+		return errors.New("timeout shutting down adapter")
 	}
+}
```
yolocs, line 111: I don't quite understand this logic. To reach `cancel()` it seems that either `Receive` returned an error or context was done. If it's an error (1st case), then it means the pull has already stopped. If the context was done (2nd case), then it should already be handled by the `Receive` function. So it looks to me the `cancel()` is redundant and we just need to wait 30s. Thoughts?

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,202 @@ package adapter
...
 	// If a transformer has been configured, then transform the message.
 	// Note that this path in the code will be executed when using the receive adapter as part of the underlying Channel
 	// of a Broker. We currently set the TransformerURI to be the address of the Broker filter pod.
-	// TODO consider renaming transformer as it is confusing.
-	if a.transformer != nil {
-		transformedCTX, transformedEvent, err := a.transformer.Send(ctx, event)
-		rtctx := cloudevents.HTTPTransportContextFrom(transformedCTX)
+	// TODO consider renaming transformer as it is confusing. And see if we can get rid of this.
+	if a.transformerURI != "" {
+		resp, err := a.sendMsg(ctx, a.transformerURI, message)
```
yolocs, line 139: Are we really hitting this code path?

pkg/pubsub/adapter/adapter.go
nachocano: I think you are right, I copied/pasted this from some eventing stuff. I'll double check.
Will fix the other stuff first, and then address this.

pkg/pubsub/adapter/adapter.go
nachocano: yes, for for replies we are... it's pretty nasty. Added a couple of TODOs in the code.. This has been here forever and we should clean it up

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,218 @@ package adapter
...
+			msg.Ack()
+			return
 		}
-		// Update the event with the transformed one.
-		event = *transformedEvent
+
 		// Update the tracing information to use the span returned by the transformer.
-		ctx = trace.NewContext(ctx, trace.FromContext(transformedCTX))
+		// TODO how to get the span returned by the transformer?
+		ctx = trace.NewContext(ctx, trace.FromContext(ctx))
```
nachocano, line 190: @Harwayne @ian-mi 
I think this is where I'm messing up... Previously we had a context returned and we created a new context with the returned span... not sure what is the proper way of doing this with the new cev2 sdk

nachocano: ok, thanks for the pointer Ian, added some stuff. Let's see if that works. 
Otherwise I'll take a look next week

cmd/pubsub/receive_adapter/main.go
```diff
@@ -72,26 +143,53 @@ func main() {
...
-		logger.Fatal("failed to start adapter: ", zap.Error(err))
+	// Convert base64 encoded json map to extensions map.
+	extensions, err := utils.Base64ToMap(env.ExtensionsBase64)
+	if err != nil {
+		logger.Error("Failed to convert base64 extensions to map: %v", zap.Error(err))
 	}
+
+	ctx = context.WithProjectKey(ctx, projectID)
+	ctx = context.WithTopicKey(ctx, env.Topic)
+	ctx = context.WithSubscriptionKey(ctx, env.Subscription)
```
ian-mi, line 167: Can these be adapter args instead of passed through ctx? It might be worthwhile to create an adapter args struct since InitializeAdapter already has so many arguments.

pkg/broker/handler/fanout.go
```diff
@@ -176,7 +176,6 @@ func (p *FanoutPool) SyncOnce(ctx context.Context) error {
 
 		// Start the handler with broker key in context.
 		hc.Start(handlerctx.WithBrokerKey(ctx, b.Key()), func(err error) {
-			// We will anyway get an error because of https://github.com/cloudevents/sdk-go/issues/470
```
ian-mi, line 179: Is this change unrelated?

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,246 @@ package adapter
...
+	return &Adapter{
+		subscription:           subscription,
+		outbound:               outbound,
+		converter:              converter,
+		reporter:               reporter,
+		sinkURI:                string(sinkURI),
+		transformerURI:         string(transformerURI),
+		namespacedResourceName: types.NamespacedName{Name: string(name), Namespace: string(namespace)},
+		converterType:          converterType,
+		extensions:             extensions,
```
ian-mi, line 99: It might simplify the constructor to bundle these into a config struct.

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,246 @@ package adapter
...
+		// If there was a reply, we need to send it to the sink.
+		// We then overwrite the initial event we sent.
+		event, err = binding.ToEvent(ctx, respMsg)
 		if err != nil {
-			logger.Errorf("error transforming cloud event %q", event.ID())
-			a.reporter.ReportEventCount(args, rtctx.StatusCode)
-			return err
+			a.logger.Error("Failed to convert response message to event",
+				zap.Any("response", respMsg), zap.Error(err))
+			msg.Ack()
```
ian-mi, line 195: Shouldn't msg be nacked in this case?

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,246 @@ package adapter
...
-			a.reporter.ReportEventCount(args, rtctx.StatusCode)
-			return nil
-		}
-		// Update the event with the transformed one.
-		event = *transformedEvent
+
 		// Update the tracing information to use the span returned by the transformer.
-		ctx = trace.NewContext(ctx, trace.FromContext(transformedCTX))
+		var span *trace.Span
+		ctx, span = a.startSpan(ctx, event)
```
ian-mi, line 201: I don't think this is right place to start a new span. We should instead start a single span at the start of receive which will be the parent of both the transform and outbound transactions. The span name should probably be derived from the name of the source in this case.

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,246 @@ package adapter
...
-	logger := logging.FromContext(ctx)
-	logger.Debug("Converting event from transport.")
+	a.reporter.ReportEventCount(args, resp.StatusCode)
 
-	if msg, ok := m.(*cepubsub.Message); ok {
-		return converters.Convert(ctx, msg, a.SendMode, a.AdapterType)
+	if resp.StatusCode/100 != 2 {
+		a.logger.Error("Event delivery failed", zap.Int("StatusCode", resp.StatusCode))
+		msg.Nack()
+		return
```
ian-mi, line 227: response body also needs to be closed in this case.

cmd/pubsub/publisher/main.go
```diff
@@ -80,13 +87,21 @@ func main() {
...
+	if err != nil {
+		logger.Fatal("Unable to create publisher", zap.Error(err))
 	}
 
-	logger.Info("Starting Pub/Sub Publisher.", zap.Any("publisher", startable))
-	if err := startable.Start(ctx); err != nil {
-		logger.Fatal("failed to start publisher: ", zap.Error(err))
+	logger.Info("Starting publisher", zap.Any("publisher", publisher))
+	if err := publisher.Start(ctx); err != nil {
+		logger.Fatal("Failed to start publisher", zap.Error(err))
```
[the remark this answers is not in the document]
nachocano: Done

cmd/pubsub/receive_adapter/main.go
nachocano: yes, let me do that. It's kind of ugly as is right now. Thanks!

pkg/broker/handler/fanout.go
nachocano: unrelated. Was looking at code and saw that TODO but that bug was fixed so I removed the comment. Let me know if you want it back

pkg/pubsub/adapter/adapter.go
nachocano: Will do an AdapterArgs one and pass it here...

cmd/pubsub/receive_adapter/main.go
nachocano: agreed to add in AdapterArgs only the ones used by the adapter. In this case topic. The others are needed for initializing other objects as well

pkg/pubsub/adapter/adapter.go
nachocano: done

pkg/pubsub/adapter/adapter.go
nachocano: updated it. Thanks for the feedback on this one.
Let's see if tests pass

pkg/pubsub/adapter/adapter.go
nachocano: yes, I think so... thanks for the catch

pkg/pubsub/adapter/adapter.go
nachocano: done, thanks!

pkg/broker/handler/fanout.go
nachocano: feel free to unresolved if you want this comment back in

nachocano:
Pushed to see if tests pass. Will debug later tonight if they do not...

pkg/pubsub/adapter/adapter.go
nachocano: Done, thanks Chen!

pkg/pubsub/adapter/converters/pubsub_pull.go
```diff
@@ -0,0 +1,64 @@
...
+	"cloud.google.com/go/pubsub"
+	"context"
+	"fmt"
+	cev2 "github.com/cloudevents/sdk-go/v2"
+	. "github.com/cloudevents/sdk-go/v2/event"
+	"github.com/google/knative-gcp/pkg/apis/events/v1beta1"
+	. "github.com/google/knative-gcp/pkg/pubsub/adapter/context"
+)
+
+func convertPubSubPull(ctx context.Context, msg *pubsub.Message) (*cev2.Event, error) {
```
yolocs, line 29: This is only used for pubsub channel, right?

pkg/pubsub/adapter/converters/pubsub_pull.go
```diff
@@ -0,0 +1,64 @@
...
+	if err != nil {
+		return nil, err
+	}
+	topic, err := GetTopicKey(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	event.SetSource(v1beta1.CloudPubSubSourceEventSource(project, topic))
+	event.SetType(v1beta1.CloudPubSubSourcePublish)
```
yolocs, line 44: We can address this separately but I think this should be unset because we don't control the data in this case. Thoughts?

pkg/pubsub/adapter/converters/build_test.go
```diff
@@ -18,122 +18,117 @@ package converters
...
 		},
-		sendMode: Binary,
-		wantEventFn: func() *cloudevents.Event {
-			return buildCloudEvent(map[string]string{
-				"buildId":    buildID,
-				"status":     buildStatus,
-				"attribute1": "value1",
-				"attribute2": "value2",
-			}, buildID, buildStatus)
+		wantEventFn: func() *cev2.Event {
```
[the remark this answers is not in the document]
nachocano: nope, we are not blindly promoting attributes to extensions any longer for any of our sources. The only one I left it as is was the PullSubscription, which is technically internal and we don't advertise it as a source. 

This is part of the work @yolocs has been doing on the schema definitions for our sources... More changes are needed after this refactor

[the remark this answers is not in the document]
nachocano: and we should properly document in the release notes all those changes

pkg/pubsub/adapter/converters/pubsub.go
```diff
@@ -20,71 +20,46 @@ import (
...
-	// Make a new event and convert the message payload.
-	event := cloudevents.NewEvent(cloudevents.VersionV1)
-	event.SetID(tx.ID)
-	event.SetTime(tx.PublishTime)
-	event.SetSource(v1alpha1.CloudPubSubSourceEventSource(tx.Project, tx.Topic))
-	event.SetType(v1alpha1.CloudPubSubSourcePublish)
-	// Set the schema if it comes as an attribute.
-	if val, ok := msg.Attributes["schema"]; ok {
-		delete(msg.Attributes, "schema")
-		event.SetDataSchema(val)
```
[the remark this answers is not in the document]
nachocano: this was again an opinionated implementation we did long time back. We were saying if they send us schema as an attribute, then we use it... This if actually never went through, unless some user send a pubsub message with an attribute named schema... 
We are not doing this anymore, we decided to always put the PubSub Push message format in the payload, which a well-defined schema

pkg/pubsub/adapter/converters/pubsub_test.go
```diff
@@ -23,126 +23,54 @@ import (
...
 func TestConvertCloudPubSub(t *testing.T) {
 
 	tests := []struct {
 		name        string
-		message     *cepubsub.Message
-		sendMode    ModeType
-		wantEventFn func() *cloudevents.Event
+		message     *pubsub.Message
+		wantEventFn func() *cev2.Event
 		wantErr     bool
```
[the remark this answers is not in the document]
nachocano: my bad, it's because of the attributes change. Added an extra check for invalid context. Thanks!

pkg/pubsub/adapter/converters/scheduler.go
```diff
@@ -20,65 +20,41 @@ import (
...
-	event.SetSource(v1alpha1.CloudSchedulerSourceEventSource(parentName, schedulerName))
+	event.SetSource(v1beta1.CloudSchedulerSourceEventSource(parentName, schedulerName))
 	event.SetSubject(resources.ExtractJobID(jobName))
 
-	// Set the mode to be an extension attribute.
-	event.SetExtension("knativecemode", string(sendMode))
-	event.Data = msg.Data
-	event.DataEncoded = true
-	// Attributes are extensions.
-	if msg.Attributes != nil && len(msg.Attributes) > 0 {
```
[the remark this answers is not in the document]
nachocano: we are not... we do not blindly promote attributes to extensions. We might set some attributes. In the above case, we use some for the CE source and CE subject and then delete them. Those values were populated from our controller, so it's fine to remove... 

Why do you think is necessary to delete them?

pkg/reconciler/intevents/reconciler.go
```diff
@@ -30,7 +30,7 @@ import (
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/equality"
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
-	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1"
```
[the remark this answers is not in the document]
nachocano: yeah, some goland IDE thing... let me add it back

pkg/reconciler/intevents/pullsubscription/resources/receive_adapter_test.go
```diff
@@ -205,7 +207,7 @@ func TestMakeFullReceiveAdapter(t *testing.T) {
 				},
 			},
 			Topic:       "topic",
-			AdapterType: "adapter-type",
+			AdapterType: string(converters.PubSubPull),
```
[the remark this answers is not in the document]
nachocano: Done! thanks!

pkg/reconciler/intevents/reconciler.go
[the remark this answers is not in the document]
nachocano: added

pkg/tracing/attributes.go
```diff
@@ -30,6 +30,10 @@ var (
 	PubSubProtocolAttribute = tracing.MessagingProtocolAttribute(PubSubProtocol)
 )
 
-func ChannelMessagingDestination(c types.NamespacedName) string {
+func ChannelDestination(c types.NamespacedName) string {
 	return fmt.Sprintf("channel:%s.%s", c.Name, c.Namespace)
 }
+
+func SourceDestination(s types.NamespacedName) string {
+	return fmt.Sprintf("source:%s.%s", s.Name, s.Namespace)
```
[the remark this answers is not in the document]
nachocano: another good point. Done, thanks Ian!

pkg/pubsub/adapter/converters/scheduler.go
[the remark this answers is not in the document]
ian-mi: I don't, I'm questioning whether that code can be removed.

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	cloud.google.com/go/logging v1.0.1-0.20200331222814-69e77e66e597
 	cloud.google.com/go/pubsub v1.3.2-0.20200506222144-2c46308f8465
 	cloud.google.com/go/storage v1.8.0
-	github.com/cloudevents/sdk-go v1.2.0
+	github.com/cloudevents/sdk-go v1.0.0
```
ian-mi: Was this intentional?

pkg/pubsub/adapter/converters/scheduler.go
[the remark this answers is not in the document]
nachocano: got it, yes, the code can be removed

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,252 @@ package adapter
...
-func (a *Adapter) Start(ctx context.Context) error {
-	var err error
-
-	if a.SendMode == "" {
-		a.SendMode = converters.DefaultSendMode
-	}
+// Stop stops the adapter.
+func (a *Adapter) Stop() {
+	a.cancel()
+}
```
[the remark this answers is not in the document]
nachocano: will be once I get the UTs working. 
Mind if I leave it?

pkg/pubsub/adapter/adapter.go
```diff
@@ -18,253 +18,252 @@ package adapter
...
+	return a.outbound.Do(req)
 }
 
-func (a *Adapter) newHTTPClient(ctx context.Context, target string) (cloudevents.Client, error) {
-	tOpts := []http.Option{
-		cloudevents.WithTarget(target),
+func (a *Adapter) startSpan(ctx context.Context, event *cev2.Event) (context.Context, *trace.Span) {
+	spanName := tracing.SourceDestination(a.namespacedName)
+	// This receive adapter code is used both for Sources and Channels.
+	// An ugly way to identify whether it was created from a Channel is to look at the resourceGroup.
```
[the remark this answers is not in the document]
nachocano: yeah, this is old stuff but mainly due to metrics. At the time we wanted to know the resourceGroup of a source (e.g., cloudpubsubsources, cloudauditlogssources, etc)... so that we can report metrics from particular sources types and not just a generic `Source`, that is why this ugly resourceGroup stuff...
If you don't mind, I'd rather leave it as is for now, and then once we do a pass on metrics, we can revisit.. Sounds good?

cmd/pubsub/receive_adapter/main.go
```diff
@@ -72,26 +144,62 @@ func main() {
...
+	extensions, err := utils.Base64ToMap(env.ExtensionsBase64)
+	if err != nil {
+		logger.Error("Failed to convert base64 extensions to map: %v", zap.Error(err))
+	}
+
+	logger.Info("Initializing adapter", zap.String("projectID", projectID), zap.String("topicID", env.Topic), zap.String("subscriptionID", env.Subscription))
+
+	args := &AdapterArgs{
+		TopicID:        env.Topic,
+		ConverterType:  converters.ConverterType(env.AdapterType),
```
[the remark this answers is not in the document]
nachocano: if it's not an adapter registered (e.g., CAL, PubSub, Storage, Scheduler, or PubSubPull), we just use the default conversion, which is basically doing:

`binding.ToEvent(ctx, cepubsub.NewMessage(msg))`

pkg/pubsub/adapter/converters/pubsub_pull.go
nachocano: this is actually used for PullSubscriptions. I decided to leave the logic there. We had some E2E tests for it.
But we have our Pullsubscription resources as internal, so users should actually be using our four sources, and not this stuff....

As we don't do any opinionated conversions for the channel, the channel just uses:

`binding.ToEvent(ctx, cepubsub.NewMessage(msg))`

pkg/pubsub/adapter/converters/pubsub_pull.go
nachocano: this is for an internal opinionated Pullsubscription not the Channel. 
I think is fine to leave it...

nachocano: unless you think otherwise

pkg/pubsub/adapter/converters/build_test.go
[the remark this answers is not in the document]
ian-mi: Should this be deleted then? Or can you switch this back to using cmp.Diff.

nachocano:
Darn, sorry guys! I updated deps to try to get rid of cloudevents sdk v1.. the import is still there, not sure why...
But now there is a breaking change in eventing. The KRShaped stuff... Will fix that and continue with the review comments...

My apologies because this is actually turning into a much bigger issue

nachocano:
The genreconciler changed and uses KRShaped. Implemented KRShaped in all the types we use in reconcilers. Mainly v1beta1 types and v1alpha1 of BrokerCell.
Removed all the old generated reconciler code. 

It seems that a trigger_test is failing. 
Pinged @liu-cong as he might now better...

pkg/pubsub/adapter/adapter.go
```diff
@@ -246,11 +246,17 @@ func (a *Adapter) sendMsg(ctx context.Context, address string, msg binding.Messa
 }
 
 func (a *Adapter) startSpan(ctx context.Context, event *cev2.Event) (context.Context, *trace.Span) {
+	spanName := tracing.SourceDestination(a.namespacedName)
+	// This receive adapter code is used both for Sources and Channels.
+	// An ugly way to identify whether it was created from a Channel is to look at the resourceGroup.
+	if a.resourceGroup == messaging.ChannelsResource.String() {
+		spanName = tracing.ChannelDestination(a.namespacedName)
```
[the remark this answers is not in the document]
nachocano: Done

pkg/pubsub/adapter/adapter.go
```diff
@@ -46,87 +66,66 @@ type Adapter struct {
...
 
-	errCh := make(chan error, 1)
+	// Augment context so that we can use it to create CE attributes.
+	ctx = WithProjectKey(ctx, a.projectID)
+	ctx = WithTopicKey(ctx, a.args.TopicID)
+	ctx = WithSubscriptionKey(ctx, a.subscription.ID())
+
 	go func() {
-		errCh <- a.subscription.Receive(ctx, a.receive)
+		done(a.subscription.Receive(ctx, a.receive))
```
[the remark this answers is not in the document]
nachocano: I was following what was done in fanout https://github.com/google/knative-gcp/blob/master/pkg/broker/handler/fanout.go#L178
I had a more convoluted version before and Chen commented on..

You mean just do:
`return a.subscription.Receive(ctx, a.receive)` ?

I think it will be easier as is for writing the UTs... Unless you have a strong objection, I'd rather leave it for now... WDYT

liu-cong:
The root cause of the unit test failures is that the new genreconciler now unconditionally calls `PreProcessReconcile` which sets the init conditions if not set. Previously the conditions are only set by our own reconcilers.

nachocano:
> The root cause of the unit test failures is that the new genreconciler now unconditionally calls `PreProcessReconcile` which sets the init conditions if not set. Previously the conditions are only set by our own reconcilers.

Thanks @liu-cong !!!

nachocano:
pull-google-knative-gcp-wi-tests is failing pretty much all the time. I'm thinking of disabling them if folks are having the same issues in other PRs

nachocano:
Seems WI tests are not required to merge...
Waiting until they run one more time, and the UTs as well (there was a flake on the configMap assertion). 
I think we can get this in to unblock people and I can focus on the follow ups... @yolocs @ian-mi

pkg/pubsub/adapter/adapter.go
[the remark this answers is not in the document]
ian-mi: Yes, async callbacks are often an antipattern in go. A synchronous api can always be easily converted to an async callback by the caller using a goroutine.

pkg/pubsub/adapter/converters/storage_test.go
```diff
@@ -19,171 +19,151 @@ package converters
...
 
 func TestConvertCloudStorageSource(t *testing.T) {
 
 	tests := []struct {
 		name        string
-		message     *cepubsub.Message
-		sendMode    ModeType
-		wantEventFn func() *cloudevents.Event
+		message     *pubsub.Message
+		wantEventFn func() *cev2.Event
```
ian-mi, line 45: This is not being used either

pkg/pubsub/adapter/stats_reporter_test.go
```diff
@@ -27,17 +27,12 @@ import (
...
-		EventType:     "dev.knative.event",
-		EventSource:   "unit-test",
-		Name:          "testobject",
-		ResourceGroup: "testresourcegroup",
+		EventType:   "dev.knative.event",
+		EventSource: "unit-test",
 	}
 
-	r := NewStatsReporter()
+	r, _ := NewStatsReporter("testobject", "testns", "testresourcegroup")
```
ian-mi, line 35: err here should be fatal

pkg/pubsub/adapter/adapter.go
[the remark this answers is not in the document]
nachocano: SG, done, thanks for the feeback

pkg/pubsub/adapter/converters/storage_test.go
nachocano: Mind if we do in a follow up?

pkg/pubsub/adapter/converters/storage_test.go
```diff
@@ -19,171 +19,151 @@ package converters
...
-				"knative-gcp": "com.google.cloud.storage",
-				"bucketId":    "my-bucket",
+				"bucketId": bucket,
+				"objectId": objectId,
 			},
 		},
-		sendMode: Binary,
-		wantEventFn: func() *cloudevents.Event {
-			return storageCloudEvent(map[string]string{})
+		wantEventFn: func() *cev2.Event {
```
nachocano, line 79: it is used here?

nachocano:
Agreed with @ian-mi and @yolocs to get this in and I'll address the unresolved comments in the follow up. 
We need to unblock @yolocs work and allow @capri-xiyue to get new numbers on receive adapter resource requests and limits

pkg/pubsub/adapter/converters/storage_test.go
ian-mi: it's set here but I don't see where it's used

pkg/pubsub/adapter/stats_reporter_test.go
nachocano: ack, let me do in a follow up

pkg/pubsub/adapter/converters/storage_test.go
nachocano: actually this should be removed, you are right... 
doing it in a follow up

pkg/pubsub/adapter/converters/build_test.go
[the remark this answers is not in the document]
nachocano: Done!

pkg/pubsub/adapter/converters/scheduler.go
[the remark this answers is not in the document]
nachocano: I did remove it?

pkg/pubsub/adapter/converters/storage_test.go
nachocano: done

pkg/pubsub/adapter/stats_reporter_test.go
nachocano: done

nachocano:
@ian-mi addressed the tests comments so that I don't forget. @yolocs will have to do changes either way and I'll also improve stuff with the follow up

nachocano:
Need one more lgtm

pkg/pubsub/adapter/converters/scheduler.go
```diff
@@ -20,65 +20,41 @@ import (
...
+	event.SetType(v1beta1.CloudSchedulerSourceExecute)
+
 	// Set the source and subject if it comes as an attribute.
-	jobName, ok := msg.Attributes[v1alpha1.CloudSchedulerSourceJobName]
+	// We added this attributes so that we could identify the scheduler.
+	// We should remove them here.
+	jobName, ok := msg.Attributes[v1beta1.CloudSchedulerSourceJobName]
 	if ok {
-		delete(msg.Attributes, v1alpha1.CloudSchedulerSourceJobName)
+		delete(msg.Attributes, v1beta1.CloudSchedulerSourceJobName)
```
ian-mi, line 41: I think my comment below was ambiguous, I was thinking that the deletion of these attributes is no longer necessary.

nachocano: ohh I think they are, because with them we populate ce-source and ce-subject...

nachocano: darn, I get your point... no need to delete them argggg

nachocano: the arguments are necessary, no need to delete, let me remove that

nachocano: Done, thanks for the catch

ian-mi:
/lgtm
Thanks for the changes!

PR #1298, opened 2020-06-17
Use CloudEvents 1.0 in docs

grantr:
Found a couple references to CloudEvents 0.3 in broker examples and pubsub binding spec. Changed them all to 1.0.

## Proposed Changes

- Use CloudEvents 1.0 in broker examples and pubsub binding spec


Harwayne:
@googlebot CLA check

PR #1302, opened 2020-06-17
Set Stackdriver as default metrics backend

zargarpur:
And set sampling period to 60 seconds


ian-mi:
/lgtm
WDYT about disabling custom metrics by default?

zargarpur:
> /lgtm
> WDYT about disabling custom metrics by default?

Is it not disabled by default? I thought you had to explicitly enabled it with `metrics.allow-stackdriver-custom-metrics: "true"`

ian-mi:
Ah you're right, I had missed that we were explicitly enabling them in e2e testing but your other PR should fix that.

Issue #1306, opened 2020-06-18
Default credential doesn't correctly set for v1beta1 resource


grac3gao:
**Problem**
In v1alpha1, the default webhook will set default credential if the resource doesn't specify any credential. However, it doesn't work in v1beta1.

Reproduce:
Apply an CloudAuditLogSource with v1beta1 without specifying secret. 
After applying it, the default secret will not show up. 
It supposes to be added by the default webhook (if set the default secret in the configMap config-gcp-auth)

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

grac3gao:
Some investigation:
After testing, in v1beta1, the webhook do try to add the default secret to the source, however, it is removed somehow. Suspecting there is some conversion issues, still investigating.

grac3gao:
#1310 changes storage version from v1beta1 to v1alpha1, this temporally fix this issue. But we need to investigate the root cause for it, and change back the storage version

grac3gao:
Open #1311 as a follow-up after the quick fix

capri-xiyue:
I can't reproduce the same issue in my cluster.
I tried to create CloudAuditLogsSource with version v1eta1,  it can be ready and it is able to send events

grac3gao:
@capri-xiyue Nacho also encountered this issue. There is a new merged PR submitted by him https://github.com/google/knative-gcp/pull/1312 to fix it. The root cause is probably due to v1beta1 types not registered in the webhook. After that PR, the above issue is solved. And the storage version changed back. I'll close this issue.

capri-xiyue:
> @capri-xiyue Nacho also encountered this issue. There is a new merged PR submitted by him #1312 to fix it. The root cause is probably due to v1beta1 types not registered in the webhook. After that PR, the above issue is solved. And the storage version changed back. I'll close this issue.

After I created a new project and new cluster, I can reproduce the same error.

PR #1310, opened 2020-06-18
Change back storage version

grac3gao:
Fixes #

## Proposed Changes

- Change back storage version to temporally fix issue for #1306
-
-

**Release Note**

```release-note

```

**Docs**


capri-xiyue:
We need to figure out why e2e tests pass when we use storage version to v1beta1

capri-xiyue:
I can't reproduce the same issue in my cluster.
I tried to create CloudAuditLogsSource with version v1eta1 with crd storage version v1beta1, it can be ready and it is able to send events

Issue #1311, opened 2020-06-19
Storage version change causes setDefault failure for v1beta1 resources


grac3gao:
**Problem**
As a follow-up of https://github.com/google/knative-gcp/issues/1306.
When we use v1beta1 resource, all setDefault() functionality seems don't work. 
This caused default credential not setting up, and annotation not setting up. 

We temporarily fix it by changing back the storage version.
Investigation shows that, in v1beta1, the setDefault() is running, and setting the defaults. However, somehow, it is removed later. 
Suspecting there is some conversion issues.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

grac3gao:
Another direction worth to investigate is the storage version change doesn't affect the e2e test, v1beta1's e2e test are all working.

nachocano:
Might have something to do with k8s version? Maybe e2e run with an older
version?

PR #1312, opened 2020-06-19
Registering v1beta1 types

nachocano:
Fixes https://github.com/google/knative-gcp/issues/1311

## Proposed Changes

- 🐛 Registering v1beta1 types in the webhook
- 🧽 Rollback storage change and storing v1beta1 types.

**Release Note**

```release-note
NONE
```

**Docs**

/assign @grac3gao 
/assign @Harwayne


nachocano:
I guess we've been carrying this for a while but became a problem when we started storing v1beta1 types. After this change, I cannot reproduce the problem in my cluster anymore.

nachocano:
Note that this will break the patch we have internally 
fyi @zargarpur

grac3gao:
Just checked my cluster, issue solved after this change too.
Thanks! @nachocano 
/lgtm
/hold in case for other reviewers to take a look

PR #1317, opened 2020-06-19
Remove setDefaults when creating resources

### LINKED ISSUE #1315: Don't directly set default in NewFoo style function
**Problem**
In `NewFoo` style function, we directly setDefault. 
https://github.com/google/knative-gcp/blob/93dbbbb6c32cb68bbc567d07e030f69a1c98e80f/pkg/reconciler/testing/auditlogs.go#L34-L44

It would be better to separate it as an option. 
In UT, we need to set default for credentials. But in E2E, we wish to rely on default webhook to set credentials (setting default here will make defaulting issues in webhook invisible, like #1310 ).

**Exit Criteria**

**Time Estimate (optional):** 

cc @capri-xiyue

danyinggu:
Fixes #1315 

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/events/auditlogs/auditlogs_test.go
```diff
@@ -160,7 +160,9 @@ func TestAllCases(t *testing.T) {
 				WithCloudAuditLogsSourceServiceName(testServiceName),
 				WithCloudAuditLogsSourceAnnotations(map[string]string{
 					duckv1beta1.ClusterNameAnnotation: testingMetadataClient.FakeClusterName,
-				})),
+				}),
+				WithCloudAuditLogsSourceDefaultGCPAuth(),
```
capri-xiyue, line 164: I think we should use `WithCloudAuditLogsSourceDefaultGCPAuth()` for every previous usage when we call `NewCloudAuditLogsSource()` I know in some cases, the UT can still pass but the logic is not correct.
The `Objects` should be a CloudAuditLogsSource with correct GCPAuth.
Similar for other resources.
I'd recommend you check usages for function `NewXXXX` and add `WithXXXDefaultGCPAuth` for all those usages.

pkg/reconciler/testing/pullsubscription.go
```diff
@@ -78,7 +77,6 @@ func NewPullSubscriptionWithoutNamespace(name string, so ...PullSubscriptionOpti
 	for _, opt := range so {
 		opt(s)
 	}
-	s.SetDefaults(gcpauthtesthelper.ContextWithDefaults())
```
grac3gao, line 81: It looks like for newpullsubscription, we have two functions. One for pullsubscription with default, one for pullsubscription without default. If you remove setdefault here, then there is no differences between these two function.

pkg/reconciler/events/auditlogs/auditlogs_test.go
capri-xiyue: In addition, I found the the coding styles related to setting GCPAuth of different resources is not consistent. It will be better if you can fix the inconsistencies in this isuse

PR #1320, opened 2020-06-22
 Refactor set default when creating resources for test

### LINKED ISSUE #1315: Don't directly set default in NewFoo style function

danyinggu:
Fixes #1315 

## Proposed Changes

- 🐛 Remove SetDefault from the NewXXX function when creating resources, so that e2e tests will let webhook to set defaults
- 🧽 Refactor SetDefault as a resource create option
- 🧽 Update all UT tests to use SetDefault option
- 🧽 Fix nits and variable names to make them consistent

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/testing/auditlogs.go
```diff
@@ -200,8 +199,8 @@ func WithCloudAuditLogsSourceAnnotations(Annotations map[string]string) CloudAud
 	}
 }
 
-func WithCloudAuditLogsSourceDefaultGCPAuth() CloudAuditLogsSourceOption {
+func WithCloudAuditLogsSourceSetDefaults() CloudAuditLogsSourceOption {
```
nachocano, line 202: if you are not passing arguments to the method I think you can just do...

`func WithCloudAuditLogsSourceSetDefaults(s *v1beta1.CloudAuditLogsSource) {
	s.SetDefaults(gcpauthtesthelper.ContextWithDefaults())
}`

and the caller code will just be `WithCloudAuditLogsSourceSetDefaults` instead of `WithCloudAuditLogsSourceSetDefaults()` everywhere...

nachocano:
Looks good, just a small nit.
Can you also add a description in the PR Proposed Changes section so that folks are aware of the change...
We will have to do this for other sources as well? Or they do not have this? If so, we can do that in a follow up...

capri-xiyue:
> Looks good, just a small nit.
> Can you also add a description in the PR Proposed Changes section so that folks are aware of the change...
> We will have to do this for other sources as well? Or they do not have this? If so, we can do that in a follow up...

We will need that for other sources. And Danying will work on other sources in this PR.
This is a WIP PR(just wanna review it for one source first)

capri-xiyue:
@danyinggu  You can follow what other PRs does of how to writing "Proposed Change" session.

pkg/reconciler/testing/trigger.go
```diff
@@ -208,3 +207,7 @@ func WithTriggerFinalizers(finalizers ...string) TriggerOption {
 		t.Finalizers = finalizers
 	}
 }
+
+func WithTriggerSetDefaults(t *brokerv1beta1.Trigger) {
+	t.SetDefaults(context.Background())
+}
```
capri-xiyue, line 213: nit:add empty line

capri-xiyue:
lgtm, just have some nits

Issue #1060, opened 2020-05-13
Add readiness to Broker ingress


liu-cong:
**Problem**
- Related to #785. 
- serve health endpoints
- Also add probes in the deployment yamls (or the BrokerCell controller if it replaces the yaml)

Ref: https://blog.colinbreck.com/kubernetes-liveness-and-readiness-probes-how-to-avoid-shooting-yourself-in-the-foot/

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
Manual test error path

PR #1322, opened 2020-06-22
Add broker ingress readiness prober

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1060
It doesn't seem we need any special readiness check.

## Proposed Changes

-
-
-

**Release Note**

```release-note
Add broker ingress readiness prober
```

**Docs**


liu-cong:
I think one thing we could check is if ALL brokers in the configmap are not ready, then the ingress is not ready. But this doesn't help with a subset of brokers being unready. The ingress also checks the broker readiness anyway, so not a big deal.

/lgtm

PR #1328, opened 2020-06-23
Unpin github.com/gogo/protobuf

ian-mi:
Upgrades github.com/gogo/protobuf to v1.3.1


yolocs:
I wonder if this change will cause a dependency conflict in our internal repo?

ian-mi:
This is already unpinned internally. Removing unnecessary replace directives in this repo should help prevent it from diverging.

PR #1334, opened 2020-06-24
Add Trigger and Filter Trace Events

### LINKED ISSUE #1331: Emit Trace Event When Processing an Event Reply
**Problem**
Adding a trace event when processing an event reply would help to delineate the reply spans from the original event spans.

ian-mi:
Fixes #1331

### DIFF
--- a/pkg/broker/handler/processors/deliver/processor.go
+++ b/pkg/broker/handler/processors/deliver/processor.go
@@ -82,12 +82,20 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
 	if !ok {
 		// If the broker no longer exists, then there is nothing to process.
 		logging.FromContext(ctx).Warn("broker no longer exist in the config", zap.String("broker", bk))
+		trace.FromContext(ctx).Annotate(
+			ceclient.EventTraceAttributes(event),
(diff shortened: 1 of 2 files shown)


pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -82,12 +82,20 @@ func (p *Processor) Process(ctx context.Context, event *event.Event) error {
 	if !ok {
 		// If the broker no longer exists, then there is nothing to process.
 		logging.FromContext(ctx).Warn("broker no longer exist in the config", zap.String("broker", bk))
+		trace.FromContext(ctx).Annotate(
+			ceclient.EventTraceAttributes(event),
```
grantr, line 86: This may be a naive question, so feel free to say so :) Why is this method not used in all the Annotate calls below?

ian-mi: It's a good question, I think it can be added to the retry event but I want to avoid adding it to the the reply event for now as that would require converting the reply to an event type which we want to avoid doing in the hot path. Also the broker ingress span should add these attributes so its not strictly necessary to attach them to the event.

Issue #1147, opened 2020-05-27
Manual verification - Trace GCP Source -> GCP Broker -> Cloud Run Service


Harwayne:
**Exit Criteria**
Manual verification that a complete Trace GCP Source -> GCP Broker -> Cloud Run Service is visible in StackDriver.

Harwayne:
Helper scripts to setup and tear down auth:

```shell
#!/bin/bash

set -e -u

export GSA_PROJECT=fill_me_in
export PROJECT=fill_me_in
export CLUSTER_NAME=fill_me_in

function secrets() {
  type=$1
  secret_name=$2
  gcloud iam service-accounts create "${CLUSTER_NAME}-$type" --project $GSA_PROJECT
  gcloud iam service-accounts keys create $type.json --project $GSA_PROJECT \
    --iam-account="${CLUSTER_NAME}-$type@$GSA_PROJECT.iam.gserviceaccount.com"
  kubectl --namespace cloud-run-events create secret generic $secret_name --from-file=key.json=$type.json
  rm -f "$type.json"
}

secrets "control" "google-cloud-key"
secrets "broker" "google-cloud-broker-key"
secrets "sources" "google-cloud-sources-key"

function iamPolicy() {
  type=$1
  shift 1
  roles=("$@")
  member="serviceAccount:${CLUSTER_NAME}-$type@$GSA_PROJECT.iam.gserviceaccount.com"
  for role in "${roles[@]}"; do
    gcloud projects add-iam-policy-binding $PROJECT \
      --role "roles/$role" \
      --member="$member"
  done
}

iamPolicy "control" \
  "logging.admin" \
  "pubsub.editor" \
  "cloudscheduler.admin" \
  "storage.admin"
iamPolicy "broker" \
  "pubsub.editor"
iamPolicy "sources" \
  "pubsub.editor"
```

Clean up

```shell
#!/bin/bash

set -u

export GSA_PROJECT=fill_me_in
export PROJECT=fill_me_in
export CLUSTER_NAME=fill_me_in

function deleteSecrets() {
  type=$1
  secret_name=$2
  gcloud iam service-accounts delete "${CLUSTER_NAME}-$type@$GSA_PROJECT.iam.gserviceaccount.com" --project $GSA_PROJECT --quiet
  kubectl --namespace cloud-run-events delete secret $secret_name
}

deleteSecrets "control" "google-cloud-key"
deleteSecrets "broker" "google-cloud-broker-key"
[...]

PR #1335, opened 2020-06-24
Accept both B3 and TraceContext Tracing

Harwayne:
Helps with #1147.

## Proposed Changes

- Use the `tracecontextb3.TraceContextEgress` in all of our client code for consistency with our server code (which should have been updated when upstream knative/eventing#3388 was vendored in).
- Only TraceContext style tracing is egressed.
    - This was done for backwards compatibility with the existing system, which sends via TraceContext style. We could output both formats, but it was determined that the additional overhead of ~100 bytes per message was too high for the potential benefit. See [here](https://github.com/knative/pkg/pull/1429#pullrequestreview-435062580) for more discussion.

**Release Note**

```release-note
NONE
```


Harwayne:
@nachocano should we hold this until after #1296 is in?

ericlem:
/lgtm
/hold
Holding for nacho's reply

nachocano:
Yes. It'd be great if we can hold it.
We need the other one in asap. And this will have a bunch of conflicts.
Hopefully we will get it in later today or tomorrow.
There is a new problem with UTs with the newer eventing reconcilers. Cong
is helping investigate

nachocano:
> And this will have a bunch of conflicts.

I take this back. I was changing all e2e to start using cev2 and removing the good_client, but I rollbacked those changes as it can be done in a follow up, after mine and yours get in...
I'm currently experiencing the flakiness on the Brokercell UT. Hope we can get mine in early tomorrow

nachocano:
/hold cancel
the other PR got merged. Thanks for waiting @Harwayne !

Harwayne:
/hold

I clearly did something wrong:

```
TestGCPBrokerTracing: broker_tracing.go:45: multiple root spans: span_id:7881518364301324575 name:"Recv./test-g-c-p-broker-tracing-qwp7n/gcp-bfcyyklc" start_time:{seconds:1593035140 nanos:460429848} end_time:{seconds:1593035140 nanos:783118377} parent_span_id:22965739263654966 labels:{key:"/http/host" value:"default-brokercell-ingress.cloud-run-events.svc.cluster.local"} labels:{key:"/http/method" value:"POST"} labels:{key:"/http/path" value:"/test-g-c-p-broker-tracing-qwp7n/gcp-bfcyyklc"} labels:{key:"/http/user_agent" value:"Go-http-client/1.1"} labels:{key:"g.co/agent" value:"opencensus-go 0.23.0; stackdriver-exporter 0.10.0"} labels:{key:"http.url" value:"/test-g-c-p-broker-tracing-qwp7n/gcp-bfcyyklc"}, span_id:6804529643887419959 name:"Recv./test-g-c-p-broker-tracing-qwp7n/gcp-bfcyyklc" start_time:{seconds:1593035140 nanos:460488105} end_time:{seconds:1593035140 nanos:783081955} parent_span_id:22965739263654966 labels:{key:"/http/host" value:"default-brokercell-ingress.cloud-run-events.svc.cluster.local"} labels:{key:"/http/method" value:"POST"} labels:{key:"/http/path" value:"/test-g-c-p-broker-tracing-qwp7n/gcp-bfcyyklc"} labels:{key:"/http/user_agent" value:"Go-http-client/1.1"} labels:{key:"g.co/agent" value:"opencensus-go 0.23.0; stackdriver-exporter 0.10.0"} labels:{key:"http.url" value:"/test-g-c-p-broker-tracing-qwp7n/gcp-bfcyyklc"}
```

ericlem:
Is there a possibility the handlers were already wrapped (or internally registered traces) in the new wrapping calls you added?

pkg/broker/ingress/handler.go
```diff
@@ -99,7 +99,7 @@ func NewHandler(ctx context.Context, httpReceiver HttpMessageReceiver, decouple
 
 // Start blocks to receive events over HTTP.
 func (h *Handler) Start(ctx context.Context) error {
-	return h.httpReceiver.StartListen(ctx, h)
+	return h.httpReceiver.StartListen(ctx, kncloudevents.CreateHandler(h))
```
ericlem, line 102: I think StartListen might also wrap the handler?

pkg/utils/clients/factories.go
```diff
@@ -68,7 +67,7 @@ func NewHTTPClient(ctx context.Context, maxConnsPerHost MaxConnsPerHost) *nethtt
 				MaxConnsPerHost:     int(maxConnsPerHost),
 				IdleConnTimeout:     30 * time.Second,
 			},
-			Propagation: &tracecontext.HTTPFormat{},
+			Propagation: tracecontextb3.TraceContextEgress,
```
ian-mi, line 70: This shouldn't be necessary as the client is already configured to use tracecontext.

test/test_images/internal/knockdown/knockdown.go
```diff
@@ -42,7 +43,7 @@ type Config struct {
 // Main should be called by the process' main function. It will run the knockdown test. The return
 // value MUST be used in os.Exit().
 func Main(config Config, kdr Receiver) int {
-	client, err := cloudevents.NewDefaultClient()
+	client, err := kncloudevents.NewDefaultClient()
```
ian-mi, line 46: I'm not sure that we should change the test image as we expect to always be given tracecontext headers.

Harwayne: My reasoning is consistency. There is little downside to accepting both, so why not do it? The only real reason we wouldn't is to verify that the incoming headers truly are TraceContext, rather than B3. 

I also expect sooner or later someone will copy-paste this code and want the 'best practice' to be what they copy.

pkg/utils/clients/factories.go
Harwayne: I think you're right, this isn't really needed as this function returns an `http.Client`, which is egress only. However, I would like to change it to be consistent with all other places in the repo. The actual difference in runtime will be miniscule.

test/test_images/internal/knockdown/knockdown.go
ian-mi: Right my concern is detecting a regression where B3 is used instead of tracecontext. I also think that for normal event consumers accepting only tracecontext is good practice as that is what they should expect the broker to give them. On the other hand broker components should always use the clients package providers.

pkg/pubsub/publisher/publisher.go
```diff
@@ -73,7 +74,7 @@ func NewPublisher(ctx context.Context, inbound HttpMessageReceiver, topic *pubsu
 
 // Start blocks to receive events over HTTP.
 func (p *Publisher) Start(ctx context.Context) error {
-	return p.inbound.StartListen(ctx, p)
+	return p.inbound.StartListen(ctx, kncloudevents.CreateHandler(p))
```
ericlem, line 77: Just from quick tracking, it looks like this is double wrapped too?   I think inbound is created via InitializePublisher(), which calls gcp/pkg/utils/clients.NewHTTPMessageReceiver, which in turn calls eventing/pkg/kncloudevents/NewHttpMessageReceiver(), which returns a kncloudevents.HttpMessageReceiver that has a call to CreateHandler() already inside it's StartListen handler?

ian-mi, line 77: Like ingress the publisher is now using the clients package providers so this can be removed. After looking closer I think since the shared providers are already using kncloudevents handler, we should have gotten the updated trace handling for free after eventing was updated.

ian-mi: I was just looking at the same thing. I think we should just get B3 handling for free since it was added to the kncloudevents message receiver.

pkg/utils/clients/factories.go
ian-mi: Can we just use the regular tracecontext format everywhere we don't need special behavior? That would maintain consistency and I feel that it is somewhat confusing to configure the client with a format that has special unneeded ingress behavior. I reviewed the tracecontextb3 PR before the formats were split but it should actually be unnecessary to have separate egress and ingress formats since the format is only every used with a handler (ingress) or a client transport (egress).

pkg/broker/ingress/handler.go
Harwayne: Yes it did, thanks!

pkg/pubsub/publisher/publisher.go
Harwayne: Yes it did, thanks!

Harwayne: You're right, reverted.

pkg/broker/handler/providers.go
```diff
@@ -45,7 +45,7 @@ var (
 				MaxConnsPerHost:     500,
 				IdleConnTimeout:     30 * time.Second,
 			},
-			Propagation: &tracecontext.HTTPFormat{},
+			Propagation: tracecontextb3.TraceContextEgress,
```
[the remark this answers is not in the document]
Harwayne: Is this the same argument as above? Namely that because this is an `http.Client`, it is inherently an egress only thing, and therefore doesn't need this new format?

pkg/utils/clients/factories.go
Harwayne: > Can we just use the regular tracecontext format everywhere we don't need special behavior? That would maintain consistency

Consistency with what? For the rest of this comment I will assume it is consistency within this repo or knative more generally. As opposed to consistency with the wider world of software using OpenCensus.

I disagree that it is simpler or more consistent to use `&tracecontext.HTTPFormat{}` for ingress and `tracecontextb3.TraceContextEgress` for egress, rather than always using `tracecontextb3.TraceContextEgress` for both. There is some runtime overhead, but it is going to be effectively imperceptible (one stack frame and an iteration of a size one slice per HTTP request, AFAICT). 

If we always use the same format, then developers don't have to understand when to use each. They just always use `tracecontextb3.TraceContextEgress`, which gives the desired behavior in both circumstances.

test/test_images/internal/knockdown/knockdown.go
Harwayne: Switched back with a comment explaining this is for tests only.

PR #1339, opened 2020-06-24
differentiate between unknown, true and false when propagate deployme…

### LINKED ISSUE #1332: Differentiate among "True", "False and "Unknown" when propagate Deployment Availability
**Problem**
Currently when we propagate deployment availability, we only set "true" or "false". https://github.com/google/knative-gcp/blob/87c676d2f0010ab6a6563829cef8c599a6254ef3/pkg/apis/intevents/v1beta1/pullsubscription_lifecycle.go#L88

**Exit Criteria**
Differentiate among "True", "False and "Unknown" when propagate Deployment Availability

capri-xiyue:
Fixes #1332 

## Proposed Changes
- 🎁 differentiate between unknown, true and false when propagate deployment status

### DIFF
--- a/pkg/apis/intevents/v1alpha1/brokercell_lifecycle.go
+++ b/pkg/apis/intevents/v1alpha1/brokercell_lifecycle.go
@@ -91,10 +91,21 @@ func (bs *BrokerCellStatus) MarkIngressFailed(reason, format string, args ...int
 // to determine if BrokerCellConditionFanout should be marked as true or
 // false.
 func (bs *BrokerCellStatus) PropagateFanoutAvailability(d *appsv1.Deployment) {
-	if duck.DeploymentIsAvailable(&d.Status, false) {
-		brokerCellCondSet.Manage(bs).MarkTrue(BrokerCellConditionFanout)
-	} else {
-		brokerCellCondSet.Manage(bs).MarkFalse(BrokerCellConditionFanout, "DeploymentUnavailable", "Deployment %q is unavailable.", d.Name)
+	deploymentAvailableFound := false
+	for _, cond := range d.Status.Conditions {
+		if cond.Type == appsv1.DeploymentAvailable {
+			deploymentAvailableFound = true
+			if cond.Status == corev1.ConditionTrue {
+				brokerCellCondSet.Manage(bs).MarkTrue(BrokerCellConditionFanout)
+			} else if cond.Status == corev1.ConditionFalse {
+				brokerCellCondSet.Manage(bs).MarkFalse(BrokerCellConditionFanout, cond.Reason, cond.Message)
+			} else if cond.Status == corev1.ConditionUnknown {
+				brokerCellCondSet.Manage(bs).MarkUnknown(BrokerCellConditionFanout, cond.Reason, cond.Message)
+			}
+		}
+	}
+	if !deploymentAvailableFound {
+		brokerCellCondSet.Manage(bs).MarkUnknown(BrokerCellConditionFanout, "DeploymentUnavailable", "The Deployment %q is unavailable.", d.Name)
 	}
 }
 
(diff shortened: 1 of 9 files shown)


pkg/apis/intevents/v1beta1/pullsubscription_lifecycle.go
```diff
@@ -86,11 +86,20 @@ func (s *PullSubscriptionStatus) MarkNoSubscription(reason, messageFormat string
...
+				pullSubscriptionCondSet.Manage(s).MarkTrue(PullSubscriptionConditionDeployed)
+			} else if cond.Status == corev1.ConditionFalse {
+				pullSubscriptionCondSet.Manage(s).MarkFalse(PullSubscriptionConditionDeployed, cond.Reason, cond.Message)
+			} else if cond.Status == corev1.ConditionUnknown {
+				pullSubscriptionCondSet.Manage(s).MarkUnknown(PullSubscriptionConditionDeployed, cond.Reason, cond.Message)
+			}
+		}
+	}
+	if !deploymentAvailableFound {
 		pullSubscriptionCondSet.Manage(s).MarkFalse(PullSubscriptionConditionDeployed, "DeploymentUnavailable", "The Deployment '%s' is unavailable.", d.Name)
```
nachocano, line 103: should we mark it as unknown in this case?

capri-xiyue: Yep. We should mark it to unknown

liu-cong:
The brokercell also uses similar logic for deployment status, which will be propagated to broker status.
https://github.com/google/knative-gcp/blob/4835f7b3a85e0d54e67d4633f8a4c3bfc069c343/pkg/apis/intevents/v1alpha1/brokercell_lifecycle.go#L94

Should we fix that as well?

pkg/apis/intevents/v1alpha1/brokercell_lifecycle.go
```diff
@@ -91,10 +91,21 @@ func (bs *BrokerCellStatus) MarkIngressFailed(reason, format string, args ...int
...
-		brokerCellCondSet.Manage(bs).MarkTrue(BrokerCellConditionFanout)
-	} else {
-		brokerCellCondSet.Manage(bs).MarkFalse(BrokerCellConditionFanout, "DeploymentUnavailable", "Deployment %q is unavailable.", d.Name)
+	deploymentAvailableFound := false
+	for _, cond := range d.Status.Conditions {
+		if cond.Type == appsv1.DeploymentAvailable {
+			deploymentAvailableFound = true
+			if cond.Status == corev1.ConditionTrue {
+				brokerCellCondSet.Manage(bs).MarkTrue(BrokerCellConditionFanout)
+			} else if cond.Status == corev1.ConditionFalse {
```
liu-cong, line 100: I just did a test, when you create a deployment, initially this is False until all pods are up and running for at least `minReadySeconds`. Therefore I think the behavior is still the same.

Note this is a general problem where clients give up when they see "False", not just for the deployment. I believe we could encounter this for other dependencies as well.

Maybe as a temporary solution we should just use True or Unknown.

capri-xiyue: We should reflect the real status. If that's `False`, we should set status to `False`. CLI should implement the grace period.

pkg/apis/intevents/v1alpha1/brokercell_lifecycle.go
```diff
@@ -91,10 +91,21 @@ func (bs *BrokerCellStatus) MarkIngressFailed(reason, format string, args ...int
 // to determine if BrokerCellConditionFanout should be marked as true or
 // false.
 func (bs *BrokerCellStatus) PropagateFanoutAvailability(d *appsv1.Deployment) {
-	if duck.DeploymentIsAvailable(&d.Status, false) {
-		brokerCellCondSet.Manage(bs).MarkTrue(BrokerCellConditionFanout)
-	} else {
-		brokerCellCondSet.Manage(bs).MarkFalse(BrokerCellConditionFanout, "DeploymentUnavailable", "Deployment %q is unavailable.", d.Name)
+	deploymentAvailableFound := false
```
liu-cong, line 94: nit: Can you make this a helper function and avoid duplicating it many times?
```
func PropagateDeploymentStatus(d *appsv1.Deployment, manager ConditionManager, ct ConditionType)
```

And here you can just call `PropagateDeploymentStatus(d, brokerCellCondSet.manager(bs), BrokerCellConditionFanout `

pkg/apis/intevents/v1alpha1/test_helper.go
```diff
@@ -74,6 +74,18 @@ func (t testHelper) UnavailableDeployment() *appsv1.Deployment {
 	return d
 }
 
+func (t testHelper) UnknownDeployment() *appsv1.Deployment {
+	d := &appsv1.Deployment{}
+	d.Name = "unknown"
+	d.Status.Conditions = []appsv1.DeploymentCondition{
+		{
+			Type:   appsv1.DeploymentAvailable,
+			Status: "Unknown",
```
liu-cong, line 83: nit: use `corev1.ConditionUnknown`

pkg/apis/intevents/v1alpha1/brokercell_lifecycle.go
liu-cong: SG

PR #1342, opened 2020-06-24
Emit trace event on ingress when broker is unavailable

### LINKED ISSUE #1329: Emit Trace Event When an Event is Sent to a Broker Which is Not Ready
**Problem**
If an event is sent to a broker which is not ready, broker ingress will return 503. A trace event should be emitted in this case to help identify the root cause of the error.

ian-mi:
Emits an event with error message during ingress when the target broker is not found, not ready, or
misconfigured. Also adds additional cloudevents attributes to the broker ingress span.

Fixes #1329

### DIFF
--- a/pkg/broker/ingress/handler.go
+++ b/pkg/broker/ingress/handler.go
@@ -27,6 +27,7 @@ import (
 	cev2 "github.com/cloudevents/sdk-go/v2"
 	"github.com/cloudevents/sdk-go/v2/binding"
 	"github.com/cloudevents/sdk-go/v2/binding/transformer"
+	ceclient "github.com/cloudevents/sdk-go/v2/client"
 	"github.com/cloudevents/sdk-go/v2/protocol"
 	"github.com/cloudevents/sdk-go/v2/protocol/http"
 	"github.com/google/knative-gcp/pkg/metrics"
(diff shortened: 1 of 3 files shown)


pkg/broker/ingress/handler.go
```diff
@@ -145,10 +146,13 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
 	defer span.End()
 	if span.IsRecordingEvents() {
 		span.AddAttributes(
-			kntracing.MessagingSystemAttribute,
-			tracing.PubSubProtocolAttribute,
-			kntracing.BrokerMessagingDestinationAttribute(broker),
-			kntracing.MessagingMessageIDAttribute(event.ID()),
+			append(
```
Harwayne, line 149: Is it worth it to just call `span.AddAttributes` twice, rather than creating a temporary slice?

ian-mi: Two separate calls would result two smaller slices being allocated instead of one larger slice. I also suspect in this case that go will be able to allocate the temporary slice on the stack. Either way, I'm not sure which is more performant and would be surprised if the difference between the two ends up being very significant. Additionally each call to AddAttributes must take a mutex so I slightly prefer keeping just a single call.

PR #1340, opened 2020-06-24
Update sources event CE types, sources, subjects

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1209

## Proposed Changes

- Used new values for CE type/source/subject
- Scheduler event data schema is updated
- Unbundled event schemas from API versions

**Release Note**

```release-note
BREAKING CHANGES

- All event types are updated to match event types in googleapis/google-cloudevents
- CAL/Storage/Scheduler events now also have updated CE "source" and "subject"
- Scheduler events now also have a new data schema

ACTION REQUIRED

- Users who only depend on "source", "type" and "subject" values in the trigger filters can create new triggers with new filter values (see details below).
- Users who depend on "source", "type" and "subject" values in their receiver code will have to update their code.
- Users who depend on scheduler event data will have to update their code.

DETAILED CHANGES

1. All event types and data schemas were updated to be compliant with https://github.com/googleapis/google-cloudevents/tree/master/proto/google/events/cloud. There is no real change to the CAL/PubSub/Storage event schema. Scheduler event schema was updated to be https://github.com/googleapis/google-cloudevents/blob/master/proto/google/events/cloud/scheduler/v1/data.proto
2. Source changes:
  CAL: was "//[service_name]/projects/[project_id]" => now "//cloudaudit.googleapis.com/projects/[project_id]/logs/[activity/data_access]"
  Scheduler: was "//cloudscheduler.googleapis.com/projects/[project_id]/locations/[location]/schedulers/[k8s_scheduler_source_name]" => now "//cloudscheduler.googleapis.com/projects/[project_id]/locations/[location]/jobs/[job_name]"
  Storage: was "//storage.googleapis.com/buckets/[bucket_name]" => now "//storage.googleapis.com/projects/_/buckets/[bucket_name]"
3. Subject changes:
  CAL: was "[resource_name]" => now "[service_name]/[resource_name]"
  Scheduler: was "jobs/[job_name]" => now NONE
  Storage: was "[object_id]" => now "objects/[object_id]"
4. Type changes:
  CAL: was "com.google.cloud.auditlog.event" => now "google.cloud.audit.log.v1.written"
  PubSub: was "com.google.cloud.pubsub.topic.publish" => now "google.cloud.pubsub.topic.v1.messagePublished"
  Scheduler: was "com.google.cloud.scheduler.job.execute" => now "google.cloud.scheduler.job.v1.executed"
  Storage:
[...]

PR #1352, opened 2020-06-25
Limit the values of event_type in Broker metrics

zargarpur:
With custom events, these values can be any string which impacts the overall cardinality of the metrics.

We're instead going to limit these values to be any GCP event type, as defined by the "google.cloud" prefix. And include some values used in some E2E tests. All other event types will default to being labeled as "custom"

Trigger metrics will continue to label filter type as "any" if there is no filter type defined.

```release-note
Broker metrics won't include specific custom event types in Stackdriver other than GCP defined ones, and will instead default to "custom"
```

### DIFF
--- a/pkg/broker/ingress/handler_test.go
+++ b/pkg/broker/ingress/handler_test.go
@@ -59,7 +59,7 @@ const (
 
 	traceID = "4bf92f3577b34da6a3ce929d0e0e4736"
 
-	eventType = "test-event-type"
+	eventType = "com.google.cloud.pubsub.topic.publish"
 	pod       = "testpod"
 	container = "testcontainer"
 )
--- a/pkg/broker/ingress/handler_test.go
+++ b/pkg/broker/ingress/handler_test.go
@@ -59,7 +59,7 @@ const (
 
 	traceID = "4bf92f3577b34da6a3ce929d0e0e4736"
 
-	eventType = "com.google.cloud.pubsub.topic.publish"
+	eventType = "google.cloud.scheduler.job.v1.executed"
 	pod       = "testpod"
 	container = "testcontainer"
 )
--- a/pkg/metrics/delivery_reporter.go
+++ b/pkg/metrics/delivery_reporter.go
@@ -155,11 +155,11 @@ func (r *DeliveryReporter) reportEventProcessingTime(ctx context.Context, end ti
 }
 
 func filterTypeValue(v string) string {
-	if v != "" {
-		return v
+	if v == "" {
+		// the default value if the filter attributes are empty.
+		return "any"
 	}
-	// the default value if the filter attributes are empty.
-	return "any"
+	return EventTypeMetricValue(v)
 }
 
 func (r *DeliveryReporter) AddTags(ctx context.Context) (context.Context, error) {
--- a/pkg/metrics/delivery_reporter.go
+++ b/pkg/metrics/delivery_reporter.go
@@ -157,7 +157,7 @@ func (r *DeliveryReporter) reportEventProcessingTime(ctx context.Context, end ti
 func filterTypeValue(v string) string {
 	if v == "" {
 		// the default value if the filter attributes are empty.
-		return "any"
+		return unspecifiedFilterType
 	}
 	return EventTypeMetricValue(v)
 }
--- a/pkg/metrics/delivery_reporter.go
+++ b/pkg/metrics/delivery_reporter.go
@@ -155,11 +155,11 @@ func (r *DeliveryReporter) reportEventProcessingTime(ctx context.Context, end ti
 }
 
 func filterTypeValue(v string) string {
-	if v == "" {
-		// the default value if the filter attributes are empty.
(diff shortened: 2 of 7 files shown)


pkg/broker/ingress/handler_test.go
```diff
@@ -59,7 +59,7 @@ const (
 
 	traceID = "4bf92f3577b34da6a3ce929d0e0e4736"
 
-	eventType = "test-event-type"
+	eventType = "com.google.cloud.pubsub.topic.publish"
```
nachocano, line 62: all these values will change with @yolocs PR https://github.com/google/knative-gcp/pull/1340

pkg/metrics/tags_test.go
```diff
@@ -0,0 +1,52 @@
...
+		want  string
+	}{
+		// Current Supported GCP event types
+		{"com.google.cloud.auditlog.event", "com.google.cloud.auditlog.event"},
+		{"com.google.cloud.pubsub.topic.publish", "com.google.cloud.pubsub.topic.publish"},
+		{"com.google.cloud.scheduler.job.execute", "com.google.cloud.scheduler.job.execute"},
+		{"com.google.cloud.storage.object.archive", "com.google.cloud.storage.object.archive"},
+		{"com.google.cloud.storage.object.delete", "com.google.cloud.storage.object.delete"},
+		{"com.google.cloud.storage.object.finalize", "com.google.cloud.storage.object.finalize"},
+		{"com.google.cloud.storage.object.metadataUpdate", "com.google.cloud.storage.object.metadataUpdate"},
```
nachocano, line 35: all these are changing

pkg/metrics/delivery_reporter.go
```diff
@@ -155,11 +155,11 @@ func (r *DeliveryReporter) reportEventProcessingTime(ctx context.Context, end ti
 }
 
 func filterTypeValue(v string) string {
-	if v != "" {
-		return v
+	if v == "" {
+		// the default value if the filter attributes are empty.
+		return "any"
```
nachocano, line 160: might be worth using a constant. I think we had one somewhere. 
Same for the "custom" one

ian-mi:
Is there anything preventing a user from creating custom event types with a com.google.cloud prefix?

pkg/metrics/tags.go
```diff
@@ -43,3 +46,27 @@ var (
 	PodNameKey       = tag.MustNewKey(metricskey.PodName)
 	ContainerNameKey = tag.MustNewKey(metricskey.ContainerName)
 )
+
+var (
+	allowedEventTypes = []string{
```
Harwayne, line 51: How about a map[string]struct{} instead? Then you just need to lookup membership in the map, rather than iterating.

pkg/broker/ingress/handler_test.go
zargarpur: Switched it to those newer ones, and now only accept event types matching "google.cloud" instead of "com.google.cloud"

pkg/metrics/tags_test.go
zargarpur: Done

pkg/metrics/delivery_reporter.go
zargarpur: Sure, replaced with constants

pkg/metrics/tags.go
zargarpur: Yeah, it seemed weird to me that there was no slice contains() function, this seems better

nachocano: k8s has a set in "k8s.io/apimachinery/pkg/util/sets"
no need to, just fyi

nachocano:
/approve
/lgtm
/hold
in case others have any other comment. Unhold otherwise

pkg/metrics/delivery_reporter.go
```diff
@@ -155,11 +155,11 @@ func (r *DeliveryReporter) reportEventProcessingTime(ctx context.Context, end ti
...
 func filterTypeValue(v string) string {
-	if v != "" {
-		return v
+	if v == "" {
+		// the default value if the filter attributes are empty.
+		return unspecifiedFilterType
 	}
-	// the default value if the filter attributes are empty.
-	return "any"
+	return EventTypeMetricValue(v)
```
ian-mi, line 162: It shouldn't be necessary to transform the filter type since the cardinality is already limited by the cardinality of triggers.

zargarpur: Across all GCP projects? Doesn't the same cardinality issue occur? Even if a single project has a trigger with one custom event type

ian-mi:
/lgtm
thanks!

pkg/metrics/tags.go
Harwayne: I prefer sets.String{} too, but some don't like adding the extra dependency.

pkg/metrics/delivery_reporter.go
zargarpur: Removed it

PR #1358, opened 2020-06-27
[WIP] test

grac3gao:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/e2e-wi-tests.sh
+++ b/test/e2e-wi-tests.sh
@@ -26,6 +26,7 @@ readonly PROW_SERVICE_ACCOUNT_EMAIL=$(gcloud config get-value core/account)
 # Constants used for creating ServiceAccount for Data Plane(Pub/Sub Admin) if it's not running on Prow.
 readonly PUBSUB_SERVICE_ACCOUNT_NON_PROW_KEY_TEMP="$(mktemp)"
 readonly CONFIG_GCP_AUTH="test/test_configs/config-gcp-auth-wi.yaml"
+readonly CONFIG_GCP_BROKER="test/test_configs/gcp-broker.yaml"
 readonly K8S_SERVICE_ACCOUNT_NAME="ksa-name"
 
 function export_variable() {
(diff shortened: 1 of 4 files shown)


test/e2e-wi-tests.sh
```diff
@@ -137,6 +138,8 @@ function gcp_broker_setup() {
   fi
   kubectl annotate --overwrite serviceaccount ${BROKER_SERVICE_ACCOUNT} iam.gke.io/gcp-service-account="${PUBSUB_SERVICE_ACCOUNT_EMAIL}" \
     --namespace "${CONTROL_PLANE_NAMESPACE}"
+  # Add a default broker to trigger brokercell ingress/fanout/retry components.
+  ko apply -f ${CONFIG_GCP_BROKER}
```
yolocs, line 142: It seems the intention is to use this broker to "warm up" the data plane? Maybe add a note to explain why you need this. And maybe rename the yaml to `warmup-broker.yaml`.

test/e2e/test_gcp_broker.go
```diff
@@ -177,5 +177,8 @@ func createGCPBroker(client *lib.Client) (url.URL, string) {
 	if err != nil {
 		client.T.Error(err.Error())
 	}
+	// Avoid propagation delay between the controller reconciles the broker config and
+	// the config being pushed to the configmap volume in the ingress pod.
+	time.Sleep(60 * time.Second)
```
yolocs, line 182: We really take that long?

test/e2e-wi-tests.sh
grac3gao: Done. I open a new PR which includes this change  https://github.com/google/knative-gcp/pull/1415, it is a little bit messy in this PR, so I will close it.

test/e2e/test_gcp_broker.go
grac3gao: Just think if the forcing update configmap function failed to update, it may take 1 min to sync (for the worst case). But it is true that it is too long. I changed to 15 second in  #1415. 

Also, I add a new sender pod which have retry for 404 and 503 error in #1415 (after sync with Ian)

grac3gao:
opened #1415 for the same purpose of this PR

PR #1360, opened 2020-06-29
[master] Auto-update dependencies

mattmoor:
Produced via:
  `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
/assign grantr ian-mi nachocano
/cc grantr ian-mi nachocano

PR #1366, opened 2020-06-30
Update deps

grantr:
Replaces #1360, which has a build failure due to a signature change in https://github.com/knative/eventing/pull/3430. That build failure is fixed in the second commit here.


nachocano:
/lgtm

thanks @grantr !!!

PR #1379, opened 2020-07-02
Add hpa knobs for broker-gcp

nlopezgi:
Allow for performance tests to modify (via kubectl) the brokercell (with name default) so that autoscaling params can be fine tuned for performance testing 

## Proposed Changes

- 🎁 Add new feature:
  - Adds knobs to brokercell spec to control component hpa (+defaults and default test modified accordingly)
  - modifies reconciler to read hpa attrs from brokercell spec
  - Enable control of autoscaling for broker-gcp via the brokercell.

- Note: Knobs should not be controlled by end user, thus no documentation on these is provided.
- Note: Not sure if we should add more tests to the brokercell reconciler. Please advise. 

### Context
See https://docs.google.com/document/d/1GWtAEegoz5F3ex0mok9liboGyeYFr0H54YsAa4r87BE

fyi @grantr


pkg/apis/intevents/v1alpha1/brokercell_types.go
```diff
@@ -68,8 +68,29 @@ var (
 	_ duckv1.KRShaped = (*BrokerCell)(nil)
 )
 
+// ComponentParameters specifies scaling and resource parameters to be used
```
liu-cong, line 71: The struct only has scaling parameters currently. Is the plan to add resource parameters as well?

pkg/apis/intevents/v1alpha1/brokercell_types.go
```diff
@@ -68,8 +68,29 @@ var (
...
+	MinReplicas *int32 `json:"minReplicas,omitempty"`
+
+	// MaxReplicas specifies the maximum replica count for the component.
+	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+}
+
+// ComponentsParametersSpec specifies separate parameters for each component
+// of a BrokerCell.
+type ComponentsParametersSpec struct {
+	Fanout  ComponentParameters `json:"fanout,omitempty"`
```
liu-cong, line 84: I think having independent structs for each components is probably more future proof, e.g., 
```
type FanoutParameters struct{
  ComponentParameters
  OtherParameters...
}
```

This way we can add component specific parameters such as the concurrency configuration for the fanout pool.

pkg/apis/intevents/v1alpha1/brokercell_types.go
nlopezgi: My plan is to evaluate if with the scaling parameters is enough to get reliable performance benchmark results. If it is then there should be no need, afaik, to add more parameters here.

pkg/apis/intevents/v1alpha1/brokercell_types.go
nlopezgi: Not sure we want to add that level of configurability at this point. The purpose of this PR is to get the minimum parameters needed to control for the most basic aspect of how these tests consume resources. While it might be worthwhile to have independent structs for each component, this is something that, imo, we should only do in the future if we indeed need more specific parameters. Since this is all an internal api, afaiu, then it should be fine to iteratively change it instead of trying to design for features we are not sure we need. Let me know if you think otherwise, if so, we'll likely have to start a design doc to outline use case (other than the perf benchmark improvements) that this change should cover.

liu-cong: SGTM. As you said this is an internal API we should be able to change it the way we want later on.

nlopezgi:
> New changes are detected. LGTM label has been removed.

Trying to rebase to see if that fixes the WI tests as this PR should not be interfering with those, afaiu

liu-cong:
/lgtm

> Trying to rebase to see if that fixes the WI tests as this PR should not be interfering with those, afaiu

BTW, it's recommended to merge instead of rebase as this is easier for the reviewer to find the actual changes.

nlopezgi:
> /lgtm
> 
> > Trying to rebase to see if that fixes the WI tests as this PR should not be interfering with those, afaiu
> 
> BTW, it's recommended to merge instead of rebase as this is easier for the reviewer to find the actual changes.

apologies, will avoid doing rebases.

liu-cong:
> > /lgtm
> > > Trying to rebase to see if that fixes the WI tests as this PR should not be interfering with those, afaiu
> > 
> > 
> > BTW, it's recommended to merge instead of rebase as this is easier for the reviewer to find the actual changes.
> 
> apologies, will avoid doing rebases.

No worries. I did this a few times too :)

Issue #1137, opened 2020-05-27
Upgrade Job to delete {pullsubscription,topic}.pubsub.cloud.google.com COs


Harwayne:
**Problem**
#905 switched from the old API group to the new one. The old API group will be deleted by 0.16 (see #1035). In order to not leak resources in GCP, the custom objects in the old API group need to be deleted while the controller is still reconciling them (e.g. before 0.16).

**Exit Criteria**
An upgrade Job that deletes all {pullsubscription,topic}.pubsub.cloud.google.com custom objects in the cluster.

Harwayne:
The commands to do this manually are:

```shell
kubectl delete pullsubscriptions.pubsub.cloud.google.com --all-namespaces --all
kubectl delete topics.pubsub.cloud.google.com --all-namespaces --all
```

PR #1383, opened 2020-07-06
Upgrade job (for v0.16.0) that deletes legacy {pullsubscription,topic}.pubsub.cloud.google.com COs

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1137

## Proposed Changes

- Added upgrade job for v0.16.0 release that deletes legacy COs

**Release Note**

```release-note
Added upgrade job/yaml that deletes legacy {pullsubscription,topic}.pubsub.cloud.google.com COs

##Action Required
**Before updating** to 0.16, run the upgrade Job. Only after that Job has succeeded, then update to 0.16. Failure to run the update Job first, may leak the resources associated with the Kubernetes resources that were to be deleted, namely `Topic`s and `PullSubscriptoin`s in GCP.
```

**Docs**


ian-mi:
Should this job be run using a more limited service account?
I think upgrade job is a misnomer since the job doesn't upgrade resources to the new API, it just deleted them. I might call this a cleanup instead. Should the job check that the equivalent resources exist in the new API as well?

yolocs:
@ian-mi If isolated from its context (if any), I would agree "cleanup" job is a better name.

@Harwayne for thoughts. I assumed this is part of a bigger upgrading plan. Also from the issue itself it seems I don't need to check the new API.

Harwayne:
/hold

Holding for @yolocs response to:

> Reading the code it looks correct, but just to be sure, you manually verified it doesn't delete the new Topic and PullSubscriptions, right?

yolocs:
@Harwayne you mean the Topic and PullSubscription in the intevent API group?

Harwayne:
> @Harwayne you mean the Topic and PullSubscription in the intevent API group?

Yes.

yolocs:
@Harwayne I made some changes to fix a couple of bugs and used a dedicated SA for upgrade (as suggested by Ian). Also manually tested and confirmed new COs won't be deleted.

nachocano:
Unfortunately I'll have to force the push...

Issue #1217, opened 2020-06-04
Remove old GCP resources, receive adapters, and publishers


nachocano:
**Describe the bug**
In https://github.com/google/knative-gcp/pull/1207 we renamed all GCP resources created by sources, as well as the receive adapters. We should clean up those old resources in code so that users can have a smooth upgrade experience. 
Also https://github.com/google/knative-gcp/pull/1226 avoids creating unnecessary publishers for sources. We should delete the already created ones.
We agreed on doing this in a follow up as it will get nasty and will take sometime. 
This is a 0.16 blocker.

 **Expected behavior**
No GCP resources are leaked when upgrading to 0.16.

nachocano:
fyi @Harwayne

PR #1398, opened 2020-07-07
Mark the subs for update if the Channel's topic has changed.

Harwayne:
Helps with #1217, #1399

## Proposed Changes

- Check to see if the PullSubscription's Topic should change (due to a naming change in 0.16), if so, mark it for update.
- Channels watch for changes to owned resources, regardless of the version of the owning Channel.

**Release Note**

```release-note
NONE
```


nachocano:
Unfortunately will have to force the push...

Issue #1394, opened 2020-07-07
Flakiness: E2E test of GCPBroker


capri-xiyue:
**Describe the bug**
E2E test related to GCP Broker under WI mode is flaky.
Example:
https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/1388/pull-google-knative-gcp-wi-tests/1280550413002608640
Error msg is as follows:
`TestGCPBroker: test_gcp_broker.go:59: job: error sending cloudevent: 503 Service Unavailable
TestGCPBroker: test_gcp_broker.go:59: dummy event wasn't sent to broker`

 **Expected behavior**
E2E test of TestGCPBroker consistently pass.

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative-GCP release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

grantr:
@grac3gao Do you think this issue will be fixed by your investigation of #1362?

grac3gao:
@grantr That may be the different issues. #1362's error is produced in the fanout Pod, this issue's error is produced in the ingress Pod (we handled unknown encoding in ingress Pod). But this issue may be fixed by PR #1358

grac3gao:
fixed by #1415 and #1444

Issue #1346, opened 2020-06-25
Flaky: Flaky TestGCPBrokerMetrics in WI e2e test


grac3gao:
**Describe the bug**
Checked the nightly test, there are roughly 50% chance for TestGCPBrokerMetrics failed in wi e2e test. https://testgrid.knative.dev/knative-gcp#nightly
When submitting PRs in Knative-GCP, TestGCPBrokerMetrics in wi e2e test is also easier than other test to fail.

There are following failure mode:
1.  https://storage.googleapis.com/knative-prow/logs/ci-google-knative-gcp-nightly-release/1273912172598005760/build-log.txt
```
    TestGCPBrokerMetrics: test_gcp_broker.go:68: unexpected metric "knative.dev/eventing/trigger/event_dispatch_latencies" count (-want, +got) =   map[string]int64{
        - 	"resp-broker-gcp-wxigtfbg":    1,
        + 	"resp-broker-gcp-wxigtfbg":    0,
        - 	"trigger-broker-gcp-wxigtfbg": 1,
        + 	"trigger-broker-gcp-wxigtfbg": 0,
    TestGCPBrokerMetrics: metrics.go:75: Pod "test-g-c-p-broker-metrics-4chlq/sender-mvpveqdf-pxk2m" does not have a metrics port
    TestGCPBrokerMetrics: metrics.go:75: Pod "test-g-c-p-broker-metrics-4chlq/target-oviakkzl-gjlgr" does not have a metrics port
```
2. https://storage.googleapis.com/knative-prow/pr-logs/pull/google_knative-gcp/1345/pull-google-knative-gcp-wi-tests/1275975518927196160/build-log.txt

```
    TestGCPBrokerMetrics: test_gcp_broker.go:68: unexpected broker metric count (-want, +got) =   map[string]int64{
        - 	"e2e-dummy-event-type":              1,
        + 	"e2e-dummy-event-type":              0,
        - 	"e2e-testing-resp-event-type-dummy": 1,
        + 	"e2e-testing-resp-event-type-dummy": 0,
          }
        
    TestGCPBrokerMetrics: test_gcp_broker.go:68: unexpected metric "knative.dev/eventing/trigger/event_count" count (-want, +got) =   map[string]int64{
        - 	"resp-broker-gcp-yhiwwney":    1,
        + 	"resp-broker-gcp-yhiwwney":    0,
        - 	"trigger-broker-gcp-yhiwwney": 1,
[...]

Issue #1058, opened 2020-05-13
Controller probes broker data plane for pubsub readiness


liu-cong:
**Problem**
The data plane may not be ready for a Broker even if the controller says ready.
* The `targets-config` volume may not be updated for whatever reason
* Pub/Sub may be temporarily unavailable
* There could be an auth issue preventing access to Pub/Sub

We need probes (not the K8s probes) from the Controller to the data plane components to more accurately tell if a Broker is ready or not.
Checking the configmap volume consistency is tricky when there are multiple pods.  This issue focuses on checking pubsub readiness instead. 
If there is an issue with pubsub(due to auth or simply pubsub outage), it's likely to impact all pods. We can have a readiness probe that checks if any data plane pods fails to communicate with pubsub.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
* Controller probes data plane readiness
* Write an e2e test that simulates data plane error and verify Broker is not ready

**Additional context (optional)**
* The data plane components (ingress/fanout/retry) should provide a readiness endpoint per Broker
* As part of the readiness check, the data plane can
  ~~* Check the config for the requested Broker exists~~
  * Do a read operation on PubSub (e.g., read Topic) to verify communication with PubSub

cc @capri-xiyue

grantr:
Related to #743 

I think this is fairly complex since the data plane may be serviced by several pods in multiple states of readiness. How many pods must be (un)ready for the Broker to be (un)ready?

liu-cong:
We could reduce the scope of the readiness check to only checking the auth setup. If there is an issue with auth, it's likely a global failure for all data plane pods. Therefore checking the auth of a single pod might be sufficient.

liu-cong:
We can use [TestIamPermissions](https://cloud.google.com/pubsub/docs/reference/rpc/google.iam.v1#google.iam.v1.TestIamPermissionsRequest) to check if the data plane has permissions to pubsub resources

grac3gao:
Just add an additional topic related on this:

Broker in wi mode may have 5XX problem. It fail to send event. In ingress pod, it failed to send event to the topic (error showed lack of credentials, and asked the user to do the binding, but we already did the binding). I believe the cause is https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#limitations 
[...]

PR #1415, opened 2020-07-08
Reduce flakiness caused by sending events error

grac3gao:
Related to #1346 #1394 

## Proposed Changes

When sending Event to GCP Broker, sometime the Event returns error for 4XX or 5XX. Some causes are:
- For 4XX error, config is not found, need to increase the waiting time after broker creation in order to give enough time for broker configmap sync.
- For 5XX error, one problem in WI is credential is not syncing up, so add a warm up broker.
- In the client side, add retry for 404 and 503 error.

**Release Note**

```release-note

```

**Docs**


test/test_images/sender_gcpbroker/main.go
```diff
@@ -0,0 +1,112 @@
...
+var defaultRetry = wait.Backoff{
+	Steps:    3,
+	Duration: 30 * time.Second,
+	Factor:   1.0,
+	// The sleep at each iteration is the duration plus an additional
+	// amount chosen uniformly at random from the interval between 0 and jitter*duration.
+	Jitter: 2.0,
+}
+
+func main() {
```
yolocs, line 54: Is it possible to add this retry functionality to the existing sender? I think it would allow easier reuse.

grac3gao: Make senses. I'll change it. Maybe pass a new env in the pod, indicating whether you want to use the try mode for the sender pod or not?

yolocs: Yes. Thanks!

ian-mi: I'm not sure that we have any tests where we would currently want to disable retries.

Harwayne:
> For 4XX error, config is not found, need to increase the waiting time after broker creation in order to give enough time for broker configmap sync.

Do we have any expected or guaranteed time bounds on how long this sync will take? E.g. the data plane syncs every 30 seconds, so expect this to work >99% of the time if you wait at least 30 seconds after the Broker becomes ready.

> For 5XX error, one problem in WI is credential is not syncing up, so add a warm up broker.

Is this visible in the status of the Broker itself?

capri-xiyue:
Do we have any follow up issues regarding the long term solution that WI is credential is not syncing up in time?

test/test_images/sender_gcpbroker/main.go
grac3gao: Maybe for metrics test, we don't want to enable retries? otherwise the metric count might be different from what we expected.

ian-mi: No, it should be fine. If a failure results in additional metrics it will fail the test either way.

test/test_images/sender_gcpbroker/main.go
```diff
@@ -0,0 +1,112 @@
...
+
+	if err := writeTerminationMessage(map[string]interface{}{
+		"success": success,
+		"traceid": span.SpanContext().TraceID.String(),
+	}); err != nil {
+		fmt.Printf("failed to write termination message, %s.\n", err)
+	}
+}
+
+func dummyCloudEvent() cloudevents.Event {
```
capri-xiyue, line 91: duplicate methods with existing sender. Can you refactor it or combine these two senders into one sender

test/e2e/lib/resources/kube.go
```diff
@@ -49,6 +49,10 @@ func SenderJob(name string, envVars []corev1.EnvVar) *batchv1.Job {
 	return baseJob(name, "sender", envVars)
 }
 
+func BrokerSenderJob(name string, envVars []corev1.EnvVar) *batchv1.Job {
+	return baseJob(name, "sender_gcpbroker", envVars)
```
capri-xiyue, line 53: Can we rename it to make it less confusing? Instead of "sender-gcpbroker", maybe name it to "retryablesender"

test/test_configs/warmup-broker.yaml
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+apiVersion: eventing.knative.dev/v1beta1
+kind: Broker
+metadata:
+  name: default
```
capri-xiyue, line 18: rename it to make it less confyusing like "warmup"

test/test_images/sender_gcpbroker/pod.yaml
```diff
@@ -0,0 +1,23 @@
...
+# limitations under the License.
+
+apiVersion: v1
+kind: Pod
+metadata:
+  name: sender
+spec:
+  containers:
+    - name: sender
+      image: ko://github.com/google/knative-gcp/test/test_images/sender_gcpbroker
```
capri-xiyue, line 22: similar as above. Naming is confusing.

test/e2e/lib/helpers/broker_event_transformation_test_with_source.go
```diff
@@ -87,8 +87,8 @@ func BrokerEventTransformationTestHelper(client *lib.Client, brokerURL url.URL,
 	// Just to make sure all resources are ready.
 	time.Sleep(5 * time.Second)
 
-	// Create a sender Job to sender the event.
-	senderJob := resources.SenderJob(senderName, []v1.EnvVar{{
+	// Create a sender Job to send the event with retry.
+	senderJob := resources.BrokerSenderJob(senderName, []v1.EnvVar{{
```
capri-xiyue, line 91: Add some comments that why gcp broker needs retryable sender.

test/e2e/lib/helpers/broker_event_transformation_test_with_source.go
```diff
@@ -138,8 +138,8 @@ func BrokerEventTransformationMetricsTestHelper(client *lib.Client, projectID st
...
```
capri-xiyue, line 142: Rename. Maybe "RetryableSenderJob"

test/e2e/lib/helpers/broker_event_transformation_test_with_source.go
```diff
@@ -214,8 +214,8 @@ func BrokerEventTransformationTracingTestHelper(client *lib.Client, projectID st
...
```
capri-xiyue, line 218: similar as above. Naming

grac3gao:
> Do we have any expected or guaranteed time bounds on how long this sync will take? E.g. the data plane syncs every 30 seconds, so expect this to work >99% of the time if you wait at least 30 seconds after the Broker becomes ready.

Synced with @liu-cong, In broker, when a target is added, we annotate the data plane component to trigger an immediately refresh for the configmap volume. So ideally, there should be no time (or just a couple of seconds) for this propagation sync. If that annotation failed, k8s refreshes the volume every minute. 10 seconds should be fine to cover 99% cases.

> Is this visible in the status of the Broker itself? If so, can we have the test utilize that instead? If not, can we make it appear in the status of the Broker?

It isn't, and it is hard to detect in the broker data plane side (might need to keep trying authenticating calls to verify if it is ready). I believe this is due to the limitation https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity#limitations:
> The GKE Metadata Server takes a few seconds to start to run on a newly created pod. Therefore, attempts to authenticate or authorize using Workload Identity made within the first few seconds of a pod's life may fail. Retrying the call will resolve the problem.

test/test_images/sender_gcpbroker/main.go
grac3gao: This sender part will be changed after Chen's comments

grac3gao:
> Do we have any follow up issues regarding the long term solution that WI is credential is not syncing up in time?

We have https://github.com/google/knative-gcp/issues/1058, the solution for that issue could help to avoid this.

test/test_images/sender_gcpbroker/main.go
grac3gao: Checked the usage, sender Pod is only used in three tests (I thought it will be used in more places...) As all three tests can use retry mode, I directly add the them without env.

capri-xiyue: The sender will be used for both gcp broker and pubsub channel based broker, right? Do we need the retry mode for the pubsub channel based broker?

grac3gao:
```
func SenderJob(name string, envVars []corev1.EnvVar) *batchv1.Job {
	return baseJob(name, "sender", envVars)
}
```
is only used in three broker test. Let me double check for it.

capri-xiyue: Yep. That's used in broker but that lib method is for both gcp broker and pubsub channel based broker(just currently pubsub channel based broker e2e tests get skipped because of issues in eventing)

grac3gao: Ah, we also use it in BrokerWithPubSubChannel test (so, three gcp broker tests + one pubsub channel test). I think it would be fine to also retry in pubsub channel (as we indicate only retry for specific error, so it most likely won't qualify for retry).

capri-xiyue: Retry is just a work-around for gcp broker e2e tests. Usually broker with pubsub channel should not use it and if it start to use it, it means something goes wrong and devs should investigate what happens. I'd prefer not enabling retryable sender for broker with pubsub channel.

grac3gao: Sounds good. I am find to change it.

test/test_images/sender/main.go
```diff
@@ -23,19 +23,43 @@ import (
...
+}
+
 func main() {
 	brokerURL := os.Getenv(brokerURLEnvVar)
+	retryVar := os.Getenv(retryEnvVar)
+
+	needRetry := false
+	if retryVar == "true" {
+		needRetry = true
+	}
```
yolocs, line 62: nit: could be `needRetry := (retryVar == "true")`

test/test_images/sender/main.go
```diff
@@ -45,11 +69,13 @@ func main() {
 	ctx, span := trace.StartSpan(context.Background(), "sender", trace.WithSampler(trace.AlwaysSample()))
 	defer span.End()
 
-	ctx, _, err = ceClient.Send(ctx, dummyCloudEvent())
-	rtctx := cloudevents.HTTPTransportContextFrom(ctx)
+	// If needRetry is true, repeat sending Event with exponential backoff when there is some specific errors.
+	// 404 error indicates some broker configmap sync up issue and 5XX error indicates credential sync issues.
```
yolocs, line 73: To be precise, 404 could also mean the broker was never created and 5xx could mean our code has bug. So better to document the other way around - like sync problems could cause 404 and 5xx and in e2e tests retrying those would help reduce flakiness.

test/test_images/sender/main.go
```diff
@@ -81,3 +121,8 @@ func writeTerminationMessage(result interface{}) error {
 	}
 	return ioutil.WriteFile("/dev/termination-log", b, 0644)
 }
+
+// isRetryable determines if the err is an error which is retryable
+func isRetryable(err error) bool {
+	return strings.Contains(err.Error(), "404 Not Found") || strings.Contains(err.Error(), "503 Service Unavailable") || strings.Contains(err.Error(), "500 Internal Server Error")
```
ian-mi, line 127: It should be possible to unwrap the error as an http.Result in order to access the HTTP status code.

ian-mi, line 127: I would avoid retrying 500 errors.

test/test_images/sender/main.go
```diff
@@ -23,19 +23,40 @@ import (
...
+	// The sleep at each iteration is the duration plus an additional
+	// amount chosen uniformly at random from the interval between 0 and jitter*duration.
+	Jitter: 2.0,
+}
+
 func main() {
 	brokerURL := os.Getenv(brokerURLEnvVar)
+	retryVar := os.Getenv(retryEnvVar)
+
+	needRetry := (retryVar == "true")
```
ian-mi, line 59: I would use envconfig for env parsing.

test/test_images/sender/main.go
grac3gao: 500 errors was caught in https://pantheon.corp.google.com/storage/browser/_details/knative-prow/pr-logs/pull/google_knative-gcp/1415/pull-google-knative-gcp-wi-tests/1281267387093487616/artifacts/pod-logs/test-g-c-p-broker-tracing-kqp86/sender-hyvivyza-7dl4s-sender-hyvivyza.log

Checking the broker ingress pod:
https://pantheon.corp.google.com/storage/browser/_details/knative-prow/pr-logs/pull/google_knative-gcp/1415/pull-google-knative-gcp-wi-tests/1281267387093487616/artifacts/controller-logs/cloud-run-events-default-brokercell-ingress-765fb65c68-cmxhw-ingress.txt, it showed error: 
```
{"level":"error","ts":"2020-07-09T17:05:30.178Z","logger":"broker-ingress","caller":"ingress/handler.go:168","msg":"Error publishing to PubSub for broker test-g-c-p-broker-tracing-kqp86/gcp-haoptzdy. event: Validation: valid\nContext Attributes,\n  specversion: 1.0\n  type: e2e-dummy-event-type\n  source: e2e-dummy-event-source\n  id: e2e-dummy-event-id\n  time: 2020-07-09T17:05:30.029102539Z\n  datacontenttype: application/json\nExtensions,\n  knativearrivaltime: 2020-07-09T17:05:30.052197779Z\n  traceparent: 00-a7690e3cd459d2676d06af82e3cc8123-e8526b5409da5713-01\nData,\n  \"{\\\"source\\\": \\\"sender!\\\"}\"\n, err: rpc error: code = Unauthenticated desc = transport: compute: Received 403 `\nUnable to generate token; IAM returned 403 Forbidden: The caller does not have permission\n\nThis error could be caused by a missing IAM policy binding on the target IAM service account.\n\nYou can create the necessary policy binding with:\n\n  gcloud iam service-accounts add-iam-policy-binding \\\n    --role=roles/iam.workloadIdentityUser \\\n    --member=\"serviceAccount:knative-boskos-30.svc.id.goog[cloud-run-events/broker]\" \\\n    prow-job@knative-tests.iam.gserviceaccount.com\n\nFor more information, refer to the Workload Identity documentation:\n\n    https://cloud.google.com/kubernetes-engine/docs/how-to/workload-identity\n\n`.","commit":"469cc16","stacktrace":"github.com/google/knative-gcp/pkg/broker/ingress.(*Handler).ServeHTTP\n\tgithub.com/google/knative-gcp/pkg/broker/ingress/handler.go:168\ngo.opencensus.io/plugin/ochttp.(*Handler).ServeHTTP\n\tgo.opencensus.io@v0.22.4/plugin/ochttp/server.go:92\nnet/http.serverHandler.ServeHTTP\n\tnet/http/server.go:2807\nnet/http.(*conn).serve\n\tnet/http/server.go:1895"}
```

As it is related on the flakiness, I include the 500 error here.

Also, in that Ingress Pod log, StorageSource and PubsubSource seem to have the same credential issue, but looks like the source itself will automatically retry, so they finally succeeded.

ian-mi: Thanks for adding a TODO, we should figure out how this case should be handled so that the broker does not return 500.

PR #1418, opened 2020-07-08
Update hack scripts to remove outdated curl commands & key creation

bharattkukreja:
## Proposed Changes

- Remove the outdated curl command which was used to get the service account credentials
- Remove the pubsub_service_account_key_temp which was only used by the curl command

### DIFF
--- a/hack/lib.sh
+++ b/hack/lib.sh
@@ -162,7 +162,10 @@ function storage_admin_set_up() {
     --member=serviceAccount:"${pubsub_service_account}"@"${project_id}".iam.gserviceaccount.com \
     --role roles/storage.admin
 
-  # Get service account credentials from: https://cloud.google.com/storage/docs/json_api/v1/projects/serviceAccount/get/
+  # We assume the service account's name is in the form '<project-number>@gs-project-accounts.iam.gserviceaccount.com',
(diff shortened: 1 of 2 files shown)


hack/lib.sh
```diff
@@ -163,8 +162,7 @@ function storage_admin_set_up() {
     --member=serviceAccount:"${pubsub_service_account}"@"${project_id}".iam.gserviceaccount.com \
     --role roles/storage.admin
 
-  curl -s -X GET -H "Authorization: Bearer \`GOOGLE_APPLICATION_CREDENTIALS=${pubsub_service_account_key_temp} gcloud auth application-default print-access-token\`" \
-    "https://www.googleapis.com/storage/v1/projects/${project_id}/serviceAccount"
+  # Get service account credentials from: https://cloud.google.com/storage/docs/json_api/v1/projects/serviceAccount/get/
```
Harwayne, line 165: I would phrase this something like:

```suggestion
  # We assume the service account's name is in the form '<project-number>@gs-project-accounts.iam.gserviceaccount.com',
  # because it has been for all projects we've encountered. However, nothing requires this
  # format. For the actual email, execute this API request:
  # https://cloud.google.com/storage/docs/json_api/v1/projects/serviceAccount/get/
```

Issue #1362, opened 2020-06-30
Corrupt reply may be dropped


grantr:
**Describe the bug**
Like https://github.com/knative/eventing/issues/3438, GCP broker may be dropping corrupt replies and marking the delivery successful.

I think the relevant lines are https://github.com/google/knative-gcp/blob/ebd932b0fa01e57fbd00ce2fd0070ececc5c70c6/pkg/broker/handler/processors/deliver/processor.go#L156-L159

 **Expected behavior**
If a consumer returns a reply but that reply cannot be parsed or is somehow incomplete,  the delivery is retried.

grantr:
Here's the attempt to fix this in eventing https://github.com/knative/eventing/pull/3450

grac3gao:
This may be the one of the causes for flaky GCP Broker related test:
Sender pod sends the event, however, Target pod doesn't receive that, like this:
```
    TestCloudStorageSourceWithGCPBroker: test_gcp_broker.go:101: timed out waiting for the condition
    TestCloudStorageSourceWithGCPBroker: test_gcp_broker.go:101: resp event didn't hit the target pod
```
The event was dropped during the process: Event sent to a KSVC, and resp Event sent to the target. 

After retuning error message `"unknown encoding` for unknown encoding, I caught up the error like this in brokercell-fanout:
(The Event was sent to the KSVC, and replied back.)
```
{"level":"warn","ts":"2020-07-03T05:55:52.069Z","logger":"broker-fanout","caller":"deliver/processor.go:126","msg":"target delivery failed","commit":"8678614","target":"test-cloud-storage-source-with-g-c-p-broker-c5d2j/gcp-hndtawlx/resp-broker-gcp-hndtawlx","error":"unknown encoding"}
```
Then, it goes to brokercell-retry for several rounds of re-send try with error, and finally got delivered:

```
[...]

PR #1419, opened 2020-07-08
Handle malformed replies in delivery processor

yolocs:
Related to https://github.com/google/knative-gcp/issues/1362
Hopefully this could provide some insight for some of the test flakiness.

## Proposed Changes

- Copied workaround from https://github.com/knative/eventing/pull/3450
-
-

**Release Note**

```release-note
Malformed replies will be treated as error (instead of being silently ignored).
```

**Docs**

### DIFF
--- a/pkg/broker/handler/processors/deliver/processor_test.go
+++ b/pkg/broker/handler/processors/deliver/processor_test.go
@@ -200,37 +200,69 @@ func TestDeliverSuccess(t *testing.T) {
 	}
 }
 
+type targetWithFailureHandler struct {
+	t              *testing.T
+	delay          time.Duration
+	respCode       int
+	malFormedEvent bool
+}
+
+func (h *targetWithFailureHandler) ServeHTTP(w http.ResponseWriter, req *http.Request) {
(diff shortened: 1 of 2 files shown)


pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -295,38 +324,10 @@ func TestDeliverFailure(t *testing.T) {
...
-					return
-				}
-
-				// Due to https://github.com/cloudevents/sdk-go/issues/433
-				// it's not possible to use Receive to easily return error.
-				if err := resp(rctx, nil, &cehttp.Result{StatusCode: http.StatusInternalServerError}); err != nil {
-					t.Errorf("unexpected error from target responding event: %v", err)
-				}
-			}()
-
```
grac3gao, line 324: Just curious, why we delete this part of test?

yolocs: This code was used to respond events. Now it's replaced by my own handler which is needed in order to reply specific response body.

Issue #1417, opened 2020-07-08
Export broker data plane logs as test artifacts


grantr:
**Problem**
Useful for debugging flaky delivery issues

yolocs:
It turned out we do have logs from broker data plane (_usually_).

They are dumped as part of this function: 

https://github.com/google/knative-gcp/blob/f7a6351211526cd3df0e261ad20599b484521d1c/test/e2e-common.sh#L105-L128

Such logs will be located at artifacts/controller-logs with names like `cloud-run-events-default-brokercell-fanout-86bb65f9df-v2j49-fanout.txt`

Although they don't always present for some reason. I suspect it's because sometimes broker data plane pods were deleted before it dumps.

yolocs:
On test failure, we will tear down the test namespace which will also deletes the broker. If all the brokers are deleted and the GC happens to recycle the broker data plane, we will end up not having broker data plane pods to dump logs.

yolocs:
One solution could be adding a warmup broker that exists until the end of the tests (like in https://github.com/google/knative-gcp/pull/1415). This will make sure the broker data plane is not recycled before we dump the logs.

grantr:
:+1: to adding a warmup broker. Ultimately this workaround can be replaced when we can assign Brokers to multiple BrokerCells in a cluster (see #866).

PR #1431, opened 2020-07-10
Always setup warmup broker in e2e tests to preserve broker data plane logs

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1417

## Proposed Changes

-
-
-

**Release Note**

```release-note
NONE
```

**Docs**

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -31,6 +31,8 @@ readonly VENDOR_EVENTING_TEST_IMAGES="vendor/knative.dev/eventing/test/test_imag
 # Constants used for authentication setup for GCP Broker if it's not running on Prow.
 readonly APP_ENGINE_REGION="us-central"
 
+readonly CONFIG_WARMUP_GCP_BROKER="test/test_configs/warmup-broker.yaml"
+
 # Setup Knative GCP.
 function knative_setup() {
   start_knative_gcp || return 1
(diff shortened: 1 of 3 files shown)


test/e2e-common.sh
```diff
@@ -102,6 +104,22 @@ function enable_monitoring(){
...
+# propagated. A warm-up broker will force the data plane to be created before
+# the real testing. This helps prevent the credential propagation delay causing
+# test flakiness.
+#
+# 2. The broker data plane will be GCed if there is no broker. Usually this would
+# happen before we dump all the pod logs in the cloud-run-events namespace. The
+# warm-up broker makes sure there is always one broker left and thus data plane pods
+# won't be deleted before we dump logs.
+function warmup_broker_setup(){
+  ko apply -f ${CONFIG_WARMUP_GCP_BROKER}
```
Harwayne, line 120: Should we clean this up at the end of the tests?

yolocs: We don't reuse clusters. So I think it's fine to leak in-cluster resources.

yolocs: Yes, we do. Let me do that.

yolocs: @Harwayne Just confirmed with Chi. We can't reliably do resource cleanup with scripts in prow. What he did was to always delete all pubsub topics/subs before tests (https://github.com/google/knative-gcp/blob/master/test/e2e-secret-tests.sh#L85-L89).

yolocs:
Will force push this PR as the only test failure is broker metrics and it was _not_ because target not receiving the event.

PR #1439, opened 2020-07-13
Change Github Action to Check PRs Without Commiting Changes

ian-mi:
Also use go run to prevent goimports related dependencies from being added to go.mod


ericlem:
/lgtm

In previous go codebases I've worked in this has worked reasonably well.

PR #1440, opened 2020-07-13
bind quota to the running project when running e2e

capri-xiyue:
Fixes #

## Proposed Changes

- 🐛 binded quota to the running project when running e2e tests

### DIFF
--- a/test/e2e/lib/auditlogs.go
+++ b/test/e2e/lib/auditlogs.go
@@ -97,8 +97,8 @@ func StackdriverSinkExists(t *testing.T, sinkID string) bool {
 	t.Helper()
 	ctx := context.Background()
 	project := os.Getenv(ProwProjectKey)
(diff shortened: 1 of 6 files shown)


test/e2e/lib/auditlogs.go
```diff
@@ -96,7 +97,8 @@ func StackdriverSinkExists(t *testing.T, sinkID string) bool {
 	t.Helper()
 	ctx := context.Background()
 	project := os.Getenv(ProwProjectKey)
-	client, err := logadmin.NewClient(ctx, project)
+	options := []option.ClientOption{option.WithQuotaProject(project)}
```
chizhg, line 100: You shouldn't need an array here, just:
```
option := option.WithQuotaProject(project)
client, err := logadmin.NewClient(ctx, project, option)
```

The same for the following ones.

PR #1442, opened 2020-07-13
Update some UTs

grac3gao:
Fixes #

## Proposed Changes

- Update some UTs
-
-

**Release Note**

```release-note

```

**Docs**


pkg/apis/events/v1alpha1/cloudauditlogssource_validation_test.go
```diff
@@ -57,6 +56,29 @@ var (
 			Project: "my-eventing-project",
 		},
 	}
+
+	auditLogsSourceSpecWithKSA = CloudAuditLogsSourceSpec{
+		ServiceName:  "foo",
```
capri-xiyue, line 61: Can we use constants here for all similar tests to avoid magic string for each sources?

PR #1444, opened 2020-07-14
Update sender Pod image 

grac3gao:
Fixes #

## Proposed Changes

- 🧽  Changed to use cloudevent v2 client, in order to get the correct statusCode. 
-
-

**Release Note**

```release-note

```

**Docs**


yolocs:
I think you're using v1 sdk. If you switch to v2 sdk, would it work? You can quickly test it locally I think.

grac3gao:
@yolocs Thanks! I have changed to v2 sdk, and it works

test/test_images/sender/main.go
```diff
@@ -92,30 +86,26 @@ func main() {
...
-func dummyCloudEvent() cloudevents.Event {
-	event := cloudevents.NewEvent(cloudevents.VersionV1)
+func dummyCloudEvent() cev2.Event {
+	event := cev2.NewEvent(cev2.VersionV1)
 	event.SetID(lib.E2EDummyEventID)
 	event.SetType(lib.E2EDummyEventType)
 	event.SetSource(lib.E2EDummyEventSource)
-	event.SetDataContentType(cloudevents.ApplicationJSON)
-	event.SetData(`{"source": "sender!"}`)
+	event.SetDataContentType(cev2.ApplicationJSON)
```
yolocs, line 107: No need to call this one now.

test/test_images/sender/main.go
```diff
@@ -129,15 +119,18 @@ func writeTerminationMessage(result interface{}) error {
 
 // isRetryable determines if the err is an error which is retryable
 func isRetryable(err error) bool {
-	var httpResult *cehttp.Result
-	if errors.As(err, &httpResult) {
+	var result *cehttp.Result
+	if errors.As(err, &result) {
```
yolocs, line 123: Although it's the same, maybe better to use `protocol.ResultAs`

test/test_images/sender/main.go
```diff
@@ -129,15 +119,18 @@ func writeTerminationMessage(result interface{}) error {
...
 		// - 500 Internal Server Error, it is currently for reducing flakiness caused by Workload Identity credential sync up.
 		// We should remove it after https://github.com/google/knative-gcp/issues/1058 lands, as 500 error may indicate bugs in our code.
 		// - 503 Service Unavailable (with or without Retry-After) (IGNORE Retry-After)
-		sc := httpResult.StatusCode
-		return sc == 404 || sc == 500 || sc == 503
+		sc := result.StatusCode
+		if sc == 404 || sc == 500 || sc == 503 {
+			log.Printf("got error: %s, retry sending event. \n", result.Error())
+			return true
+		}
```
yolocs, line 133: Did you verify this manually?

grac3gao: Yes, by inducing an error returned from processor.

test/test_images/sender/main.go
```diff
@@ -129,15 +121,49 @@ func writeTerminationMessage(result interface{}) error {
...
 	}
 	return false
 }
+
+func newDefaultClient(ctx context.Context, target ...string) (context.Context, cev2.Client, error) {
```
grac3gao, line 140: Would like to eventually use knclient in sender image, however, knclient is using cev1 and changing knclient to cev2 needs changes in other images (after #1335 merged, we are using knclient for most of the images). Created a follow-up issue for that: https://github.com/google/knative-gcp/issues/1467

Issue #1412, opened 2020-07-08
Promote CloudStorageSource to v1


nachocano:
**Problem**
We need to promote CloudStorageSource to v1.
Some tasks this will entail, and can probably be done in different PRs.

- [x] v1 types with UTs
- [x] add new version to CRD
- [x] defaulting and conversion webhook
- [x] Update docs and examples
- [x] Migrate reconciler to v1 with UTs
- [x] Add e2e tests to also support v1

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
Storage is in v1

nachocano:
As part of this PR, we should remove `spec.payloadFormat` from the v1 type.
We will always be sending a JSON payload, so it does not make sense to leave this configurable.
Note that the conversion should set it to be JSON_API_V1... 
The reconciler is not actually honoring whatever we configure there, so we should remove it... 
https://github.com/google/knative-gcp/blob/244ab97b1b2044fd1d3764ac649bd84c06e677d4/pkg/reconciler/events/storage/storage.go#L170

PR #1451, opened 2020-07-14
added conversion for CloudStorageSource v1

capri-xiyue:
Helps with #1412 

## Proposed Changes

- - 🎁 added conversion for CloudStorageSource v1 and UT
- - 🎁 added functions for v1 conversion in conversion helper and UT


capri-xiyue:
@danyinggu  This PR doesn't include change in webhook, you can do that in your follow up PRs for the other conversions.

danyinggu:
> @danyinggu This PR doesn't include change in webhook, you can do that in your follow up PRs for the other conversions.

Thanks for letting me know! will do.

pkg/apis/events/v1alpha1/cloudauditlogssource_conversion_test.go
```diff
@@ -35,85 +33,14 @@ import (
...
 
-	completeSecret = &v1.SecretKeySelector{
-		LocalObjectReference: v1.LocalObjectReference{
-			Name: "name",
-		},
-		Key:      "key",
-		Optional: &trueVal,
-	}
-
 	completePubSubSpec = duckv1alpha1.PubSubSpec{
```
danyinggu, line 40: I found that you defined the completePubSubSpec in v1beta1/test_helper.sh, and all conversion_test in v1beta1 can use that, which is nice. However, in v1alpha1, why we still define the completePubsubSpec in the auditlogsSource conversion_test? Will that be inconsistent with what you did in v1beta1?

pkg/apis/events/test_helper.go
```diff
@@ -0,0 +1,86 @@
...
+		Ref: &duckv1.KReference{
+			APIVersion: "apiVersion",
+			Kind:       "kind",
+			Namespace:  "namespace",
+			Name:       "name",
+		},
+		URI: &CompleteURL,
+	}
+
+	CompleteSourceSpec = duckv1.SourceSpec{
```
danyinggu, line 72: Is that the same as the `completeSourceSpec` in conversion_helper? Should we reuse that? I'm not sure... just purposal.

pkg/apis/events/v1alpha1/cloudauditlogssource_conversion_test.go
capri-xiyue: I have not refactored that part, we can do that in follow up PRs since it's only related to v1alpha1 and it's just a clean up

pkg/apis/events/test_helper.go
capri-xiyue: Yep. It's the same. I will refactor that part.

capri-xiyue:
@danyinggu  You can refer to this PR, [this doc](https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/), and this PR https://github.com/google/knative-gcp/pull/827/files#diff-72eb064d91edf0474e5af0bfe26ecfd7 when you work on the conversion of other three sources, topics and pullsubscriptions.

capri-xiyue:
The check PR failure is not because of this PR. It's because of the bad format in master branch. Submitted this PR to fix it https://github.com/google/knative-gcp/pull/1460

pkg/apis/events/v1alpha1/cloudstoragesource_conversion.go
```diff
@@ -27,7 +26,7 @@ import (
 
 // ConvertTo implements apis.Convertible.
 // Converts source (from v1beta1.CloudStorageSource) into v1alpha1.CloudStorageSource.
-func (source *CloudStorageSource) ConvertTo(_ context.Context, to apis.Convertible) error {
+func (source *CloudStorageSource) ConvertTo(ctx context.Context, to apis.Convertible) error {
```
danyinggu, line 29: Just want to double check with one thing: when we synced offline you mentioned that we need to add conversion between v1 and v1alpha1. Here I found you only added UT for conversion between v1alpha1 and v1, is that enough? Like we only need to add UT for this conversion, but no need to support the logic that converts directly between v1alpha1 and v1. (Because when we synced offline, you just mentioned we need to add conversion between v1 and v1alpha1) just want to double check specifically.

capri-xiyue: I'm kind of confused with your comment. https://github.com/capri-xiyue/knative-gcp/blob/1ed880e121c52579ac3d20be4b20dd079b1d35db/pkg/apis/events/v1alpha1/cloudstoragesource_conversion.go#L43 this line does conversion between v1alpha1 and v1 and I also added UT related to it

capri-xiyue: This is what knative serving does, the conversion v1alpha1 -> v1 is through v1alpha1 -> v1beta1 -> v1 https://github.com/knative/serving/blob/e071715848e6e8677c16c2dab129e6d131a2f196/pkg/apis/serving/v1alpha1/route_conversion.go#L36

danyinggu: My bad. I got it. Thanks for point it out!

pkg/apis/events/v1alpha1/cloudstoragesource_conversion_test.go
```diff
@@ -102,3 +104,50 @@ func TestCloudStorageSourceConversion(t *testing.T) {
 		}
 	}
 }
+
+func TestCloudStorageSourceConversionBetweenV1(t *testing.T) {
```
nachocano, line 108: can this be merged with the BetweenV1beta1 test, and you add both versions to the versions list?

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
```diff
@@ -20,15 +20,46 @@ import (
...
+		sink.Spec.PubSubSpec = convert.FromV1PubSubSpec(source.Spec.PubSubSpec)
+		sink.Spec.ServiceAccountName = source.Spec.ServiceAccountName
+		sink.Spec.Bucket = source.Spec.Bucket
+		sink.Spec.EventTypes = source.Spec.EventTypes
+		sink.Spec.ObjectNamePrefix = source.Spec.ObjectNamePrefix
+		sink.Status.PubSubStatus = convert.FromV1PubSubStatus(source.Status.PubSubStatus)
+		sink.Status.NotificationID = source.Status.NotificationID
+		return nil
+	default:
+		return fmt.Errorf("unknown conversion, got: %T", source)
```
nachocano, line 63: does this needs to convert as well?
As in https://github.com/knative/serving/blob/master/pkg/apis/serving/v1beta1/configuration_conversion.go#L35

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
```diff
@@ -20,15 +20,46 @@ import (
...
+		sink.Spec.PubSubSpec = convert.ToV1PubSubSpec(source.Spec.PubSubSpec)
+		sink.Spec.ServiceAccountName = source.Spec.ServiceAccountName
+		sink.Spec.Bucket = source.Spec.Bucket
+		sink.Spec.EventTypes = source.Spec.EventTypes
+		sink.Spec.ObjectNamePrefix = source.Spec.ObjectNamePrefix
+		sink.Status.PubSubStatus = convert.ToV1PubSubStatus(source.Status.PubSubStatus)
+		sink.Status.NotificationID = source.Status.NotificationID
+		return nil
+	default:
+		return fmt.Errorf("unknown conversion, got: %T", sink)
```
nachocano, line 43: does this needs to convert as well?
As in https://github.com/knative/serving/blob/master/pkg/apis/serving/v1beta1/configuration_conversion.go#L35

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
capri-xiyue: v1 is the highest version. So for v1beta1, it is ok that we return error here because v1beta1 only has one higher version v1. 
Either way works.
Now, it is obvious to user that  that v1 is the highest version.
If we change it to https://github.com/knative/serving/blob/master/pkg/apis/serving/v1beta1/configuration_conversion.go#L35, we can keep the code style consistent.
I'm fine with either.

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
capri-xiyue: Same above, 
v1 is the highest version. So for v1beta1, it is ok that we return error here because v1beta1 only has one higher version v1.
Either way works.
Now, it is obvious to user that that v1 is the highest version.
If we change it to https://github.com/knative/serving/blob/master/pkg/apis/serving/v1beta1/configuration_conversion.go#L35, we can keep the code style consistent.
I'm fine with either.

pkg/apis/events/v1alpha1/cloudstoragesource_conversion_test.go
capri-xiyue: If we want to combine them, I need to add a `if` to differentiate between v1 and v1beta1 since v1beat1 keeps status.ServiceAccountName and sepc.PayloadFormat. I think that may be hard to read. That's why I separate them in two methods.

capri-xiyue:
> I'm not very familiar with conversions. Starting with a SinkBinding conversion PR myself.
> Added some comments, might be worth having @Harwayne to review as well

Sure. Added @Harwayne in this PR.

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
capri-xiyue: I also modified the code in v1beta1 to keep the code style consistent

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
capri-xiyue: I also modified the code in v1beta1 to keep the code style consistent

pkg/apis/events/test_helper.go
```diff
@@ -0,0 +1,86 @@
+package events
```
Harwayne, line 1: This is now in our general, released package. My preference would be to move it somewhere away from our production code. Probably in its own package. You might try just changing this line to `package events_test`, but I'm not sure if the other files that import this will be able to.

pkg/apis/events/v1alpha1/test_helper.go
```diff
@@ -0,0 +1,49 @@
+package v1alpha1
```
Harwayne, line 1: Similar comment as above, this is now in our production code. Can we move it to a different package?

pkg/apis/events/v1beta1/test_helper.go
```diff
@@ -0,0 +1,49 @@
+package v1beta1
```
Harwayne, line 1: Same package comment.

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
```diff
@@ -18,17 +18,47 @@ package v1beta1
...
 	"knative.dev/pkg/apis"
 )
 
 // ConvertTo implements apis.Convertible.
-func (*CloudStorageSource) ConvertTo(_ context.Context, to apis.Convertible) error {
-	return fmt.Errorf("v1beta1 is the highest known version, got: %T", to)
+// Converts source (from v1beta1.CloudStorageSource) into a higher version.
+func (source *CloudStorageSource) ConvertTo(ctx context.Context, to apis.Convertible) error {
+	switch sink := to.(type) {
+	case *v1.CloudStorageSource:
```
Harwayne, line 31: Just to be sure, you've run this in a cluster and seen that you can read and write v1alpha1, v1beta1, and v1 versions of the resource? 

My guess based on the hub version would have been that everything had to be shoved into the v1alpha1 conversion code. But I've never actually tested that guess. Is the `apis.ConvertToViaProxy` call to handle this case?

capri-xiyue: `apis.ConvertToViaProxy` is to do the conversion between v1 and v1alpha1. I added UT to verify that the conversion between v1, v1beta1 and v1alpha1 works. 
I have not run this in a cluster since this PR only include conversion and relevant helpers, it does not include the webhook and yaml change.
When you say user can read and write v1alpha1, v1beta1, and v1 versions of the resource, do you mean user can run `kubecetl get xxx`?

pkg/apis/events/test_helper.go
danyinggu: I also think that might be moved to a different package, as pullSubscription conversion test also needs these constants.

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
capri-xiyue: @danyinggu I synced with Adam offline. Currently for this PR it's fine since we have UT, we can verify whether it works with webhook and yaml in your follow up PRs.

pkg/apis/events/v1beta1/test_helper.go
capri-xiyue: done

pkg/apis/events/v1alpha1/test_helper.go
capri-xiyue: done

pkg/apis/events/v1beta1/cloudstoragesource_conversion_test.go
```diff
@@ -32,3 +61,50 @@ func TestCloudStorageSourceConversionBadType(t *testing.T) {
...
+				// DeepCopy because we will edit it below.
+				in := test.in.DeepCopy()
+				if err := in.ConvertTo(context.Background(), ver); err != nil {
+					t.Errorf("ConvertTo() = %v", err)
+				}
+				got := &CloudStorageSource{}
+				if err := got.ConvertFrom(context.Background(), ver); err != nil {
+					t.Errorf("ConvertFrom() = %v", err)
+				}
+				// ServiceAccountName and PayloadFormat only exists in v1alpha1 and v1beta1, they doesn't exist in v1.
```
danyinggu, line 100: nit: they don't.

pkg/apis/events/v1beta1/cloudstoragesource_conversion_test.go
```diff
@@ -32,3 +61,50 @@ func TestCloudStorageSourceConversionBadType(t *testing.T) {
...
+				// So it won't round trip, it will be silently removed.
+				in.Status.ServiceAccountName = ""
+				in.Spec.PayloadFormat = ""
```
danyinggu, line 103: Is there a need to copy the object? Since the test.in will not be used after doing the ConvertTo, I think we can directly modify the fields here.

pkg/apis/events/v1beta1/cloudstoragesource_conversion_test.go
danyinggu: nit: So this is not a round trip, they will be silently removed.

pkg/apis/events/v1beta1/cloudstoragesource_conversion.go
danyinggu: Sounds good. Thanks for confirming!

pkg/apis/events/v1beta1/cloudstoragesource_conversion_test.go
danyinggu: No worries. I can fix them in my PR.

pkg/apis/events/v1beta1/cloudstoragesource_conversion_test.go
danyinggu: Just want to double check, I can fix them in my PR.

capri-xiyue: `in` is a pointer. If you don't deep copy it and directly modify it, it will affect other files who are using the same pointer.

danyinggu: Yes we should keep it.

PR #1449, opened 2020-07-14
to test parallelism of WI

nachocano:
Do not merge!

**Release Note**

```release-note

```

**Docs**

PR #1453, opened 2020-07-14
Reduce parallel run in WI test 

grac3gao:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/e2e-wi-tests.sh
+++ b/test/e2e-wi-tests.sh
@@ -88,6 +88,8 @@ function control_plane_setup() {
           --role roles/iam.workloadIdentityUser \
           --member "${member_name}" \
           --project "${PROW_PROJECT_NAME}" "${CONTROL_PLANE_SERVICE_ACCOUNT_EMAIL}"
+          # Add a sleep time between each get-set iam-policy-binding loop to avoid concurrency issue. Sleep time is based on the SLO.
+          sleep 10
       fi
     done <<< "$members"
     # Allow the Kubernetes service account to use Google service account.
@@ -159,6 +161,6 @@ function gcp_auth_setup() {
 initialize $@ --cluster-creation-flag "--workload-pool=\${PROJECT}.svc.id.goog"
 
 # Channel related e2e tests we have in Eventing is not running here.
-go_test_e2e -timeout=30m -parallel=6 ./test/e2e -workloadIndentity=true -serviceAccountName="${K8S_SERVICE_ACCOUNT_NAME}" || fail_test
+go_test_e2e -timeout=30m -parallel=4 ./test/e2e -workloadIndentity=true -serviceAccountName="${K8S_SERVICE_ACCOUNT_NAME}" || fail_test
 
 success


test/e2e-wi-tests.sh
```diff
@@ -159,6 +161,6 @@ function gcp_auth_setup() {
 initialize $@ --cluster-creation-flag "--workload-pool=\${PROJECT}.svc.id.goog"
 
 # Channel related e2e tests we have in Eventing is not running here.
-go_test_e2e -timeout=30m -parallel=6 ./test/e2e -workloadIndentity=true -serviceAccountName="${K8S_SERVICE_ACCOUNT_NAME}" || fail_test
+go_test_e2e -timeout=30m -parallel=4 ./test/e2e -workloadIndentity=true -serviceAccountName="${K8S_SERVICE_ACCOUNT_NAME}" || fail_test
```
ian-mi, line 164: I think we should attempt to fix the concurrency issues instead of reducing parallelism as this will mask issues with WI and slow down test runs. I think it would also be helpful to observe the effect of the previous change in isolation before making further changes.

yolocs: Something we have agreed in today's team standup is to take the best effort (it's ok to be shortcut like this) to reduce flakiness. The point is to unblock PRs with best effort. It makes sense if we want to delay this one a little bit to see if previous changes are effective or not. But I would prefer to get this in if we don't have a fix soon.

capri-xiyue:
It looks like reduce the parallel doesn't work. The e2e tests still fail with concurrency error.

test/e2e-wi-tests.sh
grac3gao: Discussed in today's standup, it seems some PRs have continuously encountered the test failure for concurrency. So want to reduce the parallel runs as a quick fix to unblock them. However, as the test result shows, and also from Nacho's test https://github.com/google/knative-gcp/pull/1449, even setting the parallel number to be 1, there are still concurrency issue (https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/1449/pull-google-knative-gcp-wi-tests/1283169781486194689) (which is weird...). Probably adding back-off sleep would be the next possible fixing... I have added the 10s sleep in test script in this PR (sleep time is according to SLO), and the while loop removing error mentioned https://github.com/google/knative-gcp/pull/1453#issuecomment-658459862 doesn't show up (but the amount of the test is limited). Probably we can add similar back-off sleep in iamPolicyManager?

capri-xiyue: Seems that exponential-backoff (https://cloud.google.com/iot/docs/how-tos/exponential-backoff)  is the official suggestions from IAM. Not sure whether that's too complicated to implement. If so, we can see whether 10s sleep based on slo works

grac3gao: Yeah, we can also try that. We used to have exponential-backoff https://github.com/grac3gao/knative-gcp/blob/13cee082f19d09ed6936ae53473e3b94625873f7/pkg/reconciler/identity/reconciler.go#L53-L62, using https://godoc.org/k8s.io/apimachinery/pkg/util/wait, it didn't have significant help though (probably the duration didn't set up proper). It might have more power together with iamPolicyManager's serialization.

grac3gao:
close, as we have #1461

PR #1457, opened 2020-07-14
Update installation docs to include additional roles needed in WI

grac3gao:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Follow this guide to install Knative-GCP components on a platform of your
 choice.
 
 1. [Installing Knative-GCP](./docs/install/install-knative-gcp.md)
-1. [Installing Pub/Sub Enabled Service Account](./docs/install/pubsub-service-account.md)
+1. [Installing Pub/Sub Enabled Service Account for the Data Plane](docs/install/dataplane-service-account.md)
 1. [Installing GCP Broker](./docs/install/install-gcp-broker.md)
 1. [Installing Broker with PubSub Channel](./docs/install/install-broker-with-pubsub-channel.md)
 1. [Managing Multiple Projects](./docs/install/managing-multiple-projects.md)
--- a/README.md
+++ b/README.md
@@ -14,7 +14,7 @@ Follow this guide to install Knative-GCP components on a platform of your
 choice.
 
 1. [Installing Knative-GCP](./docs/install/install-knative-gcp.md)
-1. [Installing Pub/Sub Enabled Service Account for the Data Plane](docs/install/dataplane-service-account.md)
+1. [Installing Pub/Sub Enabled Service Account for the Data Plane](./docs/install/dataplane-service-account.md)
 1. [Installing GCP Broker](./docs/install/install-gcp-broker.md)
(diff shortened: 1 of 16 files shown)


docs/install/pubsub-service-account.md
```diff
@@ -46,6 +46,23 @@ also need the ability to publish messages (`roles/pubsub.publisher`).
      --member=serviceAccount:cre-pubsub@$PROJECT_ID.iam.gserviceaccount.com \
      --role roles/pubsub.editor
    ```
+   
+   ***Note:***
```
liu-cong, line 50: Now that we need more permissions than pubsub, I think we should rename the SA to something like `cre-dataplane`, and rename the file as well.

docs/install/pubsub-service-account.md
```diff
@@ -46,6 +46,23 @@ also need the ability to publish messages (`roles/pubsub.publisher`).
...
+        --role roles/monitoring.metricWriter
+      ```
+   and `roles/cloudtrace.agent` for tracing functionality:
+      ```shell
+      gcloud projects add-iam-policy-binding $PROJECT_ID \
+        --member=serviceAccount:cre-pubsub@$PROJECT_ID.iam.gserviceaccount.com \
+        --role roles/cloudtrace.agent
+      ```
+   If you are going to configure kubernetes Secret for Authentication Mechanism, 
+   then no need to add `roles/monitoring.metricWriter` and `roles/cloudtrace.agent`.
```
liu-cong, line 65: Why secret mode doesn't require these?

docs/install/pubsub-service-account.md
grac3gao: I am not sure if we also want to separate the GSA used for source and broker?

docs/install/pubsub-service-account.md
grac3gao: In secret mode, it will pass the related test without those two roles, I think it is probably because that the Pod are using the node identity for metric and tracing, not the identity for secret.

docs/install/pubsub-service-account.md
liu-cong: I don't feel strongly. Since right now they share the same permissions (maybe sources don't need metrics for now? but we should probably add metrics anyway), I think it's better to keep using the same one.

README.md
```diff
@@ -14,7 +14,7 @@ Follow this guide to install Knative-GCP components on a platform of your
 choice.
 
 1. [Installing Knative-GCP](./docs/install/install-knative-gcp.md)
-1. [Installing Pub/Sub Enabled Service Account](./docs/install/pubsub-service-account.md)
+1. [Installing Pub/Sub Enabled Service Account for the Data Plane](./docs/install/dataplane-service-account.md)
```
liu-cong, line 17:
```suggestion
1. [Installing a Service Account for the Data Plane](./docs/install/dataplane-service-account.md)
```

liu-cong: Remove "Pub/Sub Enabled" in all other places.

Issue #1357, opened 2020-06-27
Make IAMPolicyManager more robust for concurrency error


grac3gao:
**Describe the bug**
The concurrency error appears in the recent PR https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/1144/pull-google-knative-gcp-wi-tests/1276548920733863938

Note that this concurrency issue doesn't happen so often. I vaguely remember it only happens for several times among all my PRs' submit tests. However,  it do happen sometimes after the PR for serializing the policy changes  https://github.com/google/knative-gcp/pull/833

One suggestion from @ian-mi is adding some backoff between changes

 **Expected behavior**

 **To Reproduce**
Steps to reproduce the behavior.

 **Knative-GCP release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

capri-xiyue:
It happened in my PR several times today https://prow.knative.dev/view/gcs/knative-prow/pr-logs/pull/google_knative-gcp/1388/pull-google-knative-gcp-wi-tests/1280584890395398145

nachocano:
This seems to be related to the CI. 
@grac3gao did some improvements, but it seems that are not enough. 
@ian-mi is this something you can take on, given that you wrote the IAMPolicyManager?

capri-xiyue:
Have we implemented exponential-backoff in WI when we do the retry of the whole read-modify-write https://cloud.google.com/iot/docs/how-tos/exponential-backoff ?

capri-xiyue:
The e2e of WI tests is quite unstable due to this concurrency error. https://prow.knative.dev/job-history/gs/knative-prow/pr-logs/directory/pull-google-knative-gcp-wi-tests.

grac3gao:
Just add the summary of discussion from the chat:
1. We used to have exponential backoff retry when setting iam policy binding in WI reconciler, but it was removed later as we have IAMPolicyManager to serialize IAM policy changes. Exponential backoff retry didn't work well though at that time in WI reconciler(it doesn't fully solve the concurrency issue). It was until we have IAMPolicyManager, the concurrency issue solved.
2. There is currently no backoff in the policy manager but there should also not be much contention since changes are serialized. Still, adding some backoff could be a good idea. Note also that it would not retry individual changes at the manager layer.

nachocano:
@grac3gao do you know if this can be closed with this PR https://github.com/google/knative-gcp/pull/1461?

[...]

PR #1461, opened 2020-07-15
Add back-off in IAMPolicyManager

grac3gao:
Fixes #
related #1357 

## Proposed Changes

- 🎁 add back-off in IAMPolicyManager to reduce flakiness caused by concurrency issue.
-
-

**Release Note**

```release-note

add back-off in IAMPolicyManager
```

**Docs**

### DIFF
--- a/pkg/gclient/iam/admin/client.go
+++ b/pkg/gclient/iam/admin/client.go
@@ -27,17 +27,18 @@ import (
 	"google.golang.org/grpc/status"
 )
 
-type client struct {
-	policies map[string]*iam.Policy
+type TestIamClient struct {
+	policies           map[string]*iam.Policy
+	setIamPolicyErrors []error
 }
 
-func NewTestClient() IamClient {
-	return client{
+func NewTestClient() *TestIamClient {
+	return &TestIamClient{
 		policies: make(map[string]*iam.Policy),
 	}
 }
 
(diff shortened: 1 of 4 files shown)


ian-mi:
@yolocs Can you please take another look? I probably shouldn't review since I authored one of the commits.

yolocs:
/lgtm

Maybe because I don't have any functional programming background. I found it's pretty hard to follow the logic in `iam_policy.go` with all these channels (despite you already have pretty thorough comments). Curious why don't we simply streamline SetIamPolicy? Is throughput critical in this case?

ian-mi:
The IAM calls are not made synchronously to avoid blocking the main goroutine. Throughput is likely not critical here, but we would want to avoid blocking an IAM policy change to one service account on a change being made to another account for example especially since a stall in the policy manager will also stall the reconciler. I'm not sure if I would call this a functional programming pattern, but goroutines and channels make it easy to design this in a way that is completely asynchronous. The client / server pattern I'm using here is described extremely well in this talk by Sameer Ajmani: https://blog.golang.org/io2013-talk-concurrency.

yolocs:
I understand that go promotes channel. But in this specific case IMO it's a little bit overused and starting to hurt readability. If it were me I would just use channel to implement a fan-in that streamlines the IAM reqs for each SA. That's just my opinion tho. Maybe others feel it's intuitive the existing way.

PR #1465, opened 2020-07-15
Updating dependencies & leader election bits

nachocano:
## Proposed Changes

- Updating dependencies, had to copy the leader election change from eventing but disable HA for our controller. See https://github.com/knative/eventing/commit/5a3dba03ad9482bb2bfb4e3f5720f21e9ac1eaa9
- Disabling HA in the controller until we get a Deployment reconciler leader-aware. See https://github.com/google/knative-gcp/issues/1466 for more context.
- Adding eventing patch for leader election bug.

**Release Note**

```release-note
NONE
```

**Docs**


nachocano:
There is a problem that the deployment reconciler is not leader-aware, so we are crash looping, now that the default behavior is to run in leader-election mode. 
Working on a separate PR to get this working

nachocano:
having a problem that every time a deployment changes, and not based on the Secret. I think is due to the PromoteFunc in LeaderAwareFuncs that enqueues all the objects. 

Also having a problem in the webhook.
E0716 06:40:18.097598       1 leaderelection.go:307] Failed to release lock: Lease.coordination.k8s.io "webhook.defaultingwebhook.00-of-01" is invalid: spec.leaseDurationSeconds: Invalid value: 0: must be greater than 0

Seems that a patch that folks added in eventing might help. Will add it here as well...

nachocano:
Decided to disable HA for our controller. We need to find a better way of dealing with the secret. 
If we enqueue the controller that is the leader, then it will get restarted, as it will be annotated, therefore losing its leadership, and things will keep on being restarted.

PR #1470, opened 2020-07-16
More detailed examples for CloudEvents Pub/Sub

cneijenhuis:
I implemented the Pub/Sub binding for CloudEvents and found the examples not very helpful. They claim to be a `publish` operation, but they're missing a lot of details to make it a real `publish`.

## Proposed Changes

- The first example shows only the mapping to attributes and data
- Add a second example that shows mapping the message onto a real publish request on the REST API

**Release Note**

```release-note

```

**Docs**


cneijenhuis:
@googlebot I signed it!

/assign @yolocs

yolocs:
Thanks for the contribution!
/lgtm

Issue #1409, opened 2020-07-08
Promote CloudAuditLogsSource to v1


nachocano:
**Problem**
We need to promote CloudAuditLogsSource to v1.
Some tasks this will entail, and can probably be done in different PRs.

- [x] v1 types with UTs
- [x] add new version to CRD
- [x] defaulting and conversion webhook
- [x] Update docs and examples
- [x] Migrate reconciler to v1 with UTs
- [x] Add e2e tests to also support v1

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
CAL supports v1

PR #1473, opened 2020-07-16
Webhook conversion and defaults of sources v1

danyinggu:
Help with #1409, #1410, #1411 and #1412

## Proposed Changes

- 🎁 Webhook conversion and defaults of CloudPubSubSource, CloudAuditLogSource and CloudSchedulerSource.
- 🧽  Small fixes and cleanup.
- 🎁 Conversion and defaults of topic and pullsubscription.
- 🎁 Update the yamls to use API version `apiextensions.k8s.io/v1` and support v1. 
(Notice that we don't change the storage version, the storage version is still v1beta1.)

**Release Note**

```release-note

```

**Docs**


config/core/resources/cloudauditlogssource.yaml
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-apiVersion: apiextensions.k8s.io/v1beta1
+apiVersion: apiextensions.k8s.io/v1
```
danyinggu, line 15: Synced with @Harwayne offline, we should eventually update the CRDs of all resources to use `apiextensions.k8s.io/v1` as apiVersion, they will be done in a separate PR.

config/core/resources/cloudauditlogssource.yaml
```diff
@@ -36,158 +36,297 @@ spec:
...
-    webhookClientConfig:
-      service:
-        name: webhook
-        namespace: cloud-run-events
-  additionalPrinterColumns:
+    webhook:
+      # conversionReviewVersions indicates what ConversionReview versions are understood/preferred by the webhook.
+      # The first version in the list understood by the API server is sent to the webhook.
+      # The webhook must respond with a ConversionReview object in the same version it received.
+      conversionReviewVersions: ["v1beta1", "v1", "v1alpha1"]
```
danyinggu, line 46: @Harwayne @nachocano I'm not quite sure about this... We should make "v1" the first one after v1 is the storage version. But currently "v1beta1" is the storage version, so it'd be safer that we keep "v1beta1" as the first one in the list?

nachocano:
let's not change the storage version in this release. I was discussing with
OSS folks, and it seems that we typically wait for one release.
We can decide to do it if we want, but for now, let's keep v1beta1 be the
stored version.
So I think we should keep "v1beta1" first there

config/core/resources/cloudauditlogssource.yaml
capri-xiyue: Based on the doc https://kubernetes.io/docs/tasks/extend-kubernetes/custom-resources/custom-resource-definition-versioning/ and the examples in knative eventing, https://github.com/knative/eventing/blob/master/config/core/resources/sinkbindings.yaml
I think it should be `conversionReviewVersions: ["v1beta1", "v1"]`.
I don't think it's related to the version of sources, it's related to the versions of  `apiextensions.k8s.io`.
@Harwayne  @nachocano  Please correct me if I'm wrong.

config/core/resources/cloudauditlogssource.yaml
```diff
@@ -36,158 +36,297 @@ spec:
...
+                type: string
+              stackdriverSink:
+                type: string
+                description: >
+                  ID of the Stackdriver sink used to publish audit log messages.
+  - << : *version
+    name: v1beta1
+    served: true
+    storage: true
+  - <<: *version
```
capri-xiyue, line 201: The yaml looks kind of weird.
As I suggested, you use yaml merge https://yaml.org/type/merge.html to try to deduplicate common fields.
But you still copy all fields under `v1` version, in this case, what's the usage of using yaml merge https://yaml.org/type/merge.html.
I think you should combine all duplicate methods with yaml merge for reuse.

config/core/resources/cloudauditlogssource.yaml
capri-xiyue: similar issues for other resources.

config/core/resources/cloudauditlogssource.yaml
capri-xiyue: Similar issues for other resoureces

pkg/apis/intevents/v1beta1/topic_conversion_test.go
```diff
@@ -18,11 +18,43 @@ package v1beta1
...
+			AddressStatus:  gcptesting.CompleteV1beta1AddressStatus,
+			ProjectID:      "projectID",
+			TopicID:        "topicID",
+		},
+	}
 )
 
 func TestTopicConversionBadType(t *testing.T) {
-	good, bad := &Topic{}, &Topic{}
+	good, bad := &Topic{}, &PullSubscription{}
```
capri-xiyue, line 57: Why it's `&Topic{}` vs `&PullSubscription{}`?
The logic is correct, just the code style is not consistent with other resources.

pkg/apis/intevents/v1beta1/pullsubscription_conversion.go
```diff
@@ -18,17 +18,57 @@ package v1beta1
...
+func (source *PullSubscription) ConvertTo(ctx context.Context, to apis.Convertible) error {
+	switch sink := to.(type) {
+	case *v1.PullSubscription:
+		sink.ObjectMeta = source.ObjectMeta
+		sink.Spec.PubSubSpec = convert.ToV1PubSubSpec(source.Spec.PubSubSpec)
+		sink.Spec.Topic = source.Spec.Topic
+		sink.Spec.AckDeadline = source.Spec.AckDeadline
+		sink.Spec.RetainAckedMessages = source.Spec.RetainAckedMessages
+		sink.Spec.RetentionDuration = source.Spec.RetentionDuration
+		sink.Spec.Transformer = source.Spec.Transformer
```
Harwayne, line 40: Add a comment about why we are skipping `spec.mode`.

config/core/resources/cloudauditlogssource.yaml
capri-xiyue: I'm not an expert of yaml. @Harwayne  Any ideas?

pkg/apis/intevents/v1beta1/topic_conversion_test.go
danyinggu: I'm actually keep the code style consistent with other resources here, as they are all using explicit different types. 
Check: https://github.com/google/knative-gcp/blob/master/pkg/apis/intevents/v1alpha1/pullsubscription_conversion_test.go#L171 and https://github.com/google/knative-gcp/blob/master/pkg/apis/events/v1alpha1/cloudschedulersource_conversion_test.go#L54

config/core/resources/cloudauditlogssource.yaml
danyinggu: This should be fixed.

config/core/resources/cloudauditlogssource.yaml
danyinggu: I see, I fixed it in the current commit.

config/core/resources/cloudauditlogssource.yaml
```diff
@@ -36,158 +36,182 @@ spec:
...
-    webhookClientConfig:
-      service:
-        name: webhook
-        namespace: cloud-run-events
-  additionalPrinterColumns:
+    webhook:
+      # conversionReviewVersions indicates what ConversionReview versions are understood/preferred by the webhook.
+      # The first version in the list understood by the API server is sent to the webhook.
+      # The webhook must respond with a ConversionReview object in the same version it received.
+      conversionReviewVersions: ["v1beta1", "v1"]
```
Harwayne, line 46: Looks like upstream has `v1` first, should we too?

config/core/resources/cloudauditlogssource.yaml
Harwayne: This looks correct to me. My memory is that merge is an optional part of YAML, but it looks like Kubernetes supports it, so we can use it.

config/core/resources/cloudauditlogssource.yaml
danyinggu: Fixed.

config/core/resources/cloudauditlogssource.yaml
```diff
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
danyinggu: Sounds good, I just changed it back.

config/core/resources/cloudauditlogssource.yaml
capri-xiyue: The yaml merge syntax is correct.  But previously we duplicated all fields. I think currently it is correct

pkg/apis/intevents/v1beta1/topic_conversion_test.go
capri-xiyue: Ok. I think the codebase uses different code style. 
Check https://github.com/google/knative-gcp/blob/e76eb0926f07aa97d59cfb66742f28361cfdc831/pkg/apis/intevents/v1beta1/pullsubscription_conversion_test.go#L25.
We can clean it in follow up PRs.

Issue #1474, opened 2020-07-16
Istio sidecar inject should be explicitly disabled on some workloads


yolocs:
E.g. controller, webhook, broker fanout/retry

PR #1476, opened 2020-07-16
Explicitly disallow istio sidecar injection for certain workloads

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1474

## Proposed Changes

- 
-
-

**Release Note**

```release-note
Explicitly disallow istio sidecar injection for controller/webhook/broker fanout/retry
```

**Docs**


yolocs:
@Harwayne Fanout/retry are pull based so no inbound policy enforcement is needed. We haven't clearly identified how to use istio for outbound traffic to other sidecar'ed pods (there are several possibilities though). So to minimize impact, I choose to disable sidecar for fanout/retry at the moment. You think it's a better idea to enable them now?

upodroid:
Hi @yolocs 

I need to have the istio sidecars enabled for all these workloads, where do I set the allowIstioSidecar: true value?

Thank you

yolocs:
@upodroid The parameter is internal. E.g. https://github.com/google/knative-gcp/blob/master/pkg/reconciler/brokercell/brokercell.go#L238

If you can create an issue, I will help make the change :)

PR #1485, opened 2020-07-21
Use GCP Pub/Sub backoff retry policy instead of client-side implementation

### LINKED ISSUE #1267: Use pubsub retry policy to backoff retries
Follow up on https://github.com/google/knative-gcp/issues/1100. As pubsub retry policy goes GA we should switch over.

tommyreddad:
Fixes #1267 

## Proposed Changes
- Remove client-side implementation of backoff, configure backoff through GCP subscription instead
- Configure backoff retry policy in the trigger controller instead of broker retry main
- Remove the backoff test since it is GCP pub/sub's responsibility


pkg/broker/handler/handler.go
```diff
@@ -115,14 +107,11 @@ func (h *Handler) receive(ctx context.Context, msg *pubsub.Message) {
 		defer cancel()
 	}
 	if err := h.Processor.Process(ctx, event); err != nil {
-		backoffPeriod := h.retryLimiter.When(msg.ID)
-		logging.FromContext(ctx).Error("failed to process event; backoff nack", zap.String("eventID", event.ID()), zap.Duration("backoffPeriod", backoffPeriod), zap.Error(err))
-		h.delayNack(backoffPeriod)
+		logging.FromContext(ctx).Error("failed to process event; backoff nack", zap.String("eventID", event.ID()), zap.Error(err))
```
yolocs, line 110: nit: remove "backoff nack"

pkg/reconciler/trigger/trigger.go
```diff
@@ -216,8 +219,9 @@ func (r *Reconciler) reconcileRetryTopicAndSubscription(ctx context.Context, tri
 	// Check if PullSub exists, and if not, create it.
 	subID := resources.GenerateRetrySubscriptionName(trig)
 	subConfig := pubsub.SubscriptionConfig{
-		Topic:  topic,
-		Labels: labels,
+		Topic:       topic,
+		Labels:      labels,
+		RetryPolicy: r.retryPolicy,
```
yolocs, line 224: Is it possible to somehow check in the unit tests that the retry policy is actually set?

Issue #1448, opened 2020-07-14
Check the correctness and consistency of immutability constraints in all of our resources


danyinggu:
**Problem**
Related to #1441, #1442 and #1433.
When promoting the sources related types to v1, we found that the checkImmutableFields methods between validations are somewhat inconsistent and incomplete (also for the UTs). We need to check them again in the following sprint to make sure they are correct.

**Exit Criteria** 
- [ ] Make sure the checkImmutability is called in validation logic for all versions of sources (v1, v1beta1 and v1alpha1). @bharattkukreja 

- [x]  Make sure in our code base, the immutable fields that are included in the checkImmutableFields function are correct, complete and consistent among sources, need to add the missing fields if needed. @danyinggu 
    As mentioned in PR #1504, we add immutability check ClusterNameAnnotation and AutoscalingClass Annotation (relates to #665) for all versions of resources. We also add more fields to be immutable in all of the versions. (see the PR for details)

- [ ] Based on 1 and 2, make sure the immutable fields cannot be mutated when using kubectl edit and mutable fields can be mutated. @bharattkukreja 

cc @nachocano

nachocano:
We need to do this in order to be actually immutable in higher versions: https://github.com/knative/eventing/pull/3523, https://github.com/knative/eventing/pull/3522

nachocano:
I think that needs to be done in v1beta1 and v1...

This has just become a high priority bug

capri-xiyue:
> I think that needs to be done in v1beta1 and v1...
> 
> This has just become a high priority bug

Wondering whether we  also need to do that in v1alpha1. I checked the codebase, we write the function `CheckImmutableFields` but it is never get called in validation logic.

danyinggu:
@bharattkukreja : Just synced with @nachocano offline, we should:
1. Make sure the checkImmutability is called in validation logic for all versions of sources (v1, v1beta1 and v1alpha1). 
2. Make sure in our code base, the immutable fields that are included in the checkImmutableFields function are correct, complete and consistent among sources, need to add the missing fields if needed. 
3. Based on 1 and 2, make sure the immutable fields cannot be mutated when using `kubectl edit` and mutable fields can be mutated.

danyinggu:
I already worked on the UTs to cover all fields in our current code base, will submit a PR, and @bharattkukreja will take over this issue. He might need to add more fields if missing and add UTs to them (see point 2 in the above comment)

Issue #1494, opened 2020-07-21
Release-0.16 branch vendors an outdated Knative Eventing 0.15 dependency


Ectelion:
**Describe the bug**
In [go.mod (version 0.16)]( https://github.com/google/knative-gcp/blob/v0.16.0/go.mod), see knative.dev/eventing dependency.

 **Expected behavior**
The branch should vendor the respective 0.16 Eventing version.

Harwayne:
@ericlem We might want a new issue that addresses the fact that it seems like our update-deps.sh script doesn't always update to what we expect.

Ectelion:
@Harwayne, @ericlem I suspect it's due to calling update-deps without the "--upgrade" flag. Maybe a simple warning in such case would do.

nachocano:
This is the PR that did the update. I'm almost sure I ran it with `--upgrade` (although I might be wrong)
https://github.com/google/knative-gcp/pull/1390
If you see the go.mod there, you can see that serving went to 0.16, but not eventing. So, I'm more confident that it was run with `upgrade`
If you see this PR that move pkg to 0.16 https://github.com/google/knative-gcp/pull/1373, you can see the go.mod that doesn't have pkg in 0.16.

Something similar happened to me a few times internally. I think is worth investigating a bit... 

Maybe is something related to this: https://stackoverflow.com/questions/58364988/go-module-not-downloading-latest-minor-version/60660853#60660853

PR #1495, opened 2020-07-21
Issue #1494: Release-0.16 branch vendors an outdated Knative Eventing dependency

Ectelion:
Addresses #1494

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -31,7 +31,7 @@ require (
 	k8s.io/api v0.18.1
 	k8s.io/apimachinery v0.18.1
 	k8s.io/client-go v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible
-	knative.dev/eventing v0.15.1-0.20200707170044-e018fe794753
+	knative.dev/eventing v0.16.1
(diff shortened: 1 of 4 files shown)


nachocano:
/lgtm
/approve
/hold
until we get the other immutable fix in, so that we don't accidentally trigger two point releases

nachocano:
This is the issue we should cherry-pick to this branch, once we have the fix:
https://github.com/google/knative-gcp/issues/1448

fyi @danyinggu, @bharattkukreja

grac3gao:
Checked the log from the failed WI test, it seems that this branch doesn't use some latest main branch updates to reduce flakiness, like the test sender Pod is still using an old version which doesn't include the retry for certain error code. It might be worthwhile to do a `git merge` to include those updates for reducing the test flakiness if tests kept failing.

grac3gao:
Sorry, I didn't realize it is for release-0.16 branch not for the main branch...then it might experience more flakiness...

nachocano:
Closing this one as we already pushed that change into 0.16 branch

PR #1504, opened 2020-07-22
Update immutable fields for sources in v1alpha1 and v1beta1

### LINKED ISSUE #665: Autoscaling annotation should be made immutable
**Describe the bug**
The autoscaling annotation should be made immutable, otherwise the correct controller won't be called if changed.

 **Expected behavior**
Autoscaling annotation cannot be changed

 **To Reproduce**
Change the annotation and you can save the object

0.13

danyinggu:
Helps with #1448.
Fixes #665.

## Proposed Changes

- Update immutable fields check in v1alpha1 and v1beta1.
Due to cherry pick for 0.16 release, we need to separate the update of v1 in a different PR.
- The changes are (confirmed with @nachocano ):
1. For `Topic`, `PropagationPolicy` and `EnablePublisher` should be immutable.
2.  For `PullSubscription`, `Mode`, `AckDeadline`,` RetainAckedMessages`, `RetentionDuration` should be immutable.
3. For `CloudPubSubSource`, `AckDeadline`,` RetainAckedMessages`, `RetentionDuration` should be immutable.
- Sync with @Harwayne and @grac3gao offline, in v1alpha1 we check immutability for clusterNameAnnotation (as we setClusterNameAnnotation by defaults), in check should be consistent across all of the versions although we don't setClusterNameAnnotation by default, check only runs if the field is non-empty, so we might need to check anyway.
- Add check for AutoscalingClassAnnotation to be immutable no matter it has been set initially or not.

**Release Note**

```release-note

```

**Docs**


nachocano:
Can you also make the autoscaling annotation immutable as part of this issue?
I just remembered I created an issue sometime back https://github.com/google/knative-gcp/issues/665
Either in this one, or in a follow up. I was thinking of this one for easier cherry-picking.

danyinggu:
> Can you also make the autoscaling annotation immutable as part of this issue?
> I just remembered I created an issue sometime back #665
> Either in this one, or in a follow up. I was thinking of this one for easier cherry-picking.

Sounds good!

nachocano:
/lgtm
/approve

Thanks @danyinggu !!!

PR #1508, opened 2020-07-23
Add broker support for GCP Profiler

grantr:
Enable the GCP profiler with GCP_PROFILER=1 for any broker service.

Config variables available:
GCP_PROFILER_SERVICE_VERSION (default 0.1)
GCP_PROFILER_PROJECT (optional on GCP if metadata service is enabled)
GCP_PROFILER_DEBUG_LOGGING (default false)

This eliminates the need to maintain the patch at https://github.com/yolocs/ce-test-actor/blob/master/scripts/profiling/broker.patch. We can use env vars instead, possibly with a [PodPreset](https://kubernetes.io/docs/tasks/inject-data-application/podpreset/).

Also note that Knative pkg contains [code to set up a pprof http server](https://github.com/knative/pkg/blob/8c62412074275c22632e23fbe726555bf970ba6c/profiling/server.go). Might be a useful followup (EDIT: We already use this in mainhelper.Init)

## Proposed Changes

- Add `profiling` package containing `GCPProfilerEnvConfig` to be embedded in main envconfigs
- Add `StartGCPProfiler` to encapsulate GCPProfiler options handling

/cc @yolocs


grantr:
/hold
I haven't tested this yet. Will unhold after verifying it works.

cmd/broker/fanout/main.go
```diff
@@ -68,6 +72,8 @@ func main() {
 		logger.Fatalf("MAX_STALE_DURATION must be greater than pool resync period %v", poolResyncPeriod)
 	}
 
+	profiling.StartGCPProfiler("broker-fanout", env.GCPProfilerEnvConfig)
```
liu-cong, line 75: Can we move this to the mainhelper.Init as an option (line 67 above)?

grantr: Thanks for pointing this out! I see we already use knative/pkg's pprof handler. I'll move gcp profiler init here.

pkg/utils/mainhelper/init.go
```diff
@@ -92,9 +96,25 @@ func ProcessEnvConfigOrDie(env interface{}) {
...
+func ProcessGCPProfilerEnvConfigOrDie(component string, logger *zap.SugaredLogger) {
+	var env profiling.GCPProfilerEnvConfig
+	if err := envconfig.Process("", &env); err != nil {
+		logger.Desugar().Fatal("Failed to process GCP Profiler env config", zap.Error(err))
+	}
+	started, err := profiling.StartGCPProfiler(component, env)
+	if err != nil {
+		logger.Desugar().Fatal("Failed to start GCP Profiler", zap.Error(err))
+	}
+	if started {
```
liu-cong, line 110: nit: It seems that the `started` is only used for logging. It can be removed if you check `GCPProfilerEnabled` and call `StartGCPProfiler` if true.

cmd/broker/fanout/main.go
grantr: Ok, this should now be configured automatically for any command that uses mainhelper.Init. Since the available env vars are hidden in the profiling package, I added docs in `docs/development/profiling.md` to explain what env vars may be used.

pkg/utils/mainhelper/init.go
grantr: I switched to that style and added a method to check enablement.

pkg/utils/mainhelper/init.go
```diff
@@ -63,6 +67,9 @@ func Init(component string, opts ...InitOption) (context.Context, *InitRes) {
 	logger := logging.FromContext(ctx)
 	RunProfilingServer(ctx, logger, profilingHandler)
 
+	// Try starting the GCP Profiler if config is present.
+	ProcessGCPProfilerEnvConfigOrDie(component, logger)
```
grantr, line 71: @liu-cong can you help adding a test for this? I'm not sure how the tests in init_test.go work.

liu-cong: It seems that there is no way to check if the profiler is started or not. So I don't know how to test this.

grantr:
/hold cancel

It works. Unfortunately PodPreset is not enabled in GKE without enabling alpha features, so we can't use that. I'm currently testing by running an altered version of the controller that sets the GCP_PROFILER env var. I may have to retool the configuration to use the config-observability configmap as well.

PR #1509, opened 2020-07-23
#1484. Configurable CPU and memory params for BrokerCell

### LINKED ISSUE #1484: Allow tuning resource requests and limits for BrokerCell deployments
**Problem**
While doing performance investigations we will likely need to experiment with memory limits to understand how the broker components react. Currently there's no way to do this without editing the code, because the BrokerCell will reconcile any changes to limits made on the Deployment objects.

These fields may also be useful to the cluster operator persona, but for now I think that should be out of scope since our scaling design may change as we investigate.

This can build on the HPA tuning knobs added in #1379.

Contributor

**Exit Criteria**
I can modify the CPU and memory requests and limits for each BrokerCell deployment (ingress, fanout, retry) individually.

Ectelion:
## Proposed Changes
A follow up for the [prior work](https://github.com/google/knative-gcp/pull/1379/files) on BrokerCell configurability that makes the CPU and memory requests/limits for Fanout, Ingress and Retry configurable, along with the thresholds for their respective HorizontalPodAutoscalers. Fixes #1484


pkg/apis/intevents/v1alpha1/brokercell_types.go
```diff
@@ -71,11 +71,29 @@ var (
...
+
+	// CPULimit specifies the maximal amount of the CPU to be consumable by the deployment
+	CPULimit *string `json:"cpuLimit,omitempty"`
+
 	// MinReplicas specifies the minimum replica count for the component.
 	MinReplicas *int32 `json:"minReplicas,omitempty"`
 
 	// MaxReplicas specifies the maximum replica count for the component.
 	MaxReplicas *int32 `json:"maxReplicas,omitempty"`
+
```
liu-cong, line 91: nit: Let's move these up above the MinReplicas and below the CPULImit

pkg/apis/intevents/v1alpha1/brokercell_types.go
```diff
@@ -71,11 +71,29 @@ var (
...
+	// MemoryRequest specifies the minimal amount of the CPU for the deployment to be schedulable
```
liu-cong, line 92:
```suggestion
	// MemoryRequest specifies the minimal amount of the memory for the deployment to be schedulable
```

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -19,12 +19,30 @@ package v1alpha1
...
+	// Here we only set half of the limit so that in case of surging memory
+	// usage, HPA could have enough time to kick in.
+	// See: https://github.com/google/knative-gcp/issues/1265
+	avgMemoryUsage                         string  = "1500Mi"
+	avgMemoryUsageIngress                  string  = "700Mi"
+	cpuRequest                             string  = "1000m"
+	cpuRequestFanout                       string  = "1500m"
+	cpuLimit                               string  = ""
+	memoryRequest                          string  = "500Mi"
+	memoryLimitToRequestCoefficient        float64 = 6.0
```
liu-cong, line 40: What's the motivation to use a coefficient rather than an absolute number?

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -19,12 +19,30 @@ package v1alpha1
...
+	memoryLimitToRequestCoefficientIngress float64 = 2.0
```
liu-cong, line 41: I prefer to create independent constants for fanout and retry as well to be cleaner.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -19,12 +19,30 @@ package v1alpha1
...
-	maxReplicas int32 = 10
+	avgCPUUtilization int32 = 95
+	// The limit we set (for Fanout and Retry) is 3000Mi which is mostly used
+	// to prevent surging memory usage causing OOM.
+	// Here we only set half of the limit so that in case of surging memory
+	// usage, HPA could have enough time to kick in.
+	// See: https://github.com/google/knative-gcp/issues/1265
+	avgMemoryUsage                         string  = "1500Mi"
+	avgMemoryUsageIngress                  string  = "700Mi"
+	cpuRequest                             string  = "1000m"
```
liu-cong, line 36: nit: Create 3 constants for ingress/fanout/retry, or rename this to `commonCpuRequest`, same for other params.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
Ectelion: Thanks for reviewing! Have addressed this.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
Ectelion: Addressed

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
Ectelion: Part of this work was accounting for the HPA thresholds and handle the cases when they are not specified explicitly. For instance, if one sets "50Mi" and "300Mi" for requests and limits, respectively, then the original threshold of "1500Mi" may not be as suitable anymore -- instead, we will proportionally adjust it to a more or less suitable value of "150Mi".

pkg/apis/intevents/v1alpha1/brokercell_types.go
Ectelion: Sure

pkg/apis/intevents/v1alpha1/brokercell_types.go
Ectelion: Oh, thanks for the catch!

pkg/apis/intevents/v1alpha1/brokercell_types.go
```diff
@@ -71,6 +71,24 @@ var (
...
+
+	// CPULimit specifies the maximal amount of the CPU to be consumable by the deployment
+	CPULimit *string `json:"cpuLimit,omitempty"`
+
+	// MemoryRequest specifies the minimal amount of memory for the deployment to be schedulable
+	MemoryRequest *string `json:"memoryRequest,omitempty"`
+
+	// MemoryLimit specifies the maximal amount of memory to be consumable by the deployment
+	MemoryLimit *string `json:"memoryLimit,omitempty"`
+
```
yolocs, line 91: Does it make sense to directly use `corev1.ResourceRequirements` here?

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -35,22 +63,67 @@ func (bc *BrokerCell) SetDefaults(ctx context.Context) {
...
-		bcs.Components.Ingress.MinReplicas = ptr.Int32(minReplicas)
+	if componentParams.AvgCPUUtilization == nil {
+		componentParams.AvgCPUUtilization = ptr.Int32(avgCPUUtilization)
 	}
-	if bcs.Components.Ingress.MaxReplicas == nil {
-		bcs.Components.Ingress.MaxReplicas = ptr.Int32(maxReplicas)
+	// If target average consumption for the auto-scaler is not explicitly specified or falls beyond reasonable range,
+	// default it based on the memory limit
+	memoryLimitQuantity := resource.MustParse(*componentParams.MemoryLimit)
+	if componentParams.AvgMemoryUsage == nil || memoryLimitQuantity.Cmp(resource.MustParse(*componentParams.AvgMemoryUsage)) < 0 {
```
yolocs, line 125: I think the second part belongs to validation.

Ectelion: We can have a separate "validation" section, though to me it sounds more suitable for the cases when we won't proceed with given values and rather throw an exception immediately, whereas here it seemed fine to keep it with auto-selection/defaulting. Thoughts?

pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -64,13 +64,16 @@ func MakeIngressDeployment(args IngressArgs) *appsv1.Deployment {
 	}
 	container.Resources = corev1.ResourceRequirements{
 		Limits: corev1.ResourceList{
-			corev1.ResourceMemory: resource.MustParse("1000Mi"),
+			corev1.ResourceMemory: resource.MustParse(args.MemoryLimit),
```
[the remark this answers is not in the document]
Ectelion: Yes. However, this won't happen due to defaulting logic, where memory limit is defaulted based on memory request, if not explicitly set.

pkg/apis/intevents/v1alpha1/brokercell_types.go
Ectelion: It should be doable and I see value in doing that. It also seems natural to separate input params from the requirements structure, especially that we've got a mix of autoscaling and resource params here. I guess the main consideration here is how the product and that internal structure may evolve over time. A typical example would be: if there will be an update in the format of the internal structure, during upgrade, this may possibly require an update of existing deployment yamls (i.e. customer side), whereas by having those separated, we may limit the change to the product update (i.e. our side).

pkg/reconciler/brokercell/resources/deployments.go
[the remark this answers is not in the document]
yolocs: I think it's better to code defensively here. We may change the logic of defaulting and forget about updating this. Also try avoid using `MustParse`. Previously it's ok because we hardcoded the values and we know they are valid. Now the values could be specified by users, and from what I can tell we don't do properly validation in the webhook.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -35,22 +63,67 @@ func (bc *BrokerCell) SetDefaults(ctx context.Context) {
...
+}
+
+// SetCPUDefaults sets the CPU consumption related default field values for ComponentParameters.
+func (componentParams *ComponentParameters) SetCPUDefaults(defaultCPURequest, defaultCPULimit string) {
+	if componentParams.CPURequest == nil {
+		componentParams.CPURequest = ptr.String(defaultCPURequest)
+	}
+	if componentParams.CPULimit == nil {
+		componentParams.CPULimit = ptr.String(defaultCPULimit)
 	}
```
yolocs, line 105: This is problematic. If I explicitly set a limit (e.g. 500m) but without request. Then this logic would default the request to something like 1500m which is not correct.

I suggest group request/limit together like

```
type ComponentParameters struct {
  CPU *SystemResource
  Memory *SystemResource
  // ...
}

type SystemResource struct {
  Request, Limit string
}
```

If `SystemResource` is specified for CPU, then we don't set its value anymore. We only set defaults if `SystemResource` is nil. That way we know if a request/limit is explicitly set to empty or it's something we need to set default.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -35,22 +63,67 @@ func (bc *BrokerCell) SetDefaults(ctx context.Context) {
...
-	if bcs.Components.Fanout.MaxReplicas == nil {
-		bcs.Components.Fanout.MaxReplicas = ptr.Int32(maxReplicas)
+}
+
+// SetAutoScalingDefaults sets the autoscaling-related default field values for ComponentParameters.
+func (componentParams *ComponentParameters) SetAutoScalingDefaults(targetMemoryUsageCoefficient float64, avgCPUUtilization int32) {
+	if componentParams.MemoryLimit == nil {
+		panic("Memory limit should be specified on the component before setting auto-scaling parameters")
 	}
```
yolocs, line 112: Try avoid `panic` as much as possible. In this case we definitely don't want to crash the webhook for a single invalid value. Instead, valid it in `*_validation.go`.

yolocs: Also notice `MustParse` will also cause panic.

yolocs: The order of webhook is defaulting first then validation. So if you see an invalid value here, simply pass it along and let the validation webhook reject the request.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
yolocs: See my other comment. We should rely on validation webhook. Auto correcting values provided by users would be confusing.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
Ectelion: A bit more context here: this was used in a 'code contract' way, i.e. the methods were structured such that this condition is not triggerable externally and was rather meant to set input expectations internally. I see that there are more suitable practices here and the latest updates should be more aligned with those.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
Ectelion: Thanks, the re-factored version now has a structure with groupings similar to what's suggested above, with slight variation: to make the data contract match the corev1 requests and limits structure.

pkg/apis/intevents/v1alpha1/brokercell_types.go
```diff
@@ -68,9 +68,33 @@ var (
...
+}
+
+// ResourceSpecification defines requests and limits for the system resources
+type ResourceSpecification struct {
+	// Request specifies the minimal amount of the resource for the deployment to be schedulable
+	Requests SystemResource `json:"requests,omitempty"`
+
+	// Limit specifies the maximal amount of the resource that the deployment is allowed to consume
+	Limits SystemResource `json:"limits,omitempty"`
+}
```
yolocs, line 84: Sorry if I'm appearing to be back and forth. I had a second thought on the defaulting. I found the current logic could still be confusing. E.g. If I (as a user) set the CPU requirements but not memory, then it will use the default memory requirements. But such defaulting could somehow "invalidate" my CPU requirements as the memory defaults could be much lower than my expectation.

What if we do it with a more blunt approach? Either you (as a user) don't specify the `ComponentParameters` and we default everything for you, or you need to configure the whole `ComponentParameters`. In the latter case, it doesn't mean you have to specify everything in `ComponentParameters`. It just means we won't do any defaulting in that case and will use whatever value you provide, e.g. if memory requirement is missing, then we don't specify memory requirement in the deployment.

We would still need to do proper validation.

Thoughts?

pkg/apis/intevents/v1alpha1/brokercell_validation.go
```diff
@@ -18,12 +18,119 @@ package v1alpha1
...
+	resourceutil "github.com/google/knative-gcp/pkg/utils/resource"
+	"k8s.io/apimachinery/pkg/api/resource"
 	"knative.dev/pkg/apis"
 )
 
+const (
+	avgMemoryUsageOutOfRangeErrorDetail  string = "AvgMemoryUsage should not exceed the memory limit"
+	resourceRequestOutOfRangeErrorDetail string = "Resource request should not exceed the resource limit"
+	unexpectedQuantityFormatErrorDetail  string = "The quantity is specified in an unexpected format"
+)
```
yolocs, line 32: nit: IMO it's pretty rare in Go to define constants like these. Usually for strings that are only used once, my preference is to make them inline.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -35,22 +63,81 @@ func (bc *BrokerCell) SetDefaults(ctx context.Context) {
...
+// SetAutoScalingDefaults sets the autoscaling-related default field values for ComponentParameters.
+func (componentParams *ComponentParameters) SetAutoScalingDefaults(targetMemoryUsageCoefficient float64, avgCPUUtilization int32) {
+	if componentParams.MinReplicas == nil {
+		componentParams.MinReplicas = ptr.Int32(minReplicas)
 	}
-	if bcs.Components.Retry.MaxReplicas == nil {
-		bcs.Components.Retry.MaxReplicas = ptr.Int32(maxReplicas)
+	if componentParams.MaxReplicas == nil {
+		componentParams.MaxReplicas = ptr.Int32(maxReplicas)
 	}
```
yolocs, line 123: I know this is existing logic. But it has the same problem - if I (as a user) only set the min, it's possible my min is greater than the default max, which is wrong. As I mentioned in the other comment, if we have a HPA spec, then we can easily tell if a value is explicitly left blank by the user or not.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -35,22 +63,81 @@ func (bc *BrokerCell) SetDefaults(ctx context.Context) {
...
+	if componentParams.AvgMemoryUsage == nil {
+		anchorValue := ""
+		if componentParams.Resources.Limits.Memory != nil {
+			anchorValue = *componentParams.Resources.Limits.Memory
+		} else if componentParams.Resources.Requests.Memory != nil {
+			anchorValue = *componentParams.Resources.Requests.Memory
+		}
+		if len(anchorValue) > 0 {
+			memoryAnchorQuantity, parsingError := resource.ParseQuantity(anchorValue)
+			if parsingError == nil {
```
yolocs, line 137: nit: just use `err`. Go usually prefers less verbose names.

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -35,22 +63,81 @@ func (bc *BrokerCell) SetDefaults(ctx context.Context) {
...
-		bcs.Components.Ingress.MaxReplicas = ptr.Int32(maxReplicas)
+	// If target average consumption for the auto-scaler is not explicitly specified, default it
+	if componentParams.AvgMemoryUsage == nil {
+		anchorValue := ""
+		if componentParams.Resources.Limits.Memory != nil {
+			anchorValue = *componentParams.Resources.Limits.Memory
+		} else if componentParams.Resources.Requests.Memory != nil {
+			anchorValue = *componentParams.Resources.Requests.Memory
+		}
+		if len(anchorValue) > 0 {
```
yolocs, line 135: nit: `anchoValue == ""`

pkg/utils/resource/resource.go
```diff
@@ -0,0 +1,83 @@
...
+	return resource.NewQuantity(rawResult, quantity.Format)
+}
+
+// BuildResourceRequirements constructs the core resource requirements structure based on specified resource requests and limits
+func BuildResourceRequirements(cpuRequest, cpuLimit, memoryRequest, memoryLimit string) corev1.ResourceRequirements {
+	resourceRequirements := corev1.ResourceRequirements{
+		Limits:   corev1.ResourceList{},
+		Requests: corev1.ResourceList{},
+	}
+	if len(cpuRequest) > 0 {
```
yolocs, line 39: `== ""`. Similarly for others.

pkg/utils/resource/resource.go
```diff
@@ -0,0 +1,83 @@
...
+		cpuRequestQuantity, err := resource.ParseQuantity(cpuRequest)
+		if err == nil {
+			resourceRequirements.Requests[corev1.ResourceCPU] = cpuRequestQuantity
+		}
```
yolocs, line 43: Could do one-liner.

```
if q, err := resource.ParseQuantity(cpuRequest); err == nil {
  resourceRequirements.Requests[corev1.ResourceCPU] = q
}
```

pkg/utils/resource/resource.go
```diff
@@ -0,0 +1,83 @@
...
+
+func IsValidQuantityValue(value string) bool {
+	if len(value) == 0 {
+		return true
+	}
+	_, parsingErr := resource.ParseQuantity(value)
+	if parsingErr != nil {
+		return false
+	}
+	return true
```
yolocs, line 82: Use `err`.

And `return err == nil`

pkg/apis/intevents/v1alpha1/brokercell_types.go
Ectelion: Thanks for following up! That would indeed have been a simpler approach and I think that’s the approach I was inclined to choose in the beginning. Since some considerable amount of development and testing effort was put into the current approach, I believe it’s best to keep with it, at least for now. I’ve been playing with it and believe that current version is actually pretty usable even for external users. The defaulting result will be visible on the next “kubectl edit” of BrokerCell, which I think makes it pretty discoverable. Documenting the process would further assist the experience. But at this point, it is not even clear if we want to expose this to external users, as the primary goal was to assist with the perf efforts. If we eventually decide to expose it externally and determine that the UX can be improved, we can have a follow up work item for that.

pkg/apis/intevents/v1alpha1/brokercell_validation.go
Ectelion: Some of these are used more than once (for both CPU and memory) and all of them are referred to from unit tests. Is it better to duplicate strings or any other good practice for that?

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
Ectelion: I've noticed that and thought it might be best not to change the existing structure -- to keep changes focused (already quite many) and, AFAIK, there is an ongoing work that is dependent on current format in perf/test area. But I have added a validation to account for possible min-max inconsistency.

pkg/apis/intevents/v1alpha1/brokercell_types.go
Ectelion: As for the rest of suggestions, have accounted for them in the latest update.

Ectelion: FYI, I have logged a task for the suggested approach: #1603. I'll follow up on this.

PR #1510, opened 2020-07-23
Sum data points in case there are multiple in metrics e2e test

liu-cong:
Fixes #

## Proposed Changes

- Previously the test only looks at the first data point. But in case there are multiple, we should sum them.
-
-

**Release Note**

```release-note
NONE
```

**Docs**

### DIFF
--- a/test/e2e/lib/lifecycle.go
+++ b/test/e2e/lib/lifecycle.go
@@ -206,6 +206,7 @@ func (c *Client) LogsFor(namespace, name string, tm *metav1.TypeMeta) (string, e
 }
 
 // TODO make this function more generic.
+// The caller is expected to provide a valid filter that returns exact one time series.
 func (c *Client) StackDriverEventCountMetricFor(namespace, projectID, filter string) (int64, error) {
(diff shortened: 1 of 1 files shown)


test/e2e/lib/lifecycle.go
```diff
@@ -206,6 +206,7 @@ func (c *Client) LogsFor(namespace, name string, tm *metav1.TypeMeta) (string, e
 }
 
 // TODO make this function more generic.
+// The caller is expected to provide a valid filter that returns exact one time series.
```
Harwayne, line 209:
```suggestion
// The caller is expected to provide a valid filter that returns exactly one time series.
```

nachocano:
Folks, feel free to remove/change all that old code. It was something I did before I went on leave without much idea of what I was doing. All the tests using this were skipped because they weren't working on PROW at the time. So this might be pretty buggy actually... 
I thought @ian-mi made a nice refactor and allowed for better assertions of metrics work in other PRs (I think @zargarpur already used that), so I'd think we should try to leverage that if possible.

PR #1514, opened 2020-07-24
Use transformer to avoid making unnecessary event copy in delivery transformer

### LINKED ISSUE #1511: Delivery processor event.Clone consumes too much unnecessary memroy
Problematic code:

https://github.com/google/knative-gcp/blob/73a3d4a3ca48aad0d7b6c2345012f59c203677d9/pkg/broker/handler/processors/deliver/processor.go#L106

Revealed from profiling, `event.Clone()` dominates the memory consumption (which potentially causes a serials of stability issues).

With clone.
with_clone

Without clone.
without_clone

This `event.Clone()` could be avoided with an alternative way to handle event hops.

ian-mi:
When sending to a retry topic it is still necessary to clone EventContext since the retry client
does not yet use message bindings.

Fixes #1511

Benchmark results show significant improvement for large payload sizes since this avoids buffering payload data:
> DeliveryNoReply/0_bytes-16                 9.68µs ± 1%  9.26µs ± 2%   -4.28%  (p=0.000 n=9+10)
> DeliveryNoReply/1000_bytes-16              10.4µs ± 2%   9.5µs ± 1%   -8.42%  (p=0.000 n=9+10)
> DeliveryNoReply/1000000_bytes-16            294µs ± 2%   146µs ± 0%  -50.28%  (p=0.000 n=10+9)
> DeliveryNoReplyFakeClient/0_bytes-16       2.63µs ± 1%  2.45µs ± 1%   -6.70%  (p=0.000 n=8+10)
> DeliveryNoReplyFakeClient/1000_bytes-16    2.81µs ± 3%  2.47µs ± 1%  -12.14%  (p=0.000 n=10+10)
> DeliveryNoReplyFakeClient/100000_bytes-16  14.9µs ± 1%   2.5µs ± 2%  -83.51%  (p=0.000 n=10+9)

### DIFF
--- a/pkg/broker/eventutil/hops.go
+++ b/pkg/broker/eventutil/hops.go
@@ -29,7 +29,7 @@ import (
 const (
 	// Intentionally make it short because the additional attribute
 	// increases Pubsub message size and could incur additional cost.
-	hopsAttribute = "kgcphops"
+	HopsAttribute = "kgcphops"
 )
 
 // UpdateRemainingHops update an event with proper remaining hops.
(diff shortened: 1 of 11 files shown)


pkg/broker/eventutil/hops.go
```diff
@@ -48,39 +48,39 @@ func UpdateRemainingHops(ctx context.Context, event *event.Event, preemptiveHops
...
-	event.SetExtension(hopsAttribute, hops)
+	event.SetExtension(HopsAttribute, hops)
 }
 
 // SetRemainingHops sets the remaining hops in the event.
 // It ignores any existing hops value.
 func SetRemainingHops(ctx context.Context, event *event.Event, hops int32) {
-	event.SetExtension(hopsAttribute, hops)
+	event.SetExtension(HopsAttribute, hops)
 }
```
yolocs, line 63: I think this can be removed now.

ian-mi:
Event messages are not immutable so I had to introduce an immutable version to avoid the event copy. This happened to expose a bug in the HTTP message implementation which I fixed upstream in https://github.com/cloudevents/sdk-go/pull/557.

pkg/broker/eventutil/immutable_event_message.go
```diff
@@ -0,0 +1,37 @@
...
+	*binding.EventMessage
+}
+
+func NewImmutableEventMessage(e *event.Event) ImmutableEventMessage {
+	return ImmutableEventMessage{(*binding.EventMessage)(e)}
+}
+
+func (m ImmutableEventMessage) ReadEncoding() binding.Encoding {
+	return binding.EncodingBinary
+}
```
yolocs, line 37: What if you don't have this? What would happen?

ian-mi: If ReadEncoding is not overridden it will return EncodingEvent which will cause message writes to first convert the message to an event. For EventMessages this would cause the original event to be unwrapped and potentially mutated in-place. Since ImmutableEventMessage is a new type this will instead cause an unnecessary event copy to be made. To prevent this from happening ImmutableEventMessage forces binary encoding to be used.

yolocs: Could you put a comment above about this? Thanks.

PR #1523, opened 2020-07-28
Add immutable fields check in sources validation

bharattkukreja:
Fixes https://github.com/google/knative-gcp/issues/1448

## Proposed Changes

- Call CheckImmutableFields inside Validate() in sources validation to make sure only mutable fields can be modified with kubectl edit.

### DIFF
--- a/pkg/apis/events/v1/cloudauditlogssource_validation.go
+++ b/pkg/apis/events/v1/cloudauditlogssource_validation.go
@@ -29,7 +29,13 @@ import (
 )
 
 func (current *CloudAuditLogsSource) Validate(ctx context.Context) *apis.FieldError {
-	return current.Spec.Validate(ctx).ViaField("spec")
+	err := current.Spec.Validate(ctx).ViaField("spec")
+
+	if apis.IsInUpdate(ctx) {
+		original := apis.GetBaseline(ctx).(*CloudAuditLogsSource)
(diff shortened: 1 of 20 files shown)


danyinggu:
I think something is missing: 1. You need to add this to all of the versions of sources but not only for v1. 2. You need to add that to topic and pullSubscription Validation as well.

bharattkukreja:
> I think something is missing: 1. You need to add this to all of the versions of sources but not only for v1. 2. You need to add that to topic and pullSubscription Validation as well.

I did make the changes for all versions of sources (v1, v1alpha1 and v1beta1). I'll add the check to topic and pullsubscription as well

danyinggu:
> > I think something is missing: 1. You need to add this to all of the versions of sources but not only for v1. 2. You need to add that to topic and pullSubscription Validation as well.
> 
> I did make the changes for all versions of sources (v1, v1alpha1 and v1beta1). I'll add the check to topic and pullsubscription as well.

Sorry, my bad. You need to add the change to CloudStorageSource v1 as well.

nachocano:
@bharattkukreja can you add the same for Channel in a separate PR? and maybe Broker as well, if it's not there?

bharattkukreja:
> @bharattkukreja can you add the same for Channel in a separate PR? and maybe Broker as well, if it's not there?

Yep, I'll make a separate PR for Channel and Broker.

PR #1535, opened 2020-07-30
Add Handler Benchmark and Update Cloudevents SDK

ian-mi:
Pulls in https://github.com/cloudevents/sdk-go/pull/561 which eliminates the copying of message data when converting a binary pubsub message to an event. The newly added benchmark shows a performance improvement in the broker handler with large payload sizes and the performance becomes nearly entirely dominated by the overhead of gRPC.

### DIFF
--- a/pkg/broker/handler/handler_test.go
+++ b/pkg/broker/handler/handler_test.go
@@ -18,6 +18,8 @@ package handler
 
 import (
 	"context"
+	"fmt"
+	"runtime"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


pkg/broker/handler/handler_test.go
```diff
@@ -156,6 +160,64 @@ func TestHandler(t *testing.T) {
...
+	msg := new(pubsub.Message)
+	if err := cepubsub.WritePubSubMessage(ctx, binding.ToMessage(testEvent), msg); err != nil {
+		b.Fatal(err)
+	}
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		processor.processed.Acquire(ctx, 1)
+		topic.Publish(ctx, msg)
+	}
+	processor.processed.Acquire(ctx, maxMsgs)
```
grantr, line 218: IIUC, the semaphore is limiting the maximum number of outstanding messages to `2*cores`, and waiting until the topic is empty before moving on to the next bench run. I'd love to see a quick overview of the theory behind this in a comment. Why `2*cores` as the max concurrency? Is that chosen to normalize resource differences between environments?

ian-mi: It wasn't chosen very scientifically, I wanted to choose a number which would buffer enough messages to keep all threads busy while ensuring that the memory footprint used to store published messages does not grow with b.N.

PR #1536, opened 2020-07-30
Implement the Broker delivery spec

### LINKED ISSUE #1072: Implement broker retry spec
**Problem**
There are known cases like: if a Knative function was throwing 500 errors, it will infinitely retry until the message expired from the topic, and users can’t drop it during that time, causing the log full of this error message.

In such case, Broker retry needs to figure out ways to handle it gracefully.

Possible Solutions:
1. Rely on cloud event SDK to handle response
2. Check if response code is retriable or not (difference between 400 and 500)
3. Clear application guidance for users to return error code

We can also check deadLetterQueue design in Knative Eventing for more ideas.

**Exit Criteria**

**Time Estimate (optional):** 

tommyreddad:
Fixes #1072 

## Proposed Changes

- Specify the Broker retry and dead letter policies through the Broker delivery spec
- Implement the Broker retry and dead letter policies by translating them to pubsub policies
- Set default backoff retry policy in the Broker SetDefaults


go.mod
```diff
@@ -16,6 +16,7 @@ require (
 	github.com/google/wire v0.4.0
 	github.com/googleapis/gax-go/v2 v2.0.5
 	github.com/kelseyhightower/envconfig v1.4.0
+	github.com/rickb777/date v1.13.0
```
grantr, line 19: Is this library also the one used by Eventing for parsing these durations?

pkg/apis/broker/v1beta1/broker_defaults.go
```diff
@@ -18,10 +18,29 @@ package v1beta1
...
+var (
+	// backoffDelay is the default backoff delay used in the backoff retry policy
+	// for the Broker delivery spec.
+	backoffDelay = "PT1S"
+	// backoffPolicy is the default backoff policy type used in the backoff retry
+	// policy for the Broker delivery spec.
+	backoffPolicy = eventingduckv1beta1.BackoffPolicyExponential
+	// retry is the default number of maximum delivery attempts for unacked messages
+	// before they are sent to a dead letter topic in the Broker delivery spec.
+	retry int32 = 6
```
grantr, line 34: We should only set a default retry if a dead letter sink is present. Without a dead letter sink, the default retry count should be infinite.

pkg/apis/broker/v1beta1/broker_defaults.go
```diff
@@ -18,10 +18,29 @@ package v1beta1
...
 
 // SetDefaults sets the default field values for a Broker.
 func (b *Broker) SetDefaults(ctx context.Context) {
-	// The Google Cloud Broker doesn't have any custom defaults. The
-	// eventing webhook will add the usual defaults.
+	// Default delivery spec fields.
+	b.Spec.Delivery = &eventingduckv1beta1.DeliverySpec{
+		BackoffDelay:  &backoffDelay,
+		BackoffPolicy: &backoffPolicy,
+		Retry:         &retry,
```
grantr, line 43: As above, only default this when dead letter queue is set. Alternately, reject the update if dead letter sink is set and retry is not set. What does the pubsub api do in this case?

pkg/reconciler/trigger/controller.go
```diff
@@ -60,9 +59,6 @@ var filterBroker = pkgreconciler.AnnotationFilterFunc(eventingv1beta1.BrokerClas
 
 type envConfig struct {
 	ProjectID string `envconfig:"PROJECT_ID"`
-
-	MinRetryBackoff time.Duration `envconfig:"MIN_RETRY_BACKOFF" default:"1s"`
-	MaxRetryBackoff time.Duration `envconfig:"MAX_RETRY_BACKOFF" default:"1m"`
```
grantr, line 65: We may want to keep these as defaults for brokers that don't have values for delivery spec fields, though they don't have to be env vars. They could be moved to constants.

pkg/reconciler/trigger/trigger.go
```diff
@@ -236,6 +252,71 @@ func (r *Reconciler) reconcileRetryTopicAndSubscription(ctx context.Context, tri
 	return nil
 }
 
+// getPubsubRetryPolicy gets the eventing retry policy from the Broker delivery
+// spec and translates it to a pubsub retry policy.
+func getPubsubRetryPolicy(spec *eventingduckv1beta1.DeliverySpec) (*pubsub.RetryPolicy, error) {
+	if spec == nil || spec.BackoffDelay == nil || spec.BackoffPolicy == nil {
```
grantr, line 258: Subscriptions should always have a retry policy. If the delivery spec is empty, we should revert to the default retry policy. This will be the case for any brokers that were created before the webhook defaulting was in place.

pkg/reconciler/trigger/trigger.go
```diff
@@ -236,6 +252,71 @@ func (r *Reconciler) reconcileRetryTopicAndSubscription(ctx context.Context, tri
...
+		return nil, nil
+	}
+	// The Broker delivery spec is translated to a pubsub retry policy in the
+	// manner defined in the following post:
+	// https://github.com/google/knative-gcp/issues/1392#issuecomment-655617873
```
grantr, line 263: This should probably be documented somewhere in the docs/ directory as well.

pkg/reconciler/trigger/trigger.go
```diff
@@ -236,6 +252,71 @@ func (r *Reconciler) reconcileRetryTopicAndSubscription(ctx context.Context, tri
...
+	}
+	if scheme := sink.URI.Scheme; scheme != "pubsub" {
+		return "", fmt.Errorf("Dead letter sink URI scheme should be pubsub, instead is %v", scheme)
+	}
+	topicID := sink.URI.Host
+	if topicID == "" {
+		return "", fmt.Errorf("Dead letter topic must not be empty")
+	}
+	if len(topicID) > 255 {
+		return "", fmt.Errorf("Dead letter topic maximum length is 255 characters: %v", topicID)
```
grantr, line 295: These validations should live in pkg/apis/broker/v1beta1/broker_validation.go. That way they'll be part of the webhook and give the user immediate feedback.

Generally, if the object has invalid field values that will always be invalid, we should reject that in the webhook instead of letting it be created (and never recovering).

go.mod
tommyreddad: Yes:
- https://github.com/knative/eventing/blob/master/go.mod#L23
- https://github.com/knative/eventing/blob/master/pkg/apis/duck/v1/delivery_types.go#L72

pkg/apis/broker/v1beta1/broker_defaults.go
```diff
@@ -18,10 +18,29 @@ package v1beta1
...
+	// before they are sent to a dead letter topic in the Broker delivery spec.
+	retry int32 = 6
 )
 
 // SetDefaults sets the default field values for a Broker.
 func (b *Broker) SetDefaults(ctx context.Context) {
-	// The Google Cloud Broker doesn't have any custom defaults. The
-	// eventing webhook will add the usual defaults.
+	// Default delivery spec fields.
+	b.Spec.Delivery = &eventingduckv1beta1.DeliverySpec{
```
yolocs, line 40: You should also check if `b.Spec.Delivery` is nil/empty before setting default?

pkg/apis/broker/v1beta1/broker_defaults.go
yolocs: +1

pkg/apis/broker/v1beta1/broker_defaults.go
```diff
@@ -18,10 +18,29 @@ package v1beta1
...
+		BackoffDelay:  &backoffDelay,
+		BackoffPolicy: &backoffPolicy,
+		Retry:         &retry,
+	}
+	// Besides this, the eventing webhook will add the usual defaults.
```
yolocs, line 45: Do you think we need additional validation (in addition to the validation in eventing core)?

pkg/reconciler/trigger/controller.go
yolocs: +1

pkg/apis/broker/v1beta1/broker_defaults.go
tommyreddad: Done. Currently, the defaulting is split between the retry backoff policy (BackoffDelay, BackoffPolicy) and the dead letter policy (DeadLetterSink, Retry). It could also be handled in a few other ways, but I felt this way was natural.

pkg/apis/broker/v1beta1/broker_defaults.go
tommyreddad: Agreed. I added dead letter sink validation.

pkg/reconciler/trigger/trigger.go
tommyreddad: I added a docs file to explain this: https://github.com/google/knative-gcp/pull/1536/files#diff-86642e88be266e549139dbe5b09c2d8d

pkg/apis/broker/v1beta1/broker_validation.go
```diff
@@ -19,12 +19,42 @@ package v1beta1
...
+	duckv1 "knative.dev/pkg/apis/duck/v1"
 )
 
 // Validate verifies that the Broker is valid.
 func (b *Broker) Validate(ctx context.Context) *apis.FieldError {
-	// The Google Cloud Broker doesn't have any custom validations. The
-	// eventing webhook will run the usual validations.
+	// We validate the GCP Broker's delivery spec. The eventing webhook will run
+	// the other usual validations.
+	withNS := apis.AllowDifferentNamespace(apis.WithinParent(ctx, b.ObjectMeta))
```
yolocs, line 31: nit: doesn't seem needed.

pkg/apis/broker/v1beta1/broker_defaults.go
```diff
@@ -18,10 +18,40 @@ package v1beta1
...
+	if b.Spec.Delivery == nil {
+		b.Spec.Delivery = &eventingduckv1beta1.DeliverySpec{}
+	}
+	if b.Spec.Delivery.BackoffPolicy == nil &&
+		b.Spec.Delivery.BackoffDelay == nil {
+		b.Spec.Delivery.BackoffPolicy = &DefaultBackoffPolicy
+		b.Spec.Delivery.BackoffDelay = &DefaultBackoffDelay
+	}
+	if b.Spec.Delivery.Retry == nil &&
+		(b.Spec.Delivery.DeadLetterSink != nil && !equality.Semantic.DeepEqual(b.Spec.Delivery.DeadLetterSink, &duckv1.Destination{})) {
```
yolocs, line 53: IIUC the second check `!equality.Semantic.DeepEqual(b.Spec.Delivery.DeadLetterSink, &duckv1.Destination{})` is not needed. Because the validation should reject empty sink.

pkg/apis/broker/v1beta1/broker_defaults.go
```diff
@@ -18,10 +18,40 @@ package v1beta1
...
-	// eventing webhook will add the usual defaults.
+	// Set the default delivery spec.
+	if b.Spec.Delivery == nil {
+		b.Spec.Delivery = &eventingduckv1beta1.DeliverySpec{}
+	}
+	if b.Spec.Delivery.BackoffPolicy == nil &&
+		b.Spec.Delivery.BackoffDelay == nil {
+		b.Spec.Delivery.BackoffPolicy = &DefaultBackoffPolicy
+		b.Spec.Delivery.BackoffDelay = &DefaultBackoffDelay
+	}
```
yolocs, line 51: Just to confirm. This means we will always set delivery spec with backoff policy, right?

pkg/reconciler/trigger/trigger.go
```diff
@@ -236,6 +252,55 @@ func (r *Reconciler) reconcileRetryTopicAndSubscription(ctx context.Context, tri
...
 }
 
+// getPubsubRetryPolicy gets the eventing retry policy from the Broker delivery
+// spec and translates it to a pubsub retry policy.
+func getPubsubRetryPolicy(spec *eventingduckv1beta1.DeliverySpec) *pubsub.RetryPolicy {
+	var backoffDelay *string
+	var backoffPolicy *eventingduckv1beta1.BackoffPolicyType
+	if spec == nil {
+		backoffDelay = &brokerv1beta1.DefaultBackoffDelay
+		backoffPolicy = &brokerv1beta1.DefaultBackoffPolicy
```
yolocs, line 262: From the defaulting logic, shouldn't the spec always present?

pkg/apis/broker/v1beta1/broker_defaults.go
tommyreddad: Yes. If the user doesn't give any specific backoff policy settings, then the backoff policy will be defaulted, and vice-versa.

pkg/reconciler/trigger/trigger.go
tommyreddad: I'm not sure it is guaranteed if the webhook fails for some reason.

pkg/apis/broker/v1beta1/broker_defaults.go
tommyreddad: I think you're right, I'll fix that

pkg/reconciler/trigger/trigger.go
yolocs: In case of webhook not available, the requests would fail.
The code is fine if you prefer the current way.

PR #1538, opened 2020-07-31
 Mark status of sub-conditions in finalizer of sources

### LINKED ISSUE #1294: Marking source and channel sub-conditions in Finalizer
**Problem**
With https://github.com/google/knative-gcp/issues/1190 and https://github.com/knative/eventing/issues/3305, it would be nice if we mark source and channel sub-conditions when resource deletion returns error.

**Exit Criteria**

**Time Estimate (optional):** 

danyinggu:
Fixes #1294.

## Proposed Changes

- Mark status of subCondition in finalizer of sources. 
Generally, the idea is:
  - If the deletion is successful, do not mark the status as the discussion in the PR. (This is consistent with what the broker does)
  - If the deletion fails for some reason, mark the subCondition as Unknown and use the reason why it fails as the message.
 - This logic is consistent with the Broker finalizer. 
 - Will update the inconsistencies in PubSub in a separate PR.
- Update UTs based on the idea.
- Fix some nits. 
- It seems that there is no other sub-conditions other than WI in channel, where WI already marks sub-conditions in finalizer.

**Release Note**

```release-note
NONE
```

**Docs**


pkg/reconciler/events/auditlogs/auditlogs.go
```diff
@@ -171,16 +171,20 @@ func (c *Reconciler) ensureSinkIsPublisher(ctx context.Context, s *v1.CloudAudit
...
 		logging.FromContext(ctx).Desugar().Error("Failed to create LogAdmin client", zap.Error(err))
+		s.Status.MarkSinkNotReady(deleteSinkFailed, "Failed to create LogAdmin Client: %s", err.Error())
 		return err
 	}
-	if err = logadminClient.DeleteSink(ctx, s.Status.StackdriverSink); status.Code(err) != codes.NotFound {
+	if err = logadminClient.DeleteSink(ctx, s.Status.StackdriverSink); err != nil && status.Code(err) != codes.NotFound {
+		s.Status.MarkSinkNotReady(deleteSinkFailed, "Failed to delete Stackdriver sink: %s", err.Error())
 		return err
 	}
+	s.Status.MarkSinkNotReady("SinkDeleted", "Sink successfully deleted: %s", s.Status.StackdriverSink)
```
nachocano, line 187: I'm not entirely sure we should be setting this in the success case, as we will try to update the status of a deleted resource, which will probably give some error...
can you confirm in the controller logs that this doesn't give an error?

nachocano: if we decide not to, we should remove it in other places as well

capri-xiyue: I think we should make the behavior consistent.
When @grac3gao marked the sub-conditions in finalizer of broker, https://github.com/google/knative-gcp/pull/1295/files, we won't mark as Failed or NotReady when we successfully delete the resource.
If broker and pubSub finalizer are not consistent, we should try to reach an agreement and make it consistent among all resources.

capri-xiyue: @nachocano 
I checked this issue https://github.com/knative/eventing/issues/3305
It looks like `it was confirmed that we can update the status of a resource that is being deleted.`
@grac3gao  Can you help confirm that we can update the status of a resource that is being deleted?

nachocano: Yes, we are doing so when it fails to delete. But I wasn't sure whether we should also do it when the deletion succeeds. Might not hurt, but I thought it was not needed.

capri-xiyue: @nachocano I agree it is not needed to mark the condition when the deletion succeeds if both deletion and creation are using the same condition type. Marking it failed when the deletion succeeds is kind of confusing. 
My point is that we should make the mark behavior consistent across different resources.

grac3gao: In Broker, we didn't mark the condition if the deletion succeeds (we only mark the condition to provide information if the deletion failed for some reason, and it is confirmed that we can update the condition in this case). IMO, If the deletion succeeds, then the resource will just removed from the namespace, user doesn't need (and they are not interested) to know the condition. Only deletion failure will cause the resource hang in the namespace, and we need to let the resource communicate to user when the users are looking at object's condition.

pkg/reconciler/events/auditlogs/auditlogs.go
```diff
@@ -171,16 +171,20 @@ func (c *Reconciler) ensureSinkIsPublisher(ctx context.Context, s *v1.CloudAudit
 // previously created stackdriver sink.
 func (c *Reconciler) deleteSink(ctx context.Context, s *v1.CloudAuditLogsSource) error {
 	if s.Status.StackdriverSink == "" {
+		s.Status.MarkSinkNotReady("SinkFinalized", "No sink to delete.")
 		return nil
 	}
 	logadminClient, err := c.logadminClientProvider(ctx, s.Status.ProjectID)
 	if err != nil {
 		logging.FromContext(ctx).Desugar().Error("Failed to create LogAdmin client", zap.Error(err))
+		s.Status.MarkSinkNotReady(deleteSinkFailed, "Failed to create LogAdmin Client: %s", err.Error())
```
grac3gao, line 180: Noticed for all the cases, we mark sub-condition NotReady (MarkFalse) in this PR. I am not sure if in some cases, marking Unknown with some Reason might be more proper? @nachocano . In the broker, most deletion failure cases are marked Unknown, for example, when the notification deletion failed for Reason `PubSubClientCreationFailed`, it is marked as `			b.Status.MarkSubscriptionUnknown("FinalizeSubscriptionPubSubClientCreationFailed", "Failed to create Pub/Sub client: %w", err)
` (https://github.com/google/knative-gcp/pull/1295/files)

The reason for that is, when the deletion failed, we are not sure whether this notification still work or not (it might still work, as we didn't delete it, so it will keep working as it is. It might not work, as failure to create pubsub client may indicate some credential problem, which will also cause the notification stop working). Because of this uncertainty, we choose to use MarkUnknown instead of MarkFalse.

pkg/reconciler/events/auditlogs/auditlogs.go
danyinggu: @nachocano I checked the controller logs that it doesn't gave an error regarding the finalizer. @capri-xiyue Yes, the PubSub deletion also marks Pubsub Fail when the deletion succeed. We should keep everything consistent.
I agree with all of your idea that marking resources NotReady or Fail is not needed when the deletion is successful, so I will remove it here. I will create a separate issue to make PubSub consistent with this logic in a separate PR. Is that ok with all of you?

pkg/reconciler/events/auditlogs/auditlogs.go
danyinggu: @nachocano This is another inconsistent part between this current PR, PubSub, WorkloadIdentity and Broker. 
For example, in deleteWorkLoadIdentity:
https://github.com/google/knative-gcp/blob/master/pkg/reconciler/identity/reconciler.go#L75-L79`
	identityNames, err := i.getGoogleServiceAccountName(ctx, identifiable)
	if err != nil {
		logging.FromContext(ctx).Desugar().Error("failed to get Google service account name", zap.Error(err))
		status.MarkWorkloadIdentityFailed(identifiable.ConditionSet(), workloadIdentityFailed, err.Error())
		return nil, fmt.Errorf(`failed to get Google service account name: %w`, err)`
When it fails to get the GSA, the mark WorkloadIdentityFailed instead of marking it as Unknown. However, as Grace mentioned, in broker, we mark it as Unknown. 
We should make them consistent as well.

nachocano: I'm OK with marking things Unknown. The important thing in the WI part was to mark the overall status of the resources as Ready=False, as well. As the workloadIdentity condition is not part of the conditionSet, then we needed that extra thing. IIRC if you look at the method status.MarkWorkloadIdentityFailed, you will see that we were marking ready=False as well...
@grac3gao do you see any problem marking unknown here?

danyinggu: @capri-xiyue I'm also ok to mark things as Unknown. Just want to get your idea before I start. I just synced with Nacho offline, we agree to mark all these failure cases relate to deletion as Unknown instead of Fail/NotReady to make them consistent, for example, https://github.com/google/knative-gcp/blob/master/pkg/reconciler/intevents/reconciler.go#L267 should be changed to Unknown as well (confirmed with Nacho). Do you agree with it?

pkg/reconciler/events/auditlogs/auditlogs.go
nachocano: SGTM

pkg/reconciler/events/auditlogs/auditlogs.go
danyinggu: Just synced with Nacho offline, we decided to mark Unknown for all the cases of failure of Source and PubSub during deletion and do it. For the Workload Identity reconciler that Nacho commented, maybe waiting for @grac3gao?

pkg/reconciler/events/auditlogs/auditlogs.go
capri-xiyue: sg

pkg/reconciler/events/auditlogs/auditlogs.go
grac3gao: @nachocano I didn't see any problem, it was configured before we set the broker condition, we also need to change it to follow the broker rule for consistency. The only thing I was thinking about is if we mark the workload Identity unknown, what status should we mark Ready. I would imagine the rule might be marking Ready unknown if Ready == true before, and leave Ready as it is if Ready != true before. What do you think about this? @nachocano

nachocano:
/lgtm
/approve

Thanks @danyinggu

pkg/reconciler/events/auditlogs/auditlogs.go
nachocano: I think we can keep it simple: mark unknown here, and mark false the ready condition as we were previously doing. If that doesn't break anything, which I think it won't, I'd rather do that. WDYT?

grac3gao: I also think it doesn't break anything. However, directly mark Ready as false doesn't follow the MarkUnknown function logic showed here: https://github.com/knative/pkg/blob/f1e8ecbf78c3c1ab066c611871a6e2abff375324/apis/condition_set.go#L352-L362. I am not sure if we should keep the logic consistent when marking Workload Identity unknown, or it doesn't matter?

Issue #1475, opened 2020-07-16
Update CRDs to use `apiextensions.k8s.io/v1` 


danyinggu:
**Problem**
In [serving](https://github.com/knative/serving/tree/master/config/core/resources) and [eventing](https://github.com/knative/eventing/tree/master/config/core/resources), we already used `apiextensions.k8s.io/v1` as the CRD API version. We should eventually update all of our CRD to use `apiextensions.k8s.io/v1` as well. 

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
All of our CRDs of resources are using `apiextensions.k8s.io/v1` as well. 

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

cc @Harwayne @nachocano

Harwayne:
@danyinggu has this work started?

danyinggu:
> @danyinggu has this work started?

No It hasn't started yet. I think we can utilize your PR #919 to change it to use v1 template?

capri-xiyue:
Synced with @Harwayne  offline, Adam is ok if we use yaml merge to rewrite the yaml. So currently we can work on this issue first and Adam will work on that template change #919 later.
And here is the file of the PR that you can refer to https://github.com/google/knative-gcp/pull/1473/files#diff-03dcdd5a50d164008de2bb1c827d7cab regarding how to update CRDs to use `apiextensions.k8s.io/v1` and 
use yaml merge to rewrite the yaml.

Harwayne:
> > @danyinggu has this work started?
> 
> No It hasn't started yet. I think we can utilize your PR #919 to change it to use v1 template?

With the changes made in #1473, namely usage of `<<:` to inject the differences between versions, I think the CRD generation is less useful than it used to be. 

I think we are better off moving to `apiextensions.k8s.io/v1`, then if we still want the CRD generation in #919, we can make it match the handmade CRDs. 

If we end up having issues like incorrect schemas, then we can reprioritize and get the generation first. But until we actually encounter those problems, let's prioritize moving to `apiextensions.k8s.io/v1`.

danyinggu:
@nachocano @capri-xiyue I checked the code base, since I already update all sources' CRDs to use `apiextensions.k8s.io/v1` during the promotion of source to v1, the only resources that will need to be updated are Channel and BrokerCell.

PR #1539, opened 2020-07-31
Update brokercell and channel CRDs API version

bharattkukreja:
Fixes https://github.com/google/knative-gcp/issues/1475

## Proposed Changes

- Update brokercell and channel CRDs to use apiextensions.k8s.io/v1 as the API version

### DIFF
--- a/config/core/resources/channel.yaml
+++ b/config/core/resources/channel.yaml
@@ -50,7 +50,7 @@ spec:
           namespace: cloud-run-events
   versions:
   - &version
-    name: v1
+    name: v1alpha1
     served: true
     storage: false
     subresources:
@@ -68,21 +68,20 @@ spec:
     - name: Age
       type: date
       jsonPath: .metadata.creationTimestamp
-    # We remove status.ServiceAccountName from the schema in v1.
-    schema:
+    schema: &v1alpha1Schema
       openAPIV3Schema: &openAPIV3Schema
         type: object
         properties: &properties
           spec:
             type: object
             properties:
-              serviceAccountName:
+              serviceAccountName: &serviceAccountName
                 type: string
                 description: >
                   k8s service account used to bind to a google service account to poll the Cloud Pub/Sub Subscription.
                   The value of the k8s service account must be a valid DNS subdomain name.
                   (see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names)
-              secret:
+              secret: &secret
                 type: object
                 description: >
                   Credential to use to manage Cloud Pub/Sub. The value of the secret entry must be a service
@@ -96,7 +95,7 @@ spec:
                     type: string
                   optional:
                     type: boolean
-              project:
+              project: &project
                 type: string
                 description: >
                   ID of the Google Cloud Project to own the Pub/Sub credentials. E.g.
@@ -105,25 +104,25 @@ spec:
               subscribable:
                 type: object
                 properties:
-                  subscribers:
+                  subscribers: &subscribers
                     type: array
                     items:
                       type: object
                       required:
                         - uid
-                      properties:
-                        uid:
+                      properties: &subscribersProperties
+                        uid: &uid
                           type: string
                           minLength: 1
-                        generation:
+                        generation: &generation
                           type: integer
-                        subscriberURI:
+                        subscriberURI: &subscriberURI
                           type: string
-                        replyURI:
+                        replyURI: &replyURI
                           type: string
                         deadLetterSink:
                           type: string
-                        delivery:
+                        delivery: &delivery
                           type: object
                           properties:
                             deadLetterSink:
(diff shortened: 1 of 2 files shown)


bharattkukreja:
cc @capri-xiyue

capri-xiyue:
Besides changing the version, I think you also need to follow this PR https://github.com/google/knative-gcp/pull/1473/files#diff-03dcdd5a50d164008de2bb1c827d7cab to modify other things like adding `conversionReviewVersions: ["v1", "v1beta1"]`

capri-xiyue:
In addition, it will be better if you can use yaml merge to rewrite the channel yaml

capri-xiyue:
https://yaml.org/type/merge.html

config/core/resources/channel.yaml
```diff
@@ -36,293 +36,189 @@ spec:
...
+            <<: *status
+            properties:
+              <<: *statusProperties
+              serviceAccountName:
+                type: string
+  - <<: *version
+    name: v1beta1
+    served: true
+    storage: true
+    # the schema of v1beta1 is exactly the same as v1alpha1 schema
```
capri-xiyue, line 223: @Harwayne Is schema of channel v1alpha1 the same as channel v1beta1?

config/core/resources/channel.yaml
```diff
@@ -36,293 +36,189 @@ spec:
...
   scope: Namespaced
-  subresources:
-    status: {}
   preserveUnknownFields: false
   conversion:
     strategy: Webhook
-    webhookClientConfig:
+    # conversionReviewVersions indicates what ConversionReview versions are understood/preferred by the webhook.
+    # The first version in the list understood by the API server is sent to the webhook.
+    # The webhook must respond with a ConversionReview object in the same version it received.
```
capri-xiyue, line 44: I think you are missing something. The e2e failed with applying this yaml.  https://github.com/bharattkukreja/knative-gcp/blob/dee416986d74f3cf671ab2f7e25264b54c190c47/config/core/resources/cloudpubsubsource.yaml#L42

config/core/resources/channel.yaml
```diff
@@ -36,293 +36,190 @@ spec:
...
+      # The first version in the list understood by the API server is sent to the webhook.
+      # The webhook must respond with a ConversionReview object in the same version it received.
+      conversionReviewVersions: ["v1", "v1beta1"]
+      clientConfig:
+        service:
+          name: webhook
+          namespace: cloud-run-events
+  versions:
+  - &version
+    name: v1
```
capri-xiyue, line 53: Channel doesn't have v1

capri-xiyue:
I just found that yaml merge syntax can handle it elegantly with field removal or field addition or exact same field. But for field replacement like Bharatt did for channel in this PR, it looks like  the usage of `<<:` to inject the differences between versions doesn't reduce the needed efforts.
In addition, later when we want to deprecate older versions, the usage of  `<<:`  and `&` make things more complex when we have filed replacement scenario. It looks like in this case, CRD generation is still quite useful when we have kind of complex differences between versions. @Harwayne  What do you think of it?

danyinggu:
> I just found that yaml merge syntax can handle it elegantly with field removal or field addition or exact same field. But for field replacement like Bharatt did for channel in this PR, it looks like the usage of `<<:` to inject the differences between versions doesn't reduce the efforts.
> In addition, later when we want to deprecate older versions, the usage of `<<:` and `&` make things more complex when we have filed replacement scenario. It looks like in this case, CRD generation is still quite useful when we have kind of complex differences between versions. @Harwayne What do you think of it?

I slightly checked his PR. I think for channel CRD, maybe he could put v1beta1 which have less field as the first schema, then insert additional fields that v1alpha1 have as the second schema, maybe this will make the syntax cleaner?

bharattkukreja:
> > I just found that yaml merge syntax can handle it elegantly with field removal or field addition or exact same field. But for field replacement like Bharatt did for channel in this PR, it looks like the usage of `<<:` to inject the differences between versions doesn't reduce the efforts.
> > In addition, later when we want to deprecate older versions, the usage of `<<:` and `&` make things more complex when we have filed replacement scenario. It looks like in this case, CRD generation is still quite useful when we have kind of complex differences between versions. @Harwayne What do you think of it?
> 
> I slightly checked his PR. I think for channel CRD, maybe he could put v1beta1 which have less field as the first schema, then insert additional fields that v1alpha1 have as the second schema, maybe this will make the syntax cleaner?

So v1beta1 has a different set of properties. Instead of subscribable (which is there in v1alpha1 in spec.properties), v1beta1 has subscribers (which is a layer beneath subscribable). The same for deadLetterSink (which exists in v1alpha1 in subscribable.properties.subscribers.items.properties, but doesn't in v1beta1). Since the property keys are different, regardless of which one version is described, it will involve copying down the fields in the yaml merge.

bharattkukreja:
@capri-xiyue Integration tests are finally passing. Problem was that subscriberURI and replyURI from v1alpha1 in channel.yaml are changed to subscriberUri and replyUri in v1beta1 (different casing), which I had missed. Anyways, this is ready for review.

PR #1556, opened 2020-08-05
Update PubSub and WI finalizer sub-condition mark to make them consistent with source and broker

### LINKED ISSUE #1543: Refactor Pubsub Finalizers of mark sub-condition status
**Problem**
According to discussion PR #1538 - https://github.com/google/knative-gcp/pull/1538#discussion_r464479107, we decide to refactor some points to make the finalizer consistent between all resources:
1. Remove mark status as Fail or NotReady when the deletion of subCondition is successful, while current deletion of PubSub does this https://github.com/google/knative-gcp/blob/7bf16c6b11e72c295118cf8017e6fff882134b7a/pkg/reconciler/intevents/reconciler.go#L254. We should remove it.
2. Mark sub-condition as Unknown instead of Fail or NotReady when deletion fails, because we don't know if the sub-condition itself is ready or not when the deletion fails.

**Exit Criteria**
1. We don't mark PubSub Status as Fail if the deletion is successful.
2. Always mark status as Unknown for all cases upon a failed deletion.

**Time Estimate (optional):** 

cc @nachocano @capri-xiyue @grac3gao

danyinggu:
Fixes #1543 

## Proposed Changes
As discussed in PR #1538 
- Do not mark topic and pullSubscription sub-condition as Fail when deletion is successfully
- When deletion fails, mark the sub-condition as Unknown

**Release Note**

```release-note

```

**Docs**


pkg/apis/duck/v1/identity_lifecycle.go
```diff
@@ -33,3 +34,13 @@ func (s *IdentityStatus) MarkWorkloadIdentityFailed(cs *apis.ConditionSet, reaso
...
 	cs.Manage(s).MarkFalse(apis.ConditionReady, "WorkloadIdentityFailed", messageFormat, messageA...)
 }
+
+func (s *IdentityStatus) MarkWorkloadIdentityUnknown(cs *apis.ConditionSet, reason, messageFormat string, messageA ...interface{}) {
+	cs.Manage(s).MarkUnknown(IdentityConfigured, reason, messageFormat, messageA...)
+	// ConditionType IdentityConfigured is not included in apis.NewLivingConditionSet{}, so it is not counted for conditionReady.
+	// Set ConditionReady to be Unknown if the initial status of ConditionReady is Ready.
+	// If the initial status of ConditionReady is not Ready, we keep it as it.
+	if c := cs.Manage(s).GetCondition(apis.ConditionReady); c != nil && c.Status == corev1.ConditionTrue {
+		cs.Manage(s).MarkUnknown(apis.ConditionReady, "WorkloadIdentityUnknown", messageFormat, messageA...)
```
danyinggu, line 44: @nachocano @grac3gao There is an ongoing discussion https://github.com/google/knative-gcp/pull/1538#discussion_r465322151 about whether we should directly mark ConditionReady to be false or only mark it as Unknown when the initial condition of ConditionReady is Ready... 
Could you please check to see if this is the logic that you want, please let me know if this should be changed.

nachocano: Just synced with @grac3gao, we will do something slightly different. But basically what she said in her comment. 
Please sync with her and she can point you in the right direction

pkg/apis/duck/v1/identity_lifecycle.go
```diff
@@ -33,3 +34,13 @@ func (s *IdentityStatus) MarkWorkloadIdentityFailed(cs *apis.ConditionSet, reaso
 	// It will be counted for conditionReady only if it is failed.
 	cs.Manage(s).MarkFalse(apis.ConditionReady, "WorkloadIdentityFailed", messageFormat, messageA...)
 }
+
+func (s *IdentityStatus) MarkWorkloadIdentityUnknown(cs *apis.ConditionSet, reason, messageFormat string, messageA ...interface{}) {
+	cs.Manage(s).MarkUnknown(IdentityConfigured, reason, messageFormat, messageA...)
+	// ConditionType IdentityConfigured is not included in apis.NewLivingConditionSet{}, so it is not counted for conditionReady.
+	// Set ConditionReady to be Unknown if the initial status of ConditionReady is Ready.
+	// If the initial status of ConditionReady is not Ready, we keep it as it.
+	if c := cs.Manage(s).GetCondition(apis.ConditionReady); c != nil && c.Status == corev1.ConditionTrue {
```
grac3gao, line 43: You can probably directly use `c.IsTrue()` instead of  `c != nil && c.Status == corev1.ConditionTrue`

grac3gao: And 
```
func (s *IdentityStatus) MarkWorkloadIdentityNotConfigured(cs *apis.ConditionSet, reason, messageFormat string, messageA ...interface{}) {
	cs.Manage(s).MarkUnknown(IdentityConfigured, reason, messageFormat, messageA...)
}
``` 
can be removed, as it is duplicated with `MarkWorkloadIdentityUnknown`, and there is no other place use it except test.
By the way, could also you change `MarkWorkloadIdentityConfigured` to be `MarkWorkloadIdentityReady`, just make the style consistent.

pkg/reconciler/events/scheduler/scheduler_test.go
```diff
@@ -1266,8 +1266,8 @@ func TestAllCases(t *testing.T) {
 					reconcilertestingv1.WithCloudSchedulerSourceSchedule(onceAMinuteSchedule),
 					reconcilertestingv1.WithInitCloudSchedulerSourceConditions,
 					reconcilertestingv1.WithCloudSchedulerSourceJobReady(jobName),
-					reconcilertestingv1.WithCloudSchedulerSourceTopicFailed("TopicDeleted", fmt.Sprintf("Successfully deleted Topic: %s", schedulerName)),
-					reconcilertestingv1.WithCloudSchedulerSourcePullSubscriptionFailed("PullSubscriptionDeleted", fmt.Sprintf("Successfully deleted PullSubscription: %s", schedulerName)),
+					reconcilertestingv1.WithCloudSchedulerSourceTopicReady("", ""),
```
capri-xiyue, line 1269: Should we write something like `reconcilertestingv1.WithCloudSchedulerSourceTopicReady(testTopicID, testProject)`?

pkg/reconciler/events/storage/storage_test.go
```diff
@@ -1234,8 +1235,8 @@ func TestAllCases(t *testing.T) {
 					reconcilertestingv1.WithCloudStorageSourceSink(sinkGVK, sinkName),
 					reconcilertestingv1.WithCloudStorageSourceEventTypes([]string{schemasv1.CloudStorageObjectFinalizedEventType}),
 					reconcilertestingv1.WithCloudStorageSourceObjectMetaGeneration(generation),
-					reconcilertestingv1.WithCloudStorageSourceTopicFailed("TopicDeleted", fmt.Sprintf("Successfully deleted Topic: %s", storageName)),
-					reconcilertestingv1.WithCloudStorageSourcePullSubscriptionFailed("PullSubscriptionDeleted", fmt.Sprintf("Successfully deleted PullSubscription: %s", storageName)),
+					reconcilertestingv1.WithCloudStorageSourceTopicReady(""),
```
capri-xiyue, line 1238: similar like reconcilertestingv1.WithCloudStorageSourceTopicReady(testTopicID)?

pkg/reconciler/events/auditlogs/auditlogs_test.go
```diff
@@ -1398,8 +1398,8 @@ func TestAllCases(t *testing.T) {
 				v1.WithCloudAuditLogsSourceSink(sinkGVK, sinkName),
 				v1.WithInitCloudAuditLogsSourceConditions,
 				v1.WithCloudAuditLogsSourceSinkReady,
-				v1.WithCloudAuditLogsSourceTopicFailed("TopicDeleted", fmt.Sprintf("Successfully deleted Topic: %s", sourceName)),
-				v1.WithCloudAuditLogsSourcePullSubscriptionFailed("PullSubscriptionDeleted", fmt.Sprintf("Successfully deleted PullSubscription: %s", sourceName)),
+				v1.WithCloudAuditLogsSourceTopicReady(""),
```
capri-xiyue, line 1401: similar like "v1.WithCloudAuditLogsSourceTopicReady(testTopicID)"?

pkg/reconciler/events/scheduler/scheduler_test.go
danyinggu: No, because https://github.com/google/knative-gcp/blob/master/pkg/reconciler/intevents/reconciler.go#L271 it sets topicID and projectID to be empty when deleting the topic.

pkg/reconciler/events/storage/storage_test.go
danyinggu: Same as above.

pkg/reconciler/events/auditlogs/auditlogs_test.go
danyinggu: Same as above.

pkg/apis/duck/v1/identity_lifecycle.go
danyinggu: Sure.

pkg/reconciler/events/scheduler/scheduler_test.go
capri-xiyue: In this case, instead of using "WithCloudSchedulerSourceTopicReady", maybe write a new method like `WithCloudSchedulerSourceTopicDeleted()`which makes the topicID to be "".
`WithCloudSchedulerSourceTopicReady()` is kind of confusing.

pkg/reconciler/events/storage/storage_test.go
capri-xiyue: similar as above. Maybe write a new method like WithCloudStorageSourceTopicDeleted()

pkg/reconciler/events/auditlogs/auditlogs_test.go
capri-xiyue: similar as above. Maybe write a new method like WithCloudAuditLogsSourceTopicDeleted()

pkg/reconciler/events/scheduler/scheduler_test.go
danyinggu: Does that mean methods for `WithCloudSchedulerSourcePullSubscriptionDeleted()` is also needed? Otherwise it will be more confusing that the topic is deleted but pullSubscription is ready? I think it kind of going back to the question that my initial thought of marking status as not Ready when deletion happens successfully because I was trying to avoid this kind of confusion.

pkg/reconciler/identity/reconciler.go
```diff
@@ -116,7 +116,7 @@ func (i *Identity) DeleteWorkloadIdentity(ctx context.Context, projectID string,
 	identityNames, err := i.getGoogleServiceAccountName(ctx, identifiable)
 	if err != nil {
 		logging.FromContext(ctx).Desugar().Error("failed to get Google service account name", zap.Error(err))
-		status.MarkWorkloadIdentityFailed(identifiable.ConditionSet(), workloadIdentityFailed, err.Error())
+		status.MarkWorkloadIdentityUnknown(identifiable.ConditionSet(), workloadIdentityFailed, err.Error())
```
grac3gao, line 119:
```suggestion
		status.MarkWorkloadIdentityUnknown(identifiable.ConditionSet(), deleteWorkloadIdentityFailed, err.Error())
```

pkg/reconciler/events/scheduler/scheduler_test.go
capri-xiyue: Yep. We will also need WithCloudSchedulerSourcePullSubscriptionDeleted().
I think this is a little bit different to the issue of marking status as NotReady when deletion happens successfully.
My point here is to make it explicitly that the resources get deleted but here the status is still ready.

danyinggu: Okay. Sounds good. So basically the `WithCloudSchedulerSourcePullSubscriptionDeleted()` and `WithCloudSchedulerSourceTopicDeleted()` are just wrapper but inside it's still markStatus as Ready.

capri-xiyue: Yep.  But maybe in `WithCloudSchedulerSourcePullSubscriptionDeleted()`, write some comments to explain why we have this wrapper.

danyinggu: Will do. Yes. Because this wrapper and `WithCloudSchedulerSourcePullSubscriptionReady()` have the same implementation.

PR #1566, opened 2020-08-11
Introduced a ConfigMap to configure broker delivery spec defaults

### LINKED ISSUE #1516: Introduce a ConfigMap to configure broker delivery spec defaults
**Problem**
Followup to #1072. Add a configmap to allow the cluster admin to set defaults for delivery settings (retry, dlq) for all new brokers.

Operator

**Exit Criteria**
Operator can use a configmap to set delivery defaults for new brokers.

tommyreddad:
Fixes #1516 

## Proposed Changes
- Add a ConfigMap (config-br-delivery) to configure default Broker delivery settings per cluster/namespace.


pkg/apis/broker/v1beta1/broker_validation.go
```diff
@@ -28,14 +28,22 @@ import (
 func (b *Broker) Validate(ctx context.Context) *apis.FieldError {
 	// We validate the GCP Broker's delivery spec. The eventing webhook will run
 	// the other usual validations.
-	return ValidateDeliverySpec(ctx, b.Spec.Delivery).ViaField("spec", "delivery")
+	if b.Spec.Delivery == nil {
+		return apis.ErrMissingField("spec.delivery")
+	}
```
yolocs, line 33: Now we require delivery spec?

tommyreddad: The backoff retry policy defaulting creates the delivery spec. If we don't require delivery spec, then we don't require backoff retry policy, and the backoff retry policy defaulting/validation will be duplicated between the webhook and the trigger reconciler, and default backoff settings will need to be carried in code in addition to in the ConfigMap. I assumed that it would be best to handle all the defaulting through the ConfigMap.

yolocs: Make sense.

pkg/apis/configs/broker/store.go
```diff
@@ -0,0 +1,92 @@
...
+// FromContextOrDefaults is like FromContext, but when no Config is attached it
+// returns a Config populated with the defaults for each of the Config fields.
+func FromContextOrDefaults(ctx context.Context) *Config {
+	if cfg := FromContext(ctx); cfg != nil {
+		return cfg
+	}
+	defaults, _ := NewDefaultsConfigFromMap(map[string]string{})
+	return &Config{
+		BrokerDeliverySpecDefaults: defaults,
+	}
```
yolocs, line 51: It seems `defaults` will always be `nil`?

pkg/apis/broker/v1beta1/broker_defaults.go
```diff
@@ -20,35 +20,38 @@ import (
...
-		b.Spec.Delivery.BackoffDelay == nil {
-		b.Spec.Delivery.BackoffPolicy = &DefaultBackoffPolicy
-		b.Spec.Delivery.BackoffDelay = &DefaultBackoffDelay
+	ns := apis.ParentMeta(withNS).Namespace
+	if b.Spec.Delivery.BackoffPolicy == nil {
+		b.Spec.Delivery.BackoffPolicy = deliverySpecDefaults.BackoffPolicy(ns)
+	}
+	if b.Spec.Delivery.BackoffDelay == nil {
+		b.Spec.Delivery.BackoffDelay = deliverySpecDefaults.BackoffDelay(ns)
+	}
```
yolocs, line 48: This logic seems a bit different than before? I wonder if we should blindly set defaults if only one of these values is set? E.g. `BackoffDelay` is set but not `BackoffPolicy`. It looks to me more like a user mistake that we should fail the validation.

pkg/apis/configs/broker/store.go
tommyreddad: You're right. Definitely seems strange to me, but this is how each FromContextsOrDefaults function is written in the rest of Knative, even when nil is guaranteed output. I just followed the standard without realizing, but I can change it to be more succinct.

pkg/apis/broker/v1beta1/broker_defaults.go
tommyreddad: Good point. I'll adjust that.

tommyreddad:
> Mostly looking good to me. Left a couple more questions. Also have you tested this manually in a cluster?

@yolocs I have tested it manually in my own dev cluster and it worked. However, @grantr mentioned to me this morning that the subscription config should be updated during reconciliation if it has changed, which currently is not implemented in this PR. Maybe it should be left as a separate issue, since this PR is already big enough?

yolocs:
@tommyreddad Yes, I totally missed that. Feel free to add a TODO and address it in a separate issue.

pkg/apis/broker/v1beta1/broker_validation.go
tommyreddad: After speaking with @grantr , it seems to make more sense to allow for nil delivery spec in validation, but validate a non-nil delivery spec, and also add a Broker SetDefaults in the reconciler if the delivery spec is nil. Disallowing a nil delivery spec could break existing Brokers for users who update Knative. My latest commit reflects this, so I've unresolved the conversation to see if this is alright with you.

Thanks so much for reviewing this, by the way!

pkg/apis/configs/broker/store.go
tommyreddad: Done.

pkg/apis/broker/v1beta1/broker_defaults.go
tommyreddad: Done.

Issue #1021, opened 2020-05-07
Change the service account when running prow e2e tests from superpower service account to service account with only permissions knative-gcp needed 


capri-xiyue:
**Problem**
Currently, when we run E2E tests in prow, we use a superpower service account.
However, when user uses knative-gcp, usually the control plane service account will only have roles https://github.com/google/knative-gcp/blob/master/docs/install/authentication-mechanisms-gcp.md listed here and the data plane service account will only have the roles https://github.com/google/knative-gcp/blob/master/docs/install/pubsub-service-account.md.
When we use a superpower service account to run the E2E tests in prow, sometimes we may fail to catch some bugs related to authentication.

Therefore, we need to change the service account when running prow e2e tests from superpower service account to service account with only permissions knative-gcp needed. 

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
The service account we use when running E2E tests in prow should only have the permissions listed in https://github.com/google/knative-gcp/blob/master/docs/install/authentication-mechanisms-gcp.md  and https://github.com/google/knative-gcp/blob/master/docs/install/pubsub-service-account.md

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

grac3gao:
Put a summary for the permission I used to conclude:

For brokerGSA:
roles/pubsub.editor. (For basic broker functionality, needed in both Secret mode and WI mode)
roles/monitoring.metricWriter (For metric in WI mode, Secret mode doesn’t need it)
roles/cloudtrace.agent (For tracing in WI mode, Secret mode doesn’t need it)

For controlGSA:
roles/pubsub.editor (for CloudPubSubSource, Pullsubscription and Topic)
roles/storage.admin (for CloudStorageSource)
roles/cloudscheduler.admin (for CloudSchedulerSource)
roles/pubsub.admin, roles/logging.configWriter, roles/logging.privateLogViewer (for CloudAuditLogsSource)
roles/pubsub.subscriber (for CloudBuildSource)

For sourceGSA:
roles/pubsub.editor. (For basic source functionality)
roles/monitoring.metricWriter (For metric in WI mode, Secret mode doesn’t need it, but we disable custom metrics currently)
roles/cloudtrace.agent (For tracing in WI mode, Secret mode doesn’t need it)

capri-xiyue:
[...]

PR #1574, opened 2020-08-12
[WIP] binded quota to the running project

capri-xiyue:
Fixes #

## Proposed Changes

- binded quota to the running project
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/reconciler/events/auditlogs/auditlogs.go
+++ b/pkg/reconciler/events/auditlogs/auditlogs.go
@@ -19,6 +19,8 @@ package auditlogs
 import (
 	"context"
 
+	"google.golang.org/api/option"
+
 	"cloud.google.com/go/logging/logadmin"
 	"go.uber.org/zap"
 	"google.golang.org/grpc/codes"
@@ -116,7 +118,8 @@ func (c *Reconciler) ensureSinkCreated(ctx context.Context, s *v1.CloudAuditLogs
 	if sinkID == "" {
 		sinkID = resources.GenerateSinkName(s)
 	}
-	logadminClient, err := c.logadminClientProvider(ctx, s.Status.ProjectID)
+	opt := option.WithQuotaProject(s.Status.ProjectID)
+	logadminClient, err := c.logadminClientProvider(ctx, s.Status.ProjectID, opt)
 	if err != nil {
 		logging.FromContext(ctx).Desugar().Error("Failed to create LogAdmin client", zap.Error(err))
 		return nil, err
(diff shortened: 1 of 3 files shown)


capri-xiyue:
I binded the quota to the running project because of in prow the e2e tests are flaky because of reaching quota limit issues.
Since in prow, all e2e tests of different projects use the same service account which by default the quota will goes to the project which owns the service account.
Not sure whether it will has side effects when we bind quota to the running project.

chizhg:
What's the difference between this and https://github.com/google/knative-gcp/pull/1440? Are they just in different places?

capri-xiyue:
> What's the difference between this and #1440? Are they just in different places?

I checked the error logs.
Now the e2e test failed when we tried to create CloudAuditLogsSource in reconciler which is because when don't bind quota when we create the relevant client. Therefore, I think the quota still goes to the project which owns the service account.

Though I don't know whether we should change the logic in our source code since this e2e test failure issue is kind of related to how we run prow e2e tests.

nachocano:
> > What's the difference between this and #1440? Are they just in different places?
> 
> I checked the error logs.
> Now the e2e test failed when we tried to create CloudAuditLogsSource in reconciler which is because when don't bind quota when we create the relevant client. Therefore, I think the quota still goes to the project which owns the service account.
> 
> Though I don't know whether we should change the logic in our source code since this e2e test failure issue is kind of related to how we run prow e2e tests.

If it's an issue on how we run e2e tests in PROW, I think we shouldn't change the source code.
Do we understand what this binding quota does? Users that run this will get affected with this change?

capri-xiyue:
> > > What's the difference between this and #1440? Are they just in different places?
> > 
> > 
> > I checked the error logs.
> > Now the e2e test failed when we tried to create CloudAuditLogsSource in reconciler which is because when don't bind quota when we create the relevant client. Therefore, I think the quota still goes to the project which owns the service account.
> > Though I don't know whether we should change the logic in our source code since this e2e test failure issue is kind of related to how we run prow e2e tests.
> 
> If it's an issue on how we run e2e tests in PROW, I think we shouldn't change the source code.
> Do we understand what this binding quota does? Users that run this will get affected with this change?

Yep. I agree we should not change the logic in our source code since this e2e test failure issue is kind of related to how we run prow e2e tests.
 Regarding "what this binding quota does", 
`if you're using a service account to do stuff, some operations may count quota against the project that owns the service account rather than the project being operated on, which requires service activation in the service account's project.`
For prow e2e tests, since all e2e tests of different codebase will use the same service account, so all the quotas will go to the gcp project which owns the service account. Therefore, it's more likely to reach quota limits.

User will get  affected with this change if they use a service account which doesn't belong to the running project.
Without quota biding in this PR, the quota will go to the project which owns the service account.
With quota biding in this PR, the quota will go to the running project

capri-xiyue:
@chizhg
Is it possible that knative-gcp has its own service account which belongs to knative-gcp specific prow project so that we don't need to share quota with other teams thttps://github.com/google/knative-gcp/issues/1021

capri-xiyue:
I did some investigation of what other team does. I don't think we should blindly bind the quota to the running project. 
We should allow user to specify the project they want to use for quota and billing.
If they don't specify, we should follow the default behavior of gcp.
New issues created here: https://github.com/google/knative-gcp/issues/1579

PR #1578, opened 2020-08-13
Add timeout delivery metrics

### LINKED ISSUE #1522: Add timeout delivery metrics
Currently we have delivery dispatch time metrics with response code. But if the delivery is cancelled because of timeout, we don't report metrics which could be useful in identifying receiver issues.

grac3gao:
Fixes #1522

## Proposed Changes

-  If the delivery is cancelled because of timeout, report event dispatch time without resp status code. In metric explorer, the dispatch time will look like: (trigger-actor-0 only has timeout event, trigger-actor-1 only has 2xx event)
Screen Shot 2020-08-13 at 9 53 40 AM

- Add and update some UT
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/metrics/delivery_reporter.go
+++ b/pkg/metrics/delivery_reporter.go
@@ -22,17 +22,20 @@ import (
 	"strconv"
 	"time"
 
-	"github.com/google/knative-gcp/pkg/broker/config"
 	"go.opencensus.io/stats"
 	"go.opencensus.io/stats/view"
 	"go.opencensus.io/tag"
 	"knative.dev/pkg/metrics"
+
+	"github.com/google/knative-gcp/pkg/broker/config"
 )
 
 type DeliveryMetricsKey int
+type RespStatusCodeKey int
 
 const (
 	startDeliveryProcessingTime DeliveryMetricsKey = iota
+	respStatusCode              RespStatusCodeKey  = iota
 )
 
 type DeliveryReporter struct {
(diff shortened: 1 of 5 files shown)


pkg/metrics/delivery_reporter.go
```diff
@@ -185,3 +198,11 @@ func getStartDeliveryProcessingTime(ctx context.Context) (time.Time, error) {
 	}
 	return time.Time{}, fmt.Errorf("missing or invalid start time: %v", v)
 }
+
+func getRespStatusCode(ctx context.Context) (int, error) {
```
yolocs, line 202: nit: based on the usage, a better signature seems to be `getRespStatusCode(...) (int, bool)`

grac3gao: done

pkg/metrics/delivery_reporter.go
```diff
@@ -120,15 +123,25 @@ func NewDeliveryReporter(podName PodName, containerName ContainerName) (*Deliver
 	return r, nil
 }
 
+func WithRespStatusCode(ctx context.Context, responseCode int) context.Context {
+	return context.WithValue(ctx, respStatusCode, responseCode)
```
ian-mi, line 127: This can be simplified by just inserting the relevant tags into context.

grac3gao: done

Issue #913, opened 2020-04-21
Add generation number to broker-targets config


liu-cong:
**Problem**
Right now each data plane pod has its own generation of the configmap. The generation is incremented every time the configmap is updated. However, there is no guarantee that all data plane pods have the same generation number.

Instead,
- controller increments the gen number on each update as a field in the broker targets configmap
- controller also updates the gen number as an annotation on the data plane pods

This is helpful to debug whether the data plane has the same generation of the configmap. See https://github.com/google/knative-gcp/pull/907#discussion_r412510724

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

grantr:
Just to clarify, is this issue also suggesting that we add the generation number to the serialized targets config?

liu-cong:
> Just to clarify, is this issue also suggesting that we add the generation number to the serialized targets config?

Yes this is also a nice to have. Especially when new pods may not get the generation as annotation. Updated the description.

grantr:
Also add a timestamp to configmap so the data plane pods can calculate the delay for #1443.

yolocs:
This appears to be a little bit trickier than expected. If we blindly add generation and timestamp to the configmap, it will cause unnecessary configmap updates because timestamp (and potentially generation depends on how we generate it) will always be different from the one in the existing configmap.

One possible approach: add timestamp/generation with a separate config key in the data. When reconciling the configmap, we need to skip comparing the timestamp/generation. These values will be in the same volume mount and be accessible by the broker components.

grantr:
Seems like a good idea to separate the timestamp into a different file, since it's really intended to communicate data about the configmap mechanism rather than event routing data. Generation I could see being used either way, but we can always move it into event routing later when we have a use for it.

PR #1581, opened 2020-08-13
Metrics to monitor latencies on BrokerCell (#1443)

### LINKED ISSUE #1443: Add metrics to monitor the gcp broker configmap update delay
**Problem**

There has been constant discussions around the delay between control plan ready and data plane readiness. 

Current sequence of events to get configmap volume updated:
`Create broker/trigger -> broker/trigger controller reconciles broker/trigger to be ready -> brokercell controller gets notified -> brokercell controller updates configmap and annoates data plane pods -> kubelet refreshes data plane configmap volume`

As you can see, there are multiple async operations before the data plane actually gets the update. It will be useful if we can record the delay between broker/trigger ready to data plane ready, as well as breakdown of each step, in order for use to make a more informed decision on whether the current approach is good enough.

* ResourceUpdateToBrokerCellNotified: The delay between a broker/trigger update event and the brokercell controller gets notified. The brokercell controller [watches](https://github.com/google/knative-gcp/blob/9a89e9a9a6521a72b4c7e88bd2af3de0654c934d/pkg/reconciler/brokercell/controller.go#L83-L103) for broker/trigger updates, we can record a metric when the watch is triggered, compared to the lastUpdateTime of the resource.
* ConfigmapUpdateToVolumeUpdate: The config lib can maintain the last updated ts of the volume file, and  when it [detects](https://github.com/google/knative-gcp/blob/9a89e9a9a6521a72b4c7e88bd2af3de0654c934d/pkg/broker/config/volume/volume.go#L72) a change in the volume file, it can record the delay.

Caveats: The timestamps are saved by different pods. We assume the difference in their clocks are not noticeable.

**Exit Criteria**

**Time Estimate (optional):** 

cc @grantr

Ectelion:
Adds latency metrics to capture the delay between a resource update and the notification seen by the Informer, specifically on BrokerCell. Fixes #1443 (except the volume update part of it, which depends on [#913](https://github.com/google/knative-gcp/issues/913)).

The latency util should be extendable other scenarios (pass different value to the _process_type_ tag).

### DIFF
--- a/config/core/deployments/controller.yaml
+++ b/config/core/deployments/controller.yaml
@@ -67,7 +67,7 @@ spec:
         - name: BROKER_CELL_RETRY_IMAGE
           value: ko://github.com/google/knative-gcp/cmd/broker/retry
         - name: INTERNAL_METRICS_ENABLED
-          value: "true"
+          value: "false"
         volumeMounts:
         - name: google-cloud-key
           mountPath: /var/secrets/google
--- a/config/core/deployments/controller.yaml
+++ b/config/core/deployments/controller.yaml
@@ -66,6 +66,8 @@ spec:
           value: ko://github.com/google/knative-gcp/cmd/broker/fanout
         - name: BROKER_CELL_RETRY_IMAGE
           value: ko://github.com/google/knative-gcp/cmd/broker/retry
+        - name: INTERNAL_METRICS_ENABLED
+          value: "true"
         volumeMounts:
         - name: google-cloud-key
           mountPath: /var/secrets/google
--- a/pkg/metrics/brokercell_delay_reporter.go
+++ b/pkg/metrics/brokercell_delay_reporter.go
@@ -27,17 +27,11 @@ import (
 	"knative.dev/pkg/metrics"
 )
 
-const LatencyMetricName string = "process_latencies"
+const LatencyMetricName string = "brokercell_delay"
 
 type LatencyReporter struct {
 	durationInMsecM *stats.Float64Measure
 }
-
-type ProcessType string
-
-const (
-	ResourceUpdateToBrokerCellNotified ProcessType = "ResourceUpdateToBrokerCellNotified"
-)
 
 func (r *LatencyReporter) register() error {
 	return metrics.RegisterResourceView(
@@ -47,9 +41,8 @@ func (r *LatencyReporter) register() error {
 			Measure:     r.durationInMsecM,
 			Aggregation: view.Distribution(metrics.Buckets125(1, 100000)...), // 1, 2, 5, 10, 20, 50, 100, 1000, 5000, 10000, 20000, 50000, 1000000
 			TagKeys: []tag.Key{
-				NamespaceNameKey,
-				ProcessTypeKey,
-				EntityNameKey,
+				ResourceKindKey,
+				ResourceNameKey,
 			},
 		},
 	)
@@ -71,11 +64,11 @@ func NewLatencyReporter() (*LatencyReporter, error) {
 }
 
 // ReportLatency records the value to the latency metric
-func (r *LatencyReporter) ReportLatency(ctx context.Context, duration time.Duration, processType ProcessType, entityName string) error {
+func (r *LatencyReporter) ReportLatency(ctx context.Context, duration time.Duration, resourceKind, resourceName string) error {
 	tag, err := tag.New(
 		ctx,
-		tag.Insert(ProcessTypeKey, string(processType)),
-		tag.Insert(EntityNameKey, entityName),
+		tag.Insert(ResourceKindKey, resourceKind),
+		tag.Insert(ResourceNameKey, resourceName),
 	)
 	if err != nil {
 		return fmt.Errorf("failed to create metrics tag: %v", err)
--- a/pkg/metrics/brokercell_delay_reporter.go
+++ b/pkg/metrics/brokercell_delay_reporter.go
@@ -71,7 +71,7 @@ func (r *LatencyReporter) ReportLatency(ctx context.Context, duration time.Durat
 		tag.Insert(ResourceNameKey, resourceName),
 	)
 	if err != nil {
-		return fmt.Errorf("failed to create metrics tag: %v", err)
+		return fmt.Errorf("failed to create metrics tag: %w", err)
 	}
 	durationMetricValue := r.durationInMsecM.M(float64(duration / time.Millisecond))
 	metrics.Record(tag, durationMetricValue)
--- a/pkg/metrics/brokercell_delay_reporter.go
+++ b/pkg/metrics/brokercell_delay_reporter.go
@@ -29,11 +29,11 @@ import (
 
 const LatencyMetricName string = "brokercell_delay"
 
-type LatencyReporter struct {
+type BrokerCellLatencyReporter struct {
 	durationInMsecM *stats.Float64Measure
 }
 
-func (r *LatencyReporter) register() error {
+func (r *BrokerCellLatencyReporter) register() error {
 	return metrics.RegisterResourceView(
 		&view.View{
 			Name:        r.durationInMsecM.Name(),
@@ -42,33 +42,35 @@ func (r *LatencyReporter) register() error {
 			Aggregation: view.Distribution(metrics.Buckets125(1, 100000)...), // 1, 2, 5, 10, 20, 50, 100, 1000, 5000, 10000, 20000, 50000, 1000000
 			TagKeys: []tag.Key{
 				ResourceKindKey,
+				NamespaceNameKey,
 				ResourceNameKey,
 			},
 		},
 	)
 }
 
-// NewLatencyReporter creates a new LatencyReporter
-func NewLatencyReporter() (*LatencyReporter, error) {
-	r := &LatencyReporter{
+// NewBrokerCellLatencyReporter creates a new BrokerCellLatencyReporter
+func NewBrokerCellLatencyReporter() (*BrokerCellLatencyReporter, error) {
+	r := &BrokerCellLatencyReporter{
 		durationInMsecM: stats.Float64(
 			LatencyMetricName,
-			"Latency of a process in milliseconds",
+			"Latency of the delay between a resource update and the time the informer gets it in milliseconds",
 			stats.UnitMilliseconds,
 		),
 	}
 	if err := r.register(); err != nil {
-		return nil, fmt.Errorf("failed to register the latency reporter: %w", err)
+		return nil, fmt.Errorf("failed to register BrokerCellLatencyReporter: %w", err)
 	}
 	return r, nil
 }
 
 // ReportLatency records the value to the latency metric
-func (r *LatencyReporter) ReportLatency(ctx context.Context, duration time.Duration, resourceKind, resourceName string) error {
+func (r *BrokerCellLatencyReporter) ReportLatency(ctx context.Context, duration time.Duration, resourceKind, resourceName, namespace string) error {
 	tag, err := tag.New(
 		ctx,
 		tag.Insert(ResourceKindKey, resourceKind),
 		tag.Insert(ResourceNameKey, resourceName),
+		tag.Insert(NamespaceNameKey, namespace),
 	)
 	if err != nil {
 		return fmt.Errorf("failed to create metrics tag: %w", err)
--- a/pkg/metrics/latency_reporter.go
+++ b/pkg/metrics/latency_reporter.go
@@ -1,0 +1,86 @@
+/*
+Copyright 2020 Google LLC
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package metrics
+
+import (
+	"context"
+	"fmt"
+	"time"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+	"knative.dev/pkg/metrics"
+)
+
+const LatencyMetricName string = "process_latencies"
+
+type LatencyReporter struct {
+	durationInMsecM *stats.Float64Measure
+}
+
+type ProcessType string
+
+const (
+	ResourceUpdateToBrokerCellNotified ProcessType = "ResourceUpdateToBrokerCellNotified"
+)
+
+func (r *LatencyReporter) register() error {
+	return metrics.RegisterResourceView(
+		&view.View{
+			Name:        r.durationInMsecM.Name(),
+			Description: r.durationInMsecM.Description(),
+			Measure:     r.durationInMsecM,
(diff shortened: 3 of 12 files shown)


pkg/metrics/latency_reporter.go
```diff
@@ -0,0 +1,86 @@
...
+		return nil, fmt.Errorf("failed to register the latency reporter: %w", err)
+	}
+	return r, nil
+}
+
+// ReportLatency records the value to the latency metric
+func (r *LatencyReporter) ReportLatency(ctx context.Context, duration time.Duration, processType ProcessType, entityName string) error {
+	tag, err := tag.New(
+		ctx,
+		tag.Insert(ProcessTypeKey, string(processType)),
```
liu-cong, line 77: the namespace tag key is not used

pkg/metrics/latency_reporter.go
```diff
@@ -0,0 +1,86 @@
+/*
```
liu-cong, line 1: latency_reporter is too generic. Consider "informer_delay_reporter"

pkg/metrics/latency_reporter.go
```diff
@@ -0,0 +1,86 @@
...
+	return metrics.RegisterResourceView(
+		&view.View{
+			Name:        r.durationInMsecM.Name(),
+			Description: r.durationInMsecM.Description(),
+			Measure:     r.durationInMsecM,
+			Aggregation: view.Distribution(metrics.Buckets125(1, 100000)...), // 1, 2, 5, 10, 20, 50, 100, 1000, 5000, 10000, 20000, 50000, 1000000
+			TagKeys: []tag.Key{
+				NamespaceNameKey,
+				ProcessTypeKey,
+				EntityNameKey,
```
liu-cong, line 52: I would also add the resource name as a key to make the time series more unique. Stackdriver has a limit of only accepting a unique timeseries every 10s: https://cloud.google.com/monitoring/quotas#custom_metrics_quotas

Ectelion: Good to know, accounted for it.

pkg/metrics/latency_reporter.go
```diff
@@ -0,0 +1,86 @@
...
+	"fmt"
+	"time"
+
+	"go.opencensus.io/stats"
+	"go.opencensus.io/stats/view"
+	"go.opencensus.io/tag"
+	"knative.dev/pkg/metrics"
+)
+
+const LatencyMetricName string = "process_latencies"
```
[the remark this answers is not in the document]
Ectelion: Accounted for the name. As for the process-tag: when the volume latency metric comes in, the idea was to cover it with the same reporter, just with a different "ProcessType" tag. We can drop it for now though.

pkg/metrics/latency_reporter.go
Ectelion: Accounted for the name. It's now "brokercell_delay_reporter" -- more specific, but also let's us extend the reporter to other relevant delays within the brokercell.

pkg/utils/customresource/typegenerics.go
```diff
@@ -0,0 +1,49 @@
...
+	"fmt"
+	"time"
+
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	duckv1 "knative.dev/pkg/apis/duck/v1"
+)
+
+// RetrieveLatestUpdateTime takes an arbitrary custom resource and retrives its latest update time
+// based on the resource's metadata
+func RetrieveLatestUpdateTime(customResource interface{}) (*time.Time, error) {
```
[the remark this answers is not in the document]
Ectelion: To account for the case with errors: either value != nul and err == null or vice versa -- seems like a common pattern?

pkg/metrics/latency_reporter.go
Ectelion: Thanks!

pkg/metrics/brokercell_delay_reporter.go
```diff
@@ -0,0 +1,79 @@
...
+
+// ReportLatency records the value to the latency metric
+func (r *LatencyReporter) ReportLatency(ctx context.Context, duration time.Duration, resourceKind, resourceName string) error {
+	tag, err := tag.New(
+		ctx,
+		tag.Insert(ResourceKindKey, resourceKind),
+		tag.Insert(ResourceNameKey, resourceName),
+	)
+	if err != nil {
+		return fmt.Errorf("failed to create metrics tag: %v", err)
```
yolocs, line 74: nit: `%w`

pkg/utils/customresource/typegenerics.go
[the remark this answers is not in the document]
yolocs: You can return `time.Time{}` (zero value) instead. It's a common expectation if err != nil, the actual return val is zero value.

pkg/reconciler/brokercell/controller.go
```diff
@@ -98,6 +106,7 @@ func NewController(
 				}
 				// TODO(#866) Select the brokercell that's associated with the given broker.
 				impl.EnqueueKey(types.NamespacedName{Namespace: b.Namespace, Name: brokerresources.DefaultBrokerCellName})
+				reportLatency(&ctx, obj, metricsReporter, "Trigger", t.Name)
```
yolocs, line 109: Should we report latency in the reconciler just to be more precise? Does `enqueue` instantly trigger a reconcile?

pkg/reconciler/brokercell/controller.go
```diff
@@ -74,6 +77,10 @@ func NewController(
 		logger.Fatal("Failed to create BrokerCell reconciler", zap.Error(err))
 	}
 	impl := v1alpha1brokercell.NewImpl(ctx, r)
+	metricsReporter, err := metrics.NewLatencyReporter()
+	if err != nil {
+		logger.Error("Failed to create latency reporter", zap.Error(err))
+	}
```
yolocs, line 83: IMO a user shouldn't really care about this metric. It's more like a metric we want to collect in certain envs (integ tests mostly). I wonder if we need to make a toggle for this metric reporting?

@liu-cong @grantr

pkg/utils/customresource/typegenerics.go
```diff
@@ -0,0 +1,49 @@
...
+	metaAccessor, maAssertable := customResource.(metav1.ObjectMetaAccessor)
+	resourceKR, krAssertable := customResource.(duckv1.KRShaped)
+	if !maAssertable || !krAssertable {
+		return nil, fmt.Errorf("The specified customResource does not support an expected type assertion (ObjectMetaAccessor: %v, KRShaped: %v) %#v", maAssertable, maAssertable, customResource)
+	}
```
yolocs, line 34: It seems we can simply this logic. In the `controller.go` where we setup the informer, we already attempt to convert the interface to `*brokerv1beta1.Broker` or `*brokerv1beta1.Trigger`. And they are for sure matching these types.

pkg/utils/customresource/typegenerics.go
```diff
@@ -0,0 +1,49 @@
...
+	} else {
+		status := resourceKR.GetStatus()
+		if status != nil && status.Conditions != nil {
+			for _, cond := range status.Conditions {
+				if latestUpdateTime.Before(&cond.LastTransitionTime.Inner) {
+					latestUpdateTime = cond.LastTransitionTime.Inner
+				}
+			}
+		}
+	}
```
yolocs, line 47: What's your thought on how we should read the metric? I can think of two confusing case.

1. A trigger's latestUpdateTime was updated before we report the metric here. It would give us an illusion of shorter delays.
2. There is no update to a trigger for a long time. Would the brokercell controller still gets informed on resync period? If so, on every resync we would get longer and longer delays because latestUpdateTime of the trigger is not changing.

liu-cong: I think the resync only re-enqueues the brokercell, not the trigger. The trigger controller also has a resync. But I think it will only invoke the brokercell if there is actually an update (we should expect most resync to be noop).

pkg/metrics/latency_reporter.go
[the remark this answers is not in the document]
liu-cong: I expect the configmap volume latency to be a different metric.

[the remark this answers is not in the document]
liu-cong: e.g., you have two metrics: "informer_delay" and "volume_refresh_delay". If we are using a tag to differentiate them while they share nothing in common, it's better to just create a new metric.

pkg/reconciler/brokercell/controller.go
liu-cong: Sounds like a good idea to disable these by default

pkg/reconciler/brokercell/controller.go
Ectelion: Yeah, I have also noticed this and brought up on earlier this week. I suspect that the work-queue related delay may have even a higher chance of being the root cause behind the data plane sync delays than e.g. volume updates, because potential delays here may be expected by design. If we dig further into the implementation, there even seems to be an explicit rate-limiting logic that's meant to smooth out the load.

Speaking of this PR, I think it's probably best to keep this metric informer-focused, while also considering another one that is more specifically captures the work-queue delays.

pkg/reconciler/brokercell/controller.go
Ectelion: Makes sense. There are multiple ways to approach that: environment variable, debug mode, ConfigMap, CRD. The first one seems quite suitable here -- just in case, is everyone fine with that?

pkg/utils/customresource/typegenerics.go
Ectelion: Instead of hardcoding towards trigger and broker (which I indeed had in the first version), I made this a generic util, which can be re-used on other custom resources. The error handling above is just meant to protect against inappropriate usage e.g. when some other resource or a possibly incompatible future version of it is passed -- seems like a good practice for a util/helper/library kind of methods.

pkg/metrics/brokercell_delay_reporter.go
Ectelion: Sure.

pkg/metrics/latency_reporter.go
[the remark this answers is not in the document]
Ectelion: Ok, have accounted for that as well.

pkg/utils/customresource/typegenerics.go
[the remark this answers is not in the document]
Ectelion: Sure.

pkg/utils/customresource/typegenerics.go
Ectelion: Thanks for clarifying on re-queuing. In general, I do expect that there will be some noise in the data. This could have been reduced if we had a clear last-update timestamp on the object. But that’s not the case, thus here is the workaround. 
Considering our goals, I think it’s ok that some of the samples won’t be precise — like in other statistical experiments, with large enough sampling and a measure more robust to outliers (e.g. percentiles/median vs mean), I expect the numbers to become more representative of the actual delays, which will help us confirm/rule out this as a root cause.

pkg/reconciler/brokercell/controller.go
liu-cong: I am fine with an env var

pkg/reconciler/brokercell/controller.go
liu-cong: I am not concerned by rate limiting logic as we should expect only one brokercell object in the cluster. I would expect the reconcile to happen almost instantly.

Maybe there are other factors that would affect the work queue delay. that I don't know of.

pkg/metrics/brokercell_delay_reporter.go
```diff
@@ -0,0 +1,79 @@
...
+		},
+	)
+}
+
+// NewLatencyReporter creates a new LatencyReporter
+func NewLatencyReporter() (*LatencyReporter, error) {
+	r := &LatencyReporter{
+		durationInMsecM: stats.Float64(
+			LatencyMetricName,
+			"Latency of a process in milliseconds",
```
liu-cong, line 56:
```suggestion
			"Latency of the delay between an resource gets updated and the informer gets it in milliseconds",
```

pkg/metrics/brokercell_delay_reporter.go
```diff
@@ -0,0 +1,79 @@
...
+		return nil, fmt.Errorf("failed to register the latency reporter: %w", err)
+	}
+	return r, nil
+}
+
+// ReportLatency records the value to the latency metric
+func (r *LatencyReporter) ReportLatency(ctx context.Context, duration time.Duration, resourceKind, resourceName string) error {
+	tag, err := tag.New(
+		ctx,
+		tag.Insert(ResourceKindKey, resourceKind),
```
liu-cong, line 70: I think we should also add the namespace tag to make the timeseries more unique.

pkg/metrics/brokercell_delay_reporter.go
Ectelion: Addressed

pkg/metrics/brokercell_delay_reporter.go
Ectelion: Sure.

pkg/reconciler/brokercell/controller.go
Ectelion: Addressed.

config/core/deployments/controller.yaml
```diff
@@ -66,6 +66,8 @@ spec:
           value: ko://github.com/google/knative-gcp/cmd/broker/fanout
         - name: BROKER_CELL_RETRY_IMAGE
           value: ko://github.com/google/knative-gcp/cmd/broker/retry
+        - name: INTERNAL_METRICS_ENABLED
+          value: "true"
```
liu-cong, line 70: should default to "false".

liu-cong:
/lgtm

I will defer to @yolocs for approval.

pkg/utils/customresource/typegenerics.go
yolocs: I'm being nit picking here. I didn't mean that this should _only_ accept broker and trigger. But instead of interface, probably something more deterministic, e.g. `metav1.ObjectMetaAccessor`. Then you don't have to do the cast check in this func.

pkg/reconciler/brokercell/controller.go
```diff
@@ -125,3 +138,17 @@ func handleResourceUpdate(impl *controller.Impl) cache.ResourceEventHandler {
 	// will be skipped by the function.
 	return controller.HandleAll(impl.EnqueueLabelOfNamespaceScopedResource(namespaceLabel, resources.BrokerCellLabelKey))
 }
+
+// reportLatency estimates the time spent since the last update of the resource object and records it to the latency metric
+func reportLatency(ctx *context.Context, resourceObj interface{}, latencyReporter *metrics.BrokerCellLatencyReporter, resourceKind, resourceName, namespace string) {
```
yolocs, line 143: Why `*context.Context`?

yolocs:
/lgtm
/hold

Left a few nits in case you want to address.

pkg/utils/customresource/typegenerics.go
Ectelion: Ah, I see the point, have simplified this in the latest update. The methods come from two different interfaces, so we have one remaining casting involved.

pkg/reconciler/brokercell/controller.go
Ectelion: Yeah, passing by reference was an over-optimization here, given that Context is an interface. Have switched to pass-by-value.

config/core/deployments/controller.yaml
Ectelion: Yeah, thanks!

PR #1582, opened 2020-08-13
Add constants and methods for brokerCell args

danyinggu:
Fixes #

## Proposed Changes

- Update brokerPath as a method, refactor path related logic to methods.
-

**Release Note**

```release-note
NONE
```

**Docs**

### DIFF
--- a/pkg/broker/config/config.go
+++ b/pkg/broker/config/config.go
@@ -86,6 +86,12 @@ func BrokerKey(namespace, name string) string {
 	return namespace + "/" + name
 }
 
+// BrokerPath returns the path to be set in the status of a broker.
+// The format is brokerNamespace/brokerName
+func BrokerPath(namespace, name string) string {
+	return fmt.Sprintf("/%s/%s", namespace, name)
+}
+
 // TriggerKey returns the key of a trigger. Format is namespace/brokerName/targetName.
 func TriggerKey(namespace, broker, target string) string {
 	return fmt.Sprintf("%s/%s/%s", namespace, broker, target)
--- a/pkg/broker/config/config.go
+++ b/pkg/broker/config/config.go
@@ -86,12 +86,6 @@ func BrokerKey(namespace, name string) string {
 	return namespace + "/" + name
 }
 
-// BrokerPath returns the path to be set in the status of a broker.
-// The format is brokerNamespace/brokerName
-func BrokerPath(namespace, name string) string {
-	return fmt.Sprintf("/%s/%s", namespace, name)
-}
-
 // TriggerKey returns the key of a trigger. Format is namespace/brokerName/targetName.
 func TriggerKey(namespace, broker, target string) string {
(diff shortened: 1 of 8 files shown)


pkg/broker/config/config.go
```diff
@@ -86,6 +86,12 @@ func BrokerKey(namespace, name string) string {
 	return namespace + "/" + name
 }
 
+// BrokerPath returns the path to be set in the status of a broker.
+// The format is brokerNamespace/brokerName
+func BrokerPath(namespace, name string) string {
+	return fmt.Sprintf("/%s/%s", namespace, name)
+}
+
```
yolocs, line 94: The name of the file might appear confusing but the actual purpose of this file for broker data plane configuration.

My suggestion is to create a `pkg/broker/ingress/path.go` file. Put this func there. Also, move this logic (https://github.com/google/knative-gcp/blob/master/pkg/broker/ingress/handler.go#L123-L133) there. And change other places accordingly.

pkg/broker/ingress/path.go
```diff
@@ -0,0 +1,47 @@
...
+}
+
+// convertPathToNamespacedName converts the broker path to a NamespaceName.
+func (h *Handler) convertPathToNamespacedName(path string) (*types.NamespacedName, error) {
+	// Path should be in the form of "/<ns>/<broker>".
+	pieces := strings.Split(path, "/")
+	if len(pieces) != 3 {
+		h.logger.Debug("Malformed request path", zap.String("path", path))
+		err := fmt.Errorf("Malformed request path; expect format '/<ns>/<broker>'")
+		return nil, err
```
yolocs, line 40: nit: why not directly return `fmt.Errorf`?

pkg/broker/ingress/path.go
```diff
@@ -0,0 +1,47 @@
...
+	}
+	broker := types.NamespacedName{
+		Namespace: pieces[1],
+		Name:      pieces[2],
+	}
+	return &broker, nil
```
yolocs, line 46: nit:

```
return &types.NamespacedName{
  Namespace: pieces[1],
  Name:      pieces[2],
}, nil
```

pkg/broker/ingress/path.go
```diff
@@ -0,0 +1,47 @@
...
+)
+
+// BrokerPath returns the path to be set in the status of a broker.
+// The format is brokerNamespace/brokerName
+func BrokerPath(namespace, name string) string {
+	return fmt.Sprintf("/%s/%s", namespace, name)
+}
+
+// convertPathToNamespacedName converts the broker path to a NamespaceName.
+func (h *Handler) convertPathToNamespacedName(path string) (*types.NamespacedName, error) {
```
yolocs, line 34: nit: you don't have to return pointer here if you don't want to change from `broker` to `*broker` in other places. You can return an empty (`types.NamespacedName{}`) value when there is an error.

yolocs, line 34: No need to make it a func on `Handler`.

Issue #1506, opened 2020-07-22
Eliminate Broker Ingress HTTP Span


ian-mi:
**Problem**
Two nearly identical spans are created broker ingress: the HTTP handler span followed by the main broker ingress span. These could be combined into a single span in order to improve trace readability.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Event producers and consumers

**Exit Criteria**
Manual verification that broker ingress creates a single span.

**Time Estimate:** 
1-2 days.

**Additional context**
FormatSpanName in [ochttp.Handler](https://github.com/census-instrumentation/opencensus-go/blob/v0.22.4/plugin/ochttp/server.go#L44) can be used to format the HTTP handler span so that it matches the format of the main broker ingress span. Since this Handler is created by the eventing [HttpMessageReceiver](https://github.com/knative/eventing/blob/v0.16.1/pkg/kncloudevents/message_receiver.go#L33), this may need to be extended to support setting FormatSpanName.

PR #1585, opened 2020-08-14
Eliminate creating a new broker ingress span and add broker ingress attributes to the existing http span

bharattkukreja:
Fixes https://github.com/google/knative-gcp/issues/1506

## Proposed Changes

- Reuse http span for broker ingress attributes in broker ingress handler


bharattkukreja:
@ian-mi @Harwayne

PR #1589, opened 2020-08-17
Modify e2e tests readme to contain policy binding for cloudtrace.admin role to run trace related e2e tests

bharattkukreja:
Cloudtrace.user role is needed to run trace related e2e tests, in addition to cloudtrace.agent role. Both these roles together make the cloudtrace.admin role: https://cloud.google.com/trace/docs/iam

## Proposed Changes

- Replace cloudtrace.agent with cloudtrace.admin in e2e tests readme

### DIFF
--- a/test/e2e/README.md
+++ b/test/e2e/README.md
@@ -95,12 +95,12 @@ There are two ways to set up authentication mechanism.
     If you also plan on running tracing-related E2E tests using the StackDriver
     backend, your
     [Service Account](../../docs/install/dataplane-service-account.md) needs
-    additional `cloudtrace.agent` role:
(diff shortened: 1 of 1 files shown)


bharattkukreja:
cc @Harwayne

grac3gao:
Is `Cloudtrace.user` added because we have new feature in tracing area? The reason why I ask this is: we only have `cloudtrace.agent ` role in internal test, and it doesn't block the trace integration test.

bharattkukreja:
I am not sure if a new feature was added recently, but I wasn't able to run the e2e tests without the additional cloudtrace.user permission.

PR #1590, opened 2020-08-18
Add util function to ensure that a given environment variable is set and use it for e2e tests

bharattkukreja:
Currently if the environment variable containing the project id for e2e tests (E2E_PROJECT_ID) is not set, the tests still continue to run and fail later with "permission denied error" as an empty project has no permissions to access anything. Instead, we want the test to fail if they are not able to get the required environment variable

## Proposed Changes

- Add util function to ensure that a given environment variable is set
- Modify tests to use that util function instead of os.getEnv (used currently), which succeeds even if the environment variable doesn't exist

### DIFF
--- a/test/e2e/lib/utils.go
+++ b/test/e2e/lib/utils.go
@@ -20,16 +20,15 @@ limitations under the License.
 package lib
 
 import (
-	"fmt"
 	"os"
 	"testing"
 )
 
-// Returns the value of the environment variable if it exists, otherwise exits the test with an error
+// GetEnvOrFail gets the specified environment variable. If the variable is not set, then the test exits with an error.
 func GetEnvOrFail(t *testing.T, key string) string {
 	value, success := os.LookupEnv(key)
 	if !success {
-		t.Fatal(fmt.Sprintf("Environment variable %s not set", key))
+		t.Fatalf("Environment variable %q not set", key)
 	}
 	return value
 }
--- a/test/e2e/lib/utils.go
+++ b/test/e2e/lib/utils.go
@@ -1,0 +1,35 @@
+/*
+Copyright 2020 Google LLC.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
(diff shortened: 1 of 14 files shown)


bharattkukreja:
cc @Harwayne

test/e2e/lib/utils.go
```diff
@@ -0,0 +1,35 @@
...
+*/
+package lib
+
+import (
+	"fmt"
+	"os"
+	"testing"
+)
+
+// Returns the value of the environment variable if it exists, otherwise exits the test with an error
```
Harwayne, line 28: The doc string is supposed to start with the name of the thing being documented, see https://blog.golang.org/godoc, "Notice this comment is a complete sentence that begins with the name of the element it describes."

How about something like this?
```suggestion
// GetEnvOrFail gets the specified environment variable. If the variable is not set, then the test exits with an error.
```

test/e2e/lib/utils.go
```diff
@@ -0,0 +1,35 @@
...
+func GetEnvOrFail(t *testing.T, key string) string {
+	value, success := os.LookupEnv(key)
+	if !success {
+		t.Fatal(fmt.Sprintf("Environment variable %s not set", key))
```
Harwayne, line 32:
```suggestion
		t.Fatalf("Environment variable %q not set", key)
```

test/e2e/test_storage.go
```diff
@@ -202,7 +201,7 @@ func CloudStorageSourceWithTargetTestImpl(t *testing.T, assertMetrics bool, auth
 		time.Sleep(sleepTime)
 
 		// If we reach this point, the projectID should have been set.
-		projectID := os.Getenv(lib.ProwProjectKey)
+		projectID := lib.GetEnvOrFail(t, lib.ProwProjectKey)
```
Harwayne, line 204: Is this different than `project` defined on line 137?

bharattkukreja: You are right, it's not. Modified the function to re-use `project`

PR #1605, opened 2020-08-19
Improved UX for BrokerCell Configurability

### LINKED ISSUE #1603: Improve the UX for BrokerCell configurability before exposing it externally
**Problem**
With #1484, resource and HPA configurations for BrokerCell deployments become configurable. The original motivation was to assist with performance investigations, but later on we realized that the configurability feature can also be valuable for the end users. Speaking of the end user experience, we've determined that [current implementation](https://github.com/google/knative-gcp/pull/1509) has a usability aspect that is worth iterating over before we expose this externally.

In particular, citing the key suggestion:
_"if I (as a user) set the CPU requirements but not memory, then it will use the default memory requirements. But such defaulting could somehow "invalidate" my CPU requirements as the memory defaults could be much lower than my expectation. What if we do it with a more blunt approach? Either you (as a user) don't specify the ComponentParameters and we default everything for you, or you need to configure the whole ComponentParameters. In the latter case, it doesn't mean you have to specify everything in ComponentParameters. It just means we won't do any defaulting in that case and will use whatever value you provide, e.g. if memory requirement is missing, then we don't specify memory requirement in the deployment."_

**Exit Criteria**

Ectelion:
Fixes #1603

## Proposed Changes

- Makes BrokerCell configurability experience more intuitive, so that it is more suitable not only for our internal perf work, but also for the end users; 
- Based on ideas discussed in #1509;
- Backward compatible with [yamls for the perf work](https://github.com/google/knative-gcp/pull/1402);


pkg/apis/intevents/v1alpha1/brokercell_types.go
```diff
@@ -70,8 +70,8 @@ var (
 
 // SystemResource specifies quantities for resources such as CPU and memory
 type SystemResource struct {
-	CPU    *string `json:"cpu,omitempty"`
-	Memory *string `json:"memory,omitempty"`
+	CPU    string `json:"cpu,omitempty"`
+	Memory string `json:"memory,omitempty"`
 }
```
yolocs, line 75: We previously made this because it allows us to set CPU/memory defaults separately. Now since we either set defaults for the whole component or not, maybe we should make it flat again? Like putting every parameter in the `ComponentParameters`.

pkg/apis/intevents/v1alpha1/brokercell_validation.go
```diff
@@ -89,6 +89,13 @@ func (componentParams *ComponentParameters) ValidateAutoscalingSpecification(fie
 			}
 		}
 	}
+	// At least one of the autoscaling metrics should be specified
+	isAvgMemoryUsageSpecified := componentParams.AvgMemoryUsage != nil && *componentParams.AvgMemoryUsage != ""
+	if componentParams.AvgCPUUtilization == nil && !isAvgMemoryUsageSpecified {
+		invalidValueError := apis.ErrInvalidValue(nil, componentPath)
+		invalidValueError.Details = "At least one of the autoscaling metrics (AvgCPUUtilization, AvgMemoryUsage) should be specified"
+		fieldErrors = fieldErrors.Also(invalidValueError)
+	}
```
yolocs, line 98: Questionable. Maybe add a TODO with a link to an issue. We need to figure out whether we should allow disabling HPA.

Ectelion: Makes sense, there are few ways to approach this. I have logged [the task](https://github.com/google/knative-gcp/issues/1632) and outlined the options I see.

pkg/apis/intevents/v1alpha1/brokercell_types.go
Ectelion: I think in the previously suggested structure there is additional value: the nested structure mimics the [Kubernetes format for defining requests and limits](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#requests-and-limits), which I suspect the target audience for the feature might be more comfortable with, compared to our custom structure. No problem with reverting this though. What do you think?

yolocs: Sorry I don't recall this purpose. After all this will be a very different API (than the k8s API). The current form is also not similar to the k8s API. Not even to mention the HPA part. So I wouldn't worry about difference being a problem.

Ectelion: I meant when it comes to requests and limits for CPU and memory, the HPA format is different indeed. Ok, sounds like we'd rather switch to the flat structure, I'll follow up on this.

Ectelion: Done.

PR #1607, opened 2020-08-19
Change the doc to indicate both serving and eventing are required

### LINKED ISSUE #1468: Can't install if Knative Eventing is not installed
**Describe the bug**
Since the controller now depends on Knative Eventing CRDs, it's possible the controller will crash unless Knative Eventing is also installed. I haven't confirmed this.

 **Expected behavior**
I can install and create a knative-gcp source pointing directly at a consumer without also having Knative Eventing installed.

 **Additional context**
I'm not sure if the expected behavior above is actually expected or a high priority.

cathyzhyi:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/docs/install/install-knative-gcp.md
+++ b/docs/install/install-knative-gcp.md
@@ -55,11 +55,10 @@ export KO_DOCKER_REPO='gcr.io/my-gcloud-project-id'
    > if you are using Google Container Registry or `docker login` if you are
    > using Docker Hub.
 
-1. Install [Knative](https://knative.dev/docs/install/). Preferably, set up both
+1. Install [Knative](https://knative.dev/docs/install/). Set up
    [Serving](https://knative.dev/docs/serving/) and
-   [Eventing](https://knative.dev/docs/eventing/). The latter is only required
-   if you want to use the Pub/Sub `Channel` or a `Broker` backed by a Pub/Sub
-   `Channel`.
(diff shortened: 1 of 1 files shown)


docs/install/install-knative-gcp.md
```diff
@@ -55,11 +55,10 @@ export KO_DOCKER_REPO='gcr.io/my-gcloud-project-id'
    > if you are using Google Container Registry or `docker login` if you are
    > using Docker Hub.
 
-1. Install [Knative](https://knative.dev/docs/install/). Preferably, set up both
+1. Install [Knative](https://knative.dev/docs/install/). Set up
    [Serving](https://knative.dev/docs/serving/) and
-   [Eventing](https://knative.dev/docs/eventing/). The latter is only required
```
capri-xiyue, line 60: Do we need eventing if we don't use Pub/Sub `Channel` or a `Broker` backed by a Pub/Sub
   `Channel`?

cathyzhyi: According to https://github.com/google/knative-gcp/issues/1468#issuecomment-665286067 and https://github.com/google/knative-gcp/issues/1468#issuecomment-662111600 we need to install eventing but we might want to break this dependency in the future.

grac3gao: `gcp-broker` uses Eventing's `broker` CRD, it also needs eventing to be installed.

PR #1620, opened 2020-08-20
Add pod priority doc

### LINKED ISSUE #1525: Define Pod Priority for broker pods
**Problem**
As  https://github.com/google/knative-gcp/issues/1502#issuecomment-664793074 indicates, we might need to find a proper value for broker pods' `Priority`, in order to avoid pods will be preempted when node is overcommitted. 

**Exit Criteria**

**Time Estimate (optional):** 

grac3gao:
Fixes #1525 

## Proposed Changes

- provide a pod priority doc
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/docs/how-to/broker-priority.md
+++ b/docs/how-to/broker-priority.md
@@ -1,0 +1,83 @@
+# Configuring Pod Priority for GCP-Broker
+
+## Background
+The data plane components of `GCP-broker` are running on Kubernetes clusters in `cloud-run-events` namespace. 
+Their resource consumption depends on the number of cluster nodes and number of pods running in total. 
+
+As existing components use more resources and as we add more new components, more cluster and node resources
+are required to run Kubernetes. This [summary](https://github.com/google/knative-gcp/issues/1502#issuecomment-664793074)  
+illustrates the broker components may at risk to be OOMKilled/evicted/preempted when node is over-committed. 
+
+To protest broker components from eviction or preemption, we would like to give them higher scheduling/eviction priority.
+However, this would put other applications at risk in the case of insufficient resource. To balance the needs, 
+we have a scaling mechanism for broker components, and an on-going work to provide configurable resource request and limit.
+
+Beside scaling and config, users are always able to set up pod priority for the broker components 
+if they regard the broker components a high priority components and would like to further reduce its possibility
+to be the victim of preemption. Follow this doc to set up a Pod Priority for your broker components.
+
+## Add PriorityClass for broker
+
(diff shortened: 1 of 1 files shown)


docs/how-to/broker-priority.md
```diff
@@ -0,0 +1,83 @@
...
+
+## Background
+The data plane components of `GCP-broker` are running on Kubernetes clusters in `cloud-run-events` namespace. 
+Their resource consumption depends on the number of cluster nodes and number of pods running in total. 
+
+As existing components use more resources and as we add more new components, more cluster and node resources
+are required to run Kubernetes. This [summary](https://github.com/google/knative-gcp/issues/1502#issuecomment-664793074)  
+illustrates the broker components may at risk to be OOMKilled/evicted/preempted when node is over-committed. 
+
+To protest broker components from eviction or preemption, we would like to give them higher scheduling/eviction priority.
```
grantr, line 11: protest -> protect

docs/how-to/broker-priority.md
```diff
@@ -0,0 +1,83 @@
...
+However, this would put other applications at risk in the case of insufficient resource. To balance the needs, 
+we have a scaling mechanism for broker components, and an on-going work to provide configurable resource request and limit.
+
+Beside scaling and config, users are always able to set up pod priority for the broker components 
+if they regard the broker components a high priority components and would like to further reduce its possibility
```
grantr, line 16: a -> as

docs/how-to/broker-priority.md
```diff
@@ -0,0 +1,83 @@
...
+    
+    You are going to update `deployment`s:
+    
+    ```
+    NAME                         READY   UP-TO-DATE   AVAILABLE   AGE
+    default-brokercell-fanout    1/1     1            1           4h59m
+    default-brokercell-ingress   1/1     1            1           4h59m
+    default-brokercell-retry     1/1     1            1           4h59m
+    ```
+3) Add `PriorityClass` to each deployment. This field is located at `spec.template.spec` of the `deployment`, looks like:
```
grantr, line 60: Does the brokercell controller revert the change to this field during reconcile?

grantr, line 60: It might be useful to note that this change will cause a new pod to be created, and the existing pod to be deleted after the new pod is available.

grac3gao: The brokercell controller will not revert the change if we add new brokers to the cluster or just delete some broker Pods. But if we delete any of the deployments, the newly created deployment no longer has this field. (If delete fanout deployment, re-created fanout pod will not have expected Priority, however, retry and ingress still have it)

grac3gao: Done.

PR #1639, opened 2020-08-26
Extended the Probe Helper for use with CloudStorageSource probe

tommyreddad:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


test/test_images/probe_helper/probe_helper.go
```diff
@@ -87,51 +106,141 @@ func forwardFromProbe(ctx context.Context, brokerClient cloudevents.Client, pubs
...
+
 			// The pubsub client forwards the event as a message to a pubsub topic.
 			if res := pubsubClient.Send(ctx, event); !cloudevents.IsACK(res) {
 				log.Printf("Error when publishing event %v to pubsub topic: %+v \n", event.ID(), res)
 				return res
 			}
+		case CloudStorageSourceProbeEventType:
+			obj := bucket.Object(event.ID())
+			switch event.Subject() {
+			case CloudStorageSourceProbeCreateSubject:
```
liu-cong, line 139: Can we use the `schemasv1.CloudStorageObjectFinalizedEventType` as event type and get rid of the subject?

test/test_images/probe_helper/probe_helper.go
```diff
@@ -87,51 +106,141 @@ func forwardFromProbe(ctx context.Context, brokerClient cloudevents.Client, pubs
...
 			}
+			defer receivedEvents.deleteReceiverChannel(channelID)
+
 			// The pubsub client forwards the event as a message to a pubsub topic.
 			if res := pubsubClient.Send(ctx, event); !cloudevents.IsACK(res) {
 				log.Printf("Error when publishing event %v to pubsub topic: %+v \n", event.ID(), res)
 				return res
 			}
+		case CloudStorageSourceProbeEventType:
+			obj := bucket.Object(event.ID())
```
liu-cong, line 137: If I understand this correctly, we assume all four different types of CloudStorage probes are executed with the same event ID in a certain order, right? If so, please document it.

test/test_images/probe_helper/probe_helper.yaml
```diff
@@ -39,6 +39,8 @@ spec:
               value: "http://default-brokercell-ingress.cloud-run-events.svc.cluster.local/cloud-run-events-probe/default"
             - name: CLOUDPUBSUBSOURCE_TOPIC_ID
               value: "cloudpubsubsource-topic"
+            - name: CLOUDSTORAGESOURCE_BUCKET_ID
+              value: "project-id-cloudstoragesource-bucket"
```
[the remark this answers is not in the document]
tommyreddad: Since Cloud Storage bucket names are global (not project-specific), I felt it made sense to include the project name within. It's not really important in the default config anyway, so I can remove it, but that's the reason for the scheme as it stands.

test/test_images/probe_helper/probe_helper.go
tommyreddad: This is doable but requires some changes. Currently, the object created by the probe is named by the event ID, which is prefixed by the event type (which is the probe type), e.g. event.Type() == cloudstoragesource-probe, event.ID() == cloudstoragesource-probe-1234567890. If we want the distinct probes here to be able to reference the same object without having the same event type, then we can either change the event ID scheme to not contain the event type (not a bad idea in my opinion), or add some extra string processing in the probe helper to remove the event type prefix from the event ID (not so elegant). Thoughts?

liu-cong: Thought about this a bit more. The "Event Type" in the probe request is more like "Probe Type", so it doesn't necessarily map to the real event type.  I think what you have is OK.

test/test_images/probe_helper/probe_helper.yaml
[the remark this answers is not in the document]
liu-cong: ACK

test/test_images/probe_helper/probe_helper.go
tommyreddad: Documented the probes in [main.go](https://github.com/google/knative-gcp/pull/1639/files#diff-502ab32f599d1dd7e41dd1c389b6bf5cR16-R60)

test/test_images/probe_helper/probe_helper.go
```diff
@@ -87,51 +106,141 @@ func forwardFromProbe(ctx context.Context, brokerClient cloudevents.Client, pubs
...
 		case <-receiverChannel:
-			receivedEvents.Lock()
-			delete(receivedEvents.channels, event.ID())
-			receivedEvents.Unlock()
 			return cloudevents.ResultACK
 		case <-ctx.Done():
-			log.Printf("Timed out waiting for the event to be sent back: %v \n", event.ID())
-			return cloudevents.NewReceipt(false, "Timed out waiting for event to be sent back")
+			log.Printf("Timed out waiting for the receiver channel: %v \n", channelID)
+			return cloudevents.ResultNACK
```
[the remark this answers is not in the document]
tommyreddad: > We should also add tests

Done. Added tests for all the probes in [probe_helper_test.go](https://github.com/google/knative-gcp/blob/233fc287f76b73765942aa4ff5d1ab9025acecb2/test/test_images/probe_helper/probe_helper_test.go)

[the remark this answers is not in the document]
tommyreddad: > let's keep the message

Done.

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -29,52 +49,346 @@ const (
...
+				if res := c.Send(ctx, *deletedEvent); !cloudevents.IsACK(res) {
+					logger.Fatalf("Failed to send object deleted CloudEvent from the dummy CloudStorageSource: %v", res)
+				}
+			}
+		}
+	}
+}
+
+//
+type cloudEvent struct {
```
liu-cong, line 198: why not using CloudEvent SDK type directly?

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -29,52 +49,346 @@ const (
...
+			CeID:      name + "-1234567890",
+			CeSubject: subject,
+			CeSource:  "probe-helper-test",
+			CeType:    name,
+		},
+	)
+}
+
+func newTestServer(handler func(w http.ResponseWriter, r *http.Request)) (*http.Client, func()) {
+	ts := httptest.NewTLSServer(http.HandlerFunc(handler))
```
liu-cong, line 229: Can you use httptest.NewServer?

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -29,52 +49,346 @@ const (
...
+	}
+	return c, close
+}
+
+type eventAndResult struct {
+	event      *cloudevents.Event
+	wantResult protocol.Result
+}
+
+type requestData struct {
```
liu-cong, line 265: It seems that you can just use http.Request

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -29,52 +49,346 @@ const (
...
+		if err := sub.Receive(ctx, msgHandler); err != nil {
+			logger.Fatalf("Could not receive from subscription: %v", err)
+		}
+	}
+}
+
+// A helper function that starts a dummy CloudStorageSource which intercepts
+// Cloud Storage HTTP requests and forwards the appropriate notifications as
+// CloudEvents to the probe helper's receive port.
+func runDummyCloudStorageSource(ctx context.Context, gotRequest chan requestData) {
```
liu-cong, line 127:
```suggestion
func runTestCloudStorageSource(ctx context.Context, gotRequest chan requestData) {
```

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -29,52 +49,346 @@ const (
...
-	// start a goroutine to run the dummy probe helper
-	go runProbeHelper()
-	// start a goroutine to run the dummy broker
-	go runDummyBroker(t)
-	probeHelperURL := fmt.Sprintf("http://localhost:%d", probeHelperPort)
-	p, err := cloudevents.NewHTTP(cloudevents.WithTarget(probeHelperURL))
+// A helper function that starts a dummy CloudPubSubSource which watches a
+// pubsub Subscription for messages and delivers them as CloudEvents to the
+// probe helper's receive port.
+func runDummyCloudPubSubSource(ctx context.Context, sub *pubsub.Subscription) {
```
liu-cong, line 97:
```suggestion
func runTestCloudPubSubSource(ctx context.Context, sub *pubsub.Subscription) {
```

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -29,52 +49,346 @@ const (
...
+	testStorageArchiveBody        = fmt.Sprintf(`{"bucket":"%s","name":"cloudstoragesource-probe-1234567890","storageClass":"ARCHIVE"}`, testStorageBucket)
 )
 
-// A helper function that starts a dummy broker which receives events forwarded by the probe helper and delivers the events
-// back to the probe helper's receive port
-func runDummyBroker(t *testing.T) {
-	probeReceiverURL := fmt.Sprintf("http://localhost:%d", probeReceiverPort)
+// A helper function that starts a dummy Broker which receives events forwarded by
+// the probe helper and delivers the events back to the probe helper's receive port.
+func runDummyBroker(ctx context.Context) {
```
liu-cong, line 77:
```suggestion
func runTestBroker(ctx context.Context) {
```

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -29,52 +49,346 @@ const (
...
+}
+
+type requestData struct {
+	method string
+	url    string
+	body   string
+}
+
+func TestProbeHelper(t *testing.T) {
+	t.Skip("Skip this test from running on Prow as it is only for local development.")
```
liu-cong, line 272: Just to confirm, did you try running these tests and make sure they pass?

Context: We initially skipped the tests because currently the probeHelper uses a hard-coded port which may not be always available during tests. Ideally we should use httptest.NewServer

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: No good reason, just a mistake! Fixed.

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: Yes, fixed.

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: Yes, with some extra work to make sure the request body is not erased. Done.

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: Done.

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: Done.

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: Done.

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: I did run the tests and they work. I removed the line to skip the test in the latest commit, do you also mean I should replace the way the test Broker works (with the hardcoded port 9999) with httptest.NewServer?

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -348,31 +318,23 @@ func TestProbeHelper(t *testing.T) {
...
-				gotRequest <- requestData{
-					method: r.Method,
-					url:    r.URL.String(),
-					body:   string(body),
-				}
-				w.Write([]byte("{}"))
-			})
-			defer close()
-			storageClient, err := storage.NewClient(ctx, option.WithHTTPClient(hc))
+			storageClient, gotRequest, cancel := testStorageClient(ctx, t)
```
liu-cong, line 321: I think we can run these setup steps prior to each test case.

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -257,21 +216,32 @@ func testPubsubClient(ctx context.Context, t *testing.T, projectID string) (*pub
...
-	method string
-	url    string
-	body   string
-}
-
 func TestProbeHelper(t *testing.T) {
-	t.Skip("Skip this test from running on Prow as it is only for local development.")
-
-	os.Setenv("K_SINK", dummyBrokerURL)
+	os.Setenv("K_SINK", testBrokerURL)
```
liu-cong, line 244: We should either use t.Skip() or fix the flakiness of this test. For now we can continue using t.Skip.

test/test_images/probe_helper/probe_helper.go
```diff
@@ -352,5 +353,9 @@ func runProbeHelper(ctx context.Context, pubsubClient *pubsub.Client, storageCli
 	go sc.StartReceiver(ctx, forwardFromProbe(ctx, sc, psc, bkt, receivedEvents, timeout))
 	// Receive the event and return the result back to the probe
 	logger.Info("Starting event receiver...")
-	rc.StartReceiver(ctx, receiveEvent(ctx, receivedEvents))
+	go rc.StartReceiver(ctx, receiveEvent(ctx, receivedEvents))
+
+	if ready != nil {
```
liu-cong, line 358: Is this needed?

test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -14,67 +14,343 @@ limitations under the License.
...
+			{
+				event:      probeEvent("unrecognized-probe-type", ""),
+				wantResult: cloudevents.ResultNACK,
+			},
+		},
+	}}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			ctx := logtest.TestContextWithLogger(t)
+			ctx = WithProjectKey(ctx, testProjectID)
```
liu-cong, line 300: Where is WithTopicKey defined? I couldn't find it

tommyreddad: https://github.com/google/knative-gcp/blob/master/pkg/pubsub/adapter/context/topic.go#L27

liu-cong: Thanks.

I don't think these are useful since we only have one topic in the tests. Same for other context attributes.

tommyreddad: I needed these in order to get the converter in the fake CloudPubSubSource to work: https://github.com/google/knative-gcp/blob/master/pkg/pubsub/adapter/converters/pubsub.go

I can instead manually craft the CloudEvent, if you think that is better?

liu-cong: OK I see. Never mind then.

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: Done.

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: Done.

test/test_images/probe_helper/probe_helper.go
tommyreddad: Not needed: I was trying to figure out a way to improve test flakiness, since there seemed to be a race between the probe helper starting and the testing client sending events to the probe helper, but this channel-based approach did not even really work after testing. Fixed this race by passing explicit listeners for the probe helper to use from the probe helper tests.

test/test_images/probe_helper/probe_helper_test.go
liu-cong: A hardcoded port isn't always guaranteed to be available during test and therefore creates flakiness. That's why we skipped the tests in CI/CD and only run it manually.

I am OK if we keep the Skip or we use something like httptest.NewServer to dynamically assign a port.

PR #1655, opened 2020-08-31
Add Tracing Cloud Logging fields

### LINKED ISSUE #1577: Integrate Traces with Cloud Logging
**Problem**
Attaching trace and span IDs to log entries causes those log entries to be linked from the corresponding trace and vice-versa. Enabling this integration would allow the removal of some duplication of messages between trace events and logs and improve the usability of both logging and tracing.

Event Consumers and Producers.

**Exit Criteria**
Manual verification that log messages are correlated with traces in Cloud Logging.

**Time Estimate (optional):** 
7-10 days.

 See https://cloud.google.com/trace/docs/trace-log-integration.

ian-mi:
Fixes #1577

## Proposed Changes

- Add trace related Cloud Logging fields. Allows Cloud Logging to associate log entries with traces so that log entries show up in Cloud Trace and conversely Cloud Logging will add tracing information to the log viewer. Trace IDs are logged even when not sampled as this allows the log viewer to filter on the trace ID to display only logs belonging to a given request even when tracing is not enabled for that request.

### DIFF
--- a/pkg/broker/ingress/handler.go
+++ b/pkg/broker/ingress/handler.go
@@ -114,9 +114,7 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
 
 	ctx := request.Context()
 	ctx = logging.WithLogger(ctx, h.logger)
-	if span := trace.FromContext(ctx); span != nil {
-		ctx = tracing.WithLogging(ctx, span)
-	}
+	ctx = tracing.WithLogging(ctx, trace.FromContext(ctx))
 	logging.FromContext(ctx).Debug("Serving http", zap.Any("headers", request.Header))
 	if request.Method != nethttp.MethodPost {
 		response.WriteHeader(nethttp.StatusMethodNotAllowed)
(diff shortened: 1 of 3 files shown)


pkg/broker/ingress/handler.go
```diff
@@ -113,20 +113,24 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
 	}
 
 	ctx := request.Context()
-	h.logger.Debug("Serving http", zap.Any("headers", request.Header))
+	ctx = logging.WithLogger(ctx, h.logger)
+	if span := trace.FromContext(ctx); span != nil {
+		ctx = tracing.WithLogging(ctx, span)
+	}
+	logging.FromContext(ctx).Debug("Serving http", zap.Any("headers", request.Header))
```
Harwayne, line 120: Is this intended to stay in?

Harwayne:
/lgtm
/hold
Holding for comments.

pkg/broker/ingress/handler.go
ian-mi: Yes, this was already present, I've just changed it to use the logger from ctx.

Issue #964, opened 2020-04-28
Investigate limits with targets-config configmap 


liu-cong:
**Problem**
We use a configmap to store states of brokers. The configmap has a limit of 1MB(etcd limit). Therefore we need to investigate what our limits are.

* Determine the max size of a trigger
* Determine the max size of a broker with N triggers
* Come up with how many broker/triggers can fit in the configmap
* Depending on the numbers, it may or may not be a big issue.
  * Either implement validation logic to avoid hitting the limit or advise new approaches

PR #1660, opened 2020-09-02
Use custom equality tester for TargetsConfig in ConfigMap reconciler

### LINKED ISSUE #1611: Observed brokercell pod update rate of 40/s
**Describe the bug**
A GKE investigation identified some dev-owned GKE clusters that are updating brokercell-fanout, brokercell-ingress, and brokercell-retry pods 40 times per second. Whether this is steady state update or peak, it's very likely to be a bug in the brokercell controller somewhere. I can't think of a reason to be updating pods 40 times in a second.

Probably it's related to the annotation update code that forces the configmap volume refresh.

 **Expected behavior**
Brokercell pods are updated a few times per second, only when the configmap content changes.

 **To Reproduce**
Unknown

Unknown

 **Additional context**
This may be related to #1593.

tommyreddad:
Fixes #1611

## Proposed Changes

- Allow for custom equality testers in the ConfigMap reconciler instead of always relying on k8s equality.Semantic.DeepEqual
- Use a custom equality tester for TargetsConfig which uses proto.Equal for comparing unmarshaled binary proto data


pkg/reconciler/utils/configmap.go
```diff
@@ -57,8 +57,12 @@ func (r *ConfigMapReconciler) ReconcileConfigMap(obj runtime.Object, cm *corev1.
 	if err != nil {
 		return nil, err
 	}
-
-	if !equality.Semantic.DeepEqual(cm.BinaryData, current.BinaryData) || !equality.Semantic.DeepEqual(cm.Data, current.Data) {
+	if eqFunc == nil {
+		eqFunc = func(cm1, cm2 *corev1.ConfigMap) bool {
+			return equality.Semantic.DeepEqual(cm1.BinaryData, cm2.BinaryData) && equality.Semantic.DeepEqual(cm1.Data, cm2.Data)
```
liu-cong, line 62: The default eqFunc should only look at cm.Data only

pkg/reconciler/brokercell/broker_targets_config.go
```diff
@@ -133,6 +136,36 @@ func (r *Reconciler) addToConfig(ctx context.Context, b *brokerv1beta1.Broker, t
...
+	if !equality.Semantic.DeepEqual(cm1.Data, cm2.Data) {
+		return false
+	}
+	// The broker targets ConfigMap BinaryData holds the serialized TargetsConfig
+	// proto, and therefore cannot be safely compared with equality.Semantic.DeepEqual.
+	// Instead, use proto.Equal to compare protos.
+	if len(cm1.BinaryData) != len(cm2.BinaryData) {
+		return false
+	}
+	v1Proto := &config.TargetsConfig{}
```
liu-cong, line 149: nit: would rename to something like proto1 to not confuse with proto versions.

pkg/reconciler/utils/configmap.go
```diff
@@ -57,8 +57,12 @@ func (r *ConfigMapReconciler) ReconcileConfigMap(obj runtime.Object, cm *corev1.
 	if err != nil {
 		return nil, err
 	}
-
-	if !equality.Semantic.DeepEqual(cm.BinaryData, current.BinaryData) || !equality.Semantic.DeepEqual(cm.Data, current.Data) {
+	if eqFunc == nil {
```
liu-cong, line 60: How about exporting the default eqFunc so that callers can use it if they want, and don't allow a nil eqFunc?  I think this is cleaner to use.

pkg/reconciler/utils/configmap.go
tommyreddad: In this case, eqFunc only looking at cm.Data would also fix #1611 . Do we still want to test for a difference in cm.BinaryData for TargetsConfig, even if looking only at cm.Data suffices to fix the bug?

liu-cong: cm.Data is meant for debugging purpose and we may get rid of it in the future to reduce configmap size, see #964.

tommyreddad: Ack. Fixed.

pkg/reconciler/brokercell/broker_targets_config.go
tommyreddad: Done.

pkg/reconciler/utils/configmap.go
tommyreddad: Done. Adjusted tests accordingly.

pkg/reconciler/brokercell/broker_targets_config.go
```diff
@@ -133,6 +136,36 @@ func (r *Reconciler) addToConfig(ctx context.Context, b *brokerv1beta1.Broker, t
...
+func targetsConfigMapEqual(cm1, cm2 *corev1.ConfigMap) bool {
+	if !equality.Semantic.DeepEqual(cm1.Data, cm2.Data) {
+		return false
+	}
+	// The broker targets ConfigMap BinaryData holds the serialized TargetsConfig
+	// proto, and therefore cannot be safely compared with equality.Semantic.DeepEqual.
+	// Instead, use proto.Equal to compare protos.
+	if len(cm1.BinaryData) != len(cm2.BinaryData) {
+		return false
+	}
```
yolocs, line 148: Should we skip this? If binary aren't equal, it doesn't necessarily mean proto aren't equal, right?

tommyreddad: The length comparison just checks that the number of keys in the BinaryData maps are the same, it doesn't compare the binary data itself.

I guess since this function is specialized for the broker-targets ConfigMap, the function can be written more specifically to only consider the appropriate key. How does that sound?

If you mean the Data deep equal, I can remove it, since it is redundant.

yolocs: Both points sound good to me.

tommyreddad: Done.

pkg/reconciler/utils/configmap_test.go
```diff
@@ -70,23 +71,29 @@ type commonCase struct {
 	wantErr    bool
 }
 
+func testConfigMapBinaryDataEqual(cm1, cm2 *corev1.ConfigMap) bool {
```
liu-cong, line 74: Doesn't seem to be useful to create an eqFunc and test it. Instead, I think we should test the TargetsConfigMapEqual function.

tommyreddad: Agreed. But I think it still makes sense to test using a custom eqFunc, so I'll change it to only test the simple always-true or always-false eqFuncs. Also, since the ConfigMap reconciler is technically responsible for handling all kinds of ConfigMaps, not just the broker-targets ConfigMap, then I will add the TargetsConfigMapEqual test in a separate file.

liu-cong: Thanks for doing that!

I was wondering if we can add a test to show that the custom eq func returns different result than a deepEqual for two semantically equal protos. For example, one proto with a nil map and the other with an empty map. Instead of a made up one, it's even better if we can reproduce the deepEqual bug in our real reconciliation and prove that the custom func solves that.

tommyreddad: Makes sense to want to test it, but it seems to be a bit more subtle to recreate than simply comparing a proto with a nil map vs empty map. I'm struggling to recreate it using hand-crafted protos (it might have to do with internal proto state that seems tricky to force from the outside), but I can do it using binary data extracted from my real tests. Testing it with this data would involve hard-coding the binary data in the test file. Do you think this is alright?

liu-cong: Didn't realize how hard is it to craft the data. I am fine with nil vs empty map approach. This is good enough to demonstrate the flaw of deepEqual.

tommyreddad: The problem is that marshalling a proto containing a nil map vs empty map will actually create the same byte representation in my testing. So DeepEqual would be good enough.

After more testing, the problem seems to also be caused by the proto marshaller respecting the ordering of map keys. So, protos containing identical maps with keys listed in different order will have a different binary representation. I'll add this test, since this captures the problem we faced in the issue.

liu-cong: Good to know. Thanks for looking into this!

PR #1661, opened 2020-09-02
Reduce Metric test flakiness caused by sender pod retry sending events

grac3gao:
Fixes #

## Proposed Changes

- Reduce Metric test flakiness caused by sender pod retry sending events
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/e2e/lib/broker.go
+++ b/test/e2e/lib/broker.go
@@ -73,7 +73,7 @@ func (a BrokerMetricAssertion) Assert(client *monitoring.MetricClient) error {
 
 		// Workarounds to reduce test flakiness caused by sender pod retry sending events (which will cause unexpected response code).
 		// We should remove it after https://github.com/google/knative-gcp/issues/1058 lands
-		if code == http.StatusForbidden || code == http.StatusServiceUnavailable || code == http.StatusInternalServerError {
+		if code == http.StatusNotFound || code == http.StatusServiceUnavailable || code == http.StatusInternalServerError {
 			continue
 		}
 
--- a/test/e2e/lib/broker.go
+++ b/test/e2e/lib/broker.go
@@ -71,8 +71,9 @@ func (a BrokerMetricAssertion) Assert(client *monitoring.MetricClient) error {
 			return fmt.Errorf("metric has invalid response code label: %v", ts.GetMetric())
 		}
 
-		// Workarounds to reduce test flakiness caused by sender pod retry sending events (which will cause unexpected response code).
-		// We should remove it after https://github.com/google/knative-gcp/issues/1058 lands
+		// Workarounds to reduce test flakiness caused by sender pod retry sending events (which will cause unexpected response code 404, 503 and 500).
+		// StatusCode 500 is currently for reducing flakiness caused by Workload Identity credential sync up.
+		// We would remove it after https://github.com/google/knative-gcp/issues/1058 lands, as 500 error may indicate bugs in our code.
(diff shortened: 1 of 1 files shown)


test/e2e/lib/broker.go
```diff
@@ -70,6 +70,13 @@ func (a BrokerMetricAssertion) Assert(client *monitoring.MetricClient) error {
 		if err != nil {
 			return fmt.Errorf("metric has invalid response code label: %v", ts.GetMetric())
 		}
+
+		// Workarounds to reduce test flakiness caused by sender pod retry sending events (which will cause unexpected response code).
+		// We should remove it after https://github.com/google/knative-gcp/issues/1058 lands
+		if code == http.StatusForbidden || code == http.StatusServiceUnavailable || code == http.StatusInternalServerError {
```
liu-cong, line 76: I think we would expect StatusNotFound instead of StatusForbidden?

test/e2e/lib/broker.go
```diff
@@ -70,6 +70,13 @@ func (a BrokerMetricAssertion) Assert(client *monitoring.MetricClient) error {
 		if err != nil {
 			return fmt.Errorf("metric has invalid response code label: %v", ts.GetMetric())
 		}
+
+		// Workarounds to reduce test flakiness caused by sender pod retry sending events (which will cause unexpected response code).
```
liu-cong, line 74: If I understand it correctly, the sender pod will retry if it receives non 202 code? I think the comment could clarify this.

I actually think this fix is desired even if #1058 is fixed. We are essentially saying that we only count for successful events, assuming the broker could return non 202 during tests.

test/e2e/lib/broker.go
grac3gao: Thanks! I intends to use StatusNotFound, but choose a wrong one...

test/e2e/lib/broker.go
liu-cong: If my comment above is correct, I think we can just `continue` on line #81 instead of returning an error. We rely on successful event count to verify the test correctness.

grac3gao: The sender pod will only retry for three statusCode, 404, 500 and 503. Maybe we can remove 500 if #1058 is fixed, as 500 here is mainly for reducing flakiness caused by WI credential sync (This might be fixed after #1058 lands)? And continue keep 403 and 503?

liu-cong: It depends on what the expectations are but It doesn't seem like a bad idea for the sender pod to retry on non-202, even if we fix those issues. This is not a broker availability test anyways. But I am also OK with dealing with each code separately.

grac3gao: Sounds good, I just updated the comments. Could you check if that is ok?

test/e2e/lib/broker.go
```diff
@@ -70,6 +70,14 @@ func (a BrokerMetricAssertion) Assert(client *monitoring.MetricClient) error {
 		if err != nil {
 			return fmt.Errorf("metric has invalid response code label: %v", ts.GetMetric())
 		}
+
+		// Workarounds to reduce test flakiness caused by sender pod retry sending events (which will cause unexpected response code 404, 503 and 500).
```
liu-cong, line 74: I think it's the unexpected response code that triggers the sender to retry, not the other way around. We can say something like "ignore undesired response code in metric count comparison due to flakiness in broker, the sender pod will retry until event is successfully received".

PR #1665, opened 2020-09-02
Use different TraceID for each retry in sender pod to reduce tracing test flakiness 

grac3gao:
Fixes #

## Proposed Changes

- Export sender pod span to reduce tracing test flakiness 
-
-

**Release Note**

```release-note

```

**Docs**


test/test_images/sender/main.go
```diff
@@ -43,7 +44,7 @@ type envConfig struct {
 // defaultRetry represents that there will be 3 iterations.
```
grantr, line 44: Comment now inaccurate: 4 iterations by default.

test/test_images/sender/main.go
```diff
@@ -43,7 +44,7 @@ type envConfig struct {
...
 // The duration starts from 30s and is multiplied by factor 1.0 for each iteration.
```
grantr, line 45: Same: factor is now 2.0.

grantr (approved):
/lgtm

Minor optional comment nits. Thanks @grac3gao and @ian-mi for collaborating on this!

PR #1666, opened 2020-09-02
Better ingress error message on permission denied

### LINKED ISSUE #1631: Better ingress error message on pubsub authentication problem
**Describe the bug**

On pubsub authentication failure, the ingress nowadays just returns 500 with plain pubsub authentication error. 

 **Expected behavior**
We want to provide additional context to the error message. E.g. Pointer to the init/setup doc.

 **To Reproduce**
Use broker without going through init setup

cathyzhyi:
Fixes #1631 

## Proposed Changes

-
-
-

**Release Note**

```release-note
Provide additional error info when auth with pubsub fails in broker ingress.
```

**Docs**

### DIFF
--- a/pkg/broker/ingress/handler.go
+++ b/pkg/broker/ingress/handler.go
@@ -34,6 +34,8 @@ import (
 	"github.com/google/wire"
 	"go.opencensus.io/trace"
 	"go.uber.org/zap"
+	grpccode "google.golang.org/grpc/codes"
+	grpcstatus "google.golang.org/grpc/status"
 	"k8s.io/apimachinery/pkg/types"
 	"knative.dev/eventing/pkg/kncloudevents"
 	"knative.dev/eventing/pkg/logging"
@@ -52,6 +54,11 @@ const (
 
 	// For probes.
 	heathCheckPath = "/healthz"
+
+	// for permission denied error msg
+	// TODO(cathyzhyi) point to official doc rather than github doc
+	deniedErrMsg string = `Failed to publish to PubSub because permission denied.
+Please refer to "Configure the Authentication Mechanism for GCP" at https://github.com/google/knative-gcp/blob/master/docs/install/install-gcp-broker.md`
 )
 
 // HandlerSet provides a handler with a real HTTPMessageReceiver and pubsub MultiTopicDecoupleSink.
@@ -158,10 +165,15 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
 	if res := h.decouple.Send(ctx, broker, *event); !cev2.IsACK(res) {
 		h.logger.Error("Error publishing to PubSub", zap.String("broker", broker.String()), zap.Error(res))
 		statusCode = nethttp.StatusInternalServerError
-		if errors.Is(res, ErrNotFound) {
+
+		switch {
+		case errors.Is(res, ErrNotFound):
 			statusCode = nethttp.StatusNotFound
-		} else if errors.Is(res, ErrNotReady) {
+		case errors.Is(res, ErrNotReady):
 			statusCode = nethttp.StatusServiceUnavailable
+		case grpcstatus.Code(res) == grpccode.PermissionDenied:
+			nethttp.Error(response, deniedErrMsg, statusCode)
+			return
 		}
 		nethttp.Error(response, "Failed to publish to PubSub", statusCode)
 		return


pkg/broker/ingress/handler.go
```diff
@@ -158,10 +160,18 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
...
+		switch {
+		case errors.Is(res, ErrNotFound):
 			statusCode = nethttp.StatusNotFound
-		} else if errors.Is(res, ErrNotReady) {
+		case errors.Is(res, ErrNotReady):
 			statusCode = nethttp.StatusServiceUnavailable
+		case grpcstatus.Code(res) == grpccode.PermissionDenied:
+			const msg string = "Failed to publish to PubSub because permission denied.\n" +
+			"Please refer to \"Configure the Authentication Mechanism for GCP\" at " +
+			"https://github.com/google/knative-gcp/blob/master/docs/install/install-knative-gcp.md"
```
grac3gao, line 172: It might be better to point broker installation page(https://github.com/google/knative-gcp/blob/master/docs/install/install-gcp-broker.md) rather than install knative-gcp page (this is for the Control Plane setup)

cathyzhyi: Thanks Grace! Modified accordingly.

pkg/broker/ingress/handler.go
```diff
@@ -158,10 +160,18 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
...
+		switch {
+		case errors.Is(res, ErrNotFound):
 			statusCode = nethttp.StatusNotFound
-		} else if errors.Is(res, ErrNotReady) {
+		case errors.Is(res, ErrNotReady):
 			statusCode = nethttp.StatusServiceUnavailable
+		case grpcstatus.Code(res) == grpccode.PermissionDenied:
+			const msg string = "Failed to publish to PubSub because permission denied.\n" +
+				"Please refer to \"Configure the Authentication Mechanism for GCP\" at " +
+				"https://github.com/google/knative-gcp/blob/master/docs/install/install-gcp-broker.md"
```
yolocs, line 172: At the top you can define a constant like

```
const deniedErrMsg string = `Failed to publish to PubSub because permission denied.
Please refer to "Configure the Authentication Mechanism for GCP" at https://github.com/google/knative-gcp/blob/master/docs/install/install-gcp-broker.md
`
```

And use it here.

pkg/broker/ingress/handler.go
```diff
@@ -158,10 +160,18 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
...
-		if errors.Is(res, ErrNotFound) {
+
+		switch {
+		case errors.Is(res, ErrNotFound):
 			statusCode = nethttp.StatusNotFound
-		} else if errors.Is(res, ErrNotReady) {
+		case errors.Is(res, ErrNotReady):
 			statusCode = nethttp.StatusServiceUnavailable
+		case grpcstatus.Code(res) == grpccode.PermissionDenied:
+			const msg string = "Failed to publish to PubSub because permission denied.\n" +
```
yolocs, line 170: Also add a TODO. I think eventually we want to put a link to the official doc rather than github doc.

pkg/broker/ingress/handler.go
```diff
@@ -158,10 +160,18 @@ func (h *Handler) ServeHTTP(response nethttp.ResponseWriter, request *nethttp.Re
...
 		statusCode = nethttp.StatusInternalServerError
-		if errors.Is(res, ErrNotFound) {
+
+		switch {
+		case errors.Is(res, ErrNotFound):
 			statusCode = nethttp.StatusNotFound
-		} else if errors.Is(res, ErrNotReady) {
+		case errors.Is(res, ErrNotReady):
 			statusCode = nethttp.StatusServiceUnavailable
+		case grpcstatus.Code(res) == grpccode.PermissionDenied:
```
yolocs, line 169: Just to confirm. Have you verified this in a cluster?

yolocs:
Since this has impact on user facing function, it's best to add a release note. I helped to add it for this PR.

pkg/broker/ingress/handler.go
cathyzhyi: Yes, it works in the cluster.

pkg/broker/ingress/handler.go
cathyzhyi: Added TODO. thanks for the suggestion!

pkg/broker/ingress/handler.go
cathyzhyi: Moved the const string definition to the top of the file as suggested. Thanks!

yolocs (commented):
/lgtm
/approve

PR #1674, opened 2020-09-09
Extend the Probe Helper to work with CloudSchedulerSource probe

tommyreddad:
Extended the Probe Helper to work with the CloudSchedulerSource probe.

## Proposed Changes
- The Probe Helper can now receive probe events of type cloudschedulersource-probe
- The Probe Helper can now handle events sourced from a CloudSchedulerSource
- Added a test for the CloudSchedulerSource probe


test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -284,6 +315,9 @@ func TestProbeHelper(t *testing.T) {
 	// Run the test CloudStorageSource.
 	runTestCloudStorageSource(ctx, gotCloudStorageRequest, receiverURL)
 
+	// Run the test CloudSchedulerSource.
+	runTestCloudSchedulerSource(ctx, time.Second, receiverURL)
```
liu-cong, line 319: nit: set a shorter period to speed up the test.

test/test_images/probe_helper/probe_helper.go
```diff
@@ -243,6 +246,13 @@ func (ph *ProbeHelper) forwardFromProbe(ctx context.Context, brokerClient cloude
 			default:
 				return logNACK(ctx, "Probe forwarding failed, unrecognized cloud storage probe subject: %v", event.Subject())
 			}
+		case CloudSchedulerSourceProbeEventType:
+			channelID = cloudSchedulerSourceProbeChannelID
```
liu-cong, line 250: If we receive a probe requests while there is another being processed, the new request will fail due to the duplicated channel ID.

I don't think we should rely on the channel approach for the scheduler any more. How about this - assuming a scheduler source running with a tick period of say, 1min. The receiver updates the lastUpdatedTimestamp. When we receive a probe, the elapsed time since lastUpdatedTimestamp should be within a minute (like the liveness probe).

test/test_images/probe_helper/probe_helper_test.go
tommyreddad: Done.

test/test_images/probe_helper/probe_helper.go
tommyreddad: Good idea. Done.

test/test_images/probe_helper/probe_helper.go
```diff
@@ -41,6 +41,9 @@ const (
 	CloudStorageSourceProbeUpdateMetadataSubject = "update-metadata"
 	CloudStorageSourceProbeArchiveSubject        = "archive"
 	CloudStorageSourceProbeDeleteSubject         = "delete"
+	CloudSchedulerSourceProbeEventType           = "cloudschedulersource-probe"
+
+	cloudSchedulerSourceProbeChannelID = "cloudschedulersource-probe-channel-id"
```
liu-cong, line 46: I think we don't need this any more?

test/test_images/probe_helper/main.go
```diff
@@ -88,6 +100,9 @@ type envConfig struct {
 	// Environment variable containing the CloudStorageSource Bucket ID that objects will be written to by the probeHelper for the CloudStorageSource probe
 	CloudStorageSourceBucketID string `envconfig:"CLOUDSTORAGESOURCE_BUCKET_ID" default:"cloudstoragesource-bucket"`
 
+	// Environment variable containing the duration between events emitted by the CloudSchedulerSource
+	CloudSchedulerSourcePeriod time.Duration `envconfig:"CLOUDSCHEDULERSOURCE_PERIOD" default:"1m"`
```
liu-cong, line 104: nit: if the scheduler source fires per minute, we should give it a bit buffer here. The default may be 90s. or even 2m?

test/test_images/probe_helper/probe_helper.yaml
```diff
@@ -40,6 +40,8 @@ spec:
               value: "cloudpubsubsource-topic"
             - name: CLOUDSTORAGESOURCE_BUCKET_ID
               value: "cloudstoragesource-bucket"
+            - name: CLOUDSCHEDULERSOURCE_PERIOD
+              value: "1m"
```
liu-cong, line 44: this should be a big longer than 1m

test/test_images/probe_helper/probe_helper.go
tommyreddad: Fixed.

test/test_images/probe_helper/main.go
tommyreddad: Done. Picked 90s.

liu-cong:
Feel free to unhold once description is updated.

Issue #611, opened 2020-03-06
Broker ingress available during upgrade


ericlem:
**Problem**
We want to ensure that all services in the ingress path of the broker can be upgraded through a rolling upgrade or activator model and not experience failures during upgrade.

**Exit Criteria**
Ingress path components are deployed in a way that future upgrades past the version in which this task's output has been deployed can be rolling upgraded or proxied via an activator, preventing unavailability or error returns during upgrade.  

A procedure for deploying an upgrade of these components without unavailability has been documented.

It should pass the tests developed here:
https://github.com/google/knative-gcp/issues/610

**Time Estimate (optional):** 
5d

grantr:
We should be able to use the sendtracker to determine the percentage of requests that were published successfully.
How do we determine the previous and next releases?
What is the load during upgrade?
We may need to adjust rolling update configuration of the deployment.

yolocs:
Test strategy:

- Send events to broker with a steady rate
- Make a change that will cause the controller + brokercell deployments to be updated (to mimic an upgrade)
- Wait a bit longer 
- Watch event reject rate, broker/trigger event count metrics

Results:

Under a fair rps that can be handled by a single pod (<=1000), the upgrade only caused neglectable ingress failures (<1%). Occasionally, the error was 503. In most cases, the error was something like `dial tcp 10.0.56.158:80: connect: no route to host`. In my understanding (**correct me if I'm wrong**), the error seems to be "a moment ago the pod was still alive and its IP was resolved from the ingress address, but when attempting to connect to it the pod was terminated and thus no route". If this is really the case, this is more or less a client side issue and it doesn't indicate our broker is not available.

![83SHNtWHvPAJB3x (1)](https://user-images.githubusercontent.com/14087829/92659057-7e91c300-f2ac-11ea-8de7-32a7136cda1c.png)

When the load was increased until autoscaling is triggered, there was an increased _amount_ of the same error "connect: no route to host". This is somehow expected because with increased concurrency, there is a higher chance more requests will get a previously valid IP.

However there is another problem.

![3ZCPCwXHymhJ7mK](https://user-images.githubusercontent.com/14087829/92659255-df210000-f2ac-11ea-97a1-e2ea38973151.png)

[...]

PR #1690, opened 2020-09-11
Update broker deployments spec for better availability during upgrades

yolocs:
Fixes https://github.com/google/knative-gcp/issues/611

## Proposed Changes

- Add `MinReadySeconds=60s` for broker deployments
- Add a safer rolling update strategy
- Relax probe period

**Release Note**

```release-note
Update broker deployments spec for better availability during upgrades
- Add `MinReadySeconds=60s` for broker deployments
- Add a safer rolling update strategy
- Relax probe period
```

**Docs**

### DIFF
--- a/pkg/reconciler/brokercell/resources/deployments.go
+++ b/pkg/reconciler/brokercell/resources/deployments.go
@@ -44,7 +44,7 @@ func MakeIngressDeployment(args IngressArgs) *appsv1.Deployment {
 			},
 		},
 		FailureThreshold: 5,
-		PeriodSeconds:    2,
+		PeriodSeconds:    15,
 		SuccessThreshold: 1,
 		TimeoutSeconds:   5,
 	}
@@ -57,8 +57,8 @@ func MakeIngressDeployment(args IngressArgs) *appsv1.Deployment {
 			},
 		},
 		FailureThreshold:    5,
-		InitialDelaySeconds: 5,
-		PeriodSeconds:       2,
+		InitialDelaySeconds: 15,
+		PeriodSeconds:       15,
 		SuccessThreshold:    1,
 		TimeoutSeconds:      5,
 	}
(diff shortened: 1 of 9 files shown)


pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -135,6 +135,13 @@ func deploymentTemplate(args Args, containers []corev1.Container) *appsv1.Deploy
 		},
 		Spec: appsv1.DeploymentSpec{
 			Selector: &metav1.LabelSelector{MatchLabels: Labels(args.BrokerCell.Name, args.ComponentName)},
+			Strategy: appsv1.DeploymentStrategy{
+				RollingUpdate: &appsv1.RollingUpdateDeployment{
+					MaxSurge:       &intstr.IntOrString{IntVal: 1},
+					MaxUnavailable: &intstr.IntOrString{IntVal: 0},
+				},
+			},
+			MinReadySeconds: 60,
```
Ectelion, line 144: Interesting investigation on the task page!
Regarding MinReadySeconds, I wonder if this applies to upgrade only or does it also apply to new replicas added during autoscaling?

yolocs: It should impact that too.

Ectelion: I see. And there is no such option that is upgrade-specific? It's just in the context of our recent work on fanout where we rely increasingly more on autoscaling, I was worrying if an additional the scale out delay of 45 seconds may come at some cost.

yolocs: I don't see any problem at the moment. This delay doesn't stop the pod from being ready and start pulling pubsub.

Ectelion: Cool, I have misinterpreted available vs ready then.

PR #1694, opened 2020-09-14
Fix obsolete comment in handler.go, add new comment to receive function

zhongduo:
Fixes obsolete comment in handler.go.

## Proposed Changes

- Add comment to the pubsub.Subscription in the handler struct.


yolocs (approved):
/lgtm

Thanks!

PR #1703, opened 2020-09-15
Finalize resource as best effort and always mark topic as ready

### LINKED ISSUE #1583: knative trigger status Unknown because TopicReady Unknown
**Describe the bug**
I create knative trigger with default broker(MTChannelBasedBroker) & trigger, it seems gcp controller would reconsile it and cause the trigger not ready.

```
Status:
  Conditions:
    Last Transition Time:  2020-08-14T05:47:51Z
    Status:                True
    Type:                  BrokerReady
    Last Transition Time:  2020-08-14T05:47:54Z
    Status:                True
    Type:                  DependencyReady
    Last Transition Time:  2020-08-14T05:48:02Z
    Status:                Unknown
    Type:                  Ready
    Last Transition Time:  2020-08-14T05:47:54Z
    Status:                True
    Type:                  SubscriberResolved
    Last Transition Time:  2020-08-14T05:48:02Z
    Status:                True
    Type:                  SubscriptionReady
    Last Transition Time:  2020-08-14T05:22:03Z
    Status:                Unknown
    Type:                  TopicReady
```

 **Expected behavior**
trigger ready

 **To Reproduce**

knative-gcp v0.16.0
knative-eventing v0.16.1

cathyzhyi:
Fixes #1583 

## Proposed Changes

-  For a trigger pointing to brokerclass other than `googlecloud`, finalize it as a best effort and always mark the trigger ready

**Release Note**

Before this change, a non GCP trigger readiness can be blocked by `TopicReady: Unknown` . With the change, the trigger readiness won't be blocked anymore.

```release-note

```

**Docs**

### DIFF
--- a/pkg/reconciler/trigger/trigger.go
+++ b/pkg/reconciler/trigger/trigger.go
@@ -106,7 +106,22 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, t *brokerv1beta1.Trigger
 		// Call Finalizer anyway in case the Trigger still holds GCP Broker related resources.
 		// If a Trigger used to point to a GCP Broker but now has a Broker with a different brokerclass,
 		// we should clean up resources related to GCP Broker.
-		return r.FinalizeKind(ctx, t)
+		event := r.FinalizeKind(ctx, t)
+
+		// If a trigger has never pointed to a GCP broker, topic readiness shouldn't block this trigger's
+		// readiness. However, without a reliable way to tell if the trigger has previously pointed
+		// to a GCP broker FinalizeKind called above and other code could potentially change the topic
+		// readiness to UNKNOWN even when it has never pointed to a GCP broker. Always mark the topic ready
+		// here to unblock trigger readiness.
+		// This code can potentially cause problems in cases where the trigger did refer to a GCP
+		// broker which got deleted and recreated with a new non GCP broker. It's necessary to do best
+		// effort GC but the topic is going to be marked ready even when GC fails. This can result in
+		// dangling topic without matching status.
+		// This line should be deleted once the following TODO is finished.
+		// TODO(https://github.com/knative/pkg/issues/1149) Add a FilterKind to genreconciler so it will
+		// skip a trigger if it's not pointed to a gcp broker and doesn't have googlecloud finalizer string.
+		t.Status.MarkTopicReady()
+		return event
 	}
 
 	return r.reconcile(ctx, t, b)
--- a/pkg/reconciler/trigger/trigger.go
+++ b/pkg/reconciler/trigger/trigger.go
@@ -121,7 +121,14 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, t *brokerv1beta1.Trigger
 		// TODO(https://github.com/knative/pkg/issues/1149) Add a FilterKind to genreconciler so it will
 		// skip a trigger if it's not pointed to a gcp broker and doesn't have googlecloud finalizer string.
 		t.Status.MarkTopicReady()
-		return event
+
+		if event == nil {
+			return nil
+		} else if reconcilerEvent, ok := event.(*pkgreconciler.ReconcilerEvent); ok && reconcilerEvent.EventType == corev1.EventTypeNormal {
+			return event
+		} else {
+			return fmt.Errorf("Controller won't retry to handle the error/warning and user will need to address it manually: %w", event)
+		}
 	}
 
 	return r.reconcile(ctx, t, b)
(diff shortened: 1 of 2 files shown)


pkg/reconciler/trigger/trigger.go
```diff
@@ -106,7 +106,13 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, t *brokerv1beta1.Trigger
 		// Call Finalizer anyway in case the Trigger still holds GCP Broker related resources.
 		// If a Trigger used to point to a GCP Broker but now has a Broker with a different brokerclass,
 		// we should clean up resources related to GCP Broker.
-		return r.FinalizeKind(ctx, t)
+		event := r.FinalizeKind(ctx, t)
+		// It impossible to tell if the trigger has ever pointed to a GCP broker. Mark topic ready to
+		// unblock the user of other brokers. This should be deleted once the TODO is finished
```
yolocs, line 111: "It's impossible".

It would be great to add an example here. Like previously was referencing a GCP broker, then it's deleted and recreated as non-gcp broker. We still want to best effort GC the resources but don't want to block the trigger being ready with the new broker.

yolocs (commented):
/lgtm
/hold
For Cong to take a look

pkg/reconciler/trigger/trigger.go
cathyzhyi: Added more detailed comments which the change deserves. Thanks for the suggestion!

pkg/reconciler/trigger/trigger.go
```diff
@@ -106,7 +106,22 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, t *brokerv1beta1.Trigger
...
+		// here to unblock trigger readiness.
+		// This code can potentially cause problems in cases where the trigger did refer to a GCP
+		// broker which got deleted and recreated with a new non GCP broker. It's necessary to do best
+		// effort GC but the topic is going to be marked ready even when GC fails. This can result in
+		// dangling topic without matching status.
+		// This line should be deleted once the following TODO is finished.
+		// TODO(https://github.com/knative/pkg/issues/1149) Add a FilterKind to genreconciler so it will
+		// skip a trigger if it's not pointed to a gcp broker and doesn't have googlecloud finalizer string.
+		t.Status.MarkTopicReady()
+		return event
```
grantr, line 124: I think we should indicate in this event that the error will not be retried and requires manual action to correct. Most other events are retried by their controllers and thus understood by users as informational only, not requiring action.

cathyzhyi: Modified the code according to the suggestion. Thanks!

pkg/reconciler/trigger/trigger.go
```diff
@@ -106,7 +106,29 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, t *brokerv1beta1.Trigger
...
+		// TODO(https://github.com/knative/pkg/issues/1149) Add a FilterKind to genreconciler so it will
+		// skip a trigger if it's not pointed to a gcp broker and doesn't have googlecloud finalizer string.
+		t.Status.MarkTopicReady()
+
+		if event == nil {
+			return nil
+		} else if reconcilerEvent, ok := event.(*pkgreconciler.ReconcilerEvent); ok && reconcilerEvent.EventType == corev1.EventTypeNormal {
+			return event
+		} else {
+			return fmt.Errorf("Controller won't retry to handle the error/warning and user will need to address it manually: %w", event)
```
yolocs, line 130: nit: maybe a simpler message like "Error won't be retried, please manually delete PubSub resources: %w"

pkg/reconciler/trigger/trigger.go
```diff
@@ -106,7 +106,29 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, t *brokerv1beta1.Trigger
...
+		// effort GC but the topic is going to be marked ready even when GC fails. This can result in
+		// dangling topic without matching status.
+		// This line should be deleted once the following TODO is finished.
+		// TODO(https://github.com/knative/pkg/issues/1149) Add a FilterKind to genreconciler so it will
+		// skip a trigger if it's not pointed to a gcp broker and doesn't have googlecloud finalizer string.
+		t.Status.MarkTopicReady()
+
+		if event == nil {
+			return nil
+		} else if reconcilerEvent, ok := event.(*pkgreconciler.ReconcilerEvent); ok && reconcilerEvent.EventType == corev1.EventTypeNormal {
```
yolocs, line 127: This is fine. But I think we rarely write EventTypeWarning by ourselves (we just return error). Maybe it's easier to do `else if pkgreconciler.EventAs(&pkgreconciler.ReconcilerEvent{}, &event) { return event }`

pkg/reconciler/trigger/trigger.go
cathyzhyi: changed as suggested. Thanks!

pkg/reconciler/trigger/trigger.go
cathyzhyi: Thanks for the suggestion! Modified accordingly.

grantr (approved):
Thanks @cathyzhyi!

/lgtm
/hold cancel

PR #1707, opened 2020-09-16
Refine ProbeHelper for CloudAuditLogsSource probe

tommyreddad:
Refine ProbeHelper for latest design of CloudAuditLogsSource probe

## Proposed Changes

- ProbeHelper only creates pub/sub topics instead of also deleting them.
- Add tests for unrecognized probe event subjects to avoid channel leaking in case of malformed events.

### DIFF
--- a/test/test_images/probe_helper/probe_helper.go
+++ b/test/test_images/probe_helper/probe_helper.go
@@ -45,8 +45,6 @@ const (
 	CloudStorageSourceProbeArchiveSubject        = "archive"
 	CloudStorageSourceProbeDeleteSubject         = "delete"
 	CloudAuditLogsSourceProbeEventType           = "cloudauditlogssource-probe"
-	CloudAuditLogsSourceProbeCreateSubject       = "create-topic"
-	CloudAuditLogsSourceProbeDeleteSubject       = "delete-topic"
 	CloudSchedulerSourceProbeEventType           = "cloudschedulersource-probe"
 )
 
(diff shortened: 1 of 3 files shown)


test/test_images/probe_helper/probe_helper_test.go
```diff
@@ -420,6 +409,14 @@ func TestProbeHelper(t *testing.T) {
 				wantResult: cloudevents.ResultACK,
 			},
 		},
+	}, {
+		name: "Unrecognized Broker E2E delivery probe subject",
+		steps: []eventAndResult{
+			{
+				event:      probeEvent("broker-e2e-delivery-probe", "invalid-subject"),
```
liu-cong, line 416: nice! We can continue to improve test coverage for error cases

test/test_images/probe_helper/probe_helper.go
```diff
@@ -249,17 +258,9 @@ func (ph *ProbeHelper) forwardFromProbe(ctx context.Context) cloudEventsFunc {
...
-				if _, err := ph.pubsubClient.CreateTopic(ctx, event.ID()); err != nil {
-					return logNACK(ctx, "Probe forwarding failed, error creating pubsub topic %v: %v", event.ID(), err)
-				}
-			case CloudAuditLogsSourceProbeDeleteSubject:
-				// Delete the pubsub topic with the given ID.
-				if err := ph.pubsubClient.Topic(event.ID()).Delete(ctx); err != nil {
-					return logNACK(ctx, "Probe forwarding failed, error deleting pubsub topic %v: %v", event.ID(), err)
-				}
+			// Create a pubsub topic with the given ID.
+			if _, err := ph.pubsubClient.CreateTopic(ctx, event.ID()); err != nil {
```
liu-cong, line 262: I think the topic name should be "cloudauditlogssource-probe-"+event.ID()?

tommyreddad: The event ID already includes the probe type, so event.ID() looks something like "cloudauditlogssource-probe-1234567890"

PR #1711, opened 2020-09-16
webhook reject the request when secret set to be non-nil and service account is non-empty

### LINKED ISSUE #1698: Sources allow invalid spec.secret = {} when spec.serviceAccountName != ""
**Describe the bug**
Creating a source where `spec.secret` is `{}` and `spec.serviceAccountName` is non-empty cause the bad `spec.secret` value to be used when eventually creating a Deployment, leading to:

```
Failed to reconcile Data Plane resource(s): Deployment.apps "cre-src-source-for-trigger-stor706a52b8fb71d03cdfd3c1e78e241b0f" is invalid: [spec.template.spec.volumes[0].secret.secretName: Required value, spec.template.spec.containers[0].env[16].valueFrom.secretKeyRef.name: Invalid value: "": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.template.spec.containers[0].env[16].valueFrom.secretKeyRef.key: Required value
```

Inside the Source's PullSubscription's receive adapter Deployment.

 **Expected behavior**
Either reject the request or default the field as the empty object is inherently incorrect.

 **To Reproduce**
Create the source:

```json
{
    "apiVersion": "events.cloud.google.com/v1beta1",
    "kind": "CloudStorageSource",
    "metadata": {
        "annotations": {},
        "labels": {},
        "name": "source-for-trigger-storage",
        "namespace": "events",
    },
    "spec": {
        "bucket": "my-new-test-bucket-20200914-2",
        "ceOverrides": {
            "extensions": {
                "knsourcetrigger": "link0.2549632403318268"
            }
        },
        "secret": {},
        "serviceAccountName": "foobar",
        "sink": {
            "ref": {
                "apiVersion": "eventing.knative.dev/v1beta1",
                "kind": "Broker",
                "name": "default"
            }
        }
    },
    "status": {
        "annotations": {}
    }
}
```

It will never become ready, due to #1682, the error is immediately visible, just:
```
    - lastTransitionTime: "2020-09-15T00:16:15Z"
      status: Unknown
      type: PullSubscriptionReady
```

The underlying PullSubscription has an event:

```
Failed to reconcile Data Plane resource(s): Deployment.apps "cre-src-source-for-trigger-stor706a52b8fb71d03cdfd3c1e78e241b0f" is invalid: [spec.template.spec.volumes[0].secret.secretName: Required value, spec.template.spec.containers[0].env[16].valueFrom.secretKeyRef.name: Invalid value: "": a DNS-1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character (e.g. 'example.com', regex used for validation is '[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'), spec.template.spec.containers[0].env[16].valueFrom.secretKeyRef.key: Required value
```

 **Possible Fixes**
Either of the following would work.
1. In the webhook:
    ```golang
	if equality.Semantic.DeepEqual(s.Secret, &corev1.SecretKeySelector{}) {
		s.Secret = nil
	}
    ```
1. In the webhook, calculate the default of `spec.secret` and `spec.serviceAccountName` independently.

capri-xiyue:
Fixes #1698

## Proposed Changes

- webhook reject the request of creating sources when secret set to be non-nil and service account is not empty
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/apis/duck/validations.go
+++ b/pkg/apis/duck/validations.go
@@ -119,12 +119,14 @@ func CheckImmutableAutoscalingClassAnnotations(current *metav1.ObjectMeta, origi
 
 // ValidateCredential checks secret and service account.
 func ValidateCredential(secret *corev1.SecretKeySelector, kServiceAccountName string) *apis.FieldError {
-	if secret != nil && !equality.Semantic.DeepEqual(secret, &corev1.SecretKeySelector{}) && kServiceAccountName != "" {
+	fmt.Printf("secret is empty or not %v", equality.Semantic.DeepEqual(secret, &corev1.SecretKeySelector{}))
+	fmt.Printf("secret is %v", secret)
+	if secret != nil && kServiceAccountName != "" {
 		return &apis.FieldError{
 			Message: "Can't have spec.serviceAccountName and spec.secret at the same time",
 			Paths:   []string{""},
 		}
-	} else if secret != nil && !equality.Semantic.DeepEqual(secret, &corev1.SecretKeySelector{}) {
+	} else if secret != nil {
 		return validateSecret(secret)
 	} else if kServiceAccountName != "" {
 		return validateK8sServiceAccount(kServiceAccountName)
--- a/pkg/apis/duck/v1/pubsub_defaults.go
+++ b/pkg/apis/duck/v1/pubsub_defaults.go
@@ -20,8 +20,6 @@ import (
 	"context"
 
 	"github.com/google/knative-gcp/pkg/apis/configs/gcpauth"
-	corev1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/equality"
 	"knative.dev/eventing/pkg/logging"
 	"knative.dev/pkg/apis"
 )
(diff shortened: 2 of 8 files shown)


pkg/apis/duck/validations.go
```diff
@@ -119,12 +119,14 @@ func CheckImmutableAutoscalingClassAnnotations(current *metav1.ObjectMeta, origi
 
 // ValidateCredential checks secret and service account.
 func ValidateCredential(secret *corev1.SecretKeySelector, kServiceAccountName string) *apis.FieldError {
-	if secret != nil && !equality.Semantic.DeepEqual(secret, &corev1.SecretKeySelector{}) && kServiceAccountName != "" {
+	fmt.Printf("secret is empty or not %v", equality.Semantic.DeepEqual(secret, &corev1.SecretKeySelector{}))
+	fmt.Printf("secret is %v", secret)
```
nachocano, line 123: remove these ones

nachocano:
Can you give it a try with gcloud just in case?
I recall we had some problems in the past regarding empty or nils.

capri-xiyue:
Ideally, webhook should reject the request when secret set to be non-nil and service account is non-empty.
But now if we do this, it will break CLI.
For a temporary fix, we will default empty secret to be nil.
Later after CLI fix their bug, we change the behavior of webhook to reject the request when secret set to be non-nil and service account is non-empty.

PR #1715, opened 2020-09-16
Properly set system namespace when reconciling default brokercell

ian-mi:
Backport #1705 into the release-0.15b branch.


Harwayne (commented):
/lgtm

PR #1717, opened 2020-09-16
Properly set system namespace when reconciling default brokercell (#1…

ian-mi:
Backport #1705 into release-0.17


Harwayne (commented):
/lgtm

Issue #1732, opened 2020-09-17
Allow the controller to inject istio sidecar proxy.


upodroid:
**Problem**
I need to have istio sidecar injected in to every pod but the Controller sets `sidecar.istio.io/inject` to false.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for? All

**Exit Criteria**
Istio Injection is set to true.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve? Half a day?

**Additional context (optional)**
This PR https://github.com/google/knative-gcp/pull/1476 disabled sidecar injection.

upodroid:
@yolocs ^

PR #1736, opened 2020-09-17
Allow istio sidecar injection for all broker components

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1732

## Proposed Changes

- Allows istio sidecar injection for all broker components
-
-

**Release Note**

```release-note
Allows istio sidecar injection for all broker components
```

**Docs**


danyinggu (commented):
/lgtm

PR #1737, opened 2020-09-17
Return 429 when ingress if sending too man messages, instead of 500

### LINKED ISSUE #1697: Return 429 when reaching pubsub publish buffer limit
From https://github.com/google/knative-gcp/issues/1550, when a broker ingress pod is handling more load than it's capable of and reaching pubsub publish buffer limit, we should return 429 instead of 500.

tayarani:
Fixes #1697 

## Proposed Changes

- When the ingress is overloaded, propagate a 429 back to the caller instead of masking it with a 500.

**Release Note**

```release-note
Return 429 (too many messages) when the ingress broker is sending too many messages, instead of 500 (internal server error).
```

**Docs**


pkg/broker/ingress/errors.go
```diff
@@ -28,3 +28,6 @@ var ErrIncomplete = errors.New("incomplete config")
 
 // ErrNotReady is the error when a broker is not ready.
 var ErrNotReady = errors.New("not ready")
+
+// ErrOverflow is the error when a broker ingress is sending too many requests to PubSub.
+var ErrOverflow = errors.New("bundler reached buffered byte limit")
```
Harwayne, line 33: Can you import `bundler.ErrOverflow` rather than redefine it here?

tayarani:
I'm not too familiar with the code yet, but this change doesn't seem like something easily testable in our unit tests. Is it ok for me to push this without tests? @Harwayne

Harwayne (commented):
/lgtm

I'll leave approval to someone more familiar with the Broker area.

yolocs:
You can actually test this code. You can have a fake `DecoupleSink` implementation (https://github.com/google/knative-gcp/blob/master/pkg/broker/ingress/handler.go#L76) that always returns `bundler.ErrOverflow` on `Send`. Then you can check if the resp code is 429.

PR #1733, opened 2020-09-17
Use wire-gen injection for all controllers

zhongduo:
Fixes #

Currently some controllers are not using wire-gen. By changing all controller to the same style we can achieve:
1. consistent coding style for all controllers
2. easy to support new singleton configuration maps
3. use same singleton object from the top level.

This PR will be rebased after merge of #1681 

## Proposed Changes

- Use wire-gen to generate all controllers.
-
-

PR #1747, opened 2020-09-21
Add data residency validation to webhook

### LINKED ISSUE #1741: Data residency configmap is not validated by webhook
**Describe the bug**
Followup from https://github.com/google/knative-gcp/pull/1681#issuecomment-695042860.

 **Expected behavior**
Invalid keys in the config-data-residency configmap cause a webhook error.

/assign @zhongduo

zhongduo:
Fixes #1741 

## Proposed Changes

- Add data residency configuration map to webhook validation

### DIFF
--- a/pkg/apis/configs/dataresidency/dataresidency_defaults.go
+++ b/pkg/apis/configs/dataresidency/dataresidency_defaults.go
@@ -17,6 +17,7 @@ limitations under the License.
 package dataresidency
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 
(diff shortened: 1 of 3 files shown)


zhongduo:
The fix to the unit test error will conflict with #1733, will fix it after rebase.

pkg/apis/configs/dataresidency/dataresidency_defaults.go
```diff
@@ -50,10 +51,8 @@ func NewDefaultsConfigFromMap(data map[string]string) (*Defaults, error) {
 
 	// Parse out the data residency configuration.
 	value, present := data[defaulterKey]
-	// Data residency configuration is not required, in which case it will mean
-	// allow all regions, so we simply use an empty one
 	if !present || value == "" {
-		return nc, nil
+		return nil, fmt.Errorf("ConfigMap is missing (or empty) key: %q : %v", defaulterKey, data)
```
Harwayne, line 55: Are we sure we want this to be an error?

My main concern was typos and the `DisallowUnknownFields()` below should remove that possibility, right?

zhongduo: Not really. The configmap has a weird way of parsing the file, so this is literally only checking if the key `default-dataresidency-config` exists. The function `parseEntry` actually only parse the `value` part only, which starts from `clusterDefaults`. So this conditions is triggered only when the `default-dataresidency-config` key is not present, same as the `gcp-auth` behavior.

Harwayne (commented):
/lgtm

PR #1748, opened 2020-09-21
[WIP] Return status 200 when events knockdown the target

grac3gao:
Fixes #

## Proposed Changes

- Found two metrics flakiness cases which might relate to knockdown function doesn't return any protocol result.
- In metric test, we go as: sender pod -> broker -> receiver Pod (intentionally return status code 400 two times) -> broker -> target Pod
- From the failure test log ([summary](https://docs.google.com/spreadsheets/d/1foMAzCFvOk07SEvp0uD9lsm4OC-cMJInC3wDndkf7LI/edit#gid=0)), the target Pod received the event, however, the broker didn't get the status code 200, so it will continue run the handler until EOF, and we miss a metric record.

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/test_images/internal/knockdown/knockdown.go
+++ b/test/test_images/internal/knockdown/knockdown.go
@@ -22,9 +22,13 @@ import (
 	"fmt"
 	"io/ioutil"
 	"log"
+	"net/http"
 	"time"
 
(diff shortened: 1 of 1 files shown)


ian-mi:
This should be unnecessary as StatusOK is the default status when no result is returned by the receiver.

PR #1750, opened 2020-09-21
Use errgroup to gracefully shutdown HTTP server in knockdown

### LINKED ISSUE #1751: Target Test Pod Exits Without Sending HTTP Response Status
**Describe the bug**
The target test pod will sometimes exit after successfully receiving the target event but will not send the response status causing the trigger to get an EOF error.

 **Expected behavior**
The target test pod should send status OK before exiting when receiving the target event.

 **To Reproduce**
This issue causes occasional failures in TestGCPBrokerMetrics.

ian-mi:
Fixes #1751

### DIFF
--- a/test/test_images/internal/knockdown/knockdown.go
+++ b/test/test_images/internal/knockdown/knockdown.go
@@ -83,6 +83,12 @@ func Main(config Config, kdr Receiver) int {
 				return err
 			}
 			if kdr.Knockdown(*e) {
+				// Knockdown succeeded, mark done and
+				// continue until reaching EOF. This
+				// is necessary to allow the HTTP
+				// receiver to process any queued
+				// messages so that it can shutdown
+				// gracefully.
 				done = true
 				pcancel()
 			}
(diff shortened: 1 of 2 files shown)


grac3gao (commented):
/lgtm
/hold for comments from other reviewers

test/test_images/internal/knockdown/knockdown.go
```diff
@@ -39,92 +44,84 @@ type Config struct {
...
+			if err != nil {
+				return err
+			}
+			e, err := binding.ToEvent(ctx, msg)
+			if err != nil {
+				msg.Finish(err)
+				return err
+			}
+			if kdr.Knockdown(*e) {
+				done = true
```
Harwayne, line 86: Why set `done` here, rather than:
```golang
pcancel()
msg.Finish(nil)
return nil
```

ian-mi: The receiver can't return here as the server may accept additional requests before shutting down. In order for the server to shutdown gracefully those requests need to be processed so the receiver will continue running until it gets an EOF.

Harwayne: Please add a comment to that affect. Essentially stating why we need the `done` variable.

PR #1756, opened 2020-09-22
Update subscription config during subscription reconciliation

### LINKED ISSUE #1595: Update retry subscription config during Trigger reconciling
This issue is leftover from #1566 

**Describe the bug**
When a Trigger is reconciled, we never update its retry subscription's config settings.

 **Expected behavior**
When a Trigger is reconciled, we should update its retry subscription's config settings as necessary. In particular, the Broker delivery spec might have changed, which should cause the Trigger to reconcile, updating its subscription config as well.

 **Additional context**
We should update the subscription config with this function: https://pkg.go.dev/cloud.google.com/go/pubsub?tab=doc#Subscription.Update

tommyreddad:
Fixes #1595 

## Proposed Changes

- Update subscription config during subscription reconciliation
- Upgrade pubsub module to 1.6.2, allowing for testing subscription config updates with pstest client (`go get cloud.google.com/go/pubsub && ./hack/update-deps.sh`)
- Add subscription and trigger tests


pkg/reconciler/utils/pubsub/subscription.go
```diff
@@ -64,6 +66,20 @@ func (r *Reconciler) ReconcileSubscription(ctx context.Context, id string, subCo
...
+			(subConfig.DeadLetterPolicy != nil && !equality.Semantic.DeepEqual(config.DeadLetterPolicy, subConfig.DeadLetterPolicy)) {
+			updateSubConfig := pubsub.SubscriptionConfigToUpdate{
+				RetryPolicy:      subConfig.RetryPolicy,
+				DeadLetterPolicy: subConfig.DeadLetterPolicy,
+			}
+			if _, err := sub.Update(ctx, updateSubConfig); err != nil {
+				updater.MarkSubscriptionFailed("SubscriptionConfigUpdateFailed", "Failed to update Pub/Sub subscription config: %v", err)
+				return nil, err
+			}
+			logger.Info("Updated PubSub subscription config", zap.String("name", sub.ID()))
```
grantr, line 80: Might be useful to include the new config in this log.

grantr (approved):
Thanks @tommyreddad! Sorry for the delay.

One optional logging nit. Feel free to cancel the hold if you prefer not to make the log change.

/lgtm
/hold

grantr:
Actually I'll just cancel the hold myself since you're probably out for the day. If we want to make the log change we can create a new PR.

/hold cancel

pkg/reconciler/utils/pubsub/subscription.go
tommyreddad: Good idea, I'll make a new good first issue about this

PR #1761, opened 2020-09-22
Remove gpubsub in sources

### LINKED ISSUE #1739: Switch Topic and PullSubscription to use normal pubsub client
**Problem**
Topic and PullSubscription reconcilers still use the gpubsub client, making their tests different from the Broker and Trigger tests which use pstest. It would be nice, especially for horizontal features like #1678, if the tests used similar helpers.

**Exit Criteria**
Topic and PullSubscription reconciler code stop using gpubsub, and tests use pstest.

Related to #1430.

/cc @nachocano @Harwayne @zhongduo

zhongduo:
Fixes #1739 

## Proposed Changes

- A PR is opened to `google-cloud-go` for error injection: https://github.com/googleapis/google-cloud-go/pull/2916
- A new client create function which returns real `pubsub.Client` is used to replace gpubsub version, replace gpubsub error injection with server error injection.
-

**Release Note**

```release-note

```

**Docs**


pkg/reconciler/intevents/topic/controller.go
```diff
@@ -85,6 +88,27 @@ func newController(
...
+		logging.FromContext(ctx).Error("Failed to create controller-wide Pub/Sub client", zap.Error(err))
+	}
+
+	if client != nil {
+		go func() {
+			<-ctx.Done()
+			client.Close()
+		}()
+	}
+
```
nachocano, line 111: this will not apply to sources as we can specify the project in the CO spec.

nachocano: the idea of having it in the spec was to support cross-project stuff.

nachocano: either way, if you are doing this in the broker, if you cannot create a client here, might be better to fail?

zhongduo: After our conversation yesterday, I actually removed this too. See the latest code.

And yes, you are right about the failing the broker. There are some legacy code, I have discussed this with others and opened an issue to revisit whether we should fail it.

pkg/reconciler/utils/pubsub/types.go
```diff
@@ -17,10 +17,17 @@ limitations under the License.
...
 
 import (
+	"context"
+
 	"cloud.google.com/go/pubsub"
+	"google.golang.org/api/option"
 	"k8s.io/client-go/tools/record"
 )
 
+// CreateFn is a factory function to create a Pub/Sub client. It is copied from gclient/pubsub
```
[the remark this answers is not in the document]
capri-xiyue: I think Jimmy is trying to remove the usage of gpubsub in source code and UT(which was written by ourselves) and change gpubsub to pbtest(written by pubsub)

PR #1762, opened 2020-09-22
Adjust broker fanout resource requirements to meet default delivery attempts per second target

### LINKED ISSUE #1713: Minimize resource consumption by a single fanout pod while still supporting the default max load targets
**Describe the bug**
In #1601, we've determined a Fanout configuration that can support our default max load target (#1597). While it works reliably, the resulting configuration has quite high resource requirements per individual pod. For better resource utilization and being considerate of feedback like [this](https://github.com/google/knative-gcp/issues/1671), it's best to reduce the initial resource request and make it more demand-driven. This may possibly be achieved by relying increasingly more on autoscaling, tuning HPA thresholds and ReceiverSetting parameters.

 **Exit criteria**
- Considerably lower resource requests for a single fanout; 
- Default max load targets are still supported;

Ectelion:
* Fixes #1713 
* Details are available on the task page and in "Fanout Throughput with Autoscaling" doc.

### DIFF
--- a/cmd/broker/fanout/main.go
+++ b/cmd/broker/fanout/main.go
@@ -52,6 +52,10 @@ type envConfig struct {
 	// continuous sync failures (or no sync at all) to be stale.
 	MaxStaleDuration time.Duration `envconfig:"MAX_STALE_DURATION" default:"1m"`
 
+	// MaxOutstandingBytes is the maximum size of unprocessed messages (unacknowledged but not yet expired).
+	// Default is 800Mb (~763Mi).
+	MaxOutstandingBytes int `envconfig:"OUTSTANDING_BYTES_LIMIT" default:"799999525"`
+
 	// Max to 10m.
 	TimeoutPerEvent time.Duration `envconfig:"TIMEOUT_PER_EVENT"`
 }
@@ -138,6 +142,9 @@ func buildHandlerOptions(env envConfig) []handler.Option {
 	}
 	if env.TimeoutPerEvent > 0 {
 		opts = append(opts, handler.WithTimeoutPerEvent(env.TimeoutPerEvent))
+	}
+	if env.MaxOutstandingBytes > 0 {
+		rs.MaxOutstandingBytes = env.MaxOutstandingBytes
 	}
 	opts = append(opts, handler.WithPubsubReceiveSettings(rs))
 	// The default CeClient is good?
--- a/cmd/broker/fanout/main.go
+++ b/cmd/broker/fanout/main.go
@@ -54,7 +54,7 @@ type envConfig struct {
 
 	// MaxOutstandingBytes is the maximum size of unprocessed messages (unacknowledged but not yet expired).
 	// Default is 800Mb (~763Mi)
-	MaxOutstandingBytes int `envconfig:"MAX_OUTSTANDING_BYTES" default:"799999525"`
+	MaxOutstandingBytes int `envconfig:"MAX_OUTSTANDING_BYTES" default:"800000000"`
 
 	// Max to 10m.
 	TimeoutPerEvent time.Duration `envconfig:"TIMEOUT_PER_EVENT"`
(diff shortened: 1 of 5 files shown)


yolocs (approved):
/lgtm
/hold
For others to take a look

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
```diff
@@ -23,15 +23,15 @@ import (
 )
 
 const (
-	avgCPUUtilizationFanout  int32 = 95
+	avgCPUUtilizationFanout  int32 = 30
```
liu-cong, line 26: I am concerned if we are overfitting to solve rapid request spikes. If I have a gradual request increase that just causes fanout to use CPU at, say, 40%, this will result in 2 pods with low CPU utilization. Is my understanding correct?

Ectelion: Good point, underutilization of resources in such cases was my major concern as well. There are some thoughts on this in the "Fanout with Autoscaling" doc. Overall, with limited control over scale out velocity, we run into this trade-off between performance at load and utilization of resources under lower load.  In 1.18+, this can be better addressed with recently introduced [scale out behavior](https://github.com/kubernetes/enhancements/blob/master/keps/sig-autoscaling/20190307-configurable-scale-velocity-for-hpa.md), but for now, which of these would you rather prefer:
1. 50% threshold with ~6-7 minutes cold start, but lower resource utilization at lower load;
1. 30% threshold with ~2-3 minutes cold start (~50% better), but considerably higher resource utilization at lower load;

    - (For a light load, like under 10Mb/s delivery rate, the scale out is not triggered, so, both options are on par;

I guess it's all application/customer dependent, but I would love to hear more thoughts on our expected default/common case.

liu-cong: We don't know for sure until we have more customer data. I personally prefer a higher number as I think use cases that require aggressive scaling is less common. 

@grantr @yolocs thoughts on this?

yolocs: When doing load test, we tend to think more about worst case scenarios, which could be a lot of worse than the real cases. I agree a less aggressive scaling could be a good starting point.

Ectelion: Makes sense. Ok, I think we can proceed as following:
1) For the default configuration, I'll proceed with option #1 with higher threshold;
2) Those seeking better performance at high RPS would have two workarounds: lowering this threshold or increasing minReplicas. We may provide guidance on configurability and these trade-offs in the documentation;

Sounds good?

liu-cong: I hope we can achieve a higher CPU utilization like 70. But I can be convinced with your proposal given the constraints we have today.

grantr: Now that we know the effect of changing this number, it will be relatively simple to tune in response to user feedback.

cmd/broker/fanout/main.go
```diff
@@ -52,6 +52,10 @@ type envConfig struct {
 	// continuous sync failures (or no sync at all) to be stale.
 	MaxStaleDuration time.Duration `envconfig:"MAX_STALE_DURATION" default:"1m"`
 
+	// MaxOutstandingBytes is the maximum size of unprocessed messages (unacknowledged but not yet expired).
+	// Default is 800Mb (~763Mi)
+	MaxOutstandingBytes int `envconfig:"MAX_OUTSTANDING_BYTES" default:"799999525"`
```
grantr, line 57: Why the odd number here?

pkg/apis/intevents/v1alpha1/brokercell_defaults.go
Ectelion: We've experimented with higher thresholds -- other factors start weighting in, e.g. there are increasing risks that autoscaling occasionally won't be triggered at all, which would result in the throughput considerably below the target. If we could include an additional robust metric like the number of messages in PubSub backlog, I guess, that would have addressed that risk.

cmd/broker/fanout/main.go
Ectelion: No strong reason, other than myself tying to account for inconsistencies between Mebibytes and Megabytes used across different settings, and converting these back and forth in my notes. The difference is due to rounding and I've now rounded it up more precisely (the difference is negligible and does not affect the results).

grantr (approved):

PR #1770, opened 2020-09-25
[Automated] Fix misspellings

mattmoor:
Produced via:
```shell
export FILES=( $(find . -type f -not -path './vendor/*' -not -path './third_party/*' -not -path './.git/*') )
misspell -w "${FILES[@]}"
```
/cc nachocano grantr ian-mi
/assign nachocano grantr ian-mi


tommyreddad:
I noticed this was failing due to the configmap checksum being updated after the misspell fix, this requires running codegen again to fix the checksum. I will submit a PR in a moment with the spelling mistakes fixed and the right checksums.

Issue #1600, opened 2020-08-18
Adjust broker ingress resource requirements to meet default events per second target


grantr:
**Problem**
Once we have an events per second target #1596 and a default event payload size #1599, we can measure the resource requirements for a single ingress pod to meet the target. That's our default resource requirement value.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
User

**Exit Criteria**
The broker ingress deployment is created with default resource requirements allowing it to meet the default events per second target.

**Additional context (optional)**
Part of #1552.

yolocs:
To support the target [load](https://github.com/google/knative-gcp/issues/1596#issuecomment-679309679), here are proposed values:

- CPU 2000m
- Mem 2Gi
- Pubsub publish buffer 200Mi
- HPA CPU 90%, Mem 1800Mi

Here is the reasoning.

### CPU

Between 1000-2000 qps, the CPU ranges between 1800m to 3000m. Event size has very little impact on the CPU usage. Setting 2000m CPU allow HPA to kick in near 1000 qps.

### Mem

With 1000 qps and 256k event size, the stable memory usage is about ranges between 300Mi to 1Gi with most of the time stay below 500Mi. _However_, when there is sudden increase on qps or event size, there is always a surge of memory usage which gradually smooths out. The surge sometimes could reach 2Gi. Setting the memory limit to 2Gi helps prevent OOM (HPA on memory has been proved inefficient with memory surges), and it also leaves room for events with greater sizes.

![c86f0910-3519-445f-b9c4-2c613a04c89d](https://user-images.githubusercontent.com/14087829/91091183-159f1f80-e60b-11ea-8b98-55a5528c750f.png)

### Publish buffer

With 100Mi buffer size (default value), we have started to see significant more 500s with 800 qps 256k size. 

![2c2fc1e9-01af-4776-9ed0-dabd33f35226](https://user-images.githubusercontent.com/14087829/91091803-ff459380-e60b-11ea-8ff0-6434fe004213.png)

Setting 200Mi mitigates this problem (under the same load). I also want to avoid setting a higher buffer size as it might allow a lot more mem accumulation and cause OOM. See [comment](https://github.com/google/knative-gcp/issues/1550#issuecomment-669551433).

### HPA

I want to have CPU being the main factor for auto scaling. I'm trying to set the mem limit being the upper bound I've seen (with the target load). So hopefully in practice it will never use mem to scale up.

liu-cong:
[...]

PR #1772, opened 2020-09-25
Handle requests with very large payload

### LINKED ISSUE #1754: Requests with very large payload (>1Gb) need to be gracefully handled
**Describe the bug**
Currently, there seems to be no validation that gracefully handles the case with very large payload (>1Gb are possible). Regardless of payload size, the Broker attempts to load it in-memory, which in some cases makes it prone to out-of-memory issues.

 **Expected behavior**
There should be a cap on payload. The Broker shouldn't attempt to process requests with payload exceeding the cap, but raise a client error (i.e. a response code like 4xx). It probably makes sense to make this cap configurable.

 **To Reproduce**
Both of the following tests resulted in a repro, although this must also be reproducible under a lower RPS.
**Test 1:** Generate the load of ~100 RPS with 3.5Gb request payload.
**Test 2:** Generate the load of ~1000 RPS with 256Mb request payload.

 **Exit critera**
A unit test validating this behavior.

Ectelion:
Fixes #1754. In particular:
- Prevents potential out-of-memory issues caused by requests with very large  payload;
- Addresses a functional issue that occurs even for relatively small payload sizes that exceeds PubSub limit on message size, which is 10Mb (non-extendable). Requests with payloads above that are not passing the publishing stage;
    * You may find it on [PubSub quotas page](https://cloud.google.com/pubsub/quotas): after the main table, there is a section with non-extendable quotas;
 

### On memory consumption
Memory stress test was re-run several times. Here's is how a typical memory consumption pattern looks like:
image
In both cases, there is an initial spike varying around 600-800MiB, which then gradually smoothes out. There appears to be little to no impact on memory caused by the MaxBytesReader wrapper. This can also be seen from [MaxBytesReader's implementation](https://github.com/golang/go/blob/2cd2ff6f564dce5be0c4fb7f06338ff7af3fc9a9/src/net/http/request.go#L1121), which turns out to be lightweight and does not have any notable memory overheads.

### DIFF
--- a/pkg/broker/ingress/handler.go
+++ b/pkg/broker/ingress/handler.go
@@ -47,8 +47,8 @@ const (
 	// TODO(liu-cong) configurable timeout
 	decoupleSinkTimeout = 30 * time.Second
 
-	// Limit for request payload in bytes (100Mb)
-	maxRequestBodyBytes = 100000000
+	// Limit for request payload in bytes (10Mb -- corresponds to message size limit on PubSub as of 09/2020)
+	maxRequestBodyBytes = 10000000
 
 	// EventArrivalTime is used to access the metadata stored on a
 	// CloudEvent to measure the time difference between when an event is
--- a/pkg/broker/ingress/handler.go
+++ b/pkg/broker/ingress/handler.go
@@ -47,6 +47,9 @@ const (
 	// TODO(liu-cong) configurable timeout
 	decoupleSinkTimeout = 30 * time.Second
 
+	// Limit for request payload in bytes (100Mb)
+	maxRequestBodyBytes = 100000000
+
 	// EventArrivalTime is used to access the metadata stored on a
 	// CloudEvent to measure the time difference between when an event is
 	// received on a broker and before it is dispatched to the trigger function.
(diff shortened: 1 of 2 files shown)


pkg/broker/ingress/handler.go
```diff
@@ -47,6 +47,9 @@ const (
 	// TODO(liu-cong) configurable timeout
 	decoupleSinkTimeout = 30 * time.Second
 
+	// Limit for request payload in bytes (100Mb)
+	maxRequestBodyBytes = 100000000
```
Ectelion, line 51: Let me know if we'd rather have this as an env variable.

liu-cong: Can you document how this number is picked? Is this the same as the pubsub message size limit?

ian-mi:
I recommend also wrapping the body in an io.LimitedReader.

pkg/broker/ingress/handler.go
Ectelion: Sure, I have updated the description.

Ectelion:
> I recommend also wrapping the body in an io.LimitedReader.

Some more context on this: I think this is indeed a more reliable approach. I myself considered MaxBytesReader, which is similar. In the context of recently completed perf tuning work (#1600), I wondered if the influence on memory consumption can be assumed to be negligible or whether the memory analysis in #1600 would need to be re-evaluated. Given this and that the malicious use case implies a wider problem (unauthorized access to the Broker first place), I thought it may be fine to keep it with current approach. What do you think?

grantr (approved):
Thanks @Ectelion!

/lgtm

PR #1775, opened 2020-09-28
Add rollout restart time annotation to Brokercell and update Brokercell deployments 

danyinggu:
Fixes #

## Proposed Changes

- Add rollout restart time annotation to brokercell
- Update the annotation ingress, fanout and retry deployments based on the Brokercell annotation 
-

**Release Note**

```release-note

```

**Docs**


yolocs (commented):
/lgtm
/hold
In case Grant or Cong need to take a look.

danyinggu:
FYI: I want to do some further research about the length to be set in deployment's progressDeadlineSeconds, so I didn't change by intention. Will change that in a separate PR.

pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -127,6 +127,12 @@ func MakeRetryDeployment(args RetryArgs) *appsv1.Deployment {
 
 // deploymentTemplate creates a template for data plane deployments.
 func deploymentTemplate(args Args, containers []corev1.Container) *appsv1.Deployment {
+	annotation := map[string]string{
+		"sidecar.istio.io/inject": strconv.FormatBool(args.AllowIstioSidecar),
+	}
+	if args.RolloutRestartTime != "" {
```
liu-cong, line 133: Looks like removing this condition is still fine.

danyinggu: Thanks for the comment! If removing this condition and there is no annotation for rollout restart time added the Brokercell resource, the deployment will have `"events.cloud.google.com/RestartedAt": ""`. It is functional logically, but not that meaningful, so we want to avoid this case.

pkg/reconciler/brokercell/resources/deployments.go
```diff
@@ -127,6 +127,12 @@ func MakeRetryDeployment(args RetryArgs) *appsv1.Deployment {
...
+		annotation["events.cloud.google.com/RestartedAt"] = args.RolloutRestartTime
```
grantr, line 134: I think this annotation key should be a constant.

liu-cong: I think "RestartedAt" could be misleading - the deployment could be restarted for other reasons, right? Basically I don't want this annotation to be considered the authority of last restart time.

Can we call this something like "RestartRequestedAt"?

pkg/reconciler/brokercell/resources/deployments.go
liu-cong: I am fine with either way.

pkg/reconciler/brokercell/resources/args.go
```diff
@@ -32,6 +32,11 @@ const (
 	// RetryName is the name used for the retry container.
 	RetryName          = "retry"
 	BrokerCellLabelKey = "brokerCell"
+
+	// the annotation key of the time to trigger rollout restart of the deployments
+	IngressRestartTimeAnnotationKey = "events.cloud.google.com/ingressRestartedAt"
```
liu-cong, line 37: See my comment below, I suggest "ingressRestartRequestedAt".

PR #1777, opened 2020-09-29
Increase e2e test parallel limit to 24

ian-mi:
This significantly speeds up e2e test runs. In my testing, full test runs complete in approximately 400 seconds with this change and I have not seen this cause an e2e test failure. Regardless, e2e test flakiness should be monitored after this change is submitted.

### DIFF
--- a/test/e2e-tests.sh
+++ b/test/e2e-tests.sh
@@ -25,7 +25,7 @@ if [ "${SKIP_TESTS:-}" == "true" ]; then
   exit 0
 fi
 
-go_test_e2e -timeout=30m -parallel=12 ./test/e2e \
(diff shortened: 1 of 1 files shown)


ian-mi:
/hold
This appears to result in problems on Prow. Keeping this open while I try to figure out why.

zhongduo:
The test `pull-google-knative-gcp-integration-tests` failed quite randomly to me, so not sure if it is related at all. The speedup should be closely monitored in Prow instead of your machine. As we do notice there is a big gap in the running speed in Prow machine and local machine. Eg, with this change the test might indeed be slower in Prow though faster in your local test.

Harwayne:
Should we close this?

PR #1788, opened 2020-09-30
Issue 928 filter in ingress

### LINKED ISSUE #928: Prefilter in broker ingress
**Problem**
Do not persist messages that don't match any triggers.
Also consider tagging events with filter result, which can be used by fanout and retry.

**Exit Criteria**

**Time Estimate (optional):** 

tayarani:
Fixes #928 

## Proposed Changes
Pre-filter events at the ingress level when there are no targets interested in it. This will help save pub/sub resources by testing each event against all target tilters *before* sending the event to pub/sub.


pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -69,6 +70,12 @@ type multiTopicDecoupleSink struct {
 
 // Send sends incoming event to its corresponding pubsub topic based on which broker it belongs to.
 func (m *multiTopicDecoupleSink) Send(ctx context.Context, broker types.NamespacedName, event cev2.Event) protocol.Result {
+	// Check to see if there are any triggers interested in this event. If not, no need to send this
```
tayarani, line 73: I have thought about the positioning of this block wrt to the block under it.

Right now if a broker is setup (even incorrectly, e.g., no topic) without targets, ingress will accept and acknowledge all incoming events.

If we move this block under the one below, it will at least check to see the topic exists before accepting events.

I'm open to either approach, and would appreciate it if anyone else has other thoughts.

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -90,6 +97,26 @@ func (m *multiTopicDecoupleSink) Send(ctx context.Context, broker types.Namespac
...
 }
 
+// EventFilterFunc is used to see if a target is interested in an event.
+// It is used as a vaiable to allow stubbing out in unit tests.
+var EventFilterFunc = filter.PassFilter
+
+// hasTrigger checks given event against all targets to see if it will pass any of their filters.
+// If one is fouund, hasTrigger returns true.
+func (m *multiTopicDecoupleSink) hasTrigger(ctx context.Context, event *cev2.Event) bool {
+	var hasTrigger bool
```
zhongduo, line 107: IMO it will be more readable to have `hasTrigger := false` instead.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -54,7 +57,7 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
 			name: "happy path single broker",
 			brokerConfig: &config.TargetsConfig{
 				Brokers: map[string]*config.Broker{
-					"test_ns_1/test_broker_1": {DecoupleQueue: &config.Queue{Topic: "test_topic_1", State: config.State_READY}},
+					"test_ns_1/test_broker_1": {DecoupleQueue: &config.Queue{Topic: "test_topic_1", State: config.State_READY}, Targets: brokerTargets},
```
zhongduo, line 60: The `Targets` and `DecoupleQueue` fields can be in their own new line.

pkg/broker/ingress/multi_topic_decouple_sink.go
tayarani: In go all variables are zero initialized: https://golang.org/ref/spec#The_zero_value.

zhongduo: Yes, I do know that, that's why I said readable not bug :). IMO [short variable declaration](https://golang.org/ref/spec#Short_variable_declarations) is a cleaner and easier to understand syntax, especially for these primary types.

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -90,6 +97,26 @@ func (m *multiTopicDecoupleSink) Send(ctx context.Context, broker types.Namespac
...
+	m.brokerConfig.RangeAllTargets(func(target *config.Target) bool {
+		if EventFilterFunc(ctx, target.FilterAttributes, event) {
+			hasTrigger = true
+			return true
```
zhongduo, line 111: is this required? It seems to me that without this the function will still return true later.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +228,157 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+			}
+
+			brokerConfig := memory.NewTargets(testBrokerConfig)
+			sink := NewMultiTopicDecoupleSink(ctx, brokerConfig, psClient, pubsub.DefaultPublishSettings)
+
+			event := createTestEvent(uuid.New().String())
+			fmt.Printf("Event type: %s\n", event.Type())
+
+			hasTrigger := sink.hasTrigger(context.Background(), event)
+			if hasTrigger != test.hasTrigger {
```
zhongduo, line 320: `hasTrigger` can be inlined into the `if` statement.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +228,157 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+
+	testTopic := "test_topic_1"
+	topic := psClient.Topic(testTopic)
+	if exists, err := topic.Exists(ctx); err != nil {
+		t.Fatal(err)
+	} else if !exists {
+		if topic, err = psClient.CreateTopic(ctx, testTopic); err != nil {
+			t.Fatal(err)
+		}
+	}
```
zhongduo, line 360: Not sure if this complexity is needed here, since you always create a new server, is there any chance that there will be topic already in the test server? If so you should handle `Exists` properly. Otherwise you can safely call CreateTopic.

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -90,6 +97,26 @@ func (m *multiTopicDecoupleSink) Send(ctx context.Context, broker types.Namespac
 	return err
 }
 
+// EventFilterFunc is used to see if a target is interested in an event.
+// It is used as a vaiable to allow stubbing out in unit tests.
+var EventFilterFunc = filter.PassFilter
```
zhongduo, line 102: If this is for testing purpose only, do you have to export the name?

pkg/broker/ingress/multi_topic_decouple_sink_test.go
zhongduo: Any reason you use `context.Background()` instead of `ctx`?

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +228,157 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+						Targets:       test.brokerTargets,
+					},
+				},
+			}
+
+			brokerConfig := memory.NewTargets(testBrokerConfig)
+			sink := NewMultiTopicDecoupleSink(ctx, brokerConfig, psClient, pubsub.DefaultPublishSettings)
+
+			event := createTestEvent(uuid.New().String())
+			fmt.Printf("Event type: %s\n", event.Type())
```
zhongduo, line 317: You should never used `fmt.Printf` in test code except for your own debugging purpose. Use `t.Logf` instead.

pkg/broker/ingress/multi_topic_decouple_sink.go
tayarani: That was intended to be false. It's for early breaking out of the 'RangeAllTargets' loop. Fixed.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
tayarani: Good catch.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
tayarani: I took this block of code from 'TestMultiTopicDecoupleSink'. Seems like the correct way to ensure the test is setup properly.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
tayarani: Left over from manual debugging. Removed.

pkg/broker/ingress/multi_topic_decouple_sink.go
liu-cong: I think putting this below the `topic, err := m.getTopicForBroker(ctx, broker)` is better. `getTopicForBroker` should usually be a faster operation than applying the filters.

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -90,6 +97,26 @@ func (m *multiTopicDecoupleSink) Send(ctx context.Context, broker types.Namespac
...
+
+// hasTrigger checks given event against all targets to see if it will pass any of their filters.
+// If one is fouund, hasTrigger returns true.
+func (m *multiTopicDecoupleSink) hasTrigger(ctx context.Context, event *cev2.Event) bool {
+	var hasTrigger bool
+	m.brokerConfig.RangeAllTargets(func(target *config.Target) bool {
```
liu-cong, line 108: I think we should gate this feature behind a flag and disable it by default. Alternatively we can make sure we do some performance testing before we release. However, the latter could potentially delay our release.

Thoughts? @grantr

grantr: IMO we don't need a flag for this, and we should test the performance impact anyway.

If we're concerned about filtering taking too long in edge cases, we can add a timeout after which the event is sent to the decoupling topic (and a prefilter timeout metric is incremented). This would be similar to fanout's initial delivery timeout to retry (but with a much shorter timeout, maybe 100ms).

tayarani: As discussed, I added a feature flag for this until we test the performance change. I've added TODOs to clean up the feature gate as part of #1804.

pkg/broker/ingress/multi_topic_decouple_sink.go
tayarani: Done.

pkg/broker/ingress/multi_topic_decouple_sink.go
zhongduo: It only works when the user is in the same package.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +226,219 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+
+	tests := []struct {
+		name          string
+		brokerTargets map[string]*config.Target
+		hasTrigger    bool
+	}{
+		{
+			name:       "broker with no target",
+			hasTrigger: false,
+		},
```
zhongduo, line 240: You might want to add a similar test but with an empty map (not nil).

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +226,219 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+			},
+			hasTrigger: true,
+		},
+	}
+	for _, test := range tests {
+		t.Run(test.name, func(t *testing.T) {
+			ctx := logtest.TestContextWithLogger(t)
+			psSrv := pstest.NewServer()
+			defer psSrv.Close()
+			psClient := createPubsubClient(ctx, t, psSrv)
```
zhongduo, line 300: `defer psClient.close()`

zhongduo: Alternatively, you can use https://github.com/google/knative-gcp/blob/eb5ac1b2f0231ba7bb1681858f3478a001938341/pkg/reconciler/testing/pstest.go#L200

pkg/broker/ingress/multi_topic_decouple_sink_test.go
zhongduo: I don't think line length is the concern here, in our code base, you do normally have multiple line for struct initialization. That's why `Brokers:` is in its own line.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
tayarani: Done

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +226,224 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+					FilterAttributes: map[string]string{
+						"type":   eventType,
+						"source": "some-random-source",
+					},
+				},
+			},
+			hasTrigger: false,
+		},
+		{
+			name: "broker with multiple targets with one matching target filter",
```
grantr, line 278: Can you add a test of multiple targets with none matching?

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +226,224 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+			hasTrigger := sink.hasTrigger(ctx, event)
+			if hasTrigger != test.hasTrigger {
+				t.Errorf("Sink says event has trigger %t which should be %t", hasTrigger, test.hasTrigger)
+			}
+		})
+	}
+}
+
+func TestMultiTopicDecoupleSinkSendChecksFilter(t *testing.T) {
+	origEnableEventFilterFunc := EnableEventFilterFunc
```
grantr, line 330: Add a `TODO(#1804)` here to delete lines 330-338

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +226,224 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+	defer func() { EnableEventFilterFunc = origEnableEventFilterFunc }()
+	EnableEventFilterFunc = func() bool {
+		return true
+	}
+
+	filterCalled := false
+	origEventFilterFunc := EventFilterFunc
+	defer func() { EventFilterFunc = origEventFilterFunc }()
+	EventFilterFunc = func(ctx context.Context, attrs map[string]string, event *event.Event) bool {
```
grantr, line 339: Can we remove the mock and check whether the message made it to the topic or not? Then we test that `EventFilterFunc` is set correctly in the tested code. (Arguably E2E tests would also cover this, but having it in a unit test makes identifying a bug much quicker)

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -220,6 +226,224 @@ func TestMultiTopicDecoupleSink(t *testing.T) {
...
+	}
+}
+
+// Temoporary test to ensure the filtering feature is disabled by default.
+// TODO(#1804): remove this test when enabling the feature by default.
+func TestMultiTopicDecoupleSinkSendDoesNotChecksFilterWhenFeatureDisabled(t *testing.T) {
+	filterCalled := false
+	origEventFilterFunc := EventFilterFunc
+	defer func() { EventFilterFunc = origEventFilterFunc }()
+	EventFilterFunc = func(ctx context.Context, attrs map[string]string, event *event.Event) bool {
```
grantr, line 396: I'm fine with using a mock here for the purpose of the test, but if we're changing the above test it might be useful to make this one consistent. Consider this suggestion optional since I expect this is a temporary test.

pkg/broker/handler/processors/filter/processor.go
```diff
@@ -93,7 +93,9 @@ func startSpan(ctx context.Context, trigger types.NamespacedName, event *event.E
 	return tracing.WithLogging(ctx, span), span
 }
 
-func (p *Processor) passFilter(ctx context.Context, attrs map[string]string, event *event.Event) bool {
+// PassFilter checks given event against attributes available in the attrs map to determine
+// if the event should pass or not.
+func PassFilter(ctx context.Context, attrs map[string]string, event *event.Event) bool {
```
grantr, line 98: Making this an exported function seems a bit odd to me in terms of encapsulation and modularity. Don't take this as an objection though. We may want to consider a followup that refactors the ingress to use processors in the same style as fanout and retry.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
```diff
@@ -40,6 +40,9 @@ import (
 )
 
 func TestMultiTopicDecoupleSink(t *testing.T) {
+	// Add a target to the broker to ensure all events are sent to pub/sub.
```
grantr, line 43: Can you elaborate on this comment to explain why adding a target ensures events are sent to pubsub? The reader here may not be familiar with the ingress filtering behavior.

grantr (commented):
Thanks @tayarani! I have a few nits on tests and docs and other non-blocking comments.

pkg/broker/ingress/multi_topic_decouple_sink_test.go
yolocs: This change would potentially make e2e test harder for filtering in fanout/retry. Today, we just need to send an event that doesn't match any trigger and expect not receiving it. We know for sure it's the filtering in fanout/retry makes it (not) happen. But with this change, the event would be filtered on ingress.

yolocs:
We don't have to wait until this gets merged to know the performance impact. I think a micro benchmark should do it. @ian-mi has done this for multiple critical parts, e.g. https://github.com/google/knative-gcp/blob/master/pkg/broker/handler/processors/deliver/processor_test.go#L367

I don't mean to block this change. But I would prefer getting a rough grasp on the performance impact upfront (instead of merging it and later find out it has a huge impact and revert the change).

grantr:
Good idea to add a micro benchmark @yolocs. @tayarani does that seem reasonable to add to this PR?

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -80,6 +86,13 @@ func (m *multiTopicDecoupleSink) Send(ctx context.Context, broker types.Namespac
 		return err
 	}
 
+	// Check to see if there are any triggers interested in this event. If not, no need to send this
+	// to the decouple topic.
+	// TODO(#1804): remove first check when enabling the feature by default.
+	if m.enableEventFiltering && !m.hasTrigger(ctx, &event) {
```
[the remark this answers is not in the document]
Harwayne: I don't think this needs a metric. 

It might be helpful for us to be able to say 'look how much we saved in money and latency (by sending the upstream ack faster)', but I don't see much utility for users, other than when investigating different Broker implementations.

If users are concerned that events are being dropped when they shouldn't be, then I think our answer is that you need to compare Broker `event_count` and Trigger `event_count`. Because the general case of the problem is 'I am sending events to the Broker, but my Trigger is not receiving them'. And that is a problem regardless of whether other Triggers happen to include that event or not. So distinguishing between it being dropped here or at Fanout shouldn't matter.

If you think it is likely this area will have a bug, then a metric may make it easier to track down. But I would argue adding a debug level log statement would likely be better, as we could include the attributes of the event being dropped and investigate if we think it was a mistake or not.

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -90,6 +103,36 @@ func (m *multiTopicDecoupleSink) Send(ctx context.Context, broker types.Namespac
...
+var EventFilterFunc = filter.PassFilter
+
+// EnableEventFilterFunc is a temporary function to control enabling and
+// disabling trigger-less event filtering in ingress.
+// TODO(#1804): remove this variable when enabling the feature by default.
+var EnableEventFilterFunc = isEventFilteringEnabled
+
+// TODO(#1804): remove this method when enabling the feature by default.
+func isEventFilteringEnabled() bool {
+	return os.Getenv("ENABLE_EVENT_FILTERING") == "true"
```
Harwayne, line 117: How will this be set?

IIUC, this Deployment is created by the BrokerCell reconciler. How will that reconciler choose to set or not set this env var?

pkg/broker/ingress/multi_topic_decouple_sink.go
Harwayne: I agree with @zhongduo, explicitly setting to false is quicker to understand.

pkg/broker/ingress/multi_topic_decouple_sink.go
```diff
@@ -90,6 +103,36 @@ func (m *multiTopicDecoupleSink) Send(ctx context.Context, broker types.Namespac
 	return err
 }
 
+// EventFilterFunc is used to see if a target is interested in an event.
+// It is used as a vaiable to allow stubbing out in unit tests.
```
Harwayne, line 107: Can be injected instead?

pkg/broker/ingress/multi_topic_decouple_sink.go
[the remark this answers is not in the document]
tayarani: Thanks for weighing in. I added a log message to help debugging just in case it's needed.

pkg/broker/ingress/multi_topic_decouple_sink.go
tayarani: Can u elaborate please?

pkg/broker/ingress/multi_topic_decouple_sink_test.go
tayarani: Sure. done.

pkg/broker/ingress/multi_topic_decouple_sink.go
tayarani: Good point. I have added annotations to the broker cell now which will set this env variable on ingress pods if they are set.

tayarani:
Micro benchmark results:

Running 3 times with the feature *disabled* (i.e., current default behaviour):
```pkg: github.com/google/knative-gcp/pkg/broker/ingress
BenchmarkIngressHandler/0_bytes-12         	  119655	      9726 ns/op
BenchmarkIngressHandler/1000_bytes-12      	   95484	     17588 ns/op
BenchmarkIngressHandler/100000_bytes-12    	    8024	    130719 ns/op
PASS
ok  	github.com/google/knative-gcp/pkg/broker/ingress	12.013s

pkg: github.com/google/knative-gcp/pkg/broker/ingress
BenchmarkIngressHandler/0_bytes-12         	  103238	      9874 ns/op
BenchmarkIngressHandler/1000_bytes-12      	  108384	     16846 ns/op
BenchmarkIngressHandler/100000_bytes-12    	    7946	    126279 ns/op
PASS
ok  	github.com/google/knative-gcp/pkg/broker/ingress	12.782s

pkg: github.com/google/knative-gcp/pkg/broker/ingress
BenchmarkIngressHandler/0_bytes-12         	  120055	     10090 ns/op
BenchmarkIngressHandler/1000_bytes-12      	   96837	     17096 ns/op
BenchmarkIngressHandler/100000_bytes-12    	    8575	    365640 ns/op
PASS
ok  	github.com/google/knative-gcp/pkg/broker/ingress	12.959s```

**Comment by tayarani** [comment]
New benchmark results with updated benchmark test which sweeps target counts as well:

With the feature disabled (i.e., current default behaviour):
```
pkg: github.com/google/knative-gcp/pkg/broker/ingress
BenchmarkIngressHandler/event_size_0_bytes_-_1_targets-12         	  116500	     10273 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_1_targets-12      	  106446	     10558 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_1_targets-12    	    7412	    363537 ns/op
BenchmarkIngressHandler/event_size_0_bytes_-_5_targets-12         	  114288	     10198 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_5_targets-12      	   95527	     12822 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_5_targets-12    	   10372	    108859 ns/op
BenchmarkIngressHandler/event_size_0_bytes_-_10_targets-12        	  120920	     10303 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_10_targets-12     	  120037	     17044 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_10_targets-12   	   11582	    107944 ns/op
BenchmarkIngressHandler/event_size_0_bytes_-_50_targets-12        	  114660	     10488 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_50_targets-12     	  111990	     12593 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_50_targets-12   	   11056	    101499 ns/op
BenchmarkIngressHandler/event_size_0_bytes_-_100_targets-12       	  108790	     10072 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_100_targets-12    	  112933	     17784 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_100_targets-12  	    8248	    621597 ns/op
PASS
ok  	github.com/google/knative-gcp/pkg/broker/ingress	37.846s
```

With the feature enabled (i.e., new behaviour):
```
pkg: github.com/google/knative-gcp/pkg/broker/ingress
BenchmarkIngressHandler/event_size_0_bytes_-_1_targets-12         	  115342	     10853 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_1_targets-12      	   73153	     15202 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_1_targets-12    	    7351	    415066 ns/op
BenchmarkIngressHandler/event_size_0_bytes_-_5_targets-12         	   86582	     12181 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_5_targets-12      	   80860	     18619 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_5_targets-12    	   10368	    115420 ns/op
BenchmarkIngressHandler/event_size_0_bytes_-_10_targets-12        	   79036	     13373 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_10_targets-12     	   91483	     17765 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_10_targets-12   	   11096	    299258 ns/op
BenchmarkIngressHandler/event_size_0_bytes_-_50_targets-12        	   49312	     23270 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_50_targets-12     	   53337	     21697 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_50_targets-12   	    7082	    409847 ns/op
BenchmarkIngressHandler/event_size_0_bytes_-_100_targets-12       	   33829	     41354 ns/op
BenchmarkIngressHandler/event_size_1000_bytes_-_100_targets-12    	   31771	     34050 ns/op
BenchmarkIngressHandler/event_size_100000_bytes_-_100_targets-12  	    6237	    261891 ns/op
PASS
ok  	github.com/google/knative-gcp/pkg/broker/ingress	37.085s
```

File: pkg/broker/handler/processors/filter/processor.go (hunk 16 above)
Comment (tayarani) [reply pkg/broker/handler/processors/filter/processor.go:98]: I do believe the code that decides if an event should be sent out by fan out, should be the exact same as the code that decides to drop events at ingress. Any discrepancy between those two methods would be problematic. So I opted to re-use the same method.

We could refactor into a common place where they both call a helper if that helps.

File: pkg/broker/ingress/multi_topic_decouple_sink_test.go (hunk 14 above)
Comment (tayarani) [reply pkg/broker/ingress/multi_topic_decouple_sink_test.go]: I decided to keep the mock. If we remove it, testing the case where an event shouldn't make it down stream is difficult with current pub/sub. It would make the tests overly complicated imo. so i didn't spend any more time on it. Hope that's okay

File: pkg/broker/ingress/multi_topic_decouple_sink_test.go (hunk 17 above)
Comment (tayarani) [reply pkg/broker/ingress/multi_topic_decouple_sink_test.go]: Done

File: pkg/broker/ingress/multi_topic_decouple_sink.go (hunk 20 above)
Comment (tayarani) [reply pkg/broker/ingress/multi_topic_decouple_sink.go:108]: I did consider that, but it seemed a little more complicated than the var. I'm open to it if we feel the function ptr won't work.

**Comment by grantr** [comment]
According to [benchstat](https://godoc.org/golang.org/x/perf/cmd/benchstat) the difference is not statistically significant in any of the benchmarks. I'm not sure exactly how that's calculated though, and in some cases it seems pretty significant to me (event_size_100000_bytes_-_5_targets-12 is 109us vs 326us). I'm fine with calling it equivalent performance, though anyone with more expertise in microbenchmarks please feel free to disagree.

```
$ benchstat nofilter.txt prefilter.txt
name                                                     old time/op  new time/op  delta
IngressHandler/event_size_0_bytes_-_1_targets-12         10.3µs ± 0%  10.3µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_1000_bytes_-_1_targets-12      10.6µs ± 0%  16.0µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_100000_bytes_-_1_targets-12     364µs ± 0%   132µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_0_bytes_-_5_targets-12         10.2µs ± 0%  11.6µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_1000_bytes_-_5_targets-12      12.8µs ± 0%  13.7µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_100000_bytes_-_5_targets-12     109µs ± 0%   326µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_0_bytes_-_10_targets-12        10.3µs ± 0%  12.6µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_1000_bytes_-_10_targets-12     17.0µs ± 0%  12.8µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_100000_bytes_-_10_targets-12    108µs ± 0%   411µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_0_bytes_-_50_targets-12        10.5µs ± 0%  22.2µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_1000_bytes_-_50_targets-12     12.6µs ± 0%  23.3µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_100000_bytes_-_50_targets-12    101µs ± 0%   114µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_0_bytes_-_100_targets-12       10.1µs ± 0%  39.0µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_1000_bytes_-_100_targets-12    17.8µs ± 0%  41.3µs ± 0%   ~     (p=1.000 n=1+1)
IngressHandler/event_size_100000_bytes_-_100_targets-12   622µs ± 0%   698µs ± 0%   ~     (p=1.000 n=1+1)

grantr (approved):
Nice work @tayarani!

/lgtm
/hold
Holding in case others who reviewed have comments.

liu-cong:
I am going to cancel the hold since this has been approved for a week.

/unhold

PR #1790, opened 2020-10-01
Adjusting the large payload unit test to prevent occasional timeouts

Ectelion:
Reducing payload size from 100Mb to 10mb and increasing timeout to 10 seconds for the unit test _"malicious requests or requests with unknown Content-Length and a very large payload"_ to prevent occasional timeouts. This does not reproduce on local environments, but occasionally happens on the remote test server.

### DIFF
--- a/pkg/broker/ingress/handler_test.go
+++ b/pkg/broker/ingress/handler_test.go
@@ -179,14 +179,14 @@ func TestHandler(t *testing.T) {
 			name:     "an event with a very large payload",
 			method:   "POST",
 			path:     "/ns1/broker1",
-			event:    createTestEventWithPayloadSize("test-event", 110000000),
+			event:    createTestEventWithPayloadSize("test-event", 10000001), // 10Mb + extra byte
 			wantCode: nethttp.StatusRequestEntityTooLarge,
 		},
 		{
 			name:          "malicious requests or requests with unknown Content-Length and a very large payload",
 			method:        "POST",
 			path:          "/ns1/broker1",
-			event:         createTestEventWithPayloadSize("test-event", 110000000),
+			event:         createTestEventWithPayloadSize("test-event", 10000001), // 10Mb + extra byte
 			contentLength: ptr.Int64(-1),
 			wantCode:      nethttp.StatusRequestEntityTooLarge,
 		},
--- a/pkg/broker/ingress/handler_test.go
+++ b/pkg/broker/ingress/handler_test.go
@@ -119,7 +119,7 @@ type testCase struct {
 	eventAssertions []eventAssertion
 	decouple        DecoupleSink
 	contentLength   *int64
-	timeout time.Duration
+	timeout         time.Duration
 }
 
 type fakeOverloadedDecoupleSink struct{}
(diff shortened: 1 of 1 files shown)


pkg/broker/ingress/handler_test.go
```diff
@@ -179,14 +179,14 @@ func TestHandler(t *testing.T) {
 			name:     "an event with a very large payload",
 			method:   "POST",
 			path:     "/ns1/broker1",
-			event:    createTestEventWithPayloadSize("test-event", 110000000),
+			event:    createTestEventWithPayloadSize("test-event", 10000001), // 10Mb + extra byte
```
zhongduo, line 182: I felt we don't need to set it that strict, like if somehow pubsub happens to allow extra 20 bytes for metadata, this test won't work. We would suggest to use 11MB instead.

Ectelion: Sure, I believe that won't affect the test though: it tests not PubSub itself, but our request validator on Ingress.

zhongduo:
Let's re trigger the unit test several times to make sure the flakiness is indeed resolved.

Ectelion:
Succeeded once. Checking out again.
/retest

zhongduo:
> Succeeded once. Checking out again.
> /retest

This won't work..... You have to use sth like `/test pull-google-knative-gcp-unit-tests `

Ectelion:
Succeeded the second time. Trying out again.
/test pull-google-knative-gcp-unit-tests

Ectelion:
Ok, it did succeed in all the latest re-runs ([1](https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/google_knative-gcp/1790/pull-google-knative-gcp-unit-tests/1311783536335785985), [2](https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/google_knative-gcp/1790/pull-google-knative-gcp-unit-tests/1311787184876949504), [3](https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/google_knative-gcp/1790/pull-google-knative-gcp-unit-tests/1311793476731408384), [4](https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/google_knative-gcp/1790/pull-google-knative-gcp-unit-tests/1311795238494277632)) and all took around 2-3 seconds to complete -- safely below current timeout. So, I think we are good here.

yolocs:
If you make the max size configurable, we can have a much lower limit in test.

pkg/broker/ingress/handler_test.go
```diff
@@ -295,7 +298,11 @@ func TestHandler(t *testing.T) {
 		t.Run(tc.name, func(t *testing.T) {
 			reportertest.ResetIngressMetrics()
 			ctx := logging.WithLogger(context.Background(), logtest.TestLogger(t))
-			ctx, cancel := context.WithTimeout(ctx, 3*time.Second)
+			timeout := 3 * time.Second
```
zhongduo, line 301: As discussed, increasing this default might reduce the flakiness in the future as we have more tests closed to 3s. Maybe change it to 5?

Ectelion: Yeah, makes sense, given recent observations (slightly above 3 seconds even for the HealthCheck test). Have increased to 5 for the rest.

Ectelion:
> If you make the max size configurable, we can have a much lower limit in test.

Nice idea! From recent observations, it seems it is not so much about the 10Mb payload (another timeout observed on a test with a lightweight payload), but strict timeout of 3 seconds and some sort of a slowdown with these tests in general.

PR #1792, opened 2020-10-02
Code cleanup in trigger and broker controller

zhongduo:
Remove duplicate metadataclient query in both broker and trigger
controllers. Use single ProjectIDEnvConfig for parsing project
id from env, to avoid inconsistency in the future.

Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/reconciler/broker/controller.go
+++ b/pkg/reconciler/broker/controller.go
@@ -18,9 +18,9 @@ package broker
 
 import (
 	"context"
-	"os"
 
 	"github.com/google/knative-gcp/pkg/apis/configs/dataresidency"
+	"github.com/kelseyhightower/envconfig"
 
 	"cloud.google.com/go/pubsub"
 	"go.uber.org/zap"
@@ -50,6 +50,10 @@ const (
 	controllerAgentName = "broker-controller"
 )
 
+type envConfig struct {
+	utils.ProjectIDEnvConfig
+}
+
 type Constructor injection.ControllerConstructor
 
 // NewConstructor creates a constructor to make a Broker controller.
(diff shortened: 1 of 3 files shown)


pkg/reconciler/broker/controller.go
```diff
@@ -124,16 +135,3 @@ func newController(ctx context.Context, cmw configmap.Watcher, drs *dataresidenc
 
 	return impl
 }
-
```
capri-xiyue, line 127: It looks like we deleted this helper method and put the logic to each caller? Instead of deleting this helper method, can we put it in a common package so that everyone can call it and we also can get rid of the duplicate?

zhongduo: That's what I was trying to do at the beginning, as I wanted to have sth similar to the sources as well. The original code actually called `utils.ProjectID` already, so this helper function is not really useful at all. Also, this helper function is too small and not generic enough to be moved to a common place.

pkg/utils/metadata.go
```diff
@@ -25,6 +25,11 @@ const (
 	ProjectIDEnvKey = "PROJECT_ID"
```
tommyreddad, line 25: Do we still need this if we're doing everything through `envconfig`?

pkg/utils/metadata.go
```diff
@@ -25,6 +25,11 @@ const (
...
 )
 
+// ProjectIDEnvConfig is a struct to parse Project id from env var
```
tommyreddad, line 28: nit: "project ID" rather than "Project id"

pkg/reconciler/trigger/controller.go
```diff
@@ -73,26 +73,28 @@ func NewConstructor(dataresidencyss *dataresidency.StoreSingleton) Constructor {
...
 func newController(ctx context.Context, cmw configmap.Watcher, drs *dataresidency.Store) *controller.Impl {
+	// Parse all env vars of interest
 	var env envConfig
 	if err := envconfig.Process("", &env); err != nil {
 		logging.FromContext(ctx).Fatal("Failed to process env var", zap.Error(err))
 	}
 
 	triggerInformer := triggerinformer.Get(ctx)
 
+	var client *pubsub.Client = nil
```
tommyreddad, line 84: nit: pointer zero value is `nil`, so `var client *pubsub.Client = nil` is redundant, you could just say `var client *pubsub.Client`

zhongduo: I tried to avoid implicit default value, as it is confusing to me. Pointer and object is OK, `int` and `bool` can be quite confusing. But if this is a Go style, I can change that. Or will `client := *pubsub.Client(nil)` works?

tommyreddad: Just looking through the code base, the convention seems to be `var name *type`. I can't find a single use of `var name *type = nil` nor `name := *type(nil)`.

tommyreddad: This is not a big deal though, it's up to you

tommyreddad:
/lgtm
/approve
/hold

Unhold if you feel like addressing the nits.

pkg/reconciler/trigger/controller.go
zhongduo: Fixed as suggested, thanks

pkg/utils/metadata.go
zhongduo: Fixed as suggested.

PR #1799, opened 2020-10-05
Use single function to resolve project ID

### LINKED ISSUE #1781: Infer project id from environment variable if not set in sources
**Problem**

The sources, such as the topic, currently expect project id to be set in the spec. If
not set, it uses the metadata client to infer the project id. In the broker and trigger,
an environmental variable is also used to work with non-gcp system. We should 
change the source logic to infer project id in the following order:

1. spec.project
2. env `PROJECT_ID`
3. gcp metadata client

**Exit Criteria**

**Time Estimate (optional):** 

zhongduo:
Before this change different reconcilers resolve project ID in different
ways, eg broker controller respect env var "PROJECT_ID" but in reconciler
it does not consider it. This kind of inconsistency will result in hard
to debug problem regarding project id resolve as not all the paths to
resolve project ID will be run regularly. In this change a new helper
function `ProjectIDOrDefault` is defined to resolve project id from
both env var and metadata client. This fix will be applied to all sources
too.

Fixes #1781 

## Proposed Changes

- Add helper function `ProjectIDOrDefault`
- Replace all occurrence of `ProjectID` with `ProjectIDOrDefault` call.
-

- 🎁 Infer project ID from environmental variable in all sources and brokers

**Release Note**

```release-note
- Infer project ID from environmental variable in all sources and brokers
```

**Docs**

### DIFF
--- a/pkg/reconciler/events/scheduler/scheduler.go
+++ b/pkg/reconciler/events/scheduler/scheduler.go
@@ -30,7 +30,6 @@ import (
 	v1 "github.com/google/knative-gcp/pkg/apis/events/v1"
 	cloudschedulersourcereconciler "github.com/google/knative-gcp/pkg/client/injection/reconciler/events/v1/cloudschedulersource"
 	listers "github.com/google/knative-gcp/pkg/client/listers/events/v1"
-	metadataClient "github.com/google/knative-gcp/pkg/gclient/metadata"
 	gscheduler "github.com/google/knative-gcp/pkg/gclient/scheduler"
 	"github.com/google/knative-gcp/pkg/reconciler/events/scheduler/resources"
 	"github.com/google/knative-gcp/pkg/reconciler/identity"
@@ -95,7 +94,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, scheduler *v1.CloudSched
 
 func (r *Reconciler) reconcileJob(ctx context.Context, scheduler *v1.CloudSchedulerSource, topic, jobName string) error {
 	if scheduler.Status.ProjectID == "" {
-		projectID, err := utils.ProjectID(scheduler.Spec.Project, metadataClient.NewDefaultMetadataClient())
+		projectID, err := utils.ProjectIDOrDefault(ctx, scheduler.Spec.Project)
 		if err != nil {
 			logging.FromContext(ctx).Desugar().Error("Failed to find project id", zap.Error(err))
 			return err
--- a/pkg/reconciler/events/scheduler/scheduler.go
+++ b/pkg/reconciler/events/scheduler/scheduler.go
@@ -94,7 +94,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, scheduler *v1.CloudSched
 
 func (r *Reconciler) reconcileJob(ctx context.Context, scheduler *v1.CloudSchedulerSource, topic, jobName string) error {
 	if scheduler.Status.ProjectID == "" {
-		projectID, err := utils.ProjectIDOrDefault(ctx, scheduler.Spec.Project)
+		projectID, err := utils.ProjectIDOrDefault(scheduler.Spec.Project)
 		if err != nil {
 			logging.FromContext(ctx).Desugar().Error("Failed to find project id", zap.Error(err))
 			return err
--- a/pkg/utils/metadata.go
+++ b/pkg/utils/metadata.go
@@ -17,7 +17,6 @@ limitations under the License.
 package utils
 
 import (
-	"context"
 	"os"
 
 	metadataClient "github.com/google/knative-gcp/pkg/gclient/metadata"
@@ -49,7 +48,7 @@ type ProjectIDEnvConfig struct {
 // 1) if the input project ID is valid, simply use it.
 // 2) if there is a PROJECT_ID environmental variable, use it.
 // 3) use metadataClient to resolve project id.
-func ProjectIDOrDefault(ctx context.Context, projectID string) (string, error) {
+func ProjectIDOrDefault(projectID string) (string, error) {
 	if projectID != "" {
 		return projectID, nil
 	}
--- a/pkg/utils/metadata.go
+++ b/pkg/utils/metadata.go
@@ -18,21 +18,27 @@ package utils
 
 import (
 	"context"
+	"os"
 
 	metadataClient "github.com/google/knative-gcp/pkg/gclient/metadata"
-	"github.com/google/knative-gcp/pkg/logging"
-	"github.com/kelseyhightower/envconfig"
-	"go.uber.org/zap"
 )
 
 const (
 	clusterNameAttr = "cluster-name"
+	// ProjectIDEnvKey is the name of environmental variable for project ID
 	ProjectIDEnvKey = "PROJECT_ID"
 )
 
 // defaultMetadataClientCreator is a create function to get a default metadata client. This can be
 // swapped during testing.
 var defaultMetadataClientCreator func() metadataClient.Client = metadataClient.NewDefaultMetadataClient
+
+// projectIDFromEnv loads project ID from env once when startup.
+var projectIDFromEnv string
+
+func init() {
+	projectIDFromEnv = os.Getenv(ProjectIDEnvKey)
+}
 
 // ProjectIDEnvConfig is a struct to parse project ID from env var
 type ProjectIDEnvConfig struct {
@@ -61,12 +67,7 @@ func ProjectIDOrDefault(ctx context.Context, projectID string) (string, error) {
 	if projectID != "" {
 		return projectID, nil
 	}
-	// Parse project id from env var, if not found we simply continue
-	var env ProjectIDEnvConfig
-	if err := envconfig.Process("", &env); err != nil {
-		logging.FromContext(ctx).Error("Failed to process env var", zap.Error(err))
-	}
-	return ProjectID(env.ProjectID, defaultMetadataClientCreator())
+	return ProjectID(projectIDFromEnv, defaultMetadataClientCreator())
 }
 
 // ClusterName returns the cluster name for a particular resource.
--- a/pkg/utils/metadata.go
+++ b/pkg/utils/metadata.go
@@ -17,13 +17,22 @@ limitations under the License.
 package utils
 
 import (
+	"context"
+
 	metadataClient "github.com/google/knative-gcp/pkg/gclient/metadata"
+	"github.com/google/knative-gcp/pkg/logging"
+	"github.com/kelseyhightower/envconfig"
+	"go.uber.org/zap"
 )
 
 const (
 	clusterNameAttr = "cluster-name"
 	ProjectIDEnvKey = "PROJECT_ID"
 )
+
+// defaultMetadataClientCreator is a create function to get a default metadata client. This can be
+// swapped during testing.
+var defaultMetadataClientCreator func() metadataClient.Client = metadataClient.NewDefaultMetadataClient
 
 // ProjectIDEnvConfig is a struct to parse project ID from env var
 type ProjectIDEnvConfig struct {
@@ -42,6 +51,22 @@ func ProjectID(project string, client metadataClient.Client) (string, error) {
 		return "", err
 	}
 	return projectID, nil
+}
+
+// ProjectIDOrDefault returns the project ID by performing the following order:
+// 1) if the input project ID is valid, simply use it.
+// 2) if there is a PROJECT_ID environmental variable, use it.
+// 3) use metadataClient to resolve project id.
+func ProjectIDOrDefault(ctx context.Context, projectID string) (string, error) {
+	if projectID != "" {
+		return projectID, nil
+	}
+	// Parse project id from env var, if not found we simply continue
+	var env ProjectIDEnvConfig
+	if err := envconfig.Process("", &env); err != nil {
+		logging.FromContext(ctx).Fatal("Failed to process env var", zap.Error(err))
+	}
+	return ProjectID(env.ProjectID, defaultMetadataClientCreator())
 }
 
 // ClusterName returns the cluster name for a particular resource.
--- a/pkg/utils/metadata.go
+++ b/pkg/utils/metadata.go
@@ -45,20 +45,6 @@ type ProjectIDEnvConfig struct {
 	ProjectID string `envconfig:"PROJECT_ID"`
 }
 
-// ProjectID returns the project ID for a particular resource.
-func ProjectID(project string, client metadataClient.Client) (string, error) {
-	// If project is set, then return that one.
-	if project != "" {
-		return project, nil
-	}
-	// Otherwise, ask GKE metadata server.
-	projectID, err := client.ProjectID()
-	if err != nil {
-		return "", err
-	}
-	return projectID, nil
-}
-
 // ProjectIDOrDefault returns the project ID by performing the following order:
 // 1) if the input project ID is valid, simply use it.
 // 2) if there is a PROJECT_ID environmental variable, use it.
@@ -67,7 +53,15 @@ func ProjectIDOrDefault(ctx context.Context, projectID string) (string, error) {
 	if projectID != "" {
 		return projectID, nil
 	}
-	return ProjectID(projectIDFromEnv, defaultMetadataClientCreator())
+	if projectIDFromEnv != "" {
+		return projectIDFromEnv, nil
+	}
+	// Otherwise, ask GKE metadata server.
+	projectGKE, err := defaultMetadataClientCreator().ProjectID()
+	if err != nil {
+		return "", err
+	}
+	return projectGKE, nil
 }
 
 // ClusterName returns the cluster name for a particular resource.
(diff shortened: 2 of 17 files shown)


pkg/utils/metadata_test.go
```diff
@@ -67,6 +70,66 @@ func TestProjectID(t *testing.T) {
...
+			},
+			input: "",
+			error: true,
+		},
+	}
+
+	for n, tc := range testCases {
+		t.Run(n, func(t *testing.T) {
+			// Save and restore changed variables
+			origEnv := os.Getenv(ProjectIDEnvKey)
```
capri-xiyue, line 113: Instead of changing and then restoring the `ProjectIDEnvKey`, is it possible that we write a fake os so that we can avoid changing the env variables in UT?

pkg/utils/metadata.go
```diff
@@ -44,6 +53,22 @@ func ProjectID(project string, client metadataClient.Client) (string, error) {
...
+// 2) if there is a PROJECT_ID environmental variable, use it.
+// 3) use metadataClient to resolve project id.
+func ProjectIDOrDefault(ctx context.Context, projectID string) (string, error) {
+	if projectID != "" {
+		return projectID, nil
+	}
+	// Parse project id from env var, if not found we simply continue
+	var env ProjectIDEnvConfig
+	if err := envconfig.Process("", &env); err != nil {
+		logging.FromContext(ctx).Fatal("Failed to process env var", zap.Error(err))
```
Harwayne, line 67: Calling fatal outside of the init path is frowned upon. Perhaps read it from the env once in an `init()` function, which can call fatal, and use that copy going forward. 

I don't think we plan to allow this value to change at run time, right?

pkg/utils/metadata_test.go
zhongduo: I like this idea. Do you know a way to do it? Let me dig into how env is handled in go UT. I think it is safe to always clear env variables in UT as there should not be such assumption anyway. See how `evnconfig` package test it: https://github.com/kelseyhightower/envconfig/blob/c974cae29cf5b543b30f7ba91f8ce9649cd69818/envconfig_test.go#L91

pkg/utils/metadata.go
zhongduo: This is the original logic in controllers, but I agree that here should be a simple log entry as we have backup plan anyway. 

I am not sure whether we should load it once or not, but it is a question to answer as this function is actually called every time it the reconcilers of all sources are run. Any thoughts @grantr ?

pkg/utils/metadata_test.go
capri-xiyue: If it's safe to change env variables in UT, this comment will just be a nit.
IMO, based on currently implementation, it will be easier if you can write a fake envConfig by yourself.

pkg/utils/metadata.go
```diff
@@ -44,6 +53,22 @@ func ProjectID(project string, client metadataClient.Client) (string, error) {
...
+// ProjectIDOrDefault returns the project ID by performing the following order:
+// 1) if the input project ID is valid, simply use it.
+// 2) if there is a PROJECT_ID environmental variable, use it.
+// 3) use metadataClient to resolve project id.
+func ProjectIDOrDefault(ctx context.Context, projectID string) (string, error) {
+	if projectID != "" {
+		return projectID, nil
+	}
+	// Parse project id from env var, if not found we simply continue
+	var env ProjectIDEnvConfig
```
grac3gao, line 65: I am not sure if this makes things more complicated. `envConfig` is used in lots of places. In most of the cases, there are multiple fields in `envConfig` struct (in addition to `ProjectID`), and the common way is using `envconfig.Process` to parse all the fields. Here, you picked up two cases whose `envConfig` happen to only have `ProjectID`, and wrapped the `envconfig.Process` in new `ProjectIDOrDefault`, what if later those two places need more fields in `envConfig` struct (may cause duplicate usage of `envconfig.Process`)?

zhongduo: Yes, I am aware of that and have thought about this very carefully. If you see my previous cleanup for broker and trigger, I kept the parsing part in the controller because of exactly what you said. Here is the reason I decided to do this now:

1. This code is not performance sensitive. And the the performance of `Process` function is similar to `os.lookupEnv` IIUC.
2. In case we want to pass the env var, very likely we will save the result, like what we do in the broker controller.
3. I am happy to use `os.Getenv("PROJECT_ID")` here if we are really concerned about performance difference, but from last conversation with @grantr envconfig a preferred way.

pkg/utils/metadata.go
zhongduo: Changed Fatal to Error.

pkg/utils/metadata.go
grac3gao: For me, it is less readable to change like this. Before this change, developers only need to know `envconfig.Process` logic in the files which has `envConfig` struct. Now, since you are changing it in all parts (include places where doesn’t need to deal with `envconfig` struct), developers need to know this parsing logic due to the usage of `ProjectIDOrDefault` function. It's an overexposure in my understanding. But that’s just my personal opinion.

zhongduo: I am actually not sure what you mean here. User will never need to use `envconfig.Process` or even import `envconfig` directly after this. The only function exposed here is `ProjectIDOrDefault`. In fact, I can simply use `os.Getenv` here, but either way user should not worry about it at all. Do I miss sth here?

pkg/utils/metadata.go
```diff
@@ -44,6 +53,22 @@ func ProjectID(project string, client metadataClient.Client) (string, error) {
...
+	if err := envconfig.Process("", &env); err != nil {
```
liu-cong, line 66: It seems that we have two use cases: 1. read a PROJECT_ID from env var, or apply default; 2. get project from spec, or apply default. I recommend that the existing `ProjectID` function to be renamed to `ProjectIDOrDefault`, and rename this one `ProjectIDFromEnvOrDefault`.

pkg/utils/metadata.go
grac3gao: I mean when developers use `ProjectIDOrDefault`, they will likely look into the logic, and found it is checking `envconfig`, if they don't know the detail of parsing logic for it (suppose they are not changing a file which has `envconfig`), they need to have additional knowledge to know whether this `ProjectIDOrDefault` is suitable for their use case. It increases cognitive load for them. It might be better to have a clear function name (as Cong suggested) and together with more clear comments.

pkg/reconciler/events/scheduler/scheduler.go
```diff
@@ -95,7 +94,7 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, scheduler *v1.CloudSched
 
 func (r *Reconciler) reconcileJob(ctx context.Context, scheduler *v1.CloudSchedulerSource, topic, jobName string) error {
 	if scheduler.Status.ProjectID == "" {
-		projectID, err := utils.ProjectID(scheduler.Spec.Project, metadataClient.NewDefaultMetadataClient())
+		projectID, err := utils.ProjectIDOrDefault(ctx, scheduler.Spec.Project)
```
capri-xiyue, line 97: Is the logic correct that Source also read projectID from env? I remember at first, we didn't have such logic for sources?

pkg/utils/metadata.go
capri-xiyue: +1, Not sure whether sources need read PROJECT_ID from env var or not. This pr changes the original business logic.

pkg/reconciler/events/scheduler/scheduler.go
zhongduo: You are right. This PR is to solve #1799 which is to use environment variable in sources. That's why I want to have one single function to do it. The motivation is that sources should have the same logic as brokers in terms of resolving project id. Please check the sources part to see if you are seeing any place that we shouldn't. The issue comes out from my previous discussion with Nacho and Grant.

pkg/utils/metadata.go
zhongduo: Sorry about the confusion here. This PR is not for clean up. It is mainly to introduce env var to sources #1781 , or in another word it is to have one single way of resolving project id in `knative-gcp`. @liu-cong That's why I used `ProjectIDOrDefault` instead of using implementation details as the name. The idea is that user shouldn't know the implementation details, or change their code when we decide to change the logic of resolving project id.

pkg/utils/metadata.go
zhongduo: Sorry about the confusion here. This PR is not for clean up. It is mainly to introduce env var to sources #1781 , or in another word it is to have one single way of resolving project id in `knative-gcp`. The idea is that user shouldn't know the implementation details, or change their code when we decide to change the logic of resolving project id. And caller shouldn't need to worry about the details, as the function is used as a golden way to resolve project id in `knative-gcp`.

zhongduo: Though I agree that different code can have different ways to resolve project id differently if it is on purpose, from my previous discussion with Grant and Nacho in `knative-gcp` the difference is more due to inconsistency of the code. Eg, in broker, why would controller consider env var but not in the reconciler? Same thing goes to sources, env var supported was added to brokers after sources, and the logic has got out of sync. So I am trying to have a golden way of resolving project id if not provided that we shouldn't need to worry about the implementation details at all.

pkg/utils/metadata_test.go
zhongduo: This code is removed, we now only infer project id once during startup.

pkg/utils/metadata.go
zhongduo: Changed as suggested, the project id from env var will only be initialized during startup.

pkg/utils/metadata.go
zhongduo: We now used `os.Getenv` in `init` function.

pkg/utils/metadata.go
liu-cong: Thanks for the explanation. Looks like we don't call the `ProjectID` any more in clients? If so, let's un-export it, or even inline the logic in the new function. I still think the `ProjectIDOrDefault` name does not reflect exactly what this function is doing. I think it should be renamed to either something verbose like `ProjectIDOrEnvVarOrDefault` or something general like `GetProjectID`

zhongduo: We definitely want to avoid having `EnvVar` as we might change the logic in the future, eg we decided to drop support for env var and use config file instead. `GetProjectID` works to me too,  but I think it is pretty much same as `ProjectIDOrDefault`. As for the original `ProjectID`, I am not sure if we should remove it here, because it is still possible that the function can be used.

liu-cong: Can you do a search to see if `ProjectID` is still used after this change? It will be confusing to have two helper functions that seem to be similar from their names. (To me, the old `ProjectID` does `ProjectIDOrDefault` logic)

zhongduo: I did a search and replacement, so there won't be any usage of `ProjectID` now, but I am waiting for @capri-xiyue to confirm this is OK in sources. Either way the problem is really whether we want to keep such a helper function or not, if not, I think removing it is better.

zhongduo: Remove the original `ProjectID` as suggested.

pkg/utils/metadata.go
```diff
@@ -17,31 +17,51 @@ limitations under the License.
...
-// ProjectID returns the project ID for a particular resource.
-func ProjectID(project string, client metadataClient.Client) (string, error) {
-	// If project is set, then return that one.
-	if project != "" {
-		return project, nil
+// ProjectIDOrDefault returns the project ID by performing the following order:
+// 1) if the input project ID is valid, simply use it.
+// 2) if there is a PROJECT_ID environmental variable, use it.
+// 3) use metadataClient to resolve project id.
+func ProjectIDOrDefault(ctx context.Context, projectID string) (string, error) {
```
grantr, line 52: Why does this function take a context? It doesn't seem to be used.

pkg/utils/metadata_test.go
```diff
@@ -17,19 +17,22 @@ limitations under the License.
...
 	testingMetadataClient "github.com/google/knative-gcp/pkg/gclient/metadata/testing"
 )
 
-func TestProjectID(t *testing.T) {
+func TestProjectIDOrDefault(t *testing.T) {
 	testCases := map[string]struct {
 		want  string
 		data  testingMetadataClient.TestClientData
 		input string
+		env   string
```
grantr, line 35: Can you add a test case that uses this? I don't see one.

grantr:
>  Not sure whether sources need read PROJECT_ID from env var or not.

@capri-xiyue Do sources always have `spec.project` filled in by the webhook even when the user doesn't specify them?

pkg/utils/metadata.go
```diff
@@ -17,31 +17,51 @@ limitations under the License.
...
+	if projectID != "" {
+		return projectID, nil
+	}
+	if projectIDFromEnv != "" {
+		return projectIDFromEnv, nil
```
grac3gao, line 57: It seems we didn't have unit test for function `init()`(maybe I missed that). Did you manually test, if env `PROJECT_ID` has value (assume it has a different value than what would get from  defaultMetadataClientCreator()), the function` init()` will parse that in?

pkg/utils/metadata_test.go
grantr: Oops I missed the test just a few lines down...

pkg/utils/metadata.go
capri-xiyue: > > Not sure whether sources need read PROJECT_ID from env var or not.
> 
> @capri-xiyue Do sources always have `spec.project` filled in by the webhook even when the user doesn't specify them?

No. Sources won't have `spec.project` filled in by the webhook when the user doesn't specify them.

capri-xiyue:
> > Not sure whether sources need read PROJECT_ID from env var or not.
> 
> @capri-xiyue Do sources always have `spec.project` filled in by the webhook even when the user doesn't specify them?

No. Sources won't have spec.project filled in by the webhook when the user doesn't specify them.

pkg/utils/metadata.go
zhongduo: Not sure if I understand the question here. We didn't test `init()` in unit test, though I actually test it manually to confirm `os.Getenv` works. The `init` is called only when the package is loaded, so after that, we will simply use the `projectIDFromEnv` directly and don't read the env again.

grant:
> > > Not sure whether sources need read PROJECT_ID from env var or not.
> > 
> > 
> > @capri-xiyue Do sources always have `spec.project` filled in by the webhook even when the user doesn't specify them?
> 
> @grant No. Sources won't have spec.project filled in by the webhook when the user doesn't specify them.

I think you mean @grantr.

pkg/utils/metadata.go
zhongduo: Removed.

pkg/utils/metadata.go
grantr: It looks like `spec.project` is translated into the `PROJECT_ID` env var for the receive adapter. So if there's an existing E2E test that verifies `spec.project` gets down to the receive adapter, that would also verify the operation of the `init()` function that reads `PROJECT_ID`.

grac3gao: @zhongduo The question is for verifying the operation of the `init()` function. 
@grantr Thank you for reminding this!
In most of the case, the env `project_id` maybe the same whether we get it from os.Getenv, or from the client. Just want to make sure we manually test that: when we set `project_id` env, it returns the value from `init()`, not from the client.

zhongduo: I see. Yes, i manually tested it by commenting out the line setting `projectIDFromEnv` in unit test, and the test now gets my `PROJECT_ID` env var.

grantr: @capri-xiyue do we have an E2E test that verifies the `spec.project` field in sources works correctly?

capri-xiyue: @grantr No, we don't. In addition, the idea of have `project` in spec is to support cross-project stuff as Nacho mentioned in this comment https://github.com/google/knative-gcp/pull/1761#discussion_r496937485. It may be difficult to do e2e test for cross-project stuff in prow. If we just verify `spec.project` when the `spec.project` is the same as the default project of gke server, the e2e test seems useless.

zhongduo: Most of the unit tests in sources rely no project field in the spec, so I think that feature is well tested. And the logic currently is: if project is not set, we use the default project. That's why you see there are lots of `ProjectIDOrDefault` calls in sources.

grantr: Ok, sounds like a manual test of the init behavior is fine for now.

grantr (approved):
Thanks @zhongduo! Holding in case you want to fix up commit messages. If not, cancel the hold.

/lgtm
/hold

zhongduo:
> Thanks @zhongduo! Holding in case you want to fix up commit messages. If not, cancel the hold.
> 
> /lgtm
> /hold

Thanks for checking. I think my first commit is done properly and nothing is wrong since then. So I after squash it will be ok. /unhold

PR #1800, opened 2020-10-05
Upgrade knative repos to bring in fix in pkg for delay informer start

### LINKED ISSUE #1749: Reconcile deleted topics created by broker or trigger
**Describe the bug**
Currently if a retry topic is deleted, the trigger reconciler cannot recreate it in minutes. I have tested with 30 mins at least.

 **Expected behavior**
When the resync period is reached, which is 5 mins currently, the deleted topic should be recreated.

 **To Reproduce**
Follow the gcp example, then delete the retry topic, and wait long enough to see if the topic is recreated.

 **Additional context**

zhongduo:
Note that the serving version is pinned to 1.8.0 as we are depending on
serving beta version. This should be removed after we upgrade all the
dependency code. Other than that, the commit is generated by:

```
./hack/update-deps.sh --upgrade
./hack/update-codegen.sh
```

There is a problem in `broker_test.go` and I have to use network package
instead of eventing/utils for `GetClusterDomainName` call.

Fixes #1749 

## Proposed Changes

-
-
-

- 🐛 Bring in bug fix in pkg to make resync period in informer setup work again
**Release Note**

```release-note

```

**Docs**


nlopezgi:
Could you point here to any tracking issue for upgrading all the dependency code so we can remove the pin later on? 
/lgtm
/approve

zhongduo:
> Could you point here to any tracking issue for upgrading all the dependency code so we can remove the pin later on?
> /lgtm
> /approve

I opened an issue in #1801 to keep tracking.

PR #1806, opened 2020-10-06
Fix prerequisite section of CloudBuildSource doc

eclipselu:
Fixes #

## Proposed Changes

- Fixes instructions on how to receive cloud build notifications. 

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/docs/examples/cloudbuildsource/README.md
+++ b/docs/examples/cloudbuildsource/README.md
@@ -13,17 +13,14 @@ build completes.
 
 1. [Create a Service Account for Data Plane](../../install/dataplane-service-account.md)
 
-1. Enable the `Cloud Build API` and `Cloud Pub/Sub API`, on your project:
+1. Enable the `Cloud Build API` on your project, this is needed to submit builds via gcloud cli.
 
    ```shell
    gcloud services enable cloudbuild.googleapis.com
-   gcloud services enable pubsub.googleapis.com
    ```
 
-   The Pub/Sub topic to which Cloud Build publishes these build update messages
-   is called `cloud-builds`. When you enable the `Cloud Build API` and
-   `Cloud Pub/Sub API`, the `cloud-builds` topic will automatically be created
-   for you.
+1. Follow the `Receiving build notifications` section in [Subscribing to build notifications](https:/cloud.google.com/cloud-build/docs/subscribe-build-notifications#receiving_build_notifications) to set up Cloud Build to emit build notifications.
+
 
 ## Deployment
 


docs/examples/cloudbuildsource/README.md
```diff
@@ -20,10 +20,13 @@ build completes.
...
-   is called `cloud-builds`. When you enable the `Cloud Build API` and
-   `Cloud Pub/Sub API`, the `cloud-builds` topic will automatically be created
-   for you.
+1. Create the `cloud-builds` topic:
+
+   ```shell
+   gcloud pubsub topics create cloud-builds
+   ```
+
+   Cloud Build publishes messages to this Pub/Sub topic when your build's state changes.
```
nlopezgi, line 29: Is it no longer necessary to enable the `Cloud Build API`, whether or not it creates the `cloud-builds` topic?

eclipselu: > Is it no longer necessary to enable the Cloud Build API

Cloud Build API is still needed because in [Publish](https://github.com/google/knative-gcp/blob/master/docs/examples/cloudbuildsource/README.md#publish) we'll need the API to submit builds via gcloud cli. 

> whether or not it creates the `cloud-builds` topic?

Enabling `Cloud Build API` and `Cloud Pub/Sub API` no longer create the `cloud-builds` topic automatically, according to https://cloud.google.com/cloud-build/docs/subscribe-build-notifications#receiving_build_notifications , I also verified this myself.  User need to create the topic manually.

Do you think if it's better to just replace this section with a link: [Setup Cloud Build to receive build notifications](https://cloud.google.com/cloud-build/docs/subscribe-build-notifications#receiving_build_notifications) ?

nlopezgi: yes, better to just link there imo.

zhongduo: I think for example purpose, it is better to have the link but at the same time put the command here. It is less mental effort for people that just want to follow the instruction and see it in action. Especially considering that fact that the doc you are linking is not very clear how it is related to this.

nlopezgi: The issue with having the command here is that it might get out of date (as it did, and that is why you're changing it now, iiuc?). If we just point to the source of truth there is no maintenance overhead.

eclipselu: > I think for example purpose, it is better to have the link but at the same time put the command here. It is less mental effort for people that just want to follow the instruction and see it in action. Especially considering that fact that the doc you are linking is not very clear how it is related to this.

I agree with @nlopezgi , here we just need to guide user to setup Cloud Build to emit events, how to setup it is the implementation details that is encapsulated by the GCP doc link and that might change in the future. 

If we have both the command and link and the command is out of date, then 1) it will be confusing to user: which doc should I follow? executing the command here or follow the linked doc. 2) We'll need to update the doc again. 

With that, I suggest that we

- state clear how the GCP doc link is related here, remove the command and keep the link
- remove `Cloud Pub/Sub API` setup in previous step, as it's covered by the GCP doc. 
- state clear in the previous step on why `Cloud Build API` setup is needed. 

It will be something like below

## Prerequisites
1. [Install Knative-GCP](../../install/install-knative-gcp.md)
1. [Create a Service Account for Data Plane](../../install/dataplane-service-account.md)
1. Enable the `Cloud Build API` on your project, this is needed to submit builds via gcloud cli.
   ```shell
   gcloud services enable cloudbuild.googleapis.com
   ```
1. [Setup Cloud Build to receive build notifications](https://cloud.google.com/cloud-build/docs/subscribe-build-notifications#receiving_build_notifications) 

What do you think @zhongduo @nlopezgi

zhongduo: I think you should at least scope it down to only `Receiving build notifications` section. The problem is that the doc is difficult to follow and it is not clear which step to do here. Eg, you probably don't want user to follow the Push section.

eclipselu: Oh sorry, the link does point to the `Receiving build notification` section but it's not clear here. Thanks for pointing out.

zhongduo: It does point to that section, but it is also difficult to know which sections to stop.

eclipselu: Thanks! updated, please take a look,

PR #1809, opened 2020-10-06
Add test with data residency configuration to topic_test.go

### LINKED ISSUE #1740: No unit tests for data residency config in Topic reconciler
We'll try to fix #1739 first. See https://github.com/google/knative-gcp/pull/1681#discussion_r491141529.

/assign @zhongduo

zhongduo:
Now that the topic reconciler is using pubsub testing server
for testing, we are using the same utility functions to test
topic configuration created.

Fixes #1740 

## Proposed Changes

- Add another test with data residency configured
-
-

### DIFF
--- a/pkg/reconciler/intevents/topic/topic_test.go
+++ b/pkg/reconciler/intevents/topic/topic_test.go
@@ -46,6 +46,7 @@ import (
 	. "knative.dev/pkg/reconciler/testing"
 	servingv1 "knative.dev/serving/pkg/apis/serving/v1"
 
+	"github.com/google/knative-gcp/pkg/apis/configs/dataresidency"
 	pubsubv1 "github.com/google/knative-gcp/pkg/apis/intevents/v1"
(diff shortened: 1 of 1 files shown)


pkg/reconciler/intevents/topic/topic_test.go
```diff
@@ -738,6 +797,12 @@ func TestAllCases(t *testing.T) {
 				}
 			}
 		}
+
+		// If we found "dataResidencyConfigMap" in OtherData, we create a store with the configmap
+		var drStore *dataresidency.Store
+		if cm, ok := testData["dataResidencyConfigMap"]; ok {
```
capri-xiyue, line 803: It looks like it doesn't match the business logic. Can drStore be un-initialized? Wondering maybe for all test cases, the drStore should be a valid `*dataresidency.Store` except for error scenario of `*dataresidency.Store`

zhongduo: The existing logic now allows `nil` value, in which case it will treat it as no special request for data residency.

PR #1836, opened 2020-10-15
added e2e test for trigger dependency annotation

### LINKED ISSUE #1829: Add E2E test for DependencyAnnotation in Triggers
**Problem**
In order not to regress the dependencyAnnotation functionality, we need an e2e test that first creates a Trigger with the dependency annotation, and then the source. Both should become ready.
Similar to this https://github.com/knative/eventing/issues/2582

**Exit Criteria**
An e2e test to avoid regressions on the dependency annotation functionality

**Time Estimate (optional):** 

capri-xiyue:
Fixes #1829 

## Proposed Changes

- added e2e test for trigger dependency annotation
-
-

**Release Note**

```release-note

```

**Docs**


grac3gao (commented):
/lgtm

PR #1838, opened 2020-10-16
Update documentation for using MTBroker with Pub/Sub channel

tommyreddad:
The documentation for using MTBroker with Pub/Sub channel was out of date.

## Proposed Changes
- Use serviceAccountName instead of googleServiceAccount in config-br-default-channel
- Update docs accordingly


danyinggu (commented):
/lgtm
/approve

PR #1839, opened 2020-10-16
Use a local version of monitoring.yaml rather than use Serving's

Harwayne:
## Proposed Changes

- Use a local version of monitoring.yaml rather than use Serving's nightly version, which no longer exists.
- Analogous to knative/eventing#4275
- Update deps.

**Release Note**

```release-note
NONE
```


test/lib.sh
```diff
@@ -19,11 +19,25 @@
 readonly CLOUD_RUN_EVENTS_CONFIG="config/"
 readonly CLOUD_RUN_EVENTS_ISTIO_CONFIG="config/istio"
 
+# Installs Zipkin for tracing tests.
+readonly KNATIVE_GCP_MONITORING_YAML="test/e2e/config/monitoring.yaml"
```
zhongduo, line 23: I wonder if this file can be put into test-infra if it is used for test purpose only and shared by both eventing and knative-gcp.

Harwayne: I think as-is, knative/eventing makes more sense, as it deploys to that NS. But, I would need a change in knative/eventing to allow us to vendor it in (put a dummy go file in that directory). So to stop the bleeding, I would rather check this in now and do other fixes later.

tayarani: that's a good idea. do u have an issue for it?

PR #1840, opened 2020-10-16
Allow the unit test to run locally.

Harwayne:
## Proposed Changes

- Allow the unit test to run locally.
    - It had been failing trying to build the various GCP clients (e.g. PubSub, Storage, etc.) when the local environment was not setup the way the standard clients expect. 

**Release Note**

```release-note
NONE
```


tommyreddad:
I can run the probe helper unit test locally fine without this PR, so I'm not totally sure what was wrong with it. But the cleaning/refactoring is definitely appreciate anyway!

/lgtm
/approve

Harwayne:
> I can run the probe helper unit test locally fine without this PR, so I'm not totally sure what was wrong with it. But the cleaning/refactoring is definitely appreciate anyway!

This is the failure I see when running locally (on a Mac if it makes a difference):

```
logger.go:130: 2020-10-16T13:20:17.700-0700	FATAL	probe_helper/probe_helper.go:549	Failed to create cloud pubsub client{error 26 0  pubsub(publisher): google: could not find default credentials. See https://developers.google.com/accounts/docs/application-default-credentials for more information.}
```

My guess is that whichever machine you are using has credentials in the expected location, namely the `GOOGLE_APPLICATION_CREDENTIALS` env var points at a credential file of some form.

tommyreddad:
@Harwayne Oh I see the problem now. The probe helper health checker test didn't use the test clients, so it didn't pass them to the probe helper, which then tried to create legitimate ones. Thanks for fixing this!

Issue #1544, opened 2020-08-04
Remove v1alpha1 support for our Resources


nachocano:
**Problem**
In https://github.com/google/knative-gcp/issues/1435, we agreed to remove v1alpha1 versions of all our resources. 
We should do this after 0.17 cuts. 
We should add a migration tool for our users to run before upgrading to 0.18, so that we can move all of their resources to v1beta1 (or v1).

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Developer

**Exit Criteria**
v1alpha1 is not served and is removed from our codebase.

** Additional Context **
See https://github.com/google/knative-gcp/issues/1435 for more context

nachocano:
This should be done before #1413
Also, v1alpha1 ducks in eventing are going to be removed. We need to remove our Channel v1alpha1 that uses those ducks fast if we want to keep on updating dependencies. See https://github.com/knative/eventing/pull/3789

danyinggu:
I did some investigation. Since we haven't had any migration tool before, there might be some customers still using the old cluster with v1alpha1 as the storage version for some sources. If we remove v1alpha1 support before these users migrate their storage version, their side will be crashed. I created a cluster with v1alpha1 as the storage version and serves v1beta1 and v1, then upgrade my cluster without v1alpha1 support (with v1alpha1.served=false, v1alpha1.storage=false), then the existing pods which stored as v1alpha1 will be crashed by
```k describe cloudpubsubsources.v1.events.cloud.google.com
Error from server: conversion webhook for events.cloud.google.com/v1alpha1, Kind=CloudPubSubSource failed: conversion not supported for type [kind=CloudPubSubSource group=events.cloud.google.com version=v1alpha1]

 k get sources
Error from server: conversion webhook for events.cloud.google.com/v1alpha1, Kind=CloudPubSubSource failed: conversion not supported for type [kind=CloudPubSubSource group=events.cloud.google.com version=v1alpha1]```

The root cause is that it needs the conversion logic from v1alpha1 to a higher version, but v1alpha1's logic is removed.
I think we should:
1. Delay the removal of v1alpha1 to the next release so that customers can migrate their storage version to get rid of v1alpha1 as the storage version first.
2. Expand the migration tool to add CloudBuildSource and Channel of v1beta1 storage version (not v1) in this release to make sure customers migrate to v1beta1 for these resources.

danyinggu:
Synced with @nachocano offline. We agreed that
[...]

PR #1853, opened 2020-10-22
Remove v1alpha1 support for Channel

eclipselu:
Part of https://github.com/google/knative-gcp/issues/1544

## Proposed Changes

- 🗑️ Remove v1alpha1 Channel
-
-

**Release Note**

```release-note
Dropped v1alpha1 support for the following resources:

- Channel
```

**Docs**


eclipselu:
Running `./hack/update-codegen.sh` after removing v1alpha1 api for channels does not remove the generated client code, I manually removed them.

cmd/webhook/main.go
```diff
@@ -191,8 +188,6 @@ func newConversionController(ctx context.Context, cmw configmap.Watcher, brokers
 		eventsv1alpha1_    = eventsv1alpha1.SchemeGroupVersion.Version
 		eventsv1beta1_     = eventsv1beta1.SchemeGroupVersion.Version
 		eventsv1_          = eventsv1.SchemeGroupVersion.Version
-		messagingv1alpha1_ = messagingv1alpha1.SchemeGroupVersion.Version
-		messagingv1beta1_  = messagingv1beta1.SchemeGroupVersion.Version
```
capri-xiyue, line 195: Why do remove v1beta1?

test/e2e/main_test.go
```diff
@@ -64,14 +64,6 @@ func TestMain(m *testing.M) {
 					eventingtestlib.FeatureRedelivery,
 					eventingtestlib.FeaturePersistence,
 				},
-				{
```
capri-xiyue, line 67: Why do we move this v1beta1 version here?

cmd/webhook/main.go
```diff
@@ -59,8 +57,7 @@ var types = map[schema.GroupVersionKind]resourcesemantics.GenericCRD{
 	brokerv1beta1.SchemeGroupVersion.WithKind("Broker"): &brokerv1beta1.Broker{},
 
 	// For group messaging.cloud.google.com.
-	messagingv1alpha1.SchemeGroupVersion.WithKind("Channel"): &messagingv1alpha1.Channel{},
-	messagingv1beta1.SchemeGroupVersion.WithKind("Channel"):  &messagingv1beta1.Channel{},
```
capri-xiyue, line 63: Why do we remove v1beta1 here? I'm wondering now maybe the default and validation webhook won't work for v1beta1 channel since we remove v1beta1 channel here.

cmd/webhook/main.go
eclipselu, line 195: @capri-xiyue sorry I don't know why your comment is gone after my force push. 

> @capri-xiyue why do we need to remove v1beta1?

this variable is only used below in the conversion map, since we're removing v1alpha1, there's no need to convert.

cmd/webhook/main.go
```diff
@@ -274,15 +269,6 @@ func newConversionController(ctx context.Context, cmw configmap.Watcher, brokers
 					inteventsv1_:       &inteventsv1.Topic{},
 				},
 			},
-			// messaging
-			messagingv1alpha1.Kind("Channel"): {
```
capri-xiyue, line 278: Instead of removing it, maybe we keep v1beat1 channel here? I know it is meaningless since now channel only has one version v1bea1 and there is no need to do conversion webhook here. Any suggestions @Harwayne ?

cmd/webhook/main.go
capri-xiyue: I believe it is also used in validation and defaults webhook. You need to search the usage of `var types`.

cmd/webhook/main.go
eclipselu: sorry I think the diff is a bit misleading, v1beta1 is not removed.

eclipselu: > I believe it is also used in validation and defaults webhook. You need to search the usage of var types.
I think you are referring to this

test/e2e/main_test.go
eclipselu: sorry for the confusion, I changed the constant name below, maybe it's better to keep the variable name consistent. 

```diff
-	MessagingAPIVersion         = "messaging.cloud.google.com/v1alpha1"
-	MessagingV1beta1APIVersion  = "messaging.cloud.google.com/v1beta1"
+	MessagingAPIVersion         = "messaging.cloud.google.com/v1beta1"
```

cmd/webhook/main.go
eclipselu: see comment above.

test/e2e/main_test.go
eclipselu: Changed back to `MessagingV1beta1APIVersion  = "messaging.cloud.google.com/v1beta1"` to make it more meaningful

cmd/webhook/main.go
eclipselu: ping @Harwayne

Harwayne: I think removing it is fine. It will be easy enough to add back when we want to, as there are many other examples to draw upon.

Harwayne (commented):
/approve

PR #1855, opened 2020-10-22
Update KSA annotation in Identity reconciler

grac3gao:
Fixes #

## Proposed Changes

-  In Identity reconciler, check and update KSA annotation in case it is accidentally deleted before.
-
-

**Release Note**

```release-note

```

**Docs**


Harwayne (commented):
/lgtm

PR #1859, opened 2020-10-23
Create E2E tests that verify logs go to Cloud Logging

### LINKED ISSUE #651: E2E test for Logging
## Expected Behavior
We need to have e2e tests to verify that when we configure SD as the logging backend, that actual logs are being sent there. This can be simply done with one of our current tests, in which we additionally configure the logging configMap to point to SD, and then assert for any log.

Harwayne:
Fixes #651

## Proposed Changes

- Create E2E tests that verify logs go to Cloud Logging.
    - `controller` and `webhook` in `cloud-run-events`
    - `CloudPubSubSource`
    - `Topic`'s `publisher

**Release Note**

```release-note
NONE
```


Harwayne:
Friendly ping.

pkg/reconciler/intevents/topic/topic.go
```diff
@@ -91,6 +91,12 @@ var _ topicreconciler.Interface = (*Reconciler)(nil)
 func (r *Reconciler) ReconcileKind(ctx context.Context, topic *v1.Topic) reconciler.Event {
 	ctx = logging.WithLogger(ctx, r.Logger.With(zap.Any("topic", topic)))
 
+	if v, present := topic.Annotations[v1.LoggingE2ETestAnnotation]; present {
+		// This is added purely for the TestCloudLogging E2E tests, which verify that the log line
+		// is written if this annotation is present.
+		logging.FromContext(ctx).Desugar().Error("Adding log line for the TestCloudLogging E2E tests", zap.String(v1.LoggingE2EFieldName, v))
```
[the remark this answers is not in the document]
Harwayne: We could probably add a way to turn it off. E.g. if a specific ConfigMap is not present or a key doesn't have a specific value, then don't log anything. For this in particular, I don't think it is needed because the cost of getting it wrong is so low. 

However, if we are going to be doing experimental things, then we should probably have a framework for doing them. And these log lines could follow that framework.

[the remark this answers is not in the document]
zhongduo: How about having a helper function for all the extra code? IIUC, all the added code can be done in one or two functions, eg map vs env. Then we can use e2e build tag in the helper functions to swap implementation.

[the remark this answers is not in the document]
tayarani: I am leaning the helper function way as well. It'll save us some refactoring in the future.

tayarani, line 97: Is the Error level intentional? Does this have to be an error log?

tayarani: I do recall u mentioning this once, but I just want to double check.

vendor/google.golang.org/api/logging/v2/logging-api.json
```diff
@@ -0,0 +1,4966 @@
+{
```
tayarani, line 1: This is a massive file. Where did it come from?

test/e2e/test_logging.go
```diff
@@ -0,0 +1,266 @@
...
+		TypeMeta:   topic.TypeMeta,
+		ObjectMeta: topic.ObjectMeta,
+	})
+	if err != nil {
+		t.Fatalf("Topic did not become ready: %v", err)
+	}
+
+	// Sleep so that the logging API has a chance to receive the logs lines and process them before
+	// we try to retrieve them.
+	time.Sleep(2 * time.Minute)
```
tayarani, line 91: I understand we have no other way but to wait here. but to my experience magic lines like this are usually a source of flakiness.

have you considered putting this part in a loop to sleep and retrying a few times? that will help reduce flakiness, while avoiding long waits in the test.

pkg/reconciler/intevents/topic/topic.go
Harwayne: I had made it an error to make it less likely that we don't log this because it is to low level. For example, I know it shouldn't be debug, as by default we set to info level.

I had set it to error just to be safe, do you think info is more appropriate?

tayarani: if INFO gets the job done, i'd say that is preferred.

test/e2e/test_logging.go
Harwayne: Turned it into a loop.

pkg/reconciler/intevents/topic/topic.go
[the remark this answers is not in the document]
Harwayne: > Then we can use e2e build tag in the helper functions to swap implementation.

I don't want this to be a build time flag, because I want this test to run in production. Failure of this test will prevent a release from going out. Its whole purpose is to remove a manual test that we require today.

I did create helper functions to centralize almost all the annotation/env var manipulation.

PR #1860, opened 2020-10-26
WIP: feat: register pubsub client in broker and trigger reconciler if not passed

### LINKED ISSUE #1765: Revisit pubsub client creation behaviour in several reconcilers like broker and trigger
I understand this code is unlikely to use in reality, but I am seeing in the broker and trigger that when a pubsub client is not passed from the controller, a new pubsub client will always be created then closed during reconciliation, rather than created then save. See following code from `pkg/reconciler/broker/broker.go`.

image

This code is either always executed (created a new client every time) or never called with pubsub client in error state and never recovered, because it only checks if `r.pubsubClient == nil`. So if the client is in error state, this creation will still not happen unless we somehow set it to `nil`.

Grant: This is partially legacy (the controller used to create a client on each reconcile) and partially reasoning that if an error occurs, it makes sense to create a new client the next time in case the error is corrected. But I think that logic could be improved to only recycle the client in case of error. Also, maybe there's no real need to recycle the client in case of error because it will recover on its own. I don't think this path was well tested

zhongduo:
Before this commit, when the pubsub client creation fails in the
controller while creating the reconciler, the reconciler will
create a new pubsub client in every reconciliation. Since the broker
and trigger only works with one project, we can reuse the client.
So after this commit, the newly created client will be registered.
This commit also refactors the reconcilers so that more common
code is moved to the getClientOrCreateNew function.

Fixes #1765 

## Proposed Changes

- Create a new `getClientOrCreateNew` helper function that registers the newly created pubsub client
- 
-

**Release Note**

```release-note

```

**Docs**


grantr:
Thank you for writing an excellent commit message @zhongduo!

pkg/reconciler/broker/broker_test.go
```diff
@@ -355,11 +356,93 @@ func TestAllCases(t *testing.T) {
...
+			Eventf(corev1.EventTypeNormal, "TopicCreated", `Created PubSub topic "cre-bkr_testnamespace_test-broker_abc123"`),
+			Eventf(corev1.EventTypeNormal, "SubscriptionCreated", `Created PubSub subscription "cre-bkr_testnamespace_test-broker_abc123"`),
+			brokerReconciledEvent,
+		},
+		WantPatches: []clientgotesting.PatchActionImpl{
+			patchFinalizers(testNS, brokerName, brokerFinalizerName),
+		},
+		OtherTestData: map[string]interface{}{
+			"pre":                   []PubsubAction{},
+			"maxPSClientCreateTime": 1,
```
grantr, line 392: To confirm I understand, is this intended to check that we never attempt to create the client twice?

pkg/reconciler/broker/broker_test.go
```diff
@@ -376,13 +459,23 @@ func TestAllCases(t *testing.T) {
 			drStore = NewDataresidencyTestStore(t, cm.(*corev1.ConfigMap))
 		}
 
+		// if maxPSClientCreateTime is in testData, no pubsub client is passed to reconciler, the reconciler
+		// will create one in demand
+		testPSClient := psclient
+		if maxTime, ok := testData["maxPSClientCreateTime"]; ok {
+			// Overwrite the createPubsubClientFn to one that failed when calling more than once, this does
+			// not really work since we never test anything that reconcile twice or both reconcile and delete
```
grantr, line 467: If this doesn't work, should we keep it? Is it expected to work later, or is it fundamentally guaranteed to never work?

pkg/reconciler/broker/broker_test.go
```diff
@@ -376,13 +459,23 @@ func TestAllCases(t *testing.T) {
...
+		}
+
 		ctx = addressable.WithDuck(ctx)
 		ctx = resource.WithDuck(ctx)
 		r := &Reconciler{
 			Base:               reconciler.NewBase(ctx, controllerAgentName, cmw),
 			brokerCellLister:   listers.GetBrokerCellLister(),
 			projectID:          testProject,
-			pubsubClient:       psclient,
+			pubsubClient:       testPSClient,
```
grantr, line 478: Should we also test the case where this is nil when the Reconciler is instantiated, as in https://github.com/google/knative-gcp/blob/42a25fe3dea6ea8f19d5d12607e3e1a5a59be98f/pkg/reconciler/broker/controller.go#L75?

Or maybe that's already tested elsewhere.

pkg/reconciler/broker/broker_test.go
zhongduo: Yes.

pkg/reconciler/broker/broker_test.go
zhongduo: Though the twice check does not work, it does still test that the pubsub client is created on demand, so it is still useful anyway. If later we can find the case to make sure it is never called twice it will be perfect.

pkg/reconciler/broker/broker_test.go
zhongduo: This might also answer your previous question about why have the `maxPSClientCreateTime`. If you look at how testPSClient is set, when `maxPSClientCreateTime` exists it will set to nil instead of `psclient` so that the pubsub client will be created during reconciliation. That's why we can make sure the `pubsub.NewClient` function is called only once in the first test case and inject error in the second test case.

grantr: Thanks! I missed that on the first read through.

pkg/reconciler/broker/broker_test.go
grantr: Makes sense. Can you add this explanation to the comment, so it's clear the code has a purpose?

grantr (approved):
Thanks @zhongduo!

/lgtm
/hold

Holding for optional suggestion to a comment. Cancel if you'd rather ignore that.

pkg/reconciler/broker/broker_test.go
zhongduo: Make sense. Changed as suggested by moving the explanation to the test case.

grantr (approved):
/lgtm
/hold cancel

PR #1858, opened 2020-10-23
Remove v1alpha1 support for knative-gcp sources and Pub/Sub core resources

eclipselu:
Part of https://github.com/google/knative-gcp/issues/1544 , follow up to https://github.com/google/knative-gcp/pull/1853 , need rebase once https://github.com/google/knative-gcp/pull/1853 is merged. 

## Proposed Changes

- 🗑️ Remove v1alpha1 support for knative-gcp sources
- 🗑️ Remove v1alpha1 support for Pub/Sub core resources
-

**Release Note**

```release-note
Dropped v1alpha1 support for the following resources:

- CloudPubSubSource
- CloudStorageSource
- CloudSchedulerSource
- CloudAuditLogsSource
- CloudBuildSource
- Topic
- PullSubscription
```

**Docs**

PR #1861, opened 2020-10-26
Remove v1alpha1 versions from CRDs

eclipselu:
Part of https://github.com/google/knative-gcp/issues/1544, follow-up to https://github.com/google/knative-gcp/pull/1858

_Reviewers: `channel.yaml` needs some effort to review since v1beta1's schema was merged from v1alpha1 with additions._

## Proposed Changes

- 🗑️ Remove v1alpha1 version from CRDs
-
-

**Release Note**

```release-note
Removed v1alpha1 CRD api versions for following resources: 

- CloudPubSubSource
- CloudStorageSource
- CloudSchedulerSource
- CloudAuditLogsSource
- CloudBuildSource
- Topic
- PullSubscription
- Channel
```

**Docs**

### DIFF
--- a/config/core/resources/channel.yaml
+++ b/config/core/resources/channel.yaml
@@ -1,4 +1,4 @@
-# Copyright 2019 Google LLC
+# Copyright 2020 Google LLC
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -11,8 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
-# TODO: merge v1alpha1 schema into v1beta1 then remove v1alpha1 schema from CRD
 
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
@@ -30,11 +28,11 @@ spec:
     plural: channels
     singular: channel
     categories:
-    - all
-    - knative
-    - pubsub
-    - messaging
-    - channel
+      - all
+      - knative
+      - pubsub
+      - messaging
+      - channel
     shortNames:
       - pschan
   scope: Namespaced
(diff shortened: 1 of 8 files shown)


config/core/resources/channel.yaml
```diff
@@ -51,206 +49,147 @@ spec:
...
+                        type: string
+                      type:
+                        type: string
+                    required:
+                      - type
+                      - status
+                address:
+                  type: object
                   properties:
-                    # No deadLetterSink in subscribers properties in v1beta1
```
capri-xiyue, line 230: Looks like in the v1beta1, you added deadLetterSink.

config/core/resources/cloudschedulersource.yaml
```diff
@@ -199,12 +199,11 @@ spec:
               jobName:
                 type: string
   - << : *version
-    # TODO: remove v1alpha1 schema from CRD
-    name: v1alpha1
-    served: false
+    name: v1beta1
+    served: true
     storage: false
     # v1alpha1 and v1beta have status.properties.serviceAccountName in the schema
```
tayarani, line 205: was this missed?

config/core/resources/channel.yaml
eclipselu: The comment is a bit misleading, actually deadLetterSink exists in v1beta1: https://github.com/google/knative-gcp/blob/master/vendor/knative.dev/eventing/pkg/apis/duck/v1beta1/delivery_types.go#L28-L34 

This is a bit of a tricky CRD to modify, I'm wondering if we could have a more automated way to do this kind of deprecation, the best I can do is to use [yq](https://github.com/mikefarah/yq) and [jq](https://stedolan.github.io/jq/), the idea is to expand the [merge key](https://yaml.org/type/merge.html) in v1beta1 schema, output it in json, sort the keys and compare the output:

We use the [path expressions](https://mikefarah.gitbook.io/yq/usage/path-expressions) in yq to extract just v1beta1 part in the yaml. 
```bash
yq r -X config/core/resources/channel.yaml 'spec.versions.(name==v1beta1)'
```

Next we need to output it to json, and sort the json by key to eliminate any difference that are caused by the key order. 

```bash
yq r -X config/core/resources/channel.yaml 'spec.versions.(name==v1beta1)' -jP | jq -S
```

Lastly, we can download the old version of `channel.yaml` , do the same thing and compare them:

```bash
diff <(yq r -X old-channel.yaml 'spec.versions.(name==v1beta1)' -jP | jq -S) \
     <(yq r -X config/core/resources/channel.yaml 'spec.versions.(name==v1beta1)' -jP | jq -S)

# there's no output from my side, meaning no diff
# so the v1beta1 in this PR is equivalent to what it was before. 
```

config/core/resources/cloudschedulersource.yaml
eclipselu: Good catch, fixed.

config/core/resources/channel.yaml
capri-xiyue: Maybe we should consider this https://github.com/google/knative-gcp/issues/920

PR #1867, opened 2020-10-27
Upgrade to latest dependencies and perform manual fixes

tommyreddad:
The automated dependency update PR https://github.com/google/knative-gcp/pull/1851 has been failing for days due to an apparent issue with the `config-tracing` ConfigMap not being detected. This PR includes dependency updates and the necessary manual fixes:
- `./hack/update-deps.sh --upgrade && ./hack/update-codegen.sh`
- The `pkg/metrics` function `JsonToMetricsOptions` has been renamed to `JSONToOptions`
- The `pkg/metrics` function `MetricsOptionsToJson` has been renamed to `OptionsToJSON`
- The `eventing/pkg/reconciler/names` function `ServiceHostName` has been moved to `pkg/network` and renamed to `GetServiceHostname`
- Eventing now depends on `pkg/system` which cares about the `SYSTEM_NAMESPACE` environment variable pointing to the namespace in which Knative Eventing is installed. Set `SYSTEM_NAMESPACE=knative-eventing` in `./test/lib.sh` before running our e2e tests.


liu-cong:
/lgtm
/approve

Thanks @tommyreddad for fixing this!

liu-cong:
/hold

I wonder whether we should inject the system namespace env instead of hard coding:

```
  env:
  - name: %s
    valueFrom:
      fieldRef:
        fieldPath: metadata.namespace
```

test/lib.sh
```diff
@@ -16,6 +16,10 @@
 
 # Include after test-infra/scripts/library.sh
 
+# Set and export SYSTEM_NAMESPACE where Knative Eventing is installed for pkg/system.
+readonly SYSTEM_NAMESPACE="knative-eventing"
```
liu-cong, line 20: I guess the eventing controllers need this right? Can we update the yaml to inject the namespace? This is the recommended approach in pkg/system

tommyreddad: It's my understanding that the eventing controllers already have injected SYSTEM_NAMESPACE in the way you suggest. What has changed is that now also the eventing test library requires access to SYSTEM_NAMESPACE. Does this still mean we should be injecting via YAML?

In eventing e2e tests, SYSTEM_NAMESPACE is set in the shell script: https://github.com/knative/eventing/blob/master/test/e2e-common.sh#L68
And downstream as well: https://github.com/knative-sandbox/eventing-kafka/blob/master/test/e2e-tests.sh#L69

liu-cong: Ah I see, looks like the script actually depends on this variable. Ideally we only have one source of truth (in the yaml) but that's out of scope.

Issue #1865, opened 2020-10-27
Backport test/e2e-conformance-tests.sh to release branches


grantr:
**Describe the bug**
Adding the conformance test as a required job means it is required on all the release branches as well. Release branches don't have the job entry point script that Prow expects so the job always fails on PRs to release branches. See https://github.com/google/knative-gcp/pull/1864.

Prow doesn't seem to allow per-branch configs of which jobs are required. We'll need to backport new job entry points to all release branches when they are introduced. These backported entry point scripts can just exit success immediately.

 **Expected behavior**
PRs to release branches are not blocked by jobs that don't exist in those branches.

 **Additional context**
Technically this also applies to the upgrade tests just added (@zhongduo) but since that one isn't required yet it doesn't block PRs.

grantr:
@tommyreddad is this something you can take care of?

tommyreddad:
Which release branches should I do this for? I'm guessing 0.18, 0.17, 0.16, and 0.15b?

grantr:
Thanks! That seems sufficient to me. You should be able to commit to one of the branches, then cherry-pick that commit to all the others.

tommyreddad:
The fix has been backported into all the required branches, PRs linked above. Closing the issue.

PR #1872, opened 2020-10-27
Automatically succeed on conformance tests (release-0.16)

tommyreddad:
Part of #1865

Prow tries to run ./test/e2e-conformance-tests.sh as a required job on release branches which don't have that script. This change adds the script and automatically succeeds.


Harwayne (commented):
/approve

PR #1873, opened 2020-10-27
Automatically succeed on conformance tests (release-0.15b)

tommyreddad:
Part of #1865

Prow tries to run ./test/e2e-conformance-tests.sh as a required job on release branches which don't have that script. This change adds the script and automatically succeeds.

There is also a bug in release-0.15b where test-infra does not recognize it as a release branch due to the letter suffix, and therefore tries to install nightly releases of Knative during testing instead of 0.15 versions. This change addresses this by  overriding the relevant parts of test-infra so that the branch actually tests against 0.15.


Harwayne (commented):
/approve

Harwayne (commented):
/lgtm

PR #1887, opened 2020-11-03
Attach broker delivery ConfigMap to controller context

tommyreddad:
Without this fix, the error "Failed to get BrokerDeliverySpecDefaults" pollutes the controller logs, during failure of Broker `SetDefaults` call, since the controller has no access to the broker delivery config store. The webhook can successfully default though, so the settings are applied properly, but the logs are polluted.

## Proposed Changes
- Attach broker delivery ConfigMap to controller context using ConfigStore broker reconciler option
- Rename `pkg/apis/configs/broker` to `pkg/apis/configs/brokerdelivery`


pkg/reconciler/testing/configmap.go
```diff
@@ -94,3 +98,38 @@ func NewDataresidencyConfigMapFromRegions(regions []string) *corev1.ConfigMap {
...
 }
+
+// NewBrokerDeliveryConfigMapFromDeliverySpec creates a new cluster defaulted
+// broker delivery configuration map from a given delivery spec.
+func NewBrokerDeliveryConfigMapFromDeliverySpec(spec *eventingduckv1beta1.DeliverySpec) *corev1.ConfigMap {
+	configData := `
+  clusterDefaults:`
+	if spec != nil {
+		if spec.BackoffPolicy != nil {
+			configData += `
```
Ectelion, line 109: My first thought was that it might be worth using a string builder here, though it seems there shouldn't be use cases when this piece is executed too often?

Ectelion (approved):
/lgtm
/hold in case anyone else is going to take a look.

pkg/reconciler/testing/configmap.go
tommyreddad: I just changed it to use a string builder, I wasn't aware of the pattern when I submitted the PR :) I also adjusted the other configmap testing function `NewDataresidencyConfigMapFromRegions` to use a string builder, since that's what I was basing mine off originally.

Ectelion: I see, cool!

pkg/reconciler/broker/controller.go
```diff
@@ -91,7 +92,12 @@ func newController(ctx context.Context, cmw configmap.Watcher, drs *dataresidenc
 		dataresidencyStore: drs,
 	}
 
-	impl := brokerreconciler.NewImpl(ctx, r, brokerv1beta1.BrokerClass)
+	impl := brokerreconciler.NewImpl(ctx, r, brokerv1beta1.BrokerClass,
+		func(impl *controller.Impl) controller.Options {
+			return controller.Options{
+				ConfigStore: brds,
```
grantr, line 98: Interesting that `brds` is set as `ConfigStore` here, while `drs` is passed in as `dataresidencyStore` field above. Is this due to a limitation in the reconciler code?

grantr (approved):

pkg/reconciler/broker/controller.go
tommyreddad: I'm not sure. The data residency store is loaded manually during broker reconciling, so it is attached to the reconciler itself. I didn't need to load the delivery store manually so I used this option instead.

Issue #1862, opened 2020-10-27
[Respectful Code]Replace Disrespectful Terms


grac3gao:
**Problem**
Replace disrespectful terms.

**[Persona:](https://github.com/knative/eventing/blob/master/docs/personas.md)** 
Which persona is this feature for?

**Exit Criteria**
A measurable (binary) test that would indicate that the problem has been resolved.

**Time Estimate (optional):** 
How many developer-days do you think this may take to resolve?

**Additional context (optional)**
Add any other context about the feature request here.

zhongduo:
I think native to "" can be a problem to enforce, as our project is.... Knative

grac3gao:
@zhongduo for unavoidable terms, the guidance is to minimize the usage outside of it.

grac3gao:
Summary:

hit -> reach (17 times)
native -> "" (4 times)
dummy -> sample (55 times)
victim -> "" (1 times)
special -> "" (4 times)
health check -> probe check (72 times)

PR #1893, opened 2020-11-06
replace disrespectful terms - 2

grac3gao:
Fixes #
related #1862 

## Proposed Changes

- "victim" -> (guideline: don't use) -> ""
- "special" -> (guideline: use with caution) -> "" (remove this term in the places which are not that necessary to have it)
-

**Release Note**

```release-note

```

**Docs**


docs/spec/delivery.md
```diff
@@ -16,7 +16,7 @@ subscription configuration member `DeadLetterPolicy`.
 The Knative dead letter policy is specified through the following parameters in
 the Knative Eventing delivery spec:
 
-- `DeadLetterSink`: We only allow special URLs as the dead letter sink, of the
+- `DeadLetterSink`: We only allow specific URLs as the dead letter sink, of the
```
eclipselu, line 19:
```suggestion
- `DeadLetterSink`: We only allow URLs of the form `pubsub://[dead_letter_sink_topic]` as the dead letter sink. 
```

What do you think if we change this sentence to this?

grac3gao: Sounds good, done.

PR #1896, opened 2020-11-09
feat: Add topic and subscription check to trigger test

zhongduo:
Before this change, the trigger test only used events to verify
that the topic and subscription are deleted. However, the event
emission logic and actual pubsub operations are separate. After
this change, the stardard pubsub client will be used to make sure
the topic and subscription are indeed deleted from fake server.

Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


yolocs (approved):
/lgtm

PR #1899, opened 2020-11-09
Avoid error in test script when tear down pubsub resources

grac3gao:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


danyinggu (approved):
/lgtm

danyinggu:
Thanks for the change. I think we didn't add this initially because test-infra won't fail in knative-gcp if the deletion command fails (e.g. if there is no pubsub resource to delete)

PR #1902, opened 2020-11-10
pin to release 0.19

nlopezgi:
Pin to release-0.19 in preparation for creating release-0.19 branch

### DIFF
--- a/hack/update-deps.sh
+++ b/hack/update-deps.sh
@@ -21,7 +21,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-go_update_deps "$@"
+go_update_deps "$@" --release v0.19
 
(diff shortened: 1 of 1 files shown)


hack/update-deps.sh
```diff
@@ -21,7 +21,7 @@ set -o errexit
 set -o nounset
 set -o pipefail
 
-go_update_deps "$@"
+go_update_deps "$@" --release v0.19
```
nachocano, line 24: do you need to run the script? I don't see any files changed besides this?
Are we using pkg, serving, and eventing 0.19?

nachocano: ohh I see eventing hasn't been cut yet..

PR #1903, opened 2020-11-10
update prior to release 0.19 branch

nlopezgi:
Run instructions in https://github.com/knative/pkg/blob/8269222675d1050513fb7c72e2324171eb6abe18/RELEASE-LEADS.md prior to release: 
```
RELEASE=0.19
REPO=git@github.com:knative/example.git

tmpdir=$(dirname $(mktemp -u))
cd ${tmpdir}
git clone ${REPO}
cd "$(basename "${REPO}" .git)"

if buoy check go.mod --domain knative.dev --release ${RELEASE} --verbose; then
  git checkout -b release-${RELEASE}
  ./hack/update-deps.sh --upgrade --release ${RELEASE}
  # consistency check, shouldnt do anything
  echo "There should be no changes to git:"
  git status
fi
```
send to master as there are changes.


zhongduo:
/lgtm
/hold
I think we can wait for eventing 0.19 cut as the doc seems spend some efforts in the orders, but unhold if we want to do it now.

nlopezgi:
> /lgtm
> /hold
> I think we can wait for eventing 0.19 cut as the doc seems spend some efforts in the orders, but unhold if we want to do it now.

Buoy is checking the upstream repos are ready and its determining they are with whatever criteria it currently uses. I don't think its necessary to wait here
/unhold

PR #1910, opened 2020-11-12
running update dep on master branch after 0.19 release

nlopezgi:
Updating master after creating 0.19 branch
Produced by running: 
```
./hack/update-deps.sh --upgrade --release v9000.1 && ./hack/update-codegen.sh
```


tommyreddad:
Just noting: since this includes the knative.dev/hack upgrade, after this goes in we shouldn't need to specify the hacky fix `--release v9000.1` tag for update-deps.

PR #1911, opened 2020-11-12
Use git tag to infer latest version in upgrade test

### LINKED ISSUE #1857: Failure for upgrade test in branch
**Describe the bug**
In #1854 the upgrade test is not using the right latest version function in `test/lib.sh`, because our master does not have the right tag in `git describe`. Once we do this correctly in master from 0.19.0, we should change it back. For now, the upgrade test will always used the latest patch version of 0.18, and CI test will fail if we are committing to older release like 0.17. 

 **Expected behavior**
We should change the function back to use `git describe` to get the target feature branch version.

 **To Reproduce**

 **Additional context**

zhongduo:
Fixes #1857 

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -63,18 +63,14 @@ function cloud_run_events_setup() {
   wait_until_pods_running cloud-run-events || return 1
 }
 
+# Get the latest release version
 latest_version() {
-  # TODO: currently we don't have the right version tag in master,
-  # we should switch back to use the above logic instead in v0.19
-  # https://github.com/google/knative-gcp/issues/1857
-
-  #local semver=$(git describe --match "v[0-9]*" --abbrev=0)
-  #local major_minor=$(echo "$semver" | cut -d. -f1-2)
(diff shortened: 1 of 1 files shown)


test/lib.sh
```diff
@@ -63,18 +63,14 @@ function cloud_run_events_setup() {
...
+  local major_minor=$(echo "$semver" | cut -d. -f1-2)
 
-  #local semver=$(git describe --match "v[0-9]*" --abbrev=0)
-  #local major_minor=$(echo "$semver" | cut -d. -f1-2)
-
-  # Get the latest patch release for the major minor
-  #git tag -l "${major_minor}*" | sort -r --version-sort | head -n1
-
-  git tag | sort -r --version-sort | head -n1
+  Get the latest patch release for the major minor
```
nlopezgi, line 72: should this be a comment line?

zhongduo: Good catch! done.

PR #1917, opened 2020-11-13
Rename namespace

AlexandraRoatis:
Moving to a brand-agnostic namespace.

## Proposed Changes

- Replace `cloud-run-events` namespace and prefix with `events-system`.

**Release Note**

```release-note
Replace `cloud-run-events` namespace and prefix with `events-system`.
```


AlexandraRoatis:
Note: the failing tests are expected to fail due to this change.

pkg/apis/configs/brokerdelivery/br_delivery_defaults_test.go
```diff
@@ -116,7 +117,7 @@ func TestNewDefaultsConfigFromConfigMapWithError(t *testing.T) {
 		"empty data": {
 			config: &corev1.ConfigMap{
 				ObjectMeta: metav1.ObjectMeta{
-					Namespace: "cloud-run-events",
+					Namespace: system.Namespace(),
```
MohamedElhawaty, line 120: This will evaluate to "knative-testing" instead of "events-system" because this file includes:

`import _ "knative.dev/pkg/system/testing"`

which initialize the value to  "knative-testing".

The test seem to be passing anyway because it doesn't care about the namespace value.

AlexandraRoatis: As discussed, I removed the commit that made these changes. The use of `system.Namespace()` will be delegated to solving #1958.

AlexandraRoatis:
I'm not sure what to do about these regex:
- https://github.com/google/knative-gcp/blob/1970c131414a379b0bf74359d2991c2e9a98c73a/config/monitoring/metrics/prometheus/prometheus-scrape-kn-gcp.yaml#L13
- https://github.com/google/knative-gcp/blob/1970c131414a379b0bf74359d2991c2e9a98c73a/config/monitoring/metrics/prometheus/prometheus-scrape-kn-gcp.yaml#L29

AlexandraRoatis:
I think the lint check is incorrect. It identifies a part of the word instead of the entire word.

config/core/webhooks/config-validation.yaml
```diff
@@ -24,8 +24,8 @@ webhooks:
     clientConfig:
       service:
         name: webhook
-        namespace: cloud-run-events
-    failurePolicy: Fail
+        namespace: events-system
+    failurePolicy: Ignore
```
Harwayne, line 28: Is this intended to be in this PR? I actually like the change, but am not sure we want to include it in this large renaming PR.

AlexandraRoatis: The reason we have it here is that we couldn't get past the webhook issue when testing without it. I can also move it to a separate PR, if preferable.

Harwayne (commented):
/lgtm
/approve
/unhold

AlexandraRoatis:
Merge seemed to still be blocked so I did a rebase to see if it is finally triggered after. /assign @zhongduo

PR #1922, opened 2020-11-17
Refactor setup functions in e2e script

grac3gao:
Fixes #

## Proposed Changes

- Refactor setup functions in e2e script in order to make them shared later.
-
-

**Release Note**

```release-note

```

**Docs**


zhongduo:
IIUC, you are trying to use `if..else` to merge the same functions in secret mode and WI mode. But I don't see lots of reusable code, in this sense putting those into different function might make sense. You can still have one single entry point with argument to choose which function to invoke. WDYT?

grac3gao:
@zhongduo new functions have other usage, will sync you offline

danyinggu (commented):
/lgtm

danyinggu (commented):
/lgtm

PR #1927, opened 2020-11-18
feat: Upstream improved ce-test-actor

yolocs:
Fixes https://github.com/google/knative-gcp/issues/1609

Originated from: https://github.com/yolocs/ce-test-actor
Added an aggregator to collect results.
This might also appear useful for upgrade/perf tests in the future. @zhongduo 

**Release Note**

```release-note
NONE
```

**Docs**


test/actor/README.md
```diff
@@ -0,0 +1,83 @@
...
+```
+./ns-gen -actors=2 -elapse=2m -ns=new-actor-1 -output=$PWD/actor-new -seeders=2 -triggers=5
+```
+
+The flags:
+
+- `-actors=2` means we want 2 replicas for each actor deployment.
+- `-elapse=2m` means we want all the `seeders` to send events for 2 minutes.
+- `-ns=new-actor-1` means we want to run the test in namespace `new-actor-1`
+- `-output=$PWD/actor-new` is the output dir for the yamls
```
cathyzhyi, line 43:
```suggestion
- `-output=$PWD/actor-new` is the absolute path of the output dir for the yamls 
```

yolocs:
Lint checking is not making sense. Please ignore.

test/actor/aggregator/main.go
```diff
@@ -0,0 +1,128 @@
...
+	if role == "actor" {
+		c := req.Header.Get("count")
+		delta, err := strconv.ParseInt(c, 10, 64)
+		if err != nil {
+			w.WriteHeader(http.StatusBadRequest)
+			return
+		}
+		atomic.AddInt64(&s.receivedCount, delta)
+		atomic.AddInt64(&s.actors, 1)
+	} else {
```
zhongduo, line 109: Do we want to check `role=="seeder"` etc?

test/actor/ns-gen/main.go
```diff
@@ -0,0 +1,285 @@
...
+    metadata:
+      labels:
+        app: seeder
+    spec:
+      containers:
+      - name: seeder
+        image: ko://github.com/google/knative-gcp/test/actor/seeder
+        env:
+        - name: TARGET
+          value: http://default-brokercell-ingress.cloud-run-events.svc.cluster.local/{{.namespace}}/testbroker
```
zhongduo, line 133: Do we need to hardcode the broker address?

test/actor/ns-gen/main.go
```diff
@@ -0,0 +1,285 @@
...
+	actorMaxConn   = flag.Int("max-conn", 0, "The max conns an actor pod accepts concurrently")
+	seederReplicas = flag.Int("seeders", 1, "The number of seeder replicas")
+	seederConn     = flag.Int("seeders-conn", 1, "The seeder concurrency")
+	elapse         = flag.String("elapse", "15m", "The seeding elapse")
+)
+
+func main() {
+	flag.Parse()
+
+	namespace := strings.ReplaceAll(nsTemplate, "{{.namespace}}", *ns)
```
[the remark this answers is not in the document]
yolocs: Since the logic is pretty simple, I try not to adopt any fancy templating at the moment. On the other side, I mostly copied it over :)

test/actor/ns-gen/main.go
yolocs: I can see the potential concern but I would rather change it once we have the need to do that.

test/actor/aggregator/main.go
yolocs: I changed it to return bad request if role is neither "seeder" nor "actor" above.

test/actor/actor/main.go
```diff
@@ -0,0 +1,199 @@
...
+	// MaxConn is the max number of connections this code will handle at a time.
+	MaxConn int `envconfig:"MAX_CONN"`
+
+	// AggregatorAddr is the aggregator address.
+	// If empty, then metrics won't be sent.
+	AggregatorAddr string `envconfig:"AGGREGATOR_ADDR"`
+
+	// ReportGap is a duration within which if there is no event received,
+	// existing metrics will be reported to the aggregator.
+	ReportGap time.Duration `envconfig:"REPORT_GAP" default:"1m"`
```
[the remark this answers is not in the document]
yolocs: You mean in `ns-gen`? `ns-gen` is only meant for easing the manual testing setup. If we find the default value not good enough, we can add a flag there later.

test/actor/actor/main.go
```diff
@@ -0,0 +1,199 @@
...
+		var count int64
+		reported := false
+		for {
+			select {
+			case <-time.After(env.ReportGap):
+				if reported {
+					break
+				}
+
+				if env.AggregatorAddr == "" {
```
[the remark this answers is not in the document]
yolocs: `ns-gen` is not required to use the `actor` so I don't see that every option in `actor` or `seeder` has to appear in `ns-gen` as a flag. If there is a need to disable `aggregator` with `ns-gen`, we can always add the flag there.

PR #1934, opened 2020-11-19
Enable conformance tests

### LINKED ISSUE #1932: Conformance tests failed due to missing MT channel based broker CRD manifests
**Describe the bug**
Conformance tests failed due to missing MT channel based broker CRD manifests

 **Expected behavior**
Relates to knative/eventing#4545 and #1930 

 **To Reproduce**

 **Additional context**

cc: @tommyreddad

tommyreddad:
Fixes #1932 

Nightly eventing YAML now includes the MT broker again since merging https://github.com/knative/eventing/pull/4546, so conformance tests should work after merging https://github.com/knative/hack/pull/26

## Proposed Changes

- Enable conformance tests.

### DIFF
--- a/test/e2e-conformance-tests.sh
+++ b/test/e2e-conformance-tests.sh
@@ -16,9 +16,6 @@
 # Script entry point.
 source $(dirname "$0")/e2e-secret-tests.sh
 
-# TODO: https://github.com/google/knative-gcp/issues/1932. Until then,
-# temporarily disable conformance tests.
-SKIP_TESTS="true"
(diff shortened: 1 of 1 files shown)


danyinggu:
Thanks a lot for the quick fix. However, there was a change in hack file, where only eventing core and crds are installed. https://github.com/knative/hack/blob/master/library.sh#L419

We need to wait until they revert it back to nightly yaml.
I will keep watching it and update deps when the revert is in.

danyinggu (commented):
/lgtm

Thanks @tommyreddad for the quick fix!!!

PR #1941, opened 2020-11-20
Return true if warm up broker can't find path

grac3gao:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


danyinggu (commented):
/lgtm

PR #1951, opened 2020-11-27
Fix issue to run broker gcp perf test locally

zhongduo:
After this PR, the gcp broker perf test can be run locally with https://github.com/knative/pkg/pull/1930

## Proposed Changes

- Add zhongduo to owner list
- Fix yaml for running perf test locally
-


danyinggu (approved):
/lgtm

PR #1954, opened 2020-11-30
Refactor probe helper to allow for multi-tenancy and easier probe development

tommyreddad:
The probe helper main logic was becoming bloated. This change massively refactors the probe helper to make it easier to develop new probes. This also prepares the probe helper for multi-tenancy in many ways.

The development workflow for defining new probes is described in `probe/handler.go`:
```go
/*

The following steps should be taken when implementing a new type of probe.

1. Define a static probe handler which implements the Handler interface.
2. Declare a singleton Handler object of this kind in the probe helper initializerHandlers method.
3. Add the mapping between forward probe type and the object from step 2 to the forwardProbeHandlers map in initializerHandlers.
4. Add the mapping between receiver probe type and the object from step 2 to the receiveProbeHandlers map in initializerHandlers.

*/
```

## Proposed Changes
- Massive refactoring to allow for easier development and maintenance of probes.
- Broker e2e delivery probe now requires specification of `namespace` CloudEvent extension, in which the probed Broker should live. Broker can be specified through `broker` CloudEvent extension, and if omitted, falls back to `default`.
- CloudPubSubSource probe now requires specification of `topic` CloudEvent extension, to which the probe helper publishes a message.
- CloudStorageSource probe now requires specification of `bucket` CloudEvent extension, in which the probe helper modifies an object.
- CloudSchedulerSource probe now involves a cleanup phase triggered by the liveness probe, where stale scheduler timestamps are discarded.
- All forward probes should have a `targetpath` CloudEvent extension, indicating the request path the receiver should expect receiver probes to possess.


test/test_images/probe_helper/probe/handler.go
```diff
@@ -0,0 +1,59 @@
...
+*/
+
+// Handler is the interface which probe objects should implement.
+type Handler interface {
+	// ChannelID represents the key of the receiver channel which the probe helper
+	// should wait on after forwarding a probe request. You should ensure that
+	// each distinct probe request maps to a unique ChannelID, and that when a
+	// forward probe is matched with a receiver probe, that both handlers produce
+	// the same ChannelID.
+	ChannelID() string
```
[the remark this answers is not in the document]
liu-cong: Thinking about this a bit more, I think we can hide the details of the channel. To be more specific, I am considering something like this:
```
type Handler interface {
     Forward(ctx context.Context, event Event)
     Receive(ctx context.Context, event Event)
}
```

The channel becomes an implementation detail  of each handler (some handlers don't even need the channel maybe). There will a single handler for each event type.

test/test_images/probe_helper/probe/helper.go
```diff
@@ -0,0 +1,374 @@
...
+	"knative.dev/pkg/logging"
+
+	schemasv1 "github.com/google/knative-gcp/pkg/schemas/v1"
+	"github.com/google/knative-gcp/test/test_images/probe_helper/utils"
+)
+
+const (
+	probeEventTimeoutExtension     = "timeout"
+	probeEventRequestHostExtension = "requesthost"
+	probeEventRequestHostHeader    = "Ce-RequestHost"
```
liu-cong, line 42: I think we need to better document this. We are relying on the host name of the received event to identify which trigger delivered the event. I think something like `targetTrigger` is easier to understand. Then the receiver can translate the host into a trigger, assuming there is a 1:1 mapping between a trigger type and host (1 trigger per type in a namespace).

And this only applies to the AuditLogSource. Maybe we can move this logic to the AudigLogSourceHandler?

test/test_images/probe_helper/probe/handler.go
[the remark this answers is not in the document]
tommyreddad: I agree with creating a static handler for each event type, that's definitely a better idea than what I currently have.

I'm not sure it is a good idea to have the channel be an implementation detail of each handler. This would mean duplicating the channel creation, waiting, and deletion for each handler. This could increase the likelihood of subtle channel manipulation mistakes between handlers, but also would dirty up each of the handler implementations. I think one of the main points of this refactoring is precisely about gathering the common logic between probes (like the channel manipulation) in a generic handler.

test/test_images/probe_helper/probe/handler.go
```diff
@@ -0,0 +1,59 @@
...
+	// Handle is the handler function which is called for probe forwarders before
+	// waiting on the receiver channel. Typically, Handle should forward an event
+	// to some other service.
+	Handle(context.Context) error
+}
+
+// ShouldWaitOnReceiver determines if the given probe handler should wait on a
+// receiver channel. By convention, whenever the probe's channel ID is empty, we
+// do not wait on a receiver channel.
+func ShouldWaitOnReceiver(p Handler) bool {
```
liu-cong, line 57: In the current implementation, it looks like all handlers return true for this?

test/test_images/probe_helper/probe/cloudscheduler.go
```diff
@@ -0,0 +1,89 @@
...
+
+const (
+	// CloudSchedulerSourceProbeEventType is the CloudEvent type of forward
+	// CloudSchedulerSource probes.
+	CloudSchedulerSourceProbeEventType = "cloudschedulersource-probe"
+)
+
+// CloudSchedulerSourceForwardProbe is the probe handler for forward probe requests
+// in the CloudSchedulerSource probe.
+type CloudSchedulerSourceForwardProbe struct {
```
liu-cong, line 34: In the new design, cloud scheduler source probe will wait until the receiver receives a scheduler event, right? The drawback of this design is that the probe will need to wait for a long time to return (between 0-60s).

test/test_images/probe_helper/probe/handler.go
tommyreddad: Yes, for a moment I was experimenting with different scheduler probe implementations which would use this, for now it is useless so I can remove it.

test/test_images/probe_helper/probe/cloudscheduler.go
tommyreddad: That's right. On the other hand, with the previous design, we still had to wait some amount of time like 90s after creating the scheduler object in order to guarantee a tick. With this design, that wait is not necessary. So the total waiting time is identical in either case.

The biggest problem with a receiver-based scheduler probe is that the helper would receive ticks from every scheduler in the entire cluster, and it would need to keep the latest timestamps from all of those in memory in case they would be referenced in a later probe, without knowing whether or not they would ever be probed, or when to free that memory. The logic for handling that seems much more complicated than with this forward approach, and I stated, I believe the delay necessary in both cases to be similar.

liu-cong: How about the delay for the static scheduler source?

tommyreddad: The current delay with the lone static scheduler is basically nil, the probe is instantaneous. With the new approach, I imagine it will average out to 30s. So yes, the delay for the static scheduler will increase substantially. I think this is an acceptable price to pay for cleaner unified code between the static and dynamic scheduler probes.

I did forget to address one subtlety, which is that with the increased delay, we need to worry about two scheduler probes happening at the same time for the static scheduler. I will try to fix this.

test/test_images/probe_helper/probe/broker_e2e_delivery.go
```diff
@@ -0,0 +1,108 @@
...
+	// delivery probes.
+	BrokerE2EDeliveryProbeEventType = "broker-e2e-delivery-probe"
+
+	namespaceExtension = "namespace"
+)
+
+// BrokerE2EDeliveryForwardProbe is the probe handler for forward probe requests
+// in the broker e2e delivery probe.
+type BrokerE2EDeliveryForwardProbe struct {
+	Handler
```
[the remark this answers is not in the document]
tommyreddad: I believe this comment is no longer relevant with the current changes.

test/test_images/probe_helper/probe/handler.go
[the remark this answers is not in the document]
tommyreddad: I modified the Handler to be as you suggested, let me know what you think.

test/test_images/probe_helper/probe/broker_e2e_delivery.go
```diff
@@ -0,0 +1,108 @@
...
+// probe handler from a given CloudEvent.
+func BrokerE2EDeliveryForwardProbeConstructor(ph *Helper, event cloudevents.Event, requestHost string) (Handler, error) {
+	namespace, ok := event.Extensions()[namespaceExtension]
+	if !ok {
+		return nil, fmt.Errorf("Broker e2e delivery probe event has no '%s' extension", namespaceExtension)
+	}
+	probe := &BrokerE2EDeliveryForwardProbe{
+		event:     event,
+		channelID: fmt.Sprintf("%s/%s", requestHost, event.ID()),
+		target:    fmt.Sprintf("%s/%s/default", ph.BrokerCellIngressBaseURL, namespace),
```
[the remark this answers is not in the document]
tommyreddad: Done.

test/test_images/probe_helper/probe/helper.go
```diff
@@ -0,0 +1,327 @@
...
+
+	"knative.dev/pkg/logging"
+
+	schemasv1 "github.com/google/knative-gcp/pkg/schemas/v1"
+	"github.com/google/knative-gcp/test/test_images/probe_helper/utils"
+)
+
+const (
+	probeEventTimeoutExtension       = "timeout"
+	probeEventTargetServiceExtension = "targetservice"
```
liu-cong, line 41: Just came to my mind - if we are going to use URI to target the event consumer, then we don't have the "same namespace" constraint any more. Hence instead of creating a new ExternalName service, we can just use a path in the probe helper receiver, like so: <probe_helper_receiver_uri>/<namespace>/<trigger>. So by parsing this path we know exactly the expected trigger name and its namespace.

test/test_images/probe_helper/probe/broker_e2e_delivery.go
```diff
@@ -0,0 +1,88 @@
...
+		return fmt.Errorf("Broker e2e delivery probe event has no '%s' extension", namespaceExtension)
+	}
+	broker, ok := event.Extensions()[brokerExtension]
+	if !ok {
+		broker = "default"
+	}
+
+	// Get the receiver channel
+	channelID := fmt.Sprintf("%s/%s", event.Extensions()[probeEventTargetServiceExtension], event.ID())
+	receiverChannel, cleanupFunc, err := CreateReceiverChannel(ctx, ph, channelID)
```
liu-cong, line 54: Here, I think the broker e2e handler should just take a dependency on the ReceivedEventsMap directly instead of depending ProbeHelper.

test/test_images/probe_helper/probe/handler.go
```diff
@@ -0,0 +1,76 @@
...
+type Handler interface {
+	// Forward is the handler function which is called for probe requests which come
+	// from PROBE_PORT.
+	Forward(context.Context, *Helper, cloudevents.Event) error
+	// Receive is the handler function which is called for probe requests which come
+	// from RECEIVER_PORT.
+	Receive(context.Context, *Helper, cloudevents.Event) error
+}
+
+func CreateReceiverChannel(ctx context.Context, ph *Helper, channelID string) (chan bool, func(), error) {
```
liu-cong, line 47: It seems the clients who call this can just directly call the util functions on ReceivedEventsMap.

test/test_images/probe_helper/probe/cloudscheduler.go
```diff
@@ -0,0 +1,66 @@
...
+
+// CloudSchedulerSourceProbe is the probe handler for probe requests in the
+// CloudSchedulerSource probe.
+type CloudSchedulerSourceProbe struct{}
+
+var cloudSchedulerSourceProbe Handler = &CloudSchedulerSourceProbe{}
+
+func (p *CloudSchedulerSourceProbe) Forward(ctx context.Context, ph *Helper, event cloudevents.Event) error {
+	// Get the receiver channel
+	channelID := fmt.Sprintf("%s/%s", event.Extensions()[probeEventTargetServiceExtension], "cloudschedulersource-probe")
```
liu-cong, line 40: If this handler maintains its own ReceivedEventsMap, it won't need to worry about potential channelID collision. So don't really need "cloudschedulersource-probe"

test/test_images/probe_helper/probe/broker_e2e_delivery.go
```diff
@@ -0,0 +1,88 @@
...
+	if !ok {
+		return fmt.Errorf("Broker e2e delivery probe event has no '%s' extension", namespaceExtension)
+	}
+	broker, ok := event.Extensions()[brokerExtension]
+	if !ok {
+		broker = "default"
+	}
+
+	// Get the receiver channel
+	channelID := fmt.Sprintf("%s/%s", event.Extensions()[probeEventTargetServiceExtension], event.ID())
```
liu-cong, line 53: nit: Let's make channelID a helper function to guarantee the same channelID implementation in both Forward and Receive.

test/test_images/probe_helper/probe/handler.go
```diff
@@ -0,0 +1,76 @@
...
+	if err != nil {
+		return nil, nil, err
+	}
+	cleanupFunc := func() {
+		ph.ReceivedEvents.DeleteReceiverChannel(channelID)
+	}
+	return receiverChannel, cleanupFunc, nil
+}
+
+func CloseReceiverChannel(ctx context.Context, ph *Helper, channelID string) error {
```
liu-cong, line 58: nit: This function doesn't really "close" the channel, maybe call it something like "signalReceiverChannel". Also the logic here can be moved to the ReceivedEventMap

test/test_images/probe_helper/probe/cloudscheduler.go
```diff
@@ -0,0 +1,66 @@
...
+	receiverChannel, cleanupFunc, err := CreateReceiverChannel(ctx, ph, channelID)
```
liu-cong, line 41: How do we deal with the static scheduler source now? So it will create and destroy the channels? What happens in case of concurrent probing?

tommyreddad: This is a work in progress, the current implementation on this PR is not complete. I think it makes sense to keep a map of timestamps in the receiver, check that map during probes in the forward handlers, and leverage the liveness probe to garbage collect the stale timestamps. I will work on this.

test/test_images/probe_helper/probe/helper.go
tommyreddad: Good point! Maybe the extension carrying this information can become "targetpath" instead then?

liu-cong: We are already using "namespace" and "broker" for the e2e delivery case. I wonder if we can just do similarly: just use "namespace" and "trigger".

test/test_images/probe_helper/probe/helper.go
tommyreddad: I added some documentation about this, and changed to a targetpath instead of service or trigger:
```golang
	// This is the CloudEvent extension which holds the target path of probe requests.
	// The path of receiver probe requests is injected into this extension by the
	// receiver client. The corresponding forward probe requests should include
	// this extension with the same path. For example:
	//
	// If a source is configured to sink events to:
	//   http://probe-helper-receiver.cloud-run-events-probe.svc.cluster.local/some-path-goes-here
	//
	// then, each event generated by the source and passed to the probe helper
	// receiver client will include the header 'Ce-Targetpath: /some-path-goes-here'.
	// And in order for the probe to succeed, the corresponding forward probe should
	// also include the header 'Ce-Targetpath: /some-path-goes-here'.
	probeEventTargetPathExtension = "targetpath"
```

test/test_images/probe_helper/probe/helper.go
```diff
@@ -0,0 +1,374 @@
...
+
+	// initialize the probe checker
+	if ph.ProbeChecker == nil {
+		logger.Fatalw("Unspecified probe checker")
+	}
+	ph.ProbeChecker.LastForwardEventTime.SetNow()
+	ph.ProbeChecker.LastReceiverEventTime.SetNow()
+
+	// create pubsub client
+	if ph.PubsubClient == nil {
```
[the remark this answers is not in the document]
tommyreddad: Added an `Initialize` method and emphasized it should be called before `Run`.

test/test_images/probe_helper/probe/helper.go
```diff
@@ -0,0 +1,327 @@
...
+	// The pubsub client wrapped by a CloudEvents client for the CloudPubSubSource
+	// probe and used for the CloudAuditLogsSource probe
+	PubsubClient *pubsub.Client
+
+	// The CloudEvents client responsible for forwarding events as messages to a
+	// topic for the CloudPubSubSource probe.
+	CePubsubClient cloudevents.Client
+
+	// The storage client used in the CloudStorageSource
+	StorageClient *storage.Client
```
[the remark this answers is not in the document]
tommyreddad: It is useful for the probe helper to manage & be aware of the shared resources between Handlers. With that in mind, now each Handler refers to its own pointer to the clients etc.

test/test_images/probe_helper/probe/handler.go
```diff
@@ -0,0 +1,76 @@
...
+3. Add the mapping between forward probe type and the object from step 2 to the forwardProbeHandlers map in probe/helper.go
+4. Add the mapping between receiver probe type and the object from step 2 to the receiveProbeHandlers map in probe/helper.go
+
+*/
+
+// Handler is the interface which static probe objects should implement.
+type Handler interface {
+	// Forward is the handler function which is called for probe requests which come
+	// from PROBE_PORT.
+	Forward(context.Context, *Helper, cloudevents.Event) error
```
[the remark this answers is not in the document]
tommyreddad: Correct, it is not needed.

test/test_images/probe_helper/probe/helper.go
tommyreddad: Settled on targetpath.

test/test_images/probe_helper/probe/broker_e2e_delivery.go
tommyreddad: Done.

test/test_images/probe_helper/probe/handler.go
tommyreddad: Done.

test/test_images/probe_helper/probe/broker_e2e_delivery.go
tommyreddad: Done.

test/test_images/probe_helper/probe/cloudscheduler.go
tommyreddad: There should be no problem with concurrent probing now. All different scheduler tick times are tracked, and during liveness probes, the stale ticks (>3 minutes old) are discarded from being tracked.

test/test_images/probe_helper/probe/helper.go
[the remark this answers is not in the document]
liu-cong: I don't see why the helper needs to maintain a reference to the StorageClient for example, only the storagehandler needs to know that. Same for other dependencies.  Unless there are things they really need to "share"

The reason why we refactor this to different handlers is separation of concern. The helper should really be a thin layer that should purely rely on the handler interface but not its implementation details. Imagine you need to add a new handler, you shouldn't need to tough Helper at all. (To achieve this, we could create a EventTypeToHandlerMap in the Helper, and move the Handler initialization logic out (maybe to main.go), and you just need to add the new handler to the map when initializing the helper).

test/test_images/probe_helper/probe/helper.go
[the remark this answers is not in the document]
liu-cong: nit: I think we can get rid of the "if" checks now.

test/test_images/probe_helper/probe/helper.go
```diff
@@ -0,0 +1,402 @@
...
+			return cloudevents.ResultNACK
+		}
+		return cloudevents.ResultACK
+	}
+}
+
+// receiveEvent is the base receiver probe request handler which is called
+// whenever the probe helper receives a CloudEvent through port RECEIVER_PORT or
+// through the specified receiver port listener.
+func (ph *Helper) receiveEvent(ctx context.Context) cloudEventsFunc {
```
liu-cong, line 140: I think the receiver needs to verify that the path matches probeEventTargetPathExtension?

test/test_images/probe_helper/utils/liveness.go
```diff
@@ -0,0 +1,89 @@
...
+	for timestampID, schedulerTime := range c.SchedulerEventTimes.Times {
+		if delay := time.Now().Sub(schedulerTime); delay.Nanoseconds() > c.SchedulerStaleDuration.Nanoseconds() {
+			logging.FromContext(ctx).Infow("Deleting stale scheduler time", zap.String("timestampID", timestampID), zap.Duration("delay", delay))
+			delete(c.SchedulerEventTimes.Times, timestampID)
+		}
+	}
+}
+
+// LivenessHandlerFunc returns the HTTP handler for probe helper liveness checks.
+func (c *LivenessChecker) LivenessHandlerFunc(ctx context.Context) nethttp.HandlerFunc {
```
liu-cong, line 62: I assume this is just a copy/paste except the addition the "cleanup" so I didn't review.

test/test_images/probe_helper/utils/liveness.go
```diff
@@ -0,0 +1,89 @@
...
+	LastForwardEventTime SyncTime
+
+	// LastReceiverEventTime is the timestamp of the last event processed by the receiver client.
+	LastReceiverEventTime SyncTime
+
+	// LivenessStaleDuration represents the tolerated staleness of events processed by the forward and receiver clients in the liveness probe.
+	LivenessStaleDuration time.Duration
+
+	// SchedulerStaleDuration represents the tolerated staleness of ticks observed by the CloudSchedulerSource probe.
+	SchedulerStaleDuration time.Duration
```
liu-cong, line 43: I'd like to avoid this dependency if possible. I suggest that the LivenessChecker maintain a list of "ActionFuncs" to run upon each liveness check. Again, this is about isolation: LivenessChecker shouldn't have knowledge about scheduler source. But it's OK for it to run a generic "ActionFunc".

test/test_images/probe_helper/probe/handler.go
```diff
@@ -0,0 +1,49 @@
...
+type Handler interface {
+	// Forward is the handler function which is called for probe requests which come
+	// from PROBE_PORT.
+	Forward(context.Context, cloudevents.Event) error
+	// Receive is the handler function which is called for probe requests which come
+	// from RECEIVER_PORT.
+	Receive(context.Context, cloudevents.Event) error
+}
+
+func channelID(namespace, eventID string) string {
```
liu-cong, line 47: nit: the first string is the targetPath. Why don't define the helper function like:
func channelID(event Event) string?

test/test_images/probe_helper/utils/liveness.go
```diff
@@ -0,0 +1,89 @@
...
+
+	// SchedulerEventTimes is the map of latest ticks observed by the CloudSchedulerSource probe.
+	SchedulerEventTimes *SyncTimesMap
+}
+
+func (c *LivenessChecker) cleanupStaleSchedulerTimes(ctx context.Context) {
```
liu-cong, line 49: If you do the "ActionFunc" change, this function can be moved to the scheduler handler.

test/test_images/probe_helper/probe/cloudscheduler.go
```diff
@@ -0,0 +1,90 @@
...
+	}
+	periodDuration, err := time.ParseDuration(fmt.Sprint(period))
+	if err != nil {
+		return fmt.Errorf("failed to parse CloudSchedulerSource probe period and time.Duration: " + fmt.Sprint(period))
+	}
+
+	p.SchedulerEventTimes.RLock()
+	defer p.SchedulerEventTimes.RUnlock()
+
+	timestampID := channelID(fmt.Sprint(event.Extensions()[probeEventTargetPathExtension]), "cloudschedulersource-probe")
```
liu-cong, line 58: nit: here you can directly use `event.Extensions()[probeEventTargetPathExtension]`

test/test_images/probe_helper/probe/helper.go
liu-cong: > each event generated by the source and passed to the probe helper receiver client will include the header 'Ce-Targetpath: /some-path-goes-here'.

We don't have control on the event headers generated by source, do we? What we can control is by pointing the source/trigger to this specific path in the receiver uri.

tommyreddad: Correct. I inject the path into the header of received CloudEvents with receiver client middleware: https://github.com/google/knative-gcp/blob/342f0e314558b88723e04ee0e81baaf088cb670d/test/test_images/probe_helper/probe/helper.go#L308-L315

This seemed like the simplest way to access the request path while still getting to use a CloudEvent client.

liu-cong: Sorry I missed that part. Two follow ups:
1. The code only sets the header if it's not empty, which I think is wrong.  I think the intention here is to **always** set the header to be the receiver path.
2. If 1 is correct, please document this behavior. Also I think it's probably cleaner to introduce a new header, something like "probeEventReceiverPathHeader" to differentiate from the initial targetPath

test/test_images/probe_helper/probe/handler.go
tommyreddad: The problem is that we want the channelID to match between forwarded and received events, but they come in different forms. In particular, the event.ID() of the forwarded event is not necessarily the event.ID() of the received event, the event must be parsed in a way specific to that event type which is specified in the Handler Receive method.

test/test_images/probe_helper/probe/helper.go
tommyreddad: One of the reasons why the clients are shared in this manner, or for the "if" checks, is that it allows us to easily override parts of the client initialization to use fake clients instead for testing purposes. The current setup is not critical, so we can definitely avoid tracking so much stuff in the helper struct, but we would still need to pass them to the helper initializer

test/test_images/probe_helper/probe/cloudscheduler.go
tommyreddad: Done.

test/test_images/probe_helper/utils/liveness.go
tommyreddad: Done.

test/test_images/probe_helper/utils/liveness.go
tommyreddad: Done.

test/test_images/probe_helper/utils/liveness.go
tommyreddad: That's correct. But it has changed now and runs a list of actionFuncs.

test/test_images/probe_helper/probe/cloudscheduler.go
tommyreddad: Done.

test/test_images/probe_helper/probe/helper.go
tommyreddad: Done.

test/test_images/probe_helper/probe/cloudscheduler.go
tommyreddad: This is now fixed with the receiver-based approach + liveness probe cleanup.

test/test_images/probe_helper/probe/handler.go
tommyreddad: We can avoid ever dealing with the channelID by just indexing into each Handler's personal receivedEvents map, at the index of the targetpath/receiverpath. In practice, this would look like, for each probe, indexing at key "/<namespace>". So as long as we can guarantee that there is only one probe of a certain kind running in a given namespace at a given time, this would still be safe. This isn't a problem with the dynamic namespace probes, but could easily be overlooked in the future for the static namespace probes, if the prober design changes.

test/test_images/probe_helper/utils/liveness.go
```diff
@@ -0,0 +1,67 @@
...
+		if req.URL.Path != "/healthz" {
+			logging.FromContext(ctx).Warnw("Invalid liveness check request path", zap.String("path", req.URL.Path))
+			w.WriteHeader(nethttp.StatusNotFound)
+			return
+		}
+		// Apply every actionFunc, do not interrupt when encountering an error
+		success := true
+		for _, f := range c.actionFuncs {
+			if err := f(ctx); err != nil {
+				logging.FromContext(ctx).Infow("Liveness check failed", zap.Error(err))
```
liu-cong, line 55: nit: you can use multerr package to aggregate errors. See https://github.com/google/knative-gcp/blob/df0ee9b094c7885a98edf9d93f15dc60eafaa561/pkg/reconciler/broker/broker.go#L29

test/test_images/probe_helper/probe/handler.go
liu-cong: Please rename "namespace" to something like "prefix"? namespace is confusing since what's passed in is not a namespace.

test/test_images/probe_helper/probe/helper.go
liu-cong: The initializer logic should live outside of helper. We can likely make this much cleaner with a dependency injection framework like wire. This can be a follow up.

liu-cong (commented):
/lgtm with a couple of nits

tommyreddad:
Thanks so much for the extensive review!!

liu-cong:
/lgtm
/approve

Thanks @tommyreddad for doing this. This makes the code structure much better!

PR #1956, opened 2020-11-30
Update out-dated installation doc for installing from a release

### LINKED ISSUE #1952: bad documentation: missing selector
**Describe the bug**
The documentation to install knative-gcp seems wrong. There is an example to pull a selector `messaging.cloud.google.com/crd-install=true`, this doesn't exist in `cloud-run-events.yaml`.

```
kubectl apply \
  --selector messaging.cloud.google.com/crd-install=true \
  -f https://github.com/google/knative-gcp/releases/download/v0.15.0/cloud-run-events.yaml
```

from https://github.com/google/knative-gcp/blob/master/docs/install/install-knative-gcp.md#option-2-install-a-release

 **Expected behavior**
This line should either be removed or `cloud-run-events.yaml` should be updated to include the proper selector.

 **To Reproduce**
Run the install steps, specifically:

```
kubectl apply \
  --selector messaging.cloud.google.com/crd-install=true \
  -f https://github.com/google/knative-gcp/releases/download/v0.15.0/cloud-run-events.yaml
```

0.19.0 and 0.15.0

Is this selector and its resources important?

grac3gao:
Fixes #1952

## Proposed Changes

- installation doc for knative-gcp is out-dated, update it with removing selector flag
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/docs/install/install-knative-gcp.md
+++ b/docs/install/install-knative-gcp.md
@@ -82,21 +82,24 @@ ko apply -f ./config
    export KGCP_VERSION=v0.19.0
    ```
 
-1. First install the CRDs by running the `kubectl apply` for `cloud-run-events-pre-install-jobs.yaml`.
-This prevents race conditions during the installation, which
-   cause intermittent errors:
+1. First install the CRDs by running the `kubectl apply` for `cloud-run-events-pre-install-jobs.yaml`
+and `cloud-run-events.yaml` with selector. This prevents race conditions during the installation, which
+cause intermittent errors:
 
    ```shell
    kubectl apply --filename https://github.com/google/knative-gcp/releases/download/${KGCP_VERSION}/cloud-run-events-pre-install-jobs.yaml
+   kubectl apply --selector events.cloud.google.com/crd-install=true \
+   --filename https://github.com/google/knative-gcp/releases/download/${KGCP_VERSION}/cloud-run-events.yaml
    ```
 
-1. To complete the installation, run the `kubectl apply` for `cloud-run-events.yaml`:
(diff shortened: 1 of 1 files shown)


docs/install/install-knative-gcp.md
```diff
@@ -79,26 +79,25 @@ ko apply -f ./config
...
    ```
 
-1. To complete the install run the `kubectl apply` command again, this time
-   without the `--selector` flags:
+1. To complete the installation, run the `kubectl apply` for `cloud-run-events.yaml`:
 
    ```shell
    kubectl apply --filename https://github.com/google/knative-gcp/releases/download/${KGCP_VERSION}/cloud-run-events.yaml
    ```
+**Notes:** For different releases, the installation files may be different (e.g. some old releases don't have `cloud-run-events-pre-install-jobs.yaml`).
```
liu-cong, line 98: So the old instruction still works for old releases, right?

grac3gao: I made a mistake before. We are still using `events.cloud.google.com/crd-install=true` so I added back this command in latest commit. But we don't use `messaging.cloud.google.com/crd-install=true`, even I went back to v0.12.0, this label is not used.

liu-cong: nit: I think it might be easier to just create another file such as `install-knative-gcp-up-to-v0-18.md` which contains the old instructions instead of the "Notes".

grac3gao: The difference between the right version of the old instruction(remove useless `messaging.cloud.google.com/crd-install=true` selector ) and the new version of the instruction is wether we have `cloud-run-events-pre-install-jobs.yam` or not. I updated one more change for that. Please check if this new change would be better.

PR #1960, opened 2020-12-01
[release-0.19] Upgrade trigger's serving subscriber version to v1

MohamedElhawaty:
Fixes #

This is to handle an issue related to setting the ApiVersion of the trigger's subscriber (knative service) to an unsupported version

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


eclipselu:
I think you'll need to also patch the github actions, see: https://github.com/google/knative-gcp/pull/1923/

PR #1968, opened 2020-12-02
Remove cluster-wide Secret reading permissions.

### LINKED ISSUE #1827: Controller should not have Cluster-wide Secret permissions
**Problem**
We currently grant the `cloud-run-events/controller` KSA cluster-wide read permissions on Secrets: 
https://github.com/google/knative-gcp/blob/8957a127be43c428dbc9f9db3e0219af8a8684a2/config/core/roles/clusterrole.yaml#L122

As far as I can tell, the controller only wants to inform about secrets in its own namespace:

https://github.com/google/knative-gcp/blob/8957a127be43c428dbc9f9db3e0219af8a8684a2/pkg/reconciler/deployment/controller.go#L81

So it doesn't use, and therefore should not be granted, access to all Secrets on the cluster.

**Exit Criteria**
The `controller` KSA in the `cloud-run-events` namespace does not have permission to read Secrets cluster wide.

**Time Estimate (optional):** 
1 day

**Potential Solution**
This might be as easy as removing the single line:
https://github.com/google/knative-gcp/blob/8957a127be43c428dbc9f9db3e0219af8a8684a2/config/core/roles/clusterrole.yaml#L122

Because we already grant the controller Secret permissions via a Role:

https://github.com/google/knative-gcp/blob/8957a127be43c428dbc9f9db3e0219af8a8684a2/config/core/roles/role.yaml#L26

However, this may cause issues with the informer, which seems to be cluster-wide.

The permission was added in #547. In that PR it is clear that the only intention was to use it within the `cloud-run-events` namespace.

Harwayne:
Fixes #1827 

## Proposed Changes

- Remove cluster-wide Secret reading permissions.
    - Only have Secret read permissions in the system namespace.

**Release Note**

```release-note
NONE
```


grac3gao:
Looks like the test failed for:
`E1202 23:34:05.519649       1 reflector.go:178] k8s.io/client-go@v11.0.1-0.20190805182717-6502b5e7b1b5+incompatible/tools/cache/reflector.go:125: Failed to list *v1.Secret: secrets is forbidden: User "system:serviceaccount:cloud-run-events:controller" cannot list resource "secrets" in API group "" at the cluster scope`

https://pantheon.corp.google.com/storage/browser/_details/knative-prow/pr-logs/pull/google_knative-gcp/1968/pull-google-knative-gcp-integration-tests/1334275752492797952/artifacts/ac9305f7-34f4-11eb-a4ce-36bcd4bdb657/controller-logs/cloud-run-events-controller-5667948c7b-566ww-controller.txt

Not sure if that means we need to change secret informer setup?

grac3gao (approved):
/lgtm

PR #1971, opened 2020-12-03
Revert "[Automated] Update actions"

tommyreddad:
Reverts google/knative-gcp#1970

The update to the `woke` script is causing failures due to the substring "native" in "knative", e.g., https://github.com/google/knative-gcp/commit/1ed02d21751b944fa5ae55467dc64a62d59fb467#annotation_661149281


eclipselu:
/lgtm

Just one question: 
seems the original PR was automated, I'm not familiar with how the automation works, how do we prevent the bot from sending the PR again? Or can we whitelist `knative`?

tommyreddad:
The PR is automatically sent by knobots every so often according to the rules here: https://github.com/knative-sandbox/knobots/blob/master/.github/workflows/auto-updates.yaml

I'm not sure about whitelisting, I don't know how the woke tool works.

PR #1975, opened 2020-12-04
Rename GSAs to be brand-agnostic

AlexandraRoatis:
## Proposed Changes

- Rename the GSAs that are not used by e2e tests to be brand-agnostic.

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/docs/examples/keda/README.md
+++ b/docs/examples/keda/README.md
@@ -55,7 +55,7 @@ the `CloudPubSubSource` scalable. Note that you could do this for any of the
 
    ```shell
    gcloud projects add-iam-policy-binding $PROJECT_ID \
-     --member=serviceAccount:cre-dataplane@$PROJECT_ID.iam.gserviceaccount.com \
+     --member=serviceAccount:events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com \
      --role roles/monitoring.viewer
    ```
 
--- a/docs/how-to/metrics.md
+++ b/docs/how-to/metrics.md
@@ -28,6 +28,12 @@ be expanded to include sources. To enable metrics for sources, the source GSA
 (e.g `events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com`) will also require
 the `monitoring.metricWriter` role.
 
+```shell
+gcloud projects add-iam-policy-binding $PROJECT_ID \
+  --member=serviceAccount:events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com \
+  --role roles/monitoring.metricWriter
+```
+
 ## Enable Metrics in the `config-observability` ConfigMap
 
 Edit the `config-observability` ConfigMap under the `cloud-run-events` namespace
--- a/docs/how-to/metrics.md
+++ b/docs/how-to/metrics.md
@@ -13,13 +13,13 @@ Determine the Google Service Account your data plane is running as. If you
 followed [Install Knative-GCP](../install/install-knative-gcp.md) or
 [Create a Service Account for the Data Plane](../install/dataplane-service-account.md),
 then the Google Service Account will be named
-`cre-dataplane@$PROJECT_ID.iam.gserviceaccount.com`. The following command uses
+`events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com`. The following command uses
 that name. If the Google Service Account you are using is different, then
 replace it before running the command.
 
 ```shell
 gcloud projects add-iam-policy-binding $PROJECT_ID \
-  --member=serviceAccount:cre-dataplane@$PROJECT_ID.iam.gserviceaccount.com \
+  --member=serviceAccount:events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com \
   --role roles/monitoring.metricWriter
 ```
 
--- a/docs/how-to/metrics.md
+++ b/docs/how-to/metrics.md
@@ -7,21 +7,26 @@ In cloud console, navigate to
 Enable the monitoring metrics API and add your project to a workspace (or create
 a new workspace).
 
-## Add the Monitoring Metric Writer Role to the Dataplane Service Account
+## Add the Monitoring Metric Writer Role to the Data Plane Service Account
 
-Determine the Google Service Account your data plane is running as. If you
-followed [Install Knative-GCP](../install/install-knative-gcp.md) or
+Determine the Google Service Account your broker data plane is running as. If
+you followed [Install Knative-GCP](../install/install-knative-gcp.md) or
 [Create a Service Account for the Data Plane](../install/dataplane-service-account.md),
 then the Google Service Account will be named
-`events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com`. The following command uses
-that name. If the Google Service Account you are using is different, then
+`events-broker-gsa@$PROJECT_ID.iam.gserviceaccount.com`. The following command
+uses that name. If the Google Service Account you are using is different, then
 replace it before running the command.
 
 ```shell
 gcloud projects add-iam-policy-binding $PROJECT_ID \
-  --member=serviceAccount:events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com \
+  --member=serviceAccount:events-broker-gsa@$PROJECT_ID.iam.gserviceaccount.com \
   --role roles/monitoring.metricWriter
 ```
+
+We currently provide metrics only for the GCP Broker, but this option will soon
+be expanded to include sources. To enable metrics for sources, the source GSA
+(e.g `events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com`) will also require
+the `monitoring.metricWriter` role.
 
 ## Enable Metrics in the `config-observability` ConfigMap
 
--- a/docs/install/authentication-mechanisms-gcp.md
+++ b/docs/install/authentication-mechanisms-gcp.md
@@ -249,7 +249,7 @@ broker data plane named `events-broker-gsa`. Follow
    **Be sure not to check this key into source control!**
 
    ```shell
-   gcloud iam service-accounts keys create cre-dataplane.json \
+   gcloud iam service-accounts keys create events-sources-key.json \
      --iam-account=events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com
    ```
 
@@ -258,7 +258,7 @@ broker data plane named `events-broker-gsa`. Follow
    below does so in the `default` namespace.
 
    ```shell
-   kubectl --namespace default create secret generic google-cloud-key --from-file=key.json=cre-dataplane.json
+   kubectl --namespace default create secret generic google-cloud-key --from-file=key.json=events-sources-key.json
    ```
 
    `google-cloud-key` and `key.json` are default values expected by our
--- a/docs/install/authentication-mechanisms-gcp.md
+++ b/docs/install/authentication-mechanisms-gcp.md
@@ -137,12 +137,14 @@ gcloud projects add-iam-policy-binding $PROJECT_ID \
    Note that `google-cloud-key` and `key.json` are default values expected by
    our control plane.
 
-## Authentication Mechanism for the Data Plane
+## Authentication Mechanism for the Sources Data Plane
 
-This is the manual auth configuration for the Data Plane and using the Google
-Cloud Service Account `events-sources-gsa` as the credential. Refer to
+This is the manual auth configuration for the sources data plane using the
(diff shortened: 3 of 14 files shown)


grac3gao:
Checking the failure log, it seems we are still setting up iam policy binding to the old GSA `prow-job@knative-tests.iam.gserviceaccount.com` rather than the newly created ones.

test/e2e-wi-tests.sh
```diff
@@ -36,10 +36,11 @@ function export_variable() {
...
     readonly PUBSUB_SERVICE_ACCOUNT_KEY_TEMP="${PUBSUB_SERVICE_ACCOUNT_NON_PROW_KEY_TEMP}"
   else
-    readonly CONTROL_PLANE_SERVICE_ACCOUNT_EMAIL=${PROW_SERVICE_ACCOUNT_EMAIL}
+    # readonly CONTROL_PLANE_SERVICE_ACCOUNT_EMAIL=${PROW_SERVICE_ACCOUNT_EMAIL}
     # Get the PROW service account.
     readonly PROW_PROJECT_NAME=$(cut -d'.' -f1 <<< "$(cut -d'@' -f2 <<< "${PROW_SERVICE_ACCOUNT_EMAIL}")")
-    readonly DATA_PLANE_SERVICE_ACCOUNT_EMAIL="cloud-run-events-source@${PROW_PROJECT_NAME}.iam.gserviceaccount.com"
+    readonly CONTROL_PLANE_SERVICE_ACCOUNT_EMAIL="${CONTROL_PLANE_SERVICE_ACCOUNT}@${PROW_PROJECT_NAME}.iam.gserviceaccount.com"
+    readonly DATA_PLANE_SERVICE_ACCOUNT_EMAIL="events-sources-gsa@${PROW_PROJECT_NAME}.iam.gserviceaccount.com"
     readonly PUBSUB_SERVICE_ACCOUNT_EMAIL=${PROW_SERVICE_ACCOUNT_EMAIL}
```
grac3gao, line 44: We are using PUBSUB_SERVICE_ACCOUNT_EMAIL to grant permission for the broker in: https://github.com/google/knative-gcp/blob/f0c6500bfbcc1488fde44c571f8442666ada9c4e/test/e2e-common.sh#L119-L126. It seems we also need to update these line.

grac3gao:
Checked the failure log from https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/google_knative-gcp/1975/pull-google-knative-gcp-wi-tests/1336050216209485824:

1. It seems we are still setting up iam policy binding to the old GSA `prow-job@knative-tests.iam.gserviceaccount.com` rather than the newly created ones for broker.
```
Authentication setup for GCP Broker
Updated IAM policy for serviceAccount [prow-job@knative-tests.iam.gserviceaccount.com].
```

2. It seems those two Google Service Account never exist? By the way, besides creating service account in `knative-tests` project, we will also need to create the same service account in `knative-nightly` and `knative-release` project, those projects are for nightly test and release test.
```
ERROR: (gcloud.iam.service-accounts.get-iam-policy) NOT_FOUND: Not found; Gaia id not found for email events-sources-gsa@knative-tests.iam.gserviceaccount.com
ERROR: (gcloud.iam.service-accounts.add-iam-policy-binding) NOT_FOUND: Not found; Gaia id not found for email events-controller-gsa@knative-tests.iam.gserviceaccount.com
```

3. If we don't have those Google service accounts at all, why secret mode doesn't fail?
I think it is because in the set up script running in prow, in secret mode, we just don't do any google service account level configuration. Checking https://github.com/google/knative-gcp/blob/f0c6500bfbcc1488fde44c571f8442666ada9c4e/test/e2e-common.sh#L138-L141, in secret mode, we don't export key from Google Service account, we just directly use a file stored in path: https://github.com/google/knative-gcp/blob/f0c6500bfbcc1488fde44c571f8442666ada9c4e/test/e2e-secret-tests.sh#L40-L41, this file is configured by test-infra. After creating the secret, when a new test starts running in a namespace, it will copy this secret to that namespace. In other word, secret mode test are just using the old credentials to run all the test if we don't request to change the credential in that pre-configured file.

AlexandraRoatis:
@grac3gao @Harwayne this PR is ready for review. The renaming mainly applies to documentation since changing the e2e test setup was very cumbersome (see initial [review](https://github.com/google/knative-gcp/pull/1975#issuecomment-740406275)).

docs/examples/keda/README.md
```diff
@@ -55,7 +55,7 @@ the `CloudPubSubSource` scalable. Note that you could do this for any of the
 
    ```shell
    gcloud projects add-iam-policy-binding $PROJECT_ID \
-     --member=serviceAccount:cre-dataplane@$PROJECT_ID.iam.gserviceaccount.com \
+     --member=serviceAccount:events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com \
```
grantr, line 58: The `-gsa` suffix seems like an unnecessary stutter to me. GSAs in APIs are rarely referred to by name alone (they usually have the domain attached), so I'm not sure when they'd be confused with any other type of identity (like KSA). I'm not an expert so I could be missing an ambiguous case.

docs/install/install-gcp-broker.md
```diff
@@ -48,14 +48,14 @@ Workload Identity see
     gcloud iam service-accounts add-iam-policy-binding \
      --role roles/iam.workloadIdentityUser \
      --member serviceAccount:$PROJECT_ID.svc.id.goog[cloud-run-events/broker] \
-     cre-dataplane@$PROJECT_ID.iam.gserviceaccount.com
+     events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com
```
grantr, line 51: It's a bit confusing that this GSA is called `events-sources-gsa` and is being bound to the broker KSA. I liked `dataplane` better. Not a big deal though.

grantr (approved):
Minor quibbles with the names, but IIUC this rename is already in progress in CI so we should get this in.

/lgtm

docs/examples/keda/README.md
bryeung: I probably don't have enough experience with these systems to fully understand your concern. I'm sympathetic to desire to avoid stutter, but I would imagine myself referring to this in most conversations as "events-sources-gsa" in general, not as "event-sources@PROJECT_ID.iam.gserviceaccount.com" (And hence I like the -gsa suffix for additional clarity.)

docs/install/install-gcp-broker.md
AlexandraRoatis: I will update the docs to use the broker GSA here (and to create one in the data plane script).

docs/examples/keda/README.md
AlexandraRoatis: Thanks for the feedback! I decided to keep the suffix.

docs/install/install-gcp-broker.md
AlexandraRoatis: I've update the docs to recommend using a separate GSA for the broker.

docs/how-to/metrics.md
```diff
@@ -7,22 +7,27 @@ In cloud console, navigate to
...
-  --member=serviceAccount:cre-dataplane@$PROJECT_ID.iam.gserviceaccount.com \
+  --member=serviceAccount:events-broker-gsa@$PROJECT_ID.iam.gserviceaccount.com \
   --role roles/monitoring.metricWriter
 ```
 
+We currently provide metrics only for the GCP Broker, but this option will soon
+be expanded to include sources. To enable metrics for sources, the source GSA
+(e.g `events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com`) will also require
+the `monitoring.metricWriter` role.
+
```
Harwayne, line 30: Add the shell command here.

docs/install/dataplane-service-account.md
```diff
@@ -28,11 +28,25 @@ In general, we would just need permissions to receive messages
...
+   If you are setting up multiple GSAs you can follow the same set of steps, but
+   take note to configure authentication only for the namespaces that are
+   intended for each account.
+
+   Additionally, while it is possible to use the same GSA for both the broker
+   and the sources, we recommend creating a dedicated Google Service Account for
+   the broker data plane (e.g. `events-broker-gsa`):
+
+   ```shell
+   gcloud iam service-accounts create events-sources-gsa
```
Harwayne, line 49: This looks identical to the previous shell command, should it be for `events-broker-gsa`?

docs/install/dataplane-service-account.md
```diff
@@ -54,20 +68,27 @@ also need the ability to publish messages (`roles/pubsub.publisher`).
...
    ```shell
    gcloud projects add-iam-policy-binding $PROJECT_ID \
-     --member=serviceAccount:cre-dataplane@$PROJECT_ID.iam.gserviceaccount.com \
+     --member=serviceAccount:events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com \
      --role roles/cloudtrace.agent
    ```
 
+   The same set of permissions should also be assigned to the broker data plane
+   account `events-broker-gsa@$PROJECT_ID.iam.gserviceaccount.com`.
+
```
Harwayne, line 85: Add the shell commands.

docs/install/dataplane-service-account.md
```diff
@@ -163,7 +184,7 @@ Plane:
   After running this, every time when you create resources, the Controller will
   create a Kubernetes service account `default-cre-dataplane` in the namespace
```
Harwayne, line 185: Not needed in this PR, but `default-cre-dataplane` should probably be changed too.

docs/install/install-gcp-broker.md
```diff
@@ -65,13 +69,14 @@ Workload Identity see
 
    ```shell
    gcloud iam service-accounts keys create cre-dataplane.json \
```
Harwayne, line 71: Doesn't have to be this PR, but `cre-dataplane.json` should be renamed too.

docs/how-to/metrics.md
AlexandraRoatis: Done.

docs/install/dataplane-service-account.md
AlexandraRoatis: Done.

docs/install/authentication-mechanisms-gcp.md
```diff
@@ -237,26 +239,26 @@ for automated scripts.
...
+   gcloud iam service-accounts keys create events-sources-key.json \
+     --iam-account=events-sources-gsa@$PROJECT_ID.iam.gserviceaccount.com
    ```
 
 1. Create a secret on the Kubernetes cluster with the downloaded key. Remember
    to create the secret in the namespace your resources will reside. The example
    below does so in the `default` namespace.
 
    ```shell
-   kubectl --namespace default create secret generic google-cloud-key --from-file=key.json=cre-dataplane.json
```
AlexandraRoatis, line 259: @grac3gao should I replace `google-cloud-key` here with `google-cloud-sources-key`?

grac3gao: Probably not. By default, the source is consuming `google-cloud-key` (webhook will add this to source).

AlexandraRoatis: Thank you!

docs/install/dataplane-service-account.md
AlexandraRoatis: Fixed.

docs/install/install-gcp-broker.md
AlexandraRoatis: Renamed to `events-broker-key.json` or `events-sources-key.json` for broker and sources respectively.

hack/init_data_plane_gke.sh
```diff
@@ -44,8 +44,8 @@ PROJECT_ID=${4:-$(gcloud config get-value project)}
 echo "PROJECT_ID ${PROJECT_ID} is used for this configuration"
 
 if [[ $MODE == "default" ]]; then
-  # Grant iam.serviceAccountAdmin permission of the Google service account cre-dataplane
```
grac3gao, line 47: we might need to change `K8S_SERVICE_ACCOUNT` in this script as well. It is using `K8S_SERVICE_ACCOUNT=${3:-$DEFAULT_K8S_SERVICE_ACCOUNT}` and `DEFAULT_K8S_SERVICE_ACCOUNT="default-cre-dataplane"`

grac3gao: Maybe in following PR

Harwayne (commented):
/lgtm

grantr (approved):
/lgtm

PR #1977, opened 2020-12-04
Rename cloud-run-events to events-system in release 0.19

AlexandraRoatis:
This change is currently applied to `release-0.19` for the diff, however the intention is to create a new branch `release-0.19-ns` that can be used in the upgrade tests for PR #1917 . 

## Proposed Changes

- cloud-run-events-probe -> events-system-probe
- cloud-run-events-example -> events-system-example
- cloud-run-events-controller -> events-system-controller
- cloud-run-events-broker -> events-system-broker
- cloud-run-events-webhook -> events-system-webhook
- cloud-run-events-source-observer -> events-system-source-observer
- cloud-run-events-pubsub-* -> events-pubsub-*
- cloud-run-events-* to events-system-* (non-namespace & not service accounts)
- cloud-run-events -> events-system (namespace only)
- test-cloud-run-events -> test-events-system

**Release Note**

```release-note

```

**Docs**


zhongduo:
I don't think we should do it at all, unless we really want this new namespace in release 0.19.

We can always just change the upgrade test so that it temporary pull yaml from `somewhere`, which can be whichever commit with the new namespace, when merging to master. IMO upgrade test shouldn't be a factor of whether the new namespace should be in 0.19 or not. It is just a ....test.

AlexandraRoatis:
@zhongduo Please read the note in the beginning of the PR. This is not intended to be merged with the release, but I cannot create a meaningful diff otherwise.

AlexandraRoatis:
Adding a hold to make it clear that this is not intended to be merged.
/hold

PR #1981, opened 2020-12-05
Remove former team members from OWNERS

grantr:
:cry: So they stop getting spammed with review requests.


zhongduo:
/lgtm
sad to do this.

Issue #1973, opened 2020-12-03
Ingress Readiness and Liveness check stop working after an upstream change


grac3gao:
**Describe the bug**
Ingress readiness and liveness (using kncloudevents library from eventing) check stop working after an upstream change: https://github.com/knative/eventing/pull/3917

 **Expected behavior**

 **To Reproduce**
change the following line to return `response.WriteHeader(nethttp.StatusMethodNotAllowed) `
https://github.com/google/knative-gcp/blob/544ec2d0b9c5ed63ca2416defb1234d65ac84a75/pkg/broker/ingress/handler.go#L123-L125

It should fail the readiness and liveness check for the ingress pod, however, ingress pod will continue work.

Revert the upstream change, then return `response.WriteHeader(nethttp.StatusMethodNotAllowed) `, readiness and liveness check will work.

 **Knative-GCP release version**

 **Additional context**
Add any other context about the problem here such as proposed priority

grac3gao:
kncloudevents are using drainer in pkg to run liveness and readiness check now (https://github.com/knative/pkg/blob/d5c09d2aef18eaabbb9ce3ac71a487288ca542aa/network/handlers/drain.go#L78), all inbound request will firstly send through drainer then to the handler we configured. Drainer will take every request from k8s probe and truncated it instead of continuing sending it to our handler, which makes all our customized configuration for liveness and readiness useless as no request from k8s probe will be send into handler.
There are two ways to fix it:
1. In the upstream drainer library, adding an interface for a customizedCheck function. Drainer will need to run customizedCheck for probe requests instead of directly returning 200. Then, change kncloudevents to also have a field for customizedCheck to initialize drainer.
2. Have our own message_receive which is mostly same as kncloudevents except wrapping the drainer to run some customized check.

grac3gao:
knative-pkg PR merged: https://github.com/knative/pkg/pull/1977

grac3gao:
Knative-eventing PR merged: knative/eventing#4707

Now we are OK to add customized liveness and readiness check

PR #1982, opened 2020-12-07
Authentication check running inside the Pod for Source and BrokerCell Fanout and Retry

grac3gao:
Fixes #

## Proposed Changes

- Phase one for the Second step to add the authentication check
1. Environment variable to differentiate authentication mode (<-this one)
2. **Authentication check running inside the Pod**(<-this one)
3. Additional k8s Event check for Secret absence

**Note** This PR only adds authentication check in Pod of Source and BrokerCell Fanout and Retry.
For Pods using kncloudevent library like Ingress, it is blocked by https://github.com/google/knative-gcp/issues/1973

- Functionality includes:
1. Running authentication check inside the Pod
2. Writing Termination log into Pod once authentication check failed
3. Remark upper-level resource with authentication related Reason and Message

- Needs to add more UTs.

**Release Note**

```release-note
🎁 Authentication check running inside the Pod for Source and BrokerCell Fanout and Retry 
```

**Docs**


pkg/apis/intevents/v1beta1/pullsubscription_lifecycle.go
```diff
@@ -85,14 +87,21 @@ func (s *PullSubscriptionStatus) MarkNoSubscription(reason, messageFormat string
...
+	replicaUnavailable := false
 	for _, cond := range d.Status.Conditions {
 		if cond.Type == appsv1.DeploymentAvailable {
 			deploymentAvailableFound = true
 			if cond.Status == corev1.ConditionTrue {
 				pullSubscriptionCondSet.Manage(s).MarkTrue(PullSubscriptionConditionDeployed)
 			} else if cond.Status == corev1.ConditionFalse {
+				if cond.Reason == replicaUnavailableReason {
+					replicaUnavailable = true
+				}
```
grac3gao, line 104: I changed lifecycle file for v1 pullsubscription, should I change v1beta1 like this?

pkg/apis/intevents/v1/pullsubscription_lifecycle.go
```diff
@@ -93,14 +95,21 @@ func (s *PullSubscriptionStatus) MarkDeployedUnknown(reason, messageFormat strin
 
 // PropagateDeploymentAvailability uses the availability of the provided Deployment to determine if
 // PullSubscriptionConditionDeployed should be marked as true or false.
-func (s *PullSubscriptionStatus) PropagateDeploymentAvailability(d *appsv1.Deployment) {
+// For authentication check purpose, this method will return true  if a false condition
```
Harwayne, line 98: I would normally expect it to be the opposite, where `true` is everything is healthy.

pkg/apis/intevents/v1beta1/pullsubscription_lifecycle.go
Harwayne: I think we normally try to keep them all in sync, so I think we should change the v1beta1 and v1alpha1 versions as well.

pkg/broker/handler/pool.go
```diff
@@ -80,6 +84,9 @@ func (c *probeChecker) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
+	// Perform Authentication check.
+	authcheck.AuthenticationCheck(req.Context(), c.logger, c.authType, w)
```
Harwayne, line 88: Should this return under some condition? Such as if the auth check fails.

pkg/broker/handler/pool_test.go
```diff
@@ -74,17 +74,16 @@ func TestSyncPool(t *testing.T) {
 
 		ch <- struct{}{}
 		syncPool.verifySyncOnceCalled(t)
-		assertProbeCheckResult(t, p, true)
-
-		// Intentionally causing a failed check.
-		time.Sleep(time.Second)
-		assertProbeCheckResult(t, p, false)
+		// False because authentication check will fail.
+		assertProbeCheckResult(t, p, false, "healthz")
```
Harwayne, line 78: Should we also have positive test cases?

pkg/utils/authcheck/authcheck.go
```diff
@@ -0,0 +1,114 @@
...
+	"io/ioutil"
+	"net/http"
+	nethttp "net/http"
+
+	"go.uber.org/zap"
+	"golang.org/x/oauth2/google"
+)
+
+const (
+	// Resource is used as the path to get the default token from metadata server.
```
Harwayne, line 35:
```suggestion
	// resource is used as the path to get the default token from metadata server.
```

pkg/utils/authcheck/authcheck.go
```diff
@@ -0,0 +1,114 @@
...
+	// In workload-identity-gsa mode, this path will return a token if
+	// corresponding k8s service account and google service account establish a correct relationship.
+	resource = "http://metadata.google.internal/computeMetadata/v1/instance/service-accounts/default/token"
+	// Scope is used as the scope to get token from default credential.
+	scope = "https://www.googleapis.com/auth/cloud-platform"
+)
+
+// AuthenticationCheck performs the authentication check running in the Pod.
+func AuthenticationCheck(ctx context.Context, logger *zap.Logger, authType AuthType, response nethttp.ResponseWriter) {
```
Harwayne, line 44: My preference is to have this not take in `response`. Instead return an error, and when non-nil, the caller modifies `response.WriteHeade(nethttp.StatusUnauthorized)`.

pkg/utils/authcheck/authcheck.go
```diff
@@ -0,0 +1,114 @@
...
+	var err error
+	if authType == Secret {
```
Harwayne, line 46: My preference is a switch.

pkg/utils/authcheck/authcheck.go
```diff
@@ -0,0 +1,114 @@
...
+	} else if authType == WorkloadIdentityGSA {
+		err = AuthenticationCheckForWorkloadIdentityGSA(resource)
+	} else {
+		logger.Error(fmt.Sprint("unknown auth type: ", authType))
+		response.WriteHeader(nethttp.StatusUnauthorized)
+		return
+	}
+
+	if err != nil {
+		// Transfer the error into a string message, otherwise, marshalling error may return nil unexpectedly.
```
Harwayne, line 57: Move this to another method, something like `writeTerminationLog`.

pkg/utils/authcheck/authchek_test.go
```diff
@@ -0,0 +1,81 @@
+/*
```
Harwayne, line 1: File name is missing the `c` in check.

pkg/utils/authcheck/authchek_test.go
```diff
@@ -0,0 +1,81 @@
...
+			authType:       "",
+			wantStatusCode: http.StatusUnauthorized,
+			setEnv:         false,
+		},
+		{
+			name:           "authentication check uses secret authType",
+			authType:       Secret,
+			wantStatusCode: http.StatusUnauthorized,
+			setEnv:         true,
+		},
```
Harwayne, line 50: Add a test for WI-gsa as well. Likely by making the http client a var and swapping it out during the test.

pkg/utils/authcheck/authchek_test.go
```diff
@@ -0,0 +1,81 @@
...
+	}
```
Harwayne, line 51: Add positive tests. Ones that pass the auth check.

pkg/utils/authcheck/authtype.go
```diff
@@ -52,12 +55,15 @@ const (
...
 
-var BrokerSecret = &corev1.SecretKeySelector{
-	LocalObjectReference: corev1.LocalObjectReference{
-		Name: "google-broker-key",
-	},
-	Key: "key.json",
-}
+var (
+	// Regex for a valid google service account email.
+	emailRegexp = regexp.MustCompile(`^[a-z][a-z0-9-]{5,29}@[a-z][a-z0-9-]{5,29}\.iam\.gserviceaccount.com$`)
```
Harwayne, line 60:
```suggestion
	emailRegexp = regexp.MustCompile(`^[a-z][a-z0-9-]{5,29}@[a-z][a-z0-9-]{5,29}\.iam\.gserviceaccount\.com$`)
```

As suggested by the bot.

pkg/utils/authcheck/authtype.go
```diff
@@ -114,7 +120,19 @@ func getAuthTypeForWorkloadIdentity(ctx context.Context, serviceAccountLister co
...
+		// Check if email is a valid google service account email.
+		if match := emailRegexp.FindStringSubmatch(email); len(match) == 0 {
+			// The format of google service account email is service-account-name@project-id.iam.gserviceaccount.com
+
+			// Service account name must be between 6 and 30 characters (inclusive),
+			// must begin with a lowercase letter, and consist of lowercase alphanumeric characters that can be separated by hyphens.
+
+			// Project IDs must start with a lowercase letter and can have lowercase ASCII letters, digits or hyphens,
+			// must be between 6 and 30 characters.
+			return "", fmt.Errorf("the annotation %s of Kubernetes Service Account %s does not contain a valid Google Service Account",
```
Harwayne, line 133: Include the email as well.

pkg/utils/authcheck/authtype.go
Harwayne, line 60: This isn't quite accurate, on one of my older projects named `example.com:harwayne-example`, I have an GSA with the email `bob123@harwayne-example.example.com.iam.gserviceaccount.com`. I don't think new projects can get this style name, but older ones can. You might want to allow anything between `@{5,29}@[a-z][a-z0-9-]{5,29}` and `.iam.gserviceaccount.com`. Maybe `^[a-z][a-z0-9-]{5,29}@[a-z][a-z0-9-]{5,29}.*\.iam\.gserviceaccount\.com$`.

pkg/utils/authcheck/authtype.go
```diff
@@ -114,7 +120,19 @@ func getAuthTypeForWorkloadIdentity(ctx context.Context, serviceAccountLister co
 				args.ServiceAccountName)
 		}
 		return "", fmt.Errorf("error getting Kubernetes Service Account: %w", err)
-	} else if kServiceAccount.Annotations[resources.WorkloadIdentityKey] != "" {
+	} else if email := kServiceAccount.Annotations[resources.WorkloadIdentityKey]; email != "" {
+		// Check if email is a valid google service account email.
+		if match := emailRegexp.FindStringSubmatch(email); len(match) == 0 {
+			// The format of google service account email is service-account-name@project-id.iam.gserviceaccount.com
```
Harwayne, line 126: Put this comment about the structure of the GSA's email on the regex, not here.

pkg/utils/authcheck/authtype_test.go
```diff
@@ -78,13 +78,25 @@ func TestGetAuthTypeForSources(t *testing.T) {
...
+			objects: []runtime.Object{
+				pkgtesting.NewServiceAccount(serviceAccountName, testNS,
+					pkgtesting.WithServiceAccountAnnotation("name"),
+				),
+			},
+			args:         serviceAccountArgs,
+			wantAuthType: "",
+			wantError: fmt.Errorf("using Workload Identity for authentication configuration: the annotation iam.gke.io/gcp-service-account "+
+				"of Kubernetes Service Account %s does not contain a valid Google Service Account", serviceAccountName),
+		},
```
Harwayne, line 99: Add more tests with emails that are just a little off, to better match accidental typos.

pkg/utils/authcheck/list.go
```diff
@@ -0,0 +1,66 @@
...
+	"go.uber.org/zap"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
+	"k8s.io/client-go/kubernetes"
+
+	"knative.dev/pkg/logging"
+)
+
+// GetPodList get a list of Pod in a certain namespace with certain label selector.
```
Harwayne, line 34:
```suggestion
// GetPodList get a list of Pods in a certain namespace with certain label selector.
```

pkg/utils/authcheck/list.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+	})
+	if err != nil {
+		logging.FromContext(ctx).Desugar().Error("Unable to list pod", zap.Error(err))
+		return nil, err
+	}
+	return pl, nil
+}
+
+// GetTerminationLogFromPodList get termination log from Pod.
```
Harwayne, line 50:
```suggestion
// GetTerminationLogFromPodList gets the termination log from Pods that failed due to auth check errors. Returns the first auth termination log from any Pods in the list.
```

pkg/utils/authcheck/list.go
```diff
@@ -0,0 +1,66 @@
...
+			} else if cs.LastTerminationState.Terminated != nil && isAuthMessage(cs.LastTerminationState.Terminated.Message) {
+				return cs.LastTerminationState.Terminated.Message
+			}
+		}
+	}
+	return ""
+}
+
+func isAuthMessage(message string) bool {
+	return strings.Contains(message, "checking authentication")
```
Harwayne, line 65: I would prefer to have a constant for this, as it must be the same between this function and the termination log printer.

pkg/utils/authcheck/probechecker.go
```diff
@@ -0,0 +1,73 @@
...
+// NewProbeChecker returns ProbeChecker with default probe checker port.
+func NewProbeChecker(logger *zap.Logger, authType AuthType) ProbeChecker {
+	return ProbeChecker{
+		logger:   logger,
+		port:     DefaultProbeCheckPort,
+		authType: authType,
+	}
+}
+
+// Start will initialize s http serve and start to listen.
```
Harwayne, line 47:
```suggestion
// Start will initialize an http server and start to listen.
```

pkg/utils/authcheck/probechecker_test.go
```diff
@@ -0,0 +1,75 @@
...
+		t.Fatal("Failed to listen TCP:", err)
+	}
+	l.Close()
+	port := l.Addr().(*net.TCPAddr).Port
+
+	logger := logging.FromContext(ctx)
+	probeChecker := ProbeChecker{
+		logger:   logger,
+		port:     port,
+		authType: "",
```
Harwayne, line 53: Add positive tests too.

pkg/apis/intevents/v1beta1/pullsubscription_lifecycle.go
grac3gao: pullsubscription does not have v1alpha1 in intevents (v1alpha1 folder only has brokercell). I changed v1 and v1beta1.

pkg/utils/authcheck/authtype.go
grac3gao: Good to know! Thanks

pkg/broker/handler/pool.go
grac3gao: It only write statue into the response, which I included in the authcheck.AuthenticationCheck previously. Based on your comment, I'll do:
```
if err := authcheck.AuthenticationCheck; err != nil {
			pc.logger.Error("authentication check failed", zap.Error(err))
			w.WriteHeader(http.StatusUnauthorized)
                         return
}
```

pkg/broker/handler/pool_test.go
grac3gao: done

pkg/utils/authcheck/authcheck_test.go
```diff
@@ -0,0 +1,85 @@
...
+		authType   AuthType
+		statusCode int
+		wantError  bool
+		setEnv     bool
+	}{
+		{
+			name:       "authentication check failed, using empty authType",
+			authType:   "empty",
+			statusCode: http.StatusUnauthorized,
+			wantError:  true,
```
grac3gao, line 42: Was trying to use specific errors, however, in Prow and local, it returns different errors. In local, it can't write error into termination file (no such file), but in Prow, it can. Tried to create file, error showed no permission to do so. Thus, I just verify whether there is an error or not.

pkg/broker/handler/pool.go
```diff
@@ -80,6 +86,13 @@ func (c *probeChecker) ServeHTTP(w http.ResponseWriter, req *http.Request) {
 		w.WriteHeader(http.StatusNotFound)
 		return
 	}
+	// Perform Authentication check.
+	if err := authcheck.AuthenticationCheck(req.Context(), c.authType, c.authCheckClient); err != nil {
+		c.logger.Error("authentication check failed", zap.Error(err))
+		w.WriteHeader(http.StatusUnauthorized)
```
yolocs, line 92: Is it a common practice to return 403 in prober? I find it's confusing because it's hard to tell whether it's "called not allowed to do the health check" or "health check failed".

pkg/apis/intevents/v1beta1/pullsubscription_lifecycle_test.go
```diff
@@ -506,3 +522,12 @@ func TestPullSubscriptionStatusGetCondition(t *testing.T) {
 		})
 	}
 }
+
+func TestPropagateDeploymentAvailability(t *testing.T) {
+	s := &PullSubscriptionStatus{}
+	got := s.PropagateDeploymentAvailability(replicaUnavailableDeployment)
+	want := false
+	if diff := cmp.Diff(got, want); diff != "" {
+		t.Error("unexpected condition (-want, +got) =", diff)
```
yolocs, line 531: Diff(want,got) then it will produce (-want,+got)

cmd/broker/fanout/main.go
```diff
@@ -99,7 +104,7 @@ func main() {
 	if err != nil {
 		logger.Fatal("Failed to create fanout sync pool", zap.Error(err))
 	}
-	if _, err := handler.StartSyncPool(ctx, syncPool, syncSignal, env.MaxStaleDuration, handler.DefaultProbeCheckPort); err != nil {
+	if _, err := handler.StartSyncPool(ctx, syncPool, syncSignal, env.MaxStaleDuration, handler.DefaultProbeCheckPort, env.AuthType, authcheckclient.NewAuthCheckClient()); err != nil {
```
yolocs, line 107: It doesn't impact the functionality for sure, but I'm curious why not wrapping all authcheck stuff into a single interface? E.g.

```
type AuthChecker interface {
	Check() error
}
```

The real impl could contain the `AuthType` and the fake impl just needs to implement the `Check()` func.

In this way, you don't have to check the `StartSyncPool` to pass along the auth type.

pkg/broker/ingress/handler.go
```diff
@@ -97,15 +99,17 @@ type Handler struct {
...
 
 // NewHandler creates a new ingress handler.
-func NewHandler(ctx context.Context, httpReceiver HttpMessageReceiver, decouple DecoupleSink, reporter *metrics.IngressReporter) *Handler {
+func NewHandler(ctx context.Context, httpReceiver HttpMessageReceiver, decouple DecoupleSink, reporter *metrics.IngressReporter, authType authcheck.AuthType) *Handler {
 	return &Handler{
 		httpReceiver: httpReceiver,
 		decouple:     decouple,
 		reporter:     reporter,
 		logger:       logging.FromContext(ctx),
+		authType:     authType,
```
yolocs, line 112: I lost track - how is `authType` used in the `Handler`?

yolocs (commented):
Sorry for the late comments

pkg/broker/handler/pool.go
grac3gao: I think the probe result is showing the pod status, for whether it is having issues while the process is running. So I choose 401 to indicate the process is running with auth problem. I think this is the most close one I could find, but I could be wrong. Which statusCode do you think is more proper?

cmd/broker/fanout/main.go
grac3gao: Are you referring to inject those parameters when InitializeSyncPool instead of passing the parameter into StartSyncPool, or referring to add an interface into startSyncPool instead of those two parameters? I still need to pass the authType into the pool in order to make the auth check work together with existing probe checkers. And even I use an interface, I still need the mock client for my other UTs (to add UT for authentication check in authcheck.go). So I just re-use the mock client for pool related UTs (If I add the interface, I'll need additional fake interface for those UTs in addition to the mock client). I can change to use an interface if you think that is more proper.

pkg/broker/ingress/handler.go
grac3gao: It should be passed into ingress's probe check. However, as our ingress liveness and readiness check has an issue #1973, in this PR, I only configure authType to be an env into ingress Pod, but didn't add the probe check for it. I will do it after unblock #1973

pkg/broker/handler/pool.go
yolocs: 4xx means client error so its natural meaning is that kubelet is not authorized to call the health check. I think 500 should be sufficient. Are you depending on the status code to do something?

grac3gao: makes sense, changed

cmd/broker/fanout/main.go
grac3gao: done. changed to use interface

pkg/utils/authcheck/authcheck.go
```diff
@@ -0,0 +1,132 @@
...
+	Check(ctx context.Context) error
+}
+
+type DefaultAuthenticationCheck struct {
+	authType AuthType
+	client   *http.Client
+	url      string
+}
+
+func NewDefaultAuthenticationCheck(authType AuthType) AuthenticationCheck {
```
yolocs, line 53: A simpler name would be `NewDefault` because the package name already indicates it's for auth check.

pkg/broker/handler/pool_test.go
```diff
@@ -138,3 +143,22 @@ func GetFreePort() (int, error) {
...
+	authType authcheck.AuthType
+	noError  bool
+}
+
+func NewFakeAuthenticationCheck(authType authcheck.AuthType, noError bool) authcheck.AuthenticationCheck {
+	return &FakeAuthenticationCheck{
+		authType: authType,
+		noError:  noError,
+	}
+}
```
yolocs, line 157: nit: you can save this method and directly use `&FakeAuthenticationCheck{}`. Also is there any need to use `FakeXxx` instead of `fakeXxx`?

pkg/utils/authcheck/probechecker_test.go
```diff
@@ -0,0 +1,114 @@
...
+		})
+	}
+}
+
+type FakeAuthenticationCheck struct {
+	authType AuthType
+	noError  bool
+}
+
+func NewFakeAuthenticationCheck(authType AuthType, noError bool) AuthenticationCheck {
```
yolocs, line 102: Similar comments for this one. Also if you want, you can provide a shared fake implementation under `pkg/utils/authcheck/fake`

pkg/utils/authcheck/probechecker_test.go
```diff
@@ -0,0 +1,114 @@
...
+			resp, err := client.Do(req)
+
+			if err != nil {
+				t.Fatal("Failed to execute probe check:", err)
+				return
+			}
+			if diff := cmp.Diff(resp.StatusCode, tc.wantStatusCode); diff != "" {
+				t.Error("unexpected probe check result (-want, +got) = ", diff)
+			}
+			cancel()
```
yolocs, line 92: Prefer `defer cancel()`?

pkg/utils/authcheck/probechecker_test.go
```diff
@@ -0,0 +1,114 @@
...
+			}
+
+			client := http.DefaultClient
+			resp, err := client.Do(req)
+
+			if err != nil {
+				t.Fatal("Failed to execute probe check:", err)
+				return
+			}
+			if diff := cmp.Diff(resp.StatusCode, tc.wantStatusCode); diff != "" {
```
yolocs, line 89: want first, got second. Also in this case you probably don't need `cmp.Diff`.

pkg/utils/authcheck/fake.go
```diff
@@ -0,0 +1,43 @@
...
+	authType AuthType
+	noError  bool
+}
+
+func NewFakeAuthenticationCheck(authType AuthType, noError bool) AuthenticationCheck {
+	return &FakeAuthenticationCheck{
+		authType: authType,
+		noError:  noError,
+	}
+}
```
yolocs, line 36: You don't have to expose such a func; instead you can expose `authType` and `noError`.

pkg/utils/authcheck/authcheck_test.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, tc := range testCases {
+		tc := tc
+		t.Run(tc.name, func(t *testing.T) {
+			t.Parallel()
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+
+			if tc.setEnv {
+				os.Setenv("GOOGLE_APPLICATION_CREDENTIALS", "empty")
+			}
```
yolocs, line 71: Will this cause any problem with `t.Parallel`? `Setenv` won't be test local right?

grac3gao: I can add a defer unsetenv, do you think that would be helpful?

yolocs (commented):
/lgtm
/hold for others to take a look

pkg/reconciler/brokercell/brokercell.go
```diff
@@ -175,7 +175,17 @@ func (r *Reconciler) ReconcileKind(ctx context.Context, bc *intv1alpha1.BrokerCe
 		bc.Status.MarkIngressFailed("IngressServiceFailed", "Failed to reconcile ingress service: %v", err)
 		return err
 	}
-	bc.Status.PropagateIngressAvailability(endpoints)
+	// If deployment has replicaUnavailable error, it potentially has authentication configuration issues.
+	if replicaAvailable := bc.Status.PropagateIngressAvailability(endpoints, ind); !replicaAvailable {
+		podList, err := authcheck.GetPodList(ctx, resources.GetLabelSelector(bc.Name, resources.IngressName), r.KubeClientSet, bc.Namespace)
+		if err != nil {
+			logging.FromContext(ctx).Error("Failed to propagate authentication check message from ingress component", zap.Any("namespace", bc.Namespace), zap.Any("name", bc.Name), zap.Error(err))
```
Harwayne, line 182:
```suggestion
			logging.FromContext(ctx).Error("Failed to propagate authentication check message from ingress component", zap.Error(err))
```

The name and namespace of the thing being reconciled are already in the logger, in the `knative.dev/key` field.

pkg/utils/authcheck/probechecker_test.go
```diff
@@ -0,0 +1,92 @@
...
+			addr, err := net.ResolveTCPAddr("tcp", "localhost:0")
+			if err != nil {
+				t.Fatal("Failed to resolve TCP address:", err)
+			}
+			l, err := net.ListenTCP("tcp", addr)
+			if err != nil {
+				t.Fatal("Failed to listen TCP:", err)
+			}
+			l.Close()
+			port := l.Addr().(*net.TCPAddr).Port
```
Harwayne, line 63: It might be worth extracting this into a method, something like `getFreePort()`.

pkg/utils/authcheck/fake.go
```diff
@@ -0,0 +1,35 @@
...
+// File authcheck contains functions to run customized checks inside of a Pod.
+package authcheck
+
+import (
+	"context"
+	"errors"
+)
+
+type FakeAuthenticationCheck struct {
+	NoError bool
```
Harwayne, line 27: I would recommend

```suggestion
	Err error
```

Then in `Check`, you can simply call `return ac.Err`.

pkg/utils/authcheck/authtype.go
```diff
@@ -114,7 +125,12 @@ func getAuthTypeForWorkloadIdentity(ctx context.Context, serviceAccountLister co
 				args.ServiceAccountName)
 		}
 		return "", fmt.Errorf("error getting Kubernetes Service Account: %w", err)
-	} else if kServiceAccount.Annotations[resources.WorkloadIdentityKey] != "" {
+	} else if email := kServiceAccount.Annotations[resources.WorkloadIdentityKey]; email != "" {
+		// Check if email is a valid google service account email.
+		if match := emailRegexp.FindStringSubmatch(email); len(match) == 0 {
```
Harwayne, line 130: Do we ever expect multiple emails in this? My guess is that we don't need Submatch.

```suggestion
		if match := emailRegexp.FindString(email); match == "" {
```

pkg/utils/authcheck/authcheck.go
```diff
@@ -0,0 +1,132 @@
...
+	scope = "https://www.googleapis.com/auth/cloud-platform"
+	// authMessage is the key words to determine if a termination log is about authentication.
+	authMessage = "checking authentication"
+)
+
+type AuthenticationCheck interface {
+	Check(ctx context.Context) error
+}
+
+type DefaultAuthenticationCheck struct {
```
Harwayne, line 47: Can this be `defaultAuthenticationCheck` instead?

pkg/utils/authcheck/authcheck.go
```diff
@@ -0,0 +1,132 @@
...
+	req.Header.Set("Metadata-Flavor", "Google")
+	resp, err := client.Do(req)
+	if err != nil {
+		return fmt.Errorf("error getting the http response: %w", err)
+	}
+	defer resp.Body.Close()
+	// Check if we can successfully get the token.
+	if resp.StatusCode < http.StatusOK || resp.StatusCode >= http.StatusBadRequest {
+		return errors.New("the Pod is not fully authenticated, " +
+			"probably due to corresponding k8s service account and google service account do not establish a correct relationship")
```
Harwayne, line 113: So that we have some pointer if we need to debug, please add the response code into the error.

```suggestion
		return fmt.Errorf("the Pod is not fully authenticated, " +
			"probably due to corresponding k8s service account and google service account do not establish a correct relationship: %d", resp.StatusCode)
```

pkg/utils/authcheck/probechecker_test.go
```diff
@@ -30,17 +31,17 @@ import (
...
 		{
 			name:           "probe check got a failure result",
-			noError:        false,
+			err:            errors.New("induced error"),
 			wantStatusCode: http.StatusInternalServerError,
 		},
 		{
 			name:           "probe check got a success result",
-			noError:        true,
+			err:            nil,
```
Harwayne, line 44: If you want, you could also omit this line.

pkg/utils/authcheck/probechecker_test.go
```diff
@@ -90,3 +85,16 @@ func TestProbeCheckResult(t *testing.T) {
 		})
 	}
 }
+
+func getFreePort() (int, error) {
```
Harwayne, line 89: Pass in `t` and have it call `t.Fatal` inside the function.

```suggestion
func getFreePort(t *testing.T) int {
```

Harwayne (commented):
/lgtm

Harwayne:
/unhold

I think only @yolocs and myself were commenting.

PR #1986, opened 2020-12-08
[WIP] Inject boilerplate into wire-generated files

tommyreddad:
Without this change, wire-generated files fail the boilerplate check.

The packages to generate are explicitly passed to wire since otherwise the generator creates wire_gen files in every package directory.

## Proposed Changes
- Inject boilerplate into wire-generated files


hack/update-codegen.sh
```diff
@@ -90,7 +90,15 @@ EXTERNAL_INFORMER_PKG="k8s.io/client-go/informers" \
...
-go generate "${REPO_ROOT_DIR}"/...
+"${GOPATH}"/bin/wire gen \
+  --header_file="${REPO_ROOT_DIR}"/hack/boilerplate/boilerplate.go.txt \
+  "${REPO_ROOT_DIR}"/cmd/broker/fanout \
+  "${REPO_ROOT_DIR}"/cmd/broker/ingress \
+  "${REPO_ROOT_DIR}"/cmd/broker/retry \
+  "${REPO_ROOT_DIR}"/cmd/controller \
+  "${REPO_ROOT_DIR}"/cmd/pubsub/publisher \
+  "${REPO_ROOT_DIR}"/cmd/pubsub/receive_adapter \
+  "${REPO_ROOT_DIR}"/pkg/broker/handler
```
zhongduo, line 101: Do we have to list all instead of `...` now?

PR #1993, opened 2020-12-14
Avoid overwriting eventing clusterrole

### LINKED ISSUE #1824: ClusterRole of  knative-eventing and knative-gcp has same name
**Describe the bug**
For example, in knative-gcp, we have clusterrole called [addressable-resolver](https://github.com/google/knative-gcp/blob/e048a4ff559a26871b287632eeeceddfdf1490b9/config/core/roles/addressable-resolvers-clusterrole.yaml#L20) ,
and in knative-eventing, we also have clusterrole called  [addressable-resolver](https://github.com/knative/eventing/blob/7d981f818ef045898581d31578226e795e26c3f6/config/core/roles/addressable-resolvers-clusterrole.yaml#L19)
Those two cluster roles have different labels. But when we install eventing and then install knative-gcp, the clusterrole in knative-gcp will overrride the clusterrole with the same name that got installed when we install knative-eventing 

 **Expected behavior**
Avoid using clusterroles with same name in knative-eventing

 **To Reproduce**

 **Additional context**

cathyzhyi:
Fixes #1824 

## Proposed Changes

- add `events-system-` prefix to the clusterrole name

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/config/core/roles/addressable-resolvers-clusterrole.yaml
+++ b/config/core/roles/addressable-resolvers-clusterrole.yaml
@@ -17,7 +17,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: addressable-resolver
+  name: googlecloud-addressable-resolver
   labels:
     events.cloud.google.com/release: devel
 aggregationRule:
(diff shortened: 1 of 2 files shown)


AlexandraRoatis:
I think the prefix should be `events-system-`, since that is what we use to replace `cloud-run-events-` in #1917 .

yolocs (approved):
/lgtm

AlexandraRoatis (approved):

PR #1997, opened 2020-12-15
feat/dataresidency: use cluster region as default setting

### LINKED ISSUE #1989: Use cluster region as the default constraint for data residency
**Problem**
As per data residency requirements, the default data residency of a GKE application should be the cluster region instead of global.

Developers and customers.

**Exit Criteria**
In case of no data residency configuration, make it the same as the cluster region.

**Time Estimate (optional):** 
2 developer-days

zhongduo:
Fixes #1989 

## Proposed Changes

- Default to cluster region when the allowed region is nil or empty
- Use a new field `global` to default to no setting.

- 🎁  use cluster region as default data residency constraints
**Release Note**

```release-note
- 🎁  use cluster region as default data residency constraints
```

**Docs**

### DIFF
--- a/config/core/configmaps/data-residency.yaml
+++ b/config/core/configmaps/data-residency.yaml
@@ -18,7 +18,7 @@ metadata:
   name: config-dataresidency
   namespace: cloud-run-events
   annotations:
-    knative.dev/example-checksum: "16b18753"
+    knative.dev/example-checksum: "3c0011c5"
 data:
   default-dataresidency-config: |
     clusterDefaults:
@@ -48,7 +48,10 @@ data:
       # clusterDefaults are the defaults to apply to every namespace in the
       # cluster
       clusterDefaults:
-        # messagestoragepolicy.global will set to the Org Policy
+        # messagestoragepolicy.global determines whether Global PubSub Topics are allowed.
+        # If set to false, then the PubSub Topic will be regional, based on the region the
+        # GKE cluster is running in. If set to true, then the PubSub Topic is allowed to be
+        # Global, as long as messagestoragepolicy.allowedpersistenceregions is empty and no GCP Org Policy forbids it.
         messagestoragepolicy.global: false
         # messagestoragepolicy.allowedpersistenceregions field specifies
         # all the allowed regions for data residency. The default or an empty value will
--- a/config/core/configmaps/data-residency.yaml
+++ b/config/core/configmaps/data-residency.yaml
@@ -18,10 +18,11 @@ metadata:
   name: config-dataresidency
   namespace: cloud-run-events
   annotations:
-    knative.dev/example-checksum: "809e8f92"
+    knative.dev/example-checksum: "f9aa8a13"
 data:
   default-dataresidency-config: |
     clusterDefaults:
+      messagestoragepolicy.global: false
       messagestoragepolicy.allowedpersistenceregions: []
   _example: |
     ################################
@@ -40,14 +41,16 @@ data:
     # to actually change the configuration.
 
     # default-dataresidency-config is the configuration for determining the default
-    # data residency to apply to all objects that require data residency.
-    # This is expected to be Channels and Sources and Brokers.
+    # data residency to apply to all objects that require data residency but,
+    # do not specify it. This is expected to be Channels and Sources.
     #
     # We only support cluster scoped now
     default-dataresidency-config: |
       # clusterDefaults are the defaults to apply to every namespace in the
       # cluster
       clusterDefaults:
+        # messagestoragepolicy.global will set to the Org Policy
+        messagestoragepolicy.global: false
         # messagestoragepolicy.allowedpersistenceregions field specifies
         # all the allowed regions for data residency. The default or an empty value will
         # mean no data residency requirement.
--- a/config/core/configmaps/data-residency.yaml
+++ b/config/core/configmaps/data-residency.yaml
@@ -18,11 +18,10 @@ metadata:
   name: config-dataresidency
   namespace: cloud-run-events
   annotations:
-    knative.dev/example-checksum: "f9aa8a13"
+    knative.dev/example-checksum: "16b18753"
 data:
   default-dataresidency-config: |
     clusterDefaults:
-      messagestoragepolicy.global: false
       messagestoragepolicy.allowedpersistenceregions: []
   _example: |
     ################################
@@ -41,8 +40,8 @@ data:
     # to actually change the configuration.
 
     # default-dataresidency-config is the configuration for determining the default
-    # data residency to apply to all objects that require data residency but,
-    # do not specify it. This is expected to be Channels and Sources.
+    # data residency to apply to all objects that require data residency.
+    # This is expected to be Channels and Sources and Brokers.
     #
     # We only support cluster scoped now
     default-dataresidency-config: |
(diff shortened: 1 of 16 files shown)


pkg/apis/configs/dataresidency/defaults.go
```diff
@@ -48,18 +53,35 @@ func (d *Defaults) AllowedPersistenceRegions() []string {
...
+}
+
 // ComputeAllowedPersistenceRegions computes the final message storage policy in
 // topicConfig. Return true if the topicConfig is updated.
-func (d *Defaults) ComputeAllowedPersistenceRegions(topicConfig *pubsub.TopicConfig) bool {
+func (d *Defaults) ComputeAllowedPersistenceRegions(topicConfig *pubsub.TopicConfig, clusterRegionGetter utils.ClusterRegionGetter) (bool, error) {
+	if topicConfig.MessageStoragePolicy.AllowedPersistenceRegions != nil {
+		// Don't try to change anything if it is not empty
+		return false, fmt.Errorf("topic config is unexpectedly set, please check the logic again")
+	}
```
Harwayne, line 67: Is this expected to happen if I

1. Set `allowedpersistenceregions`.
1. Create a Broker/Source.
1. Change my `allowedpersistenceregions`

?

pkg/reconciler/trigger/trigger.go
```diff
@@ -62,6 +62,8 @@ var (
 	// defaulting fails.
 	defaultBackoffDelay  = "PT1S"
 	defaultBackoffPolicy = eventingduckv1beta1.BackoffPolicyExponential
+	// clusterRegionGetter is a function that can get the cluster region
+	clusterRegionGetter = utils.NewClusterRegionGetter()
```
Harwayne, line 66: I suggest injecting this into the reconciler. For this and the others. Have Wire generate it and push it into the controller creation.

pkg/utils/metadata.go
```diff
@@ -75,3 +77,36 @@ func ClusterName(clusterName string, client metadataClient.Client) (string, erro
...
+	return ZoneToRegion(zone)
+}
+
+// ClusterRegionGetter is a handler that gets cluster region
+type ClusterRegionGetter func() (string, error)
+
+// NewClusterRegionGetter returns CluterRegionGetter in production code
+func NewClusterRegionGetter() ClusterRegionGetter {
+	return func() (string, error) {
+		return ClusterRegion("", defaultMetadataClientCreator)
```
Harwayne, line 110: This isn't expected to change for the lifetime of the process, so I recommend using a `sync.Once` to get the value once, cache it, and return that value from then on without calling the metadata server.

pkg/apis/configs/dataresidency/defaults.go
zhongduo: In this case, we don't update existing ones in 1.

config/core/configmaps/data-residency.yaml
```diff
@@ -48,6 +48,8 @@ data:
       # clusterDefaults are the defaults to apply to every namespace in the
       # cluster
       clusterDefaults:
+        # messagestoragepolicy.global will set to the Org Policy
```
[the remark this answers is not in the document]
zhongduo: So we want to have: 
1.  Allow all regions, default to global or Org Policy. In code the topic config filed will be nil.
2. No user setting, default to cluster region.
3. Specified regions.

Our discussion result was for 1 using user input `global`, for 2 use `[]`, for 3 use non empty setting.  so global is actually a kind of user input to avoid typing out all possible fields. We can potentially use nil as global, but think that this will be confusing.

pkg/apis/configs/dataresidency/defaults.go
Harwayne: I'm not sure I understand. I know that if the user changes the `allowedpersistencergions`, we _don't_ move anything that already exists. I think this is what your comment means, is that correct?

My overall question is: what happens to the Broker/Source that is using the older location? Does it continue to reconcile correctly? Or does it start failing reconciliation due to this error? My expectation is that it would continue to reconcile correctly, using its original location, probably with a warning in its status saying that it is using an older location (maybe a K8s event too?).

zhongduo: Your understanding about my comment is correct. Since we don't reconcile existing topics, the only way we invoke this is when we are creating topics, so this is to avoid changing anything here in case in the future we somehow allow other configuration before getting the default. Eg. if we have a data residency field in the broker spec. The errors in this function are treated as a warning in the callers. So there will never be failure because of this. The original reasoning is that resolving final AllowedPersistenceRegions is best effort, but I guess maybe we should fail in case of unable to get the cluster region. How about 1) checking the error and fail the topic creation, and 2) change this to no error?

pkg/utils/metadata.go
zhongduo: sync.Once might not work due to possible error condition, how about the new caching way?

pkg/reconciler/broker/broker.go
```diff
@@ -141,11 +143,12 @@ func (r *Reconciler) reconcileDecouplingTopicAndSubscription(ctx context.Context
 	// Check if topic exists, and if not, create it.
 	topicID := resources.GenerateDecouplingTopicName(b)
 	topicConfig := &pubsub.TopicConfig{Labels: labels}
+	fmt.Printf("Calling compute\n")
```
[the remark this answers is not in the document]
zhongduo: Fixed, thanks for catching this.

pkg/reconciler/trigger/trigger.go
zhongduo: This function is not really configurable in production. I am using it this way simply to overwrite it during testing.

zhongduo: Another concern I originally had to do it this way is to avoid failing the controller when it cannot get the cluster region, as it is not required if there are valid setting already. But I think if we don't have to worry about this, we can just have a `clusterRegion string` in the Reconciler and let the controller initiate it properly. WDYT?

pkg/apis/configs/dataresidency/defaults.go
```diff
@@ -48,18 +53,35 @@ func (d *Defaults) AllowedPersistenceRegions() []string {
...
+	if d.Global() {
+		// Not setting means same as Org Policy
+		return false, nil
+	}
```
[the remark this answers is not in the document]
zhongduo: Changed as suggested, now if allowedpersistenceregions is set to non empty, global will be ignored.

config/core/configmaps/data-residency.yaml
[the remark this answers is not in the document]
zhongduo: Fixed as suggested.

Harwayne (commented):
/lgtm
/hold

Holding just to make you answer this question (which you've already answered, but I want to be very sure before we submit):

Does this correctly handle Brokers/Sources that were created before this change? E.g. I made a Broker/Source in 0.19. Now I upgrade and this code is now in the reconciler. Will my existing Broker/Source continue to use the old Topic? And will it reconcile as Ready=True (assuming no other failures)?

zhongduo:
> /lgtm
> /hold
> 
> Holding just to make you answer this question (which you've already answered, but I want to be very sure before we submit):
> 
> Does this correctly handle Brokers/Sources that were created before this change? E.g. I made a Broker/Source in 0.19. Now I upgrade and this code is now in the reconciler. Will my existing Broker/Source continue to use the old Topic? And will it reconcile as Ready=True (assuming no other failures)?

Existing topic will never be modified at all. All the topic confi in Broker/Trigger/Topic reconciler will only be used when the topic does not exist. So during upgrade, the behavior should be exactly the same as before.

Harwayne:
> Existing topic will never be modified at all. All the topic confi in Broker/Trigger/Topic reconciler will only be used when the topic does not exist. So during upgrade, the behavior should be exactly the same as before.

/unhold

PR #1999, opened 2020-12-15
feat/perf-test: Add google-broker-key secret to perf test

zhongduo:
The performance test is currently not working due to lack of
google-broker-key secret. This commit tries to solve it by
introducing the broker key.

Fixes #

## Proposed Changes

- Add google-broker-key in performance test setup.
-
-


grac3gao (approved):
/lgtm
/approve

PR #2008, opened 2020-12-17
Add enableServiceLinks:false to webhook yaml

### LINKED ISSUE #2007: The default webhook port does not work
**Describe the bug**

Kubernets adds additional env vars for each service with name
*_SERVICE_PORT, *_SERVICE_HOST and *_PORT, this causes trouble
in the webhook as it is using WEBHOOK_PORT to configure the
webhook port. Disabling it will enable using a default port.

 **Expected behavior**
Without WEBHOOK_PORT env the default port should be used.
 **To Reproduce**
remove the WEBHOOK_PORT env var in webhook.yaml and apply

zhongduo:
Kubernets adds additional env vars for each service with name
*_SERVICE_PORT, *_SERVICE_HOST and *_PORT, this causes trouble
in the webhook as it is using WEBHOOK_PORT to configure the
webhook port. Disabling it will enable using a default port.

Fixes #2007

### DIFF
--- a/config/core/deployments/webhook.yaml
+++ b/config/core/deployments/webhook.yaml
@@ -35,6 +35,7 @@ spec:
         sidecar.istio.io/inject: "false"
     spec:
       serviceAccountName: webhook
+      enableServiceLinks: false
       containers:
         - name: webhook
           terminationMessagePolicy: FallbackToLogsOnError


yolocs (commented):
Fix the PR title

yolocs (approved):
/lgtm

PR #2014, opened 2020-12-22
Authentication check: additional k8s Event check for Secret absence in source

grac3gao:
Fixes #

PRs to add the authentication check:
1. Environment variable to differentiate authentication mode
2. Authentication check running inside the Pod (left part:  Pods using kncloudevent library like Ingress)
3. **Additional k8s Event check for Secret absence in source**(<-this PR)

## Proposed Changes
- If deployment has minimal replica unavailable error and it is secret mode, check k8s event to see if it is caused by secret absence.
- Remark upper-level source with authentication related Reason and Message.

**Release Note**

```release-note
🎁 Authentication check: additional k8s Event check for Secret absence
```

### DIFF
--- a/pkg/reconciler/intevents/pullsubscription/static/pullsubscription.go
+++ b/pkg/reconciler/intevents/pullsubscription/static/pullsubscription.go
@@ -68,11 +68,26 @@ func (r *Reconciler) ReconcileDeployment(ctx context.Context, ra *appsv1.Deploym
 			logging.FromContext(ctx).Error("Error propagating authentication check message", zap.Error(err))
 			return err
 		}
+		// If source is in secret mode, we need to check if the secret is absent.
+		if src.Spec.Secret != nil {
+			logging.FromContext(ctx).Error("Start to get pod list", zap.Error(err))
+			for _, pod := range podList.Items {
+				eventList, err := authcheck.GetEventList(ctx, r.KubeClientSet, pod.Name, src.Namespace)
+				logging.FromContext(ctx).Error("Start to get event list", zap.Error(err))
+				if err != nil {
+					logging.FromContext(ctx).Error("Error propagating authentication check message", zap.Error(err))
+					return err
+				}
+				if authenticationCheckMessage := authcheck.GetMountFailureMessageFromEventList(eventList, src.Spec.Secret); authenticationCheckMessage != "" {
+					src.Status.MarkDeployedUnknown(authcheck.AuthenticationCheckUnknownReason, authenticationCheckMessage)
+					return nil
+				}
+			}
+		}
 		if authenticationCheckMessage := authcheck.GetTerminationLogFromPodList(podList); authenticationCheckMessage != "" {
 			src.Status.MarkDeployedUnknown(authcheck.AuthenticationCheckUnknownReason, authenticationCheckMessage)
 		}
 	}
-
 	return nil
 }
 
--- a/pkg/reconciler/intevents/pullsubscription/static/pullsubscription.go
+++ b/pkg/reconciler/intevents/pullsubscription/static/pullsubscription.go
@@ -70,10 +70,8 @@ func (r *Reconciler) ReconcileDeployment(ctx context.Context, ra *appsv1.Deploym
 		}
 		// If source is in secret mode, we need to check if the secret is absent.
 		if src.Spec.Secret != nil {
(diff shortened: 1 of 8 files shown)


pkg/reconciler/intevents/pullsubscription/static/pullsubscription.go
```diff
@@ -68,11 +68,26 @@ func (r *Reconciler) ReconcileDeployment(ctx context.Context, ra *appsv1.Deploym
 			logging.FromContext(ctx).Error("Error propagating authentication check message", zap.Error(err))
 			return err
 		}
+		// If source is in secret mode, we need to check if the secret is absent.
+		if src.Spec.Secret != nil {
+			logging.FromContext(ctx).Error("Start to get pod list", zap.Error(err))
+			for _, pod := range podList.Items {
+				eventList, err := authcheck.GetEventList(ctx, r.KubeClientSet, pod.Name, src.Namespace)
+				logging.FromContext(ctx).Error("Start to get event list", zap.Error(err))
```
Harwayne, line 76: Is this meant to be checked in?

pkg/utils/authcheck/list.go
```diff
@@ -65,3 +94,17 @@ func GetTerminationLogFromPodList(pl *corev1.PodList) string {
...
 	return strings.Contains(message, authMessage)
 }
+
+// isWarningMessage checks if the message is for a specific secret's failure.
+func isWarningMessage(message, namespace string, secret *corev1.SecretKeySelector) bool {
+	return strings.Contains(message, fmt.Sprintf(`MountVolume.SetUp failed for volume "%s"`, secret.Name)) ||
+		strings.Contains(message, fmt.Sprintf(`couldn't find key %s in Secret %s/%s`, secret.Key, namespace, secret.Name))
+}
+
+func getFieldSelector(pod string) fields.Selector {
```
Harwayne, line 104:
```suggestion
func podWarningFieldSelector(pod string) fields.Selector {
```

pkg/utils/authcheck/list.go
```diff
@@ -65,3 +94,17 @@ func GetTerminationLogFromPodList(pl *corev1.PodList) string {
 func isAuthMessage(message string) bool {
 	return strings.Contains(message, authMessage)
 }
+
+// isWarningMessage checks if the message is for a specific secret's failure.
+func isWarningMessage(message, namespace string, secret *corev1.SecretKeySelector) bool {
```
Harwayne, line 99:
```suggestion
func isSecretMountFailureMessage(message, namespace string, secret *corev1.SecretKeySelector) bool {
```

pkg/reconciler/intevents/pullsubscription/static/pullsubscription.go
AlexandraRoatis: I think the same comment applies to line 73 as well.

pkg/reconciler/intevents/pullsubscription/static/pullsubscription_test.go
```diff
@@ -24,14 +24,13 @@ import (
 
 	"cloud.google.com/go/pubsub"
 	"cloud.google.com/go/pubsub/pstest"
+	"google.golang.org/grpc/codes"
```
AlexandraRoatis, line 27: This may be an automated change, but the import seems to fit better where it was previously.

pkg/reconciler/intevents/pullsubscription/static/pullsubscription.go
grac3gao: My bad, forgot to remove the debug lines after manual test...

pkg/utils/authcheck/list.go
grac3gao: This function checks message for two cases: key is absent, and secret is absent. The k8s Event Reason for those two case are different. For key is absent, the reason is just `Failed`, for secret is absent, the reason is `FailedMount`. So I think case` key is absent` is not strictly a secretMountFailure case. If you don't have mind, I'll change it to isSecretFailureMessage.

Harwayne: SGTM. I found the `isWarningMessage` name to be misleading, `isSecretFailureMessage` is much clearer.

Harwayne (commented):
/lgtm
/hold

Holding in case @AlexandraRoatis has any additional comments.

PR #2016, opened 2020-12-22
feat/trigger: enable trigger to send to different namespace

### LINKED ISSUE #2015: Trigger unable to deliver to different namespace
**Describe the bug**
Currently the trigger always set the sink namespace to its own namespace, so there is no way to deliver to sink in other namespace. Note that according to the spec, the destination can be a k8s service, which is cluster wide.

 **Expected behavior**
When the namespace is set in subscriber spec, it should be respected.

 **To Reproduce**
Try adding a namespace to the subscriber in trigger and check the resolved URI.

zhongduo:
After this change, when the namepsace is specified in the subscriber
the URI can be resolved correctly to the specified namespace instead
of forcing it to the be same as the trigger namespace.

Fixes #2015 

## Proposed Changes

- Don't over the namespace of a trigger

- 🎁 Allow triggers to send to a sink in different namespace

**Release Note**

```release-note
- 🎁 Allow triggers to send to a sink in different namespace
```

**Docs**

### DIFF
--- a/pkg/reconciler/trigger/trigger.go
+++ b/pkg/reconciler/trigger/trigger.go
@@ -180,7 +180,7 @@ func (r *Reconciler) FinalizeKind(ctx context.Context, t *brokerv1beta1.Trigger)
 }
 
 func (r *Reconciler) resolveSubscriber(ctx context.Context, t *brokerv1beta1.Trigger, b *brokerv1beta1.Broker) error {
-	if t.Spec.Subscriber.Ref != nil {
+	if t.Spec.Subscriber.Ref != nil && t.Spec.Subscriber.Ref.Namespace == "" {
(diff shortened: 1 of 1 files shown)


liu-cong:
/lgtm

BTW, can you test this in a cluster? Just to confirm this isn't something else that blocks cross-namespace. I remember there used to be some logic in eventing webhook.

zhongduo:
> /lgtm
> 
> BTW, can you test this in a cluster? Just to confirm this isn't something else that blocks cross-namespace. I remember there used to be some logic in eventing webhook.

I am actively testing it now with my experiment to see if cross namespace delivery is possible. After this change, I can now send from curl pod in default namespace to broker/trigger in cloud-run-events-example namespace, then got routed back to event-display default namespace.

PR #2021, opened 2020-12-23
Remove redundant tag keys from delivery metrics

cathyzhyi:
## Proposed Changes

- Remove metric labels that should not be present. They are not in the metric's descriptors, therefore are unexpected.
  - Current StackDriver exporter ignores them, but the otel Collector doesn't. 

**Release Note**

```release-note
NONE
```


Harwayne (commented):
/lgtm

PR #2025, opened 2020-12-28
Webhook has only namespaced Secret permissions, rather than cluster-w…

Harwayne:
## Proposed Changes

- Webhook has only namespaced Secret permissions, rather than cluster-wide.

**Release Note**

```release-note
Webhook no longer has Cluster-wide Secret permissions. Now the Webhook's Secret permissions are restricted to the `cloud-run-events` namespace.
```

### DIFF
--- a/config/201-webhook-role.yaml
+++ b/config/201-webhook-role.yaml
@@ -1,0 +1,1 @@
+core/roles/webhook-role.yaml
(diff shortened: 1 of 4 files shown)


danyinggu (commented):
/lgtm

PR #2026, opened 2020-12-29
Remove unneeded method

Harwayne:
## Proposed Changes

- Rename String() to DebugString(), make explicit that this is for debug purposes only. 
- Remove EqualsString(), which was only used in tests.
- Alter the string format to be more human readable.

**Release Note**

```release-note
NONE
```


zhongduo:
This looks cleaner to me.
/lgtm

PR #2027, opened 2020-12-29
Remove GetBroker, GetTarget, and EqualsBytes

Harwayne:
## Proposed Changes

- Remove `GetBroker` and `GetTarget`, in favor of their `Get{Broker,Target}ByKey` equivalents.
    - The non-keyed versions were used primarily in tests, not real code.
- Remove 'EqualsBytes`, it was only used in tests.

**Release Note**

```release-note
NONE
```


danyinggu (commented):
/lgtm

PR #2028, opened 2020-12-29
Replace string with {Broker,Target}Key structs in the Targets interfaces

Harwayne:
## Proposed Changes

- Replace string with {Broker,Target}Key structs in the Targets interfaces.

**Release Note**

```release-note
NONE
```


pkg/broker/config/key.go
```diff
@@ -0,0 +1,104 @@
...
+// for the portions known about by the BrokerKey.
+func (k *BrokerKey) CreateEmptyBroker() *Broker {
+	return &Broker{
+		Namespace: k.namespace,
+		Name:      k.name,
+	}
+}
+
+// Key returns the BrokerKey for this Broker.
+func (x *Broker) Key() *BrokerKey {
```
tayarani, line 54: x? I always thought we use b for broker.

pkg/broker/config/key.go
```diff
@@ -0,0 +1,104 @@
...
+// String creates a human readable version of this key. It is for debug purposes only. It is free to
+// change at any time.
+func (k *TargetKey) String() string {
+	// Note that this is explicitly different than the PersistenceString, so that we don't
+	// accidentally use String(), rather than PersistenceString().
+	return k.brokerKey.String() + "//" + k.name
+}
+
+// Key returns the TargetKey for this Target.
+func (x *Target) Key() *TargetKey {
```
tayarani, line 80: t for target?

pkg/broker/config/key.go
```diff
@@ -0,0 +1,104 @@
...
+		brokerKey: BrokerKey{
+			namespace: x.Namespace,
+			name:      x.Broker,
+		},
+		name: x.Name,
+	}
+}
+
+// KeyFromBroker creates a BrokerKey from a K8s Broker object.
+func KeyFromBroker(b *brokerv1beta1.Broker) *BrokerKey {
```
tayarani, line 91: in this file this is the only place we use "brokerv1beta1.Broker", whereas in other places we use "Broker". Is that intentional?

pkg/broker/config/key.go
```diff
@@ -0,0 +1,104 @@
...
+	return &BrokerKey{
+		namespace: b.Namespace,
+		name:      b.Name,
+	}
+}
+
+// TestOnlyBrokerKey returns the key of a broker.
+func TestOnlyBrokerKey(namespace, name string) *BrokerKey {
```
tayarani, line 99: I'm not sure I understand why this method is needed.

pkg/broker/config/key.go
Harwayne: `Broker` is a struct defined by the protobuf. As far as I can tell, the proto compiler always uses `x`. So to match the existing methods (in targets.pb.go), I am also using `x`.

pkg/broker/config/key.go
Harwayne: Same reasoning as `x` for Broker.

pkg/broker/config/key.go
Harwayne: Yes, this is intentional. 

In this package `Broker` is the protobuf defined struct. That is almost always what we want to use. This method is a helper that allows code to go from the K8s Broker object to a BrokerKey. 

Would the name `KeyFromK8sBroker` make this clearer?

pkg/broker/config/key.go
Harwayne: Added more to the function's documentation. Does that make it clearer?

pkg/broker/config/key.go
tayarani: Sounds good.

pkg/broker/config/key.go
tayarani: The comment makes things very clear. Thanks.

pkg/broker/config/key.go
tayarani: The comment makes it better. The function name doesn't have to change.

PR #2038, opened 2021-01-05
Fix probe helper client listener and logging

tommyreddad:
The probe helper is facing issues where there is some mixup between the forward client and receiver client. The relevant options are mismatched currently.

## Proposed Changes
- Use the correct options for the forward client and receiver client.
- Improved logging to reduce clutter and be more informative.
- ACK received events even if they generate an error.

### DIFF
--- a/test/test_images/probe_helper/probe/handlers/cloudpubsub.go
+++ b/test/test_images/probe_helper/probe/handlers/cloudpubsub.go
@@ -25,6 +25,8 @@ import (
 	cecontext "github.com/cloudevents/sdk-go/v2/context"
 	schemasv1 "github.com/google/knative-gcp/pkg/schemas/v1"
 	"github.com/google/knative-gcp/test/test_images/probe_helper/utils"
+	"go.uber.org/zap"
+	"knative.dev/pkg/logging"
 )
 
 const (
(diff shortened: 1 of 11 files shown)


test/test_images/probe_helper/probe/handlers/cloudpubsub.go
```diff
@@ -70,6 +72,7 @@ func (p *CloudPubSubSourceProbe) Forward(ctx context.Context, event cloudevents.
 		return fmt.Errorf("CloudPubSubSource probe event has no '%s' extension", topicExtension)
 	}
 	ctx = cecontext.WithTopic(ctx, fmt.Sprint(topic))
+	logging.FromContext(ctx).Infow("Publishing message to pubsub topic", zap.String("topic", fmt.Sprint(topic)))
```
zhongduo, line 75: I don't think you need `Sprint`, did you try not having it? Worst case you can do `topic.String()` if that does not work.

eclipselu:
Apart from the nits that Jimmy brought up.
/lgtm

eclipselu:
oops sorry for the inconvenience. forgot to hold the PR.

test/test_images/probe_helper/probe/handlers/cloudpubsub.go
tommyreddad: `topic` is of type `interface{}`, it does not implement the `String` method. So `fmt.Sprint` just performs that cast. But I should probably have used `topic.(string)` instead, I will keep this in mind in the future.

PR #2046, opened 2021-01-07
feat/upgrade-test: Use eventing prober instead of our modified version

### LINKED ISSUE #2050: Reuse the eventing prober code in upgrade test
**Problem**
We are maintaining a copy of the prober code from knative eventing to get the upgrade test ready ASAP. This will require constant syncing with knative eventing when there is change upstream. Ideally we should contribute to the OOS version to allow customized configuration we need.

Developers.

**Exit Criteria**
Not `test/upgrade/prober` in our repo.

**Time Estimate (optional):** 
1 day.

zhongduo:
After this commit, the eventing prober will be used directly, so any
changes upstream will not likely to requrie modification here any more.

Fix #2050


vendor/knative.dev/eventing/test/upgrade/prober/configuration.go
```diff
@@ -78,6 +77,8 @@ type Wathola struct {
 
 // ConfigMap represents options of wathola config toml file.
 type ConfigMap struct {
+	// ConfigTemplate is a template file that will be compiled to the configmap
```
eclipselu, line 80:
```suggestion
	// ConfigTemplate is the name of a template file that will be compiled into the configmap
```

test/upgrade/continual.go
```diff
@@ -38,6 +40,9 @@ func ContinualTest() pkgupgrade.BackgroundOperation {
 			// overwrite configuration
 			config.FailOnErrors = true
 			config.Interval = 10 * time.Millisecond
+			config.BrokerOpts = append(config.BrokerOpts, resources.WithBrokerClassForBrokerV1Beta1(v1beta1.BrokerClass))
+			config.FinishedSleep = 40 * time.Second
+			config.ConfigTemplate = "../../../../../../test/upgrade/config.toml"
```
eclipselu, line 45: maybe we can just use `../config.toml`

zhongduo: No, it is relative to the `eventing/test` directory in vendor directory. This is in `knaitve/test`.

PR #2048, opened 2021-01-07
Documentation for migrating workloads to the new namespace

AlexandraRoatis:
## Proposed Changes

* Documentation for migrating workloads to the new namespace. In the current state we assume the use is based on `knative-gcp` and Kubernetes secret-based authentication.

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/docs/how-to/namespace-transition.md
+++ b/docs/how-to/namespace-transition.md
@@ -1,4 +1,10 @@
-# Transition a project to the new `events-system` namespace
+# Transition a cluster to the new `events-system` namespace
+
+**This documentation is for clusters that already have release `v0.19` of
+eventing running in the `cloud-run-events` namespace. If you are installing
+eventing for the first time on your cluster, you do not need to follow this
+steps and can go directly to
+[Installing Knative-GCP](https://github.com/google/knative-gcp/blob/master/docs/install/install-knative-gcp.md).**
 
 The upcoming `v0.20` release of `knative-gcp` will contain changes that are not
 compatible with old releases. Namely, the namespace will change from
@@ -20,7 +26,7 @@ release from the `knative-gcp` repository. We will update the page with specific
    explain the purpose of each command in comments.
 
    ```shell script
-   # create the new namespace by copying the information from the old one
+   # create the new namespace by copying the labels and annotations from the old one
    kubectl get namespace cloud-run-events -o yaml | \
      sed 's/  name: cloud-run-events/  name: events-system/g' | \
      kubectl create -f -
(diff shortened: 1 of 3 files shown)


eclipselu:
a general comment on the structure, should we put everything under a directory `docs/how-to/namespace-transition`?

docs/how-to/namespace-transition.md
```diff
@@ -0,0 +1,286 @@
...
+   job.batch/namespace-transition                    1/1           37s        3m3s
+   ```
+
+   ```shell script
+   # delete the job
+   kubectl delete -f namespace-transition.yaml
+   ```
+
+1. Upgrade to the new namespace by applying the changes from the `master` branch
+   or any of the releases from `v0.20` onward. This instructions are identical
```
nlopezgi, line 80: knative supports upgrading by a single minor version number, so these instructions, imo, should be very specific to indicate migration from 0.19 to 0.20. They should also probably not recommend using the master branch. https://knative.dev/docs/install/upgrade-installation/

docs/how-to/namespace-transition.md
```diff
@@ -0,0 +1,286 @@
...
+   ```shell script
+   # apply changes based on the master branch
+   cd knative-gcp
+   git checkout master
+   ko apply -f ./config
+   ```
+
+   ```shell script
+   # apply changes based on the release
+   export KGCP_VERSION=v0.20.0 # or above
```
nlopezgi, line 93: Note below these will work only once 0.20 is released, remove the # or above (As per my comment above)

docs/how-to/namespace-transition.md
```diff
@@ -0,0 +1,286 @@
+# Transition a project to the new `events-system` namespace
```
grantr, line 1: Would cluster make more sense than project here?

docs/how-to/namespace-transition.md
```diff
@@ -0,0 +1,286 @@
...
+release from the `knative-gcp` repository. We will update the page with specific
+`gcloud` instructions as the release becomes available there as well.**
+
+## Recommended Migration
+
+1. Run the following commands before installing the new `v0.20` release. We
+   explain the purpose of each command in comments.
+
+   ```shell script
+   # create the new namespace by copying the information from the old one
```
grantr, line 23: Can we be more specific than "information"? Is the purpose of this line to copy labels and annotations?

AlexandraRoatis: Yes. I updated the comment.

docs/how-to/namespace-transition.md
AlexandraRoatis: Updated.

docs/how-to/namespace-transition.md
AlexandraRoatis: Done.

docs/how-to/namespace-transition.md
AlexandraRoatis: Done.

AlexandraRoatis:
@eclipselu Since the number of files in the `how-to` directory is low, my preference would be for not changing the file structure. But, if there are any strong opinions on this matter I can make the update.

PR #2057, opened 2021-01-12
Add auth check into broker ingress pod

grac3gao:
Fixes #

## Proposed Changes

- Add auth check into broker ingress pod after https://github.com/google/knative-gcp/issues/1973 closed
-
-

**Release Note**

```release-note
Add auth check into broker ingress pod
```

**Docs**

### DIFF
--- a/pkg/utils/clients/factories.go
+++ b/pkg/utils/clients/factories.go
@@ -88,7 +88,7 @@ func authChecker(authType authcheck.AuthType) func(w nethttp.ResponseWriter, req
 				w.WriteHeader(nethttp.StatusInternalServerError)
 				return
 			}
-			w.WriteHeader(nethttp.StatusOK)
 		}
+		w.WriteHeader(nethttp.StatusOK)
 	}
 }
--- a/pkg/utils/clients/factories.go
+++ b/pkg/utils/clients/factories.go
@@ -27,6 +27,8 @@ import (
 	"go.opencensus.io/plugin/ochttp"
 	"knative.dev/eventing/pkg/kncloudevents"
 	"knative.dev/pkg/tracing/propagation/tracecontextb3"
+
+	"github.com/google/knative-gcp/pkg/utils/authcheck"
 )
 
 type Port int
@@ -36,6 +38,11 @@ type MaxConnsPerHost int
 // NewHTTPMessageReceiver wraps kncloudevents.NewHttpMessageReceiver with type-safe options.
 func NewHTTPMessageReceiver(port Port) *kncloudevents.HTTPMessageReceiver {
 	return kncloudevents.NewHTTPMessageReceiver(int(port))
+}
+
+// NewHTTPMessageReceiverWithChecker wraps kncloudevents.NewHttpMessageReceiver with a authentication checker.
+func NewHTTPMessageReceiverWithChecker(port Port, authType authcheck.AuthType) *kncloudevents.HTTPMessageReceiver {
+	return kncloudevents.NewHTTPMessageReceiver(int(port), kncloudevents.WithChecker(authChecker(authType)))
 }
 
 // NewPubsubClient provides a pubsub client from PubsubClientOpts.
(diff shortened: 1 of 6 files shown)


pkg/utils/clients/factories.go
```diff
@@ -71,3 +78,16 @@ func NewHTTPClient(_ context.Context, maxConnsPerHost MaxConnsPerHost) *nethttp.
...
+
+func authChecker(authType authcheck.AuthType) func(w nethttp.ResponseWriter, req *nethttp.Request) {
+	authCheck := authcheck.NewDefault(authType)
+	return func(w nethttp.ResponseWriter, req *nethttp.Request) {
+		if req.URL.Path == "/healthz" {
+			// Perform Authentication check.
+			if err := authCheck.Check(req.Context()); err != nil {
+				w.WriteHeader(nethttp.StatusInternalServerError)
+				return
+			}
```
zhongduo, line 90: Do we need to write OK code?

grac3gao: Yes, we need. Forget to add that.... Thanks

zhongduo:
This may worth a release note, WDYT?

grac3gao:
@zhongduo done.

zhongduo:
All the code looks good to me. My only question is: is there any way to fail before this commit in testing?

pkg/utils/clients/factories.go
```diff
@@ -38,6 +40,11 @@ func NewHTTPMessageReceiver(port Port) *kncloudevents.HTTPMessageReceiver {
 	return kncloudevents.NewHTTPMessageReceiver(int(port))
 }
 
+// NewHTTPMessageReceiverWithChecker wraps kncloudevents.NewHttpMessageReceiver with a authentication checker.
+func NewHTTPMessageReceiverWithChecker(port Port, authType authcheck.AuthType) *kncloudevents.HTTPMessageReceiver {
```
Harwayne, line 44: Is it possible to unit test this?

I _think_ it is possible by passing port = 0, which should mean find any available port. And a way to inject/overwrite the auth check implementation that is used.

I probably won't hold up the PR if this turns out to be too onerous, but I would feel much more comfortable if this was unit tested.

zhongduo: I was also commenting on testing this, but more in adding some check in our e2e test.

Harwayne: E2E test would be even better, but likely difficult because we only have a single Broker data plane and testing this would require us to break it, thus the other tests could not run in parallel with this one.

grac3gao: I added UTs for this. I am trying to add E2E test for brokercell currently, but as Adam mentioned, it has some difficulty as we only have a single broker data plane (I probably will add e2e test before we run all sub-tests.)

grac3gao:
ready for review

pkg/utils/clients/factories_test.go
```diff
@@ -0,0 +1,151 @@
...
+			ctx, cancel := context.WithCancel(context.Background())
+			defer cancel()
+			// Get a free port.
+			port := pkgtesting.GetFreePort(t)
+			receiver := NewHTTPMessageReceiverWithChecker(Port(port), "")
+			go receiver.StartListen(ctx, &testHandler{})
+
+			time.Sleep(1 * time.Second)
+
+			req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://127.0.0.1:%d"+tc.path, port), nil)
```
Harwayne, line 72:
```suggestion
			req, err := http.NewRequest(http.MethodGet, fmt.Sprintf("http://127.0.0.1:%d%s, port, tc.path), nil)
```

Harwayne (commented):
/lgtm
/hold

Holding for one small suggestion.

pkg/utils/clients/factories_test.go
grac3gao: done

Harwayne (commented):
/lgtm

PR #2059, opened 2021-01-12
Always pass upgrade test

zhongduo:
The upgrade test is always failing due to the namespace renaming. Temporarily disable it now.


nlopezgi:
we're probably going to have to cherry pick this into the 0.20 branch, but lets wait until release automation runs

tayarani:
can you add a TODO or issue link, or a comment on why it's there?

PR #2068, opened 2021-01-15
feat(broker): treat non-CloudEvent reply as successful delivery

eclipselu:
Fixes #

Currently if the event consumer reply back with a non-CloudEvent response, even if it's HTTP 2xx, it will still be considered as `malformed event`. 

This PR introduces a pre-check that checks Content-Type and ce-specversion, if it's not in structured/batched mode or binary mode, it's not a CloudEvent reply and we consider the delivery successful. 

## Proposed Changes

- 🧽 treat non-CloudEvent reply as successful delivery
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/broker/handler/processors/deliver/processor.go
+++ b/pkg/broker/handler/processors/deliver/processor.go
@@ -22,6 +22,7 @@ import (
 	"fmt"
 	"net/http"
 	"net/url"
+	"strings"
 	"time"
 
 	"github.com/cloudevents/sdk-go/v2/binding"
@@ -174,6 +175,13 @@ func (p *Processor) deliver(ctx context.Context, target *config.Target, broker *
 
 	if resp.StatusCode/100 != 2 {
 		return fmt.Errorf("event delivery failed: HTTP status code %d", resp.StatusCode)
+	}
+
+	// Pre-check the reply response header, if it's not in structured mode/batched mode or binary mode,
+	// then it's not a CloudEvent, we treat the delivery as successful and ignore the response.
+	// Otherwise, we proceed with the malformed event check.
+	if !strings.HasPrefix(resp.Header.Get("Content-Type"), "application/cloudevents") && (resp.Header.Get("ce-specversion") == "") {
+		return nil
 	}
 
 	respMsg := cehttp.NewMessageFromHttpResponse(resp)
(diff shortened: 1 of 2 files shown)


pkg/broker/handler/processors/deliver/processor_test.go
```diff
@@ -200,6 +200,87 @@ func TestDeliverSuccess(t *testing.T) {
 	}
 }
 
+type targetWithNonCloudEventReplyHandler struct {
```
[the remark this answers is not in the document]
eclipselu: Thanks, updated.

grantr:
Thanks @eclipselu! We might want to backport this to 0.20.

pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -176,6 +177,13 @@ func (p *Processor) deliver(ctx context.Context, target *config.Target, broker *
 		return fmt.Errorf("event delivery failed: HTTP status code %d", resp.StatusCode)
 	}
 
+	// Pre-check the reply response header, if it's not in structured mode/batched mode or binary mode,
+	// then it's not a CloudEvent, we treat the delivery as successful and ignore the response.
+	// Otherwise, we proceed with the malformed event check.
+	if !strings.HasPrefix(resp.Header.Get("Content-Type"), "application/cloudevents") && (resp.Header.Get("ce-specversion") == "") {
```
Harwayne, line 183: I don't think this is accurate. I believe in binary mode the `Content-Type` can be just about anything, see https://github.com/cloudevents/spec/blob/v1.0.1/http-protocol-binding.md#311-http-content-type.

In fact the first example in https://github.com/cloudevents/spec/blob/v1.0.1/http-protocol-binding.md#314-examples is:

```
POST /someresource HTTP/1.1
Host: webhook.example.com
ce-specversion: 1.0
ce-type: com.example.someevent
ce-time: 2018-04-05T03:56:24Z
ce-id: 1234-1234-1234
ce-source: /mycontext/subcontext
    .... further attributes ...
Content-Type: application/json; charset=utf-8
Content-Length: nnnn

{
    ... application data ...
}
```

Note that `Content-Type` is `application/json; charset=utf-8`, which is _not_ prefixed with `application/cloudevents`.

Harwayne: I am completely wrong. Please ignore.

PR #2069, opened 2021-01-15
feat/broker: validate deadLetterSink with retry

### LINKED ISSUE #2053: Broker ignores spec.delivery.retry if spec.delivery.deadLetterSink is nil
**Describe the bug**
Broker ignores spec.delivery.retry if spec.delivery.deadLetterSink is nil.

 **Expected behavior**
Either the Broker is rejected by the webhook as having an invalid deliverySpec, or retries is respected.

 **To Reproduce**
With the default `config-br-delivery`, create

```
apiVersion: eventing.knative.dev/v1beta1
kind: Broker
metadata:
  name: test-broker
  annotations:
    "eventing.knative.dev/broker.class": "googlecloud"
spec:
  delivery:
    retry: 33
```

Observe that the GCP PubSub Subscription created does not have the retry number anywhere:

```
{
  "name": "projects/$PROJECT/subscriptions/cre-tgr_default_hello-display_d00c77d2-7464-4213-a76a-ab63c7d4af8f",
  "topic": "projects/$PROJECT/topics/cre-tgr_default_hello-display_d00c77d2-7464-4213-a76a-ab63c7d4af8f",
  "pushConfig": {},
  "ackDeadlineSeconds": 10,
  "messageRetentionDuration": "604800s",
  "labels": {
    "resource": "triggers",
    "name": "hello-display",
    "namespace": "default"
  },
  "expirationPolicy": {
    "ttl": "2678400s"
  },
  "retryPolicy": {
    "minimumBackoff": "1s",
    "maximumBackoff": "600s"
  }
}

```
 
This is the only place that the `retry` value is read, and it is ignored if `deadLetterSink` is nil.
https://github.com/google/knative-gcp/blob/87b9a08562d4019eb35522b616b67a0fbe0b9ed0/pkg/reconciler/trigger/trigger.go#L313-L321

zhongduo:
After this commit, when retry is defined in broker delivery spec,
not having deadLetterSink will fail the validation. This is necessary
because pubsub will ignore retry number without a DLQ.

Fixes #2053 

**Release Note**

```release-note
- 🎁 Reject broker with retry but no DLQ
```

### DIFF
--- a/pkg/apis/broker/v1beta1/broker_validation.go
+++ b/pkg/apis/broker/v1beta1/broker_validation.go
@@ -44,7 +44,7 @@ func ValidateDeliverySpec(ctx context.Context, spec *eventingduckv1beta1.Deliver
 		errs = errs.Also(apis.ErrMissingField("backoffPolicy"))
 	}
 	if spec.Retry != nil && spec.DeadLetterSink == nil {
-		errs = errs.Also(apis.ErrMissingField("deadLetterSink"))
+		errs = errs.Also(apis.ErrGeneric("need DeadLetterSink when retry is defined", "deadLetterSink"))
 	}
 	return errs.Also(ValidateDeadLetterSink(ctx, spec.DeadLetterSink).ViaField("deadLetterSink"))
 }
--- a/pkg/apis/broker/v1beta1/broker_validation.go
+++ b/pkg/apis/broker/v1beta1/broker_validation.go
@@ -43,6 +43,9 @@ func ValidateDeliverySpec(ctx context.Context, spec *eventingduckv1beta1.Deliver
 	if spec.BackoffPolicy == nil {
 		errs = errs.Also(apis.ErrMissingField("backoffPolicy"))
 	}
+	if spec.Retry != nil && spec.DeadLetterSink == nil {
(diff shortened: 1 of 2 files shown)


pkg/apis/broker/v1beta1/broker_validation.go
```diff
@@ -43,6 +43,9 @@ func ValidateDeliverySpec(ctx context.Context, spec *eventingduckv1beta1.Deliver
 	if spec.BackoffPolicy == nil {
 		errs = errs.Also(apis.ErrMissingField("backoffPolicy"))
 	}
+	if spec.Retry != nil && spec.DeadLetterSink == nil {
+		errs = errs.Also(apis.ErrMissingField("deadLetterSink"))
```
Harwayne, line 47: I would prefer an error message saying that if `spec.retry` is set, then `spec.deadLetterSink` must be set. 

As-is, I think this may confuse people into believing that `spec.deadLetterSink` is always required.

zhongduo: Make sense. Updated the error to: "need DeadLetterSink when retry is defined: spec.delivery.deadLetterSink"

Harwayne (commented):
/lgtm

PR #2070, opened 2021-01-16
Add authentication check e2e test for BrokerCell

grac3gao:
Fixes #

## Proposed Changes

- Add authentication check e2e test for BrokerCell
- As we can't run this test parallel with other sub-test (brokercell is shared), it is added before we configure the right authentication for brokercell (which located in test_setup script)
- Only Integration test in secret mode and integration test in wi mode need this test. Since we reuse the test_setup for upgrade test and conformance test, an extra-env ENABLE_AUTH_CHECK_TEST is added into knative/test-infra#2637 to differentiate whether a test workflow needs auth check test or not.

**Release Note**

```release-note

```

**Docs**


test/e2e-common.sh
```diff
@@ -210,6 +211,78 @@ function enable_monitoring(){
...
+function test_authentication_check_for_brokercell() {
+  echo "Starting authentication check test for brokercell."
+  local auth_mode=${1}
+  ko apply -f ${CONFIG_WARMUP_GCP_BROKER}
+
+  echo "Starting authentication check test which is running outside of the broker related Pods."
+  wait_until_brokercell_authentication_check_pending "$(non_pod_check_keywords)" || return 1
+
+  echo "Starting authentication check test which is running inside of the broker related Pods."
+  apply_invalid_resource "$auth_mode"
```
zhongduo, line 223: If I read it correctly, after this all the e2e test will have this invalid resource test including adding an invalid resource and then delete it. I am wondering if there is a way to just run it with a single test without affecting others. General heath check is OK, but this applying bad one and delete can be a little bit confusing.

grac3gao: This PR is not ready for review currently, just want to run pipeline workflows as a try. I will remove WIP once it is ready.
The tricky part is this test can't run parallel with other sub-tests in a workflow, so that it can only run before or after all the sub-tests. Thus, I add it before configuring broker credential, we need to configure the right credential after finishing this test anyway (BrokerCell is shared by broker related sub-tests).
From the pipeline test result, it seems four workflows will run auth check test (integration test for secret, integration test for wi, upgrade and conformance), we need this test for two integration workflows. Upgrade and conformance tests are sourcing the test_setup from `e2e-secret-tests.sh`, I will override their test-setup to exclude this test.

grac3gao: Or probably use some workflow env to exclude this.

test/e2e-common.sh
```diff
@@ -210,6 +211,78 @@ function enable_monitoring(){
...
+  kubectl delete brokercell default -n "${CONTROL_PLANE_NAMESPACE}"
+  kubectl wait pod --for=delete -n "${CONTROL_PLANE_NAMESPACE}" --selector=brokerCell=default --timeout=5m || return 1
+}
+
+function wait_until_brokercell_authentication_check_pending() {
+  local keywords=${1}
+  echo "Waiting until brokercell authentication check pending."
+  for i in {1..150}; do #timeout after 5 minutes
+    local reason=$(kubectl get brokercell default -n "${CONTROL_PLANE_NAMESPACE}" -o=jsonpath="{range .status.conditions[*]}{.reason}" --ignore-not-found=true)
+    local message=$(kubectl get brokercell default -n "${CONTROL_PLANE_NAMESPACE}" -o=jsonpath="{range .status.conditions[*]}{.message}" --ignore-not-found=true)
```
Harwayne, line 239: Remove `reason` and use this, which I think provides the same output as the current `message` line.

```suggestion
    local message=$(kubectl get brokercell default -n "${CONTROL_PLANE_NAMESPACE}" -o=jsonpath='{.status.conditions[?(@.reason == "AuthenticationCheckPending")].message}' --ignore-not-found=true)
```

test/e2e-common.sh
```diff
@@ -210,6 +211,78 @@ function enable_monitoring(){
...
+    if [[ "$reason" == *"AuthenticationCheckPending"* ]] && [[ "$message" == *"$keywords"* ]]; then
+      return 0
+    fi
+    sleep 2
+  done
+  echo -e "\n\nERROR: timeout waiting for brokercell authentication check pending with correct message."
```
Harwayne, line 245: It would probably be useful to print the BrokerCell object's YAML here.

test/e2e-common.sh
```diff
@@ -210,6 +211,78 @@ function enable_monitoring(){
...
+  wait_until_brokercell_authentication_check_pending "$(non_pod_check_keywords)" || return 1
+
+  echo "Starting authentication check test which is running inside of the broker related Pods."
+  apply_invalid_resource "$auth_mode"
+  wait_until_brokercell_authentication_check_pending "$(pod_check_keywords "$auth_mode")" || return 1
+
+  # Clean up all the testing resources.
+  echo "Authentication check test finished, waiting until all broker related testing resources deleted."
+  delete_invalid_resource "$auth_mode"
+  ko delete -f ${CONFIG_WARMUP_GCP_BROKER}
```
Harwayne, line 229:
```suggestion
  kubectl delete -f ${CONFIG_WARMUP_GCP_BROKER}
```

test/e2e-common.sh
```diff
@@ -210,6 +211,78 @@ function enable_monitoring(){
       --role roles/cloudtrace.agent
 }
 
+function test_authentication_check_for_brokercell() {
```
Harwayne, line 214: Do we care about leaking anything from this test? If so, we probably need something like a trap to delete the warm up Broker and BrokerCell. We may not care because we expect the cluster to be deleted after this, but just want to make sure we thought about leaking resources.

As-is, if the test fails, then it will leak the warm up Broker.

test/e2e-common.sh
```diff
@@ -210,6 +211,84 @@ function enable_monitoring(){
       --role roles/cloudtrace.agent
 }
 
+# test_authentication_check_for_brokercell tests the authentication check function for BrokerCell.
+# This test required the entire BrokerCell to be non-functional.
+# In order to avoid running in parallel with other tests, we put it in the shell script, rather than more common go tests.
+function test_authentication_check_for_brokercell() {
```
Harwayne, line 217: Add:
```
# TODO Once we support multiple BrokerCells with distinct authentication mechanisms, make this a normal test that runs in parallel with the others.
```

test/e2e-common.sh
grac3gao: If the test fails, the leaked warm up Broker will be handled when deleting cluster. 

Besides the Broker itself (as the k8s resource), we also need to take care about the PubSub subscription and PubSub topic (as the Cloud resource) which were created by the controller when reconciling the Broker. If the test fails, the deletion of remaining pubsub subscription and pubsub topic would rely on the delete_topics_and_subscriptions function (https://github.com/google/knative-gcp/blob/ce080cf13a8d42e0588684f140c08d0f94666f85/test/e2e-common.sh#L180-L195). That is, every time when start to run the test, clean up the subscription and topic from last run.

That's the two kind of the resource which might be leak, and the way to clean up them, AFAIK.

test/e2e-common.sh
```diff
@@ -210,6 +211,85 @@ function enable_monitoring(){
...
+  echo -e "brokercell object YAML: $(kubectl get brokercell default -n "${CONTROL_PLANE_NAMESPACE}" -o yaml)"
+  return 1
+}
+
+function pod_check_keywords() {
+  local auth_mode=${1}
+  if [ "${auth_mode}" == "secret" ]; then
+    echo "error getting the token, probably due to the key stored in the Kubernetes Secret is expired or revoked"
+  elif [ "${auth_mode}" == "workload_identity" ]; then
+    echo "the Pod is not fully authenticated, probably due to corresponding k8s service account and google service account do not establish a correct relationship"
```
zhongduo, line 258: nits: I don't think you need to have the full message, it is hard to maintain if we later change the message. I suggest to just keep the important info, like `error getting the token`.

test/e2e-common.sh
```diff
@@ -210,6 +211,85 @@ function enable_monitoring(){
       --role roles/cloudtrace.agent
 }
 
+# test_authentication_check_for_brokercell tests the authentication check function for BrokerCell.
+# This test required the entire BrokerCell to be non-functional.
+# In order to avoid running in parallel with other tests, we put it in the shell script, rather than more common go tests.
+# TODO Once we support multiple BrokerCells with distinct authentication mechanisms, make this a normal test that runs in parallel with the others.
```
zhongduo, line 217: What's the plan to run this test now? A new Prow job?

test/e2e-common.sh
```diff
@@ -210,6 +211,85 @@ function enable_monitoring(){
...
+}
+
+function non_pod_check_keywords() {
+  echo "authentication is not configured"
+}
+
+function apply_invalid_auth() {
+  local auth_mode=${1}
+  if [ "${auth_mode}" == "secret" ]; then
+    kubectl -n "${CONTROL_PLANE_NAMESPACE}" create secret generic "google-broker-key" --from-file=key.json=${CONFIG_INVALID_CREDENTIAL}
```
zhongduo, line 272: There is `${GCP_BROKER_SECRET_NAM}` for google-broker-key.

test/e2e-common.sh
grac3gao: This is not a full message. The full message is even longer... I would like to keep this message, to clearly test out which step it goes wrong.

test/e2e-common.sh
grac3gao: This test will only run Integration test workflow in secret mode and integration test workflow in wi mode. Since we reuse the test_setup for upgrade test and conformance test, an extra-env `ENABLE_AUTH_CHECK_TEST` is added into knative/test-infra#2637 to differentiate whether a test workflow needs auth check test or not.

test/e2e-common.sh
grac3gao: done

test/e2e-secret-tests.sh
```diff
@@ -45,6 +45,14 @@ function export_variable() {
 # Setup resources common to all eventing tests.
 function test_setup() {
   pubsub_setup "secret" || return 1
+
+  # Authentication check test for BrokerCell. It is used in integration test in secret mode.
+  # We do not put it in the same place as other integration tests, because this test can not run in parallel with others,
+  # as this test requires the entire BrokerCell to be non-functional.
+  if [[ -v ENABLE_AUTH_CHECK_TEST && $ENABLE_AUTH_CHECK_TEST == 0 ]]; then
```
zhongduo, line 52: Do I read it correctly that if ENABLE_AUTH_CHECK_TEST is 0, the test will run. I would expect 0 means disabled...

test/e2e-common.sh
zhongduo: I see. As long as we only catch important errors you think necessary, it lgtm.

test/e2e-common.sh
zhongduo: Make sense. From the change in test-infra, this will only run in integration test and workload identity. lgtm

test/e2e-common.sh
```diff
@@ -210,6 +211,85 @@ function enable_monitoring(){
...
 }
 
+# test_authentication_check_for_brokercell tests the authentication check function for BrokerCell.
+# This test required the entire BrokerCell to be non-functional.
+# In order to avoid running in parallel with other tests, we put it in the shell script, rather than more common go tests.
+# TODO Once we support multiple BrokerCells with distinct authentication mechanisms, make this a normal test that runs in parallel with the others.
+function test_authentication_check_for_brokercell() {
+  echo "Starting authentication check test for brokercell."
+  local auth_mode=${1}
+  kubectl apply -f ${CONFIG_WARMUP_GCP_BROKER}
```
zhongduo, line 221: Random nits: some shellcheck actually does `"${CONFIG_WARMUP_GCP_BROKER}"`.  It is not enforced at all, but I see you are using both with and without quotes in your code, so maybe it makes sense to always add that.

grac3gao: The place where double quote was used is because it is caught by IDE automatic syntax checking for SC2086: Double quote to prevent globbing and word splitting. 
For this command, it seems it doesn't volatile that rule so it didn't show the warning to me...anyway, I will change it with double quote to make it consistent.

test/e2e-secret-tests.sh
grac3gao: Messed up with return 0 and return 1 in the script. To avoid that, I'll change it to be "true" in test-infra.

zhongduo: For bash you can also simply check if `"$ENABLE_AUTH_CHECK_TEST" == "1"` as undefined var will be empty IIUC.

grac3gao: I would like to keep `-v ENABLE_AUTH_CHECK_TEST `, in case there might return some unexpected error

zhongduo:
So we should merge it after the Prow env var update is changed.

test/e2e-common.sh
Harwayne: Thank you for explaining it!.

Harwayne (commented):
/lgtm
/hold

Holding in case @zhongduo has any more comments or any changes are needed upstream.

grac3gao:
@zhongduo I made the new commit after the new test-infra PR merged. https://github.com/knative/test-infra/pull/2639. The most recent test in this PR should be running under the new env value.

zhongduo:
awesome, thanks. 
/unhold
/approve
/lgtm

PR #2086, opened 2021-01-21
Remove stale liveliness and readiness check in broker ingress

grac3gao:
Fixes #

## Proposed Changes

- Since broker ingress HTTP Receiver is using Drainer for liveliness and readiness check from knative/pkg, this part of the code is stale. remove that with related UT.
- New liveliness and readiness related UTs are in https://github.com/google/knative-gcp/blob/master/pkg/utils/clients/factories_test.go#L31
-

**Release Note**

```release-note

```

**Docs**


Harwayne (commented):
/lgtm

PR #2088, opened 2021-01-22
Remove latest_version as change is merged upstream

zhongduo:
The change is already merged in knative/hack#43, so remove it.


Harwayne (commented):
/lgtm
/hold

Holding in case @zhongduo wants the review specifically from @nlopezgi.

zhongduo:
The upstream process is part of the plan, and the fact that upgrade test works means it is picking up from the code upstream. 
/unhold

PR #2089, opened 2021-01-22
[WIP] Refactor e2e scripts

MohamedElhawaty:
Fixes #

## Proposed Changes

- Remove unused service account parameter from storage_admin_set_up
- Rename GSA variable names
- Add logic to support two separate data plane services (Broker & Sources) instead of one. 
- Break down setup methods in two different lib files (doesn't really have lots of reusable code)

**Release Note**

```release-note

```

**Docs**


MohamedElhawaty:
I will split this into smaller PRs.

PR #2095, opened 2021-01-26
Remove unused parameter from storage_admin_set_up

MohamedElhawaty:
Fixes #

## Proposed Changes

- This PR could be considered as a followup to https://github.com/google/knative-gcp/pull/1418, it includes further cleanup for the unused parameters in `storage_admin_set_up` which were mainly used for the outdated curl command.
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/hack/lib.sh
+++ b/hack/lib.sh
@@ -17,8 +17,6 @@
 readonly CONTROL_PLANE_SERVICE_ACCOUNT="events-controller-gsa"
 readonly CONTROL_PLANE_NAMESPACE="events-system"
 
-readonly PUBSUB_SERVICE_ACCOUNT="cre-pubsub"
-
 readonly SERVICE_ACCOUNT_EMAIL_KEY="EMAIL"
 
 readonly ZONAL_CLUSTER_LOCATION_TYPE="zonal"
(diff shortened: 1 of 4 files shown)


Harwayne:
Please fill in the PR description.

hack/lib.sh
```diff
@@ -151,16 +149,11 @@ function enable_workload_identity(){
 function storage_admin_set_up() {
   echo "Update ServiceAccount for Storage Admin"
   local project_id=${1}
-  local pubsub_service_account=${2}
```
zhongduo, line 154: The fact that this is in `hack/lib.sh` is concerning to me. Eg. it might be used in one of our examples but there is no way we can tell it. Can you provide your justification that this is safe to remove? What's the old curl command in your PR? Will this break any storage related examples? Note the storage is our official getting started guide.

MohamedElhawaty: Sorry for not making it clear in the description but the curl command was actually deleted as part of this PR https://github.com/google/knative-gcp/pull/1418 and not this one. 

I guess we can consider that this change includes two parts:

1) Removing the key `"${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}"` which is still passed to `storage_admin_set_up` in e2e-common even though this parameter was removed in the PR I mentioned.  This include removing all the logic for creating the key e.g. `create_private_key_for_pubsub_service_account` in e2e-wi-tests.

2) Remove the service account parameter which its key was used by the curl command (and I guess that's why we were adding storage admin permissions).   

I guess this script is only used here https://github.com/google/knative-gcp/blob/master/docs/examples/cloudstoragesource/README.md and I don't think this should break it.

zhongduo: Thanks for the clarification. I did a grep when reviewing and found that file too. So it does look like this service account is not used any more.

PR #2096, opened 2021-01-26
Add subscriber metrics

bharattkukreja:
We need a metric which gives us more information about the event subscribers. This helps us in monitoring the services which are receiving events.

As such, we need 3 additional labels for subscriber info:
- SubscriberType
- SubscriberNamespace
- SubscriberName

This metric is similar to the existing event_count metric and measure the same thing. The new labels are added as a new metric by design to ensure backwards compatibility and avoid breaking changes.

## Proposed Changes

- Modify targets.proto to include the additional labels
- Add subscriber_info in broker_targets_config
- Export the metric in metrics/delivery_reporter.go

### DIFF
--- a/pkg/broker/handler/processors/deliver/processor.go
+++ b/pkg/broker/handler/processors/deliver/processor.go
@@ -170,6 +170,13 @@ func (p *Processor) deliver(ctx context.Context, target *config.Target, broker *
 	if err != nil {
 		logging.FromContext(ctx).Error("failed to add status code tags to context", zap.Error(err))
 	}
+
+	// Insert subscriber info into context.
+	cctx, err = metrics.AddSubscriberTags(ctx, target)
+	if err != nil {
+		logging.FromContext(ctx).Error("failed to add subscriber tags to context", zap.Error(err))
+	}
+
 	// Report event dispatch time with resp status code.
 	p.StatsReporter.ReportEventDispatchTime(cctx, time.Since(startTime))
 
--- a/pkg/metrics/testing/helper.go
+++ b/pkg/metrics/testing/helper.go
@@ -8,12 +8,12 @@ import (
 
 func ResetIngressMetrics() {
 	// OpenCensus metrics carry global state that need to be reset between unit tests.
-	metricstest.Unregister("event_count", "event_dispatch_latencies")
+	metricstest.Unregister("event_count", "event_dispatch_latencies", "subscriber_event_count")
 }
 
 func ResetDeliveryMetrics() {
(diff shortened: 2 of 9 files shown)


pkg/broker/handler/processors/deliver/processor.go
```diff
@@ -170,6 +170,13 @@ func (p *Processor) deliver(ctx context.Context, target *config.Target, broker *
 	if err != nil {
 		logging.FromContext(ctx).Error("failed to add status code tags to context", zap.Error(err))
 	}
+
+	// Insert subscriber info into context.
+	cctx, err = metrics.AddSubscriberTags(ctx, target)
```
Harwayne, line 175:
```suggestion
	cctx, err = metrics.AddSubscriberTags(cctx, target)
```

pkg/metrics/testing/helper.go
```diff
@@ -8,12 +8,12 @@ import (
 
 func ResetIngressMetrics() {
 	// OpenCensus metrics carry global state that need to be reset between unit tests.
-	metricstest.Unregister("event_count", "event_dispatch_latencies")
+	metricstest.Unregister("event_count", "event_dispatch_latencies", "subscriber_event_count")
```
Harwayne, line 11: Is this needed?

pkg/reconciler/brokercell/broker_targets_config.go
```diff
@@ -117,6 +118,11 @@ func (r *Reconciler) addToConfig(_ context.Context, b *brokerv1beta1.Broker, tri
 						Topic:        brokerresources.GenerateRetryTopicName(t),
 						Subscription: brokerresources.GenerateRetrySubscriptionName(t),
 					},
+					SubscriberInfo: &config.SubscriberInfo{
+						SubscriberType:      getSubscriberType(t.Kind, t.APIVersion),
+						SubscriberNamespace: t.Spec.Subscriber.Ref.Namespace,
```
Harwayne, line 123: Just to be sure, what defaults a value into `t.Spec.Subscriber.Ref.Namespace`? Is it the webhook? Or the URL lookup code? If it is the code, then this might be the empty string and we would need to use `t.Namespace` when this is the empty string.

pkg/broker/handler/processors/deliver/processor.go
quentin-cha: Do we care about the timeout case above ? 
Maybe we should move the adding of the subscriber tag higher up to before we call sendMsg().

PR #2099, opened 2021-01-26
Add filter broker class when building the routing config

### LINKED ISSUE #2094: Fanout trying to create handler goroutine for knative eventing broker/trigger
**Describe the bug**
When the cluster has knative eventing broker and trigger, the broker fanout pod will try
to create handler goroutine for the knative eventing trigger, which will result in the following error
in stackdriver log.

```
{
insertId: "d224dig4e75pew"
jsonPayload: {
logger: "broker-fanout"
message: "handler for broker has stopped with error"
commit: "5e6049b"
knative.dev/pod: "default-brokercell-fanout-7cc748475c-f69cz"
caller: "handler/fanout.go:180"
logging.googleapis.com/sourceLocation: {3}
timestamp: "2021-01-26T15:11:24.507992023Z"
broker: "BROKER:event-example//default"
error: "rpc error: code = NotFound desc = Resource not found (resource=cre-bkr_event-example_default_3ac6629b-e9aa-4f78-b2e6-4233102b2a99)."
stacktrace: "github.com/google/knative-gcp/pkg/broker/handler.(*FanoutPool).SyncOnce.func2.1
	github.com/google/knative-gcp/pkg/broker/handler/fanout.go:180
github.com/google/knative-gcp/pkg/broker/handler.(*Handler).Start.func1
	github.com/google/knative-gcp/pkg/broker/handler/handler.go:76"
}
resource: {2}
timestamp: "2021-01-26T15:11:24.508537436Z"
severity: "ERROR"
labels: {5}
logName: "projects/jimmy-knative-dev/logs/stderr"
receiveTimestamp: "2021-01-26T15:11:25.014272940Z"
}
```

 **Expected behavior**
Fanout will not try to create the handler for knative eventing triggers.

 **To Reproduce**
After installing both knative eventing and knative gcp, follow https://knative.dev/docs/eventing/getting-started/ to create eventing broker and trigger. Then check the log in stackdriver.

cathyzhyi:
Fixes  #2094 

## Proposed Changes

- Add the filter as an util function to be shared by different places
- Filter the broker class when building routing configmap so that only gcp broker and trigger are included
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/reconciler/broker/controller.go
+++ b/pkg/reconciler/broker/controller.go
@@ -30,11 +30,9 @@ import (
 	"k8s.io/client-go/tools/cache"
 
 	"github.com/google/knative-gcp/pkg/logging"
-	eventingv1beta1 "knative.dev/eventing/pkg/apis/eventing/v1beta1"
 	"knative.dev/pkg/configmap"
 	"knative.dev/pkg/controller"
 	"knative.dev/pkg/injection"
-	pkgreconciler "knative.dev/pkg/reconciler"
 
 	brokerv1beta1 "github.com/google/knative-gcp/pkg/apis/broker/v1beta1"
 	inteventsv1alpha1 "github.com/google/knative-gcp/pkg/apis/intevents/v1alpha1"
@@ -42,6 +40,7 @@ import (
 	brokercellinformer "github.com/google/knative-gcp/pkg/client/injection/informers/intevents/v1alpha1/brokercell"
 	brokerreconciler "github.com/google/knative-gcp/pkg/client/injection/reconciler/broker/v1beta1/broker"
 	"github.com/google/knative-gcp/pkg/reconciler"
+	reconcilerutils "github.com/google/knative-gcp/pkg/reconciler/utils"
 	"github.com/google/knative-gcp/pkg/utils"
 )
 
@@ -108,7 +107,7 @@ func newController(ctx context.Context, cmw configmap.Watcher, brds *brokerdeliv
 	brokerInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.FilteringResourceEventHandler{
 			// Only reconcile brokers with the proper class annotation
-			FilterFunc: pkgreconciler.AnnotationFilterFunc(eventingv1beta1.BrokerClassAnnotationKey, brokerv1beta1.BrokerClass, false /*allowUnset*/),
+			FilterFunc: reconcilerutils.BrokerClassFilter,
 			Handler:    controller.HandleAll(impl.Enqueue),
 		},
 		reconciler.DefaultResyncPeriod,
--- a/pkg/reconciler/broker/controller.go
+++ b/pkg/reconciler/broker/controller.go
@@ -106,7 +106,7 @@ func newController(ctx context.Context, cmw configmap.Watcher, brds *brokerdeliv
 
 	brokerInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.FilteringResourceEventHandler{
-			// Only reconcile brokers with the proper class annotation
+			// Only reconcile brokers with proper brokerclass
 			FilterFunc: reconcilerutils.BrokerClassFilter,
 			Handler:    controller.HandleAll(impl.Enqueue),
 		},
--- a/pkg/reconciler/brokercell/broker_targets_config.go
+++ b/pkg/reconciler/brokercell/broker_targets_config.go
@@ -32,6 +32,7 @@ import (
 	"github.com/google/knative-gcp/pkg/broker/config/memory"
 	brokerresources "github.com/google/knative-gcp/pkg/reconciler/broker/resources"
 	"github.com/google/knative-gcp/pkg/reconciler/brokercell/resources"
+	"github.com/google/knative-gcp/pkg/reconciler/utils"
 	"github.com/google/knative-gcp/pkg/reconciler/utils/volume"
 )
 
@@ -73,6 +74,9 @@ func (r *Reconciler) addBrokersAndTriggersToTargets(ctx context.Context, bc *int
 		return err
 	}
 	for _, broker := range brokers {
+		if !utils.BrokerClassFilter(broker) {
+			continue
+		}
 		// Filter by `eventing.knative.dev/broker: <name>` here
 		// to get only the triggers for this broker. The trigger webhook will
 		// ensure that triggers are always labeled with their broker name.
--- a/pkg/reconciler/trigger/controller.go
+++ b/pkg/reconciler/trigger/controller.go
@@ -111,7 +111,7 @@ func newController(ctx context.Context, cmw configmap.Watcher, drs *dataresidenc
 	// Watch brokers.
 	brokerinformer.Get(ctx).Informer().AddEventHandler(
 		cache.FilteringResourceEventHandler{
-			// Only care about brokers with the proper class annotation
+			// Only care about brokers with proper brokerclass
 			FilterFunc: reconcilerutils.BrokerClassFilter,
 			Handler: controller.HandleAll(func(obj interface{}) {
 				if b, ok := obj.(*brokerv1beta1.Broker); ok {
--- a/pkg/reconciler/trigger/controller.go
+++ b/pkg/reconciler/trigger/controller.go
@@ -28,7 +28,6 @@ import (
 
 	"github.com/google/knative-gcp/pkg/logging"
 	"knative.dev/eventing/pkg/apis/eventing"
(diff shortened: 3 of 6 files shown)


pkg/reconciler/brokercell/broker_targets_config.go
```diff
@@ -54,6 +55,9 @@ func (r *Reconciler) reconcileConfig(ctx context.Context, bc *intv1alpha1.Broker
 	// maintaining 2 queues for updated brokers and triggers, and only update the config for updated brokers/triggers.
 	brokerTargets := memory.NewEmptyTargets()
 	for _, broker := range brokers {
+		if !utils.BrokerClassFilter(broker) {
```
eclipselu, line 58: I cannot find the definition of `BrokerClassFilter`, maybe I'm not familiar with this part of the code. can you point to me where this is?

cathyzhyi: sry... just added the new file.

pkg/reconciler/utils/filter.go
```diff
@@ -0,0 +1,26 @@
...
+package utils
+
+import (
+	brokerv1beta1 "github.com/google/knative-gcp/pkg/apis/broker/v1beta1"
+	eventingv1beta1 "knative.dev/eventing/pkg/apis/eventing/v1beta1"
+	reconciler "knative.dev/pkg/reconciler"
+)
+
+// BrokerClassFilter is the function to filter brokers with proper brokerclass.
+var BrokerClassFilter = reconciler.AnnotationFilterFunc(eventingv1beta1.BrokerClassAnnotationKey, brokerv1beta1.BrokerClass, false /*allowUnset*/)
```
eclipselu, line 26: nits: maybe we can make it `BrokerClassAnnotationFilter`?

pkg/reconciler/trigger/trigger.go
```diff
@@ -28,6 +28,7 @@ import (
 	apierrs "k8s.io/apimachinery/pkg/api/errors"
 
 	"github.com/google/knative-gcp/pkg/logging"
+	reconcilerutils "github.com/google/knative-gcp/pkg/reconciler/utils"
```
eclipselu, line 31: nits: can we put this together with other `pkg/reconciler/utils` imports?

cathyzhyi: done as suggested

pkg/reconciler/utils/filter.go
cathyzhyi: hmm, I personally prefer `BrokerClassFilter` because the util function name should expose what's filtered (the broker class) not how it's filtered (through an annotation). I feel adding `Annotation` into the name is exposing implementation details the caller usually don't care.

pkg/reconciler/broker/controller.go
```diff
@@ -104,7 +103,7 @@ func newController(ctx context.Context, cmw configmap.Watcher, brds *brokerdeliv
 	brokerInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.FilteringResourceEventHandler{
 			// Only reconcile brokers with the proper class annotation
```
eclipselu, line 105:
```suggestion
			// Only reconcile brokers with proper brokerclass
```

pkg/reconciler/utils/filter.go
eclipselu: Thanks, I see, that makes sense. I saw the comment in https://github.com/google/knative-gcp/pull/2099/files#diff-88052577c3d76c3adb4cda9795e0020e8a949de34b610b34dadace01fab839c4R105 so I thought the implementation detail is already leaked. Maybe we can change that comment.

pkg/reconciler/trigger/controller.go
```diff
@@ -112,7 +108,7 @@ func newController(ctx context.Context, cmw configmap.Watcher, drs *dataresidenc
 	brokerinformer.Get(ctx).Informer().AddEventHandler(
 		cache.FilteringResourceEventHandler{
 			// Only care about brokers with the proper class annotation
```
eclipselu, line 110:
```suggestion
			// Only care about brokers with proper brokerclass
```

pkg/reconciler/utils/filter.go
cathyzhyi: Thanks a lot for updating the comments!

pkg/reconciler/utils/filter.go
```diff
@@ -0,0 +1,26 @@
+/*
+Copyright 2020 Google LLC
```
zhongduo, line 2: 2021

pkg/reconciler/trigger/controller.go
```diff
@@ -111,8 +107,8 @@ func newController(ctx context.Context, cmw configmap.Watcher, drs *dataresidenc
 	// Watch brokers.
 	brokerinformer.Get(ctx).Informer().AddEventHandler(
 		cache.FilteringResourceEventHandler{
-			// Only care about brokers with the proper class annotation
-			FilterFunc: filterBroker,
+			// Only care about brokers with proper brokerclass
+			FilterFunc: reconcilerutils.BrokerClassFilter,
```
zhongduo, line 111: Can you checked with Cong about this? Considering this case:

A broker started as a gcp broker, then the broker class is changed manually to MT broker, will the pub/sub topic for gcp broker leak?

zhongduo:
There is an issue checker, can you put it to your PR?

pkg/reconciler/trigger/controller.go
cathyzhyi: This event handler makes sure triggers's pubsub topic and subscription won't be leaked. Looking at the usage of filter function, as long as the updated broker is gcp broker either before the update or after the update, the trigger will be enqueued https://github.com/kubernetes/client-go/blob/b72204b2445de5ac815ae2bb993f6182d271fdb4/tools/cache/controller.go#L268. The filter function filters out triggers that's entirely unrelated to this broker.

zhongduo: OK, it seems that both the original broker and trigger already have this filtering. The only additional filter here is in the brokercell. Makes sense to me.

eclipselu:
I believe I co-authored one of the commit (where I suggested a change) which is associated with my personal email, this violated the CLA check. sorry about that, please feel free to squash the commits to pass CLA.

pkg/reconciler/broker/controller.go
```diff
@@ -103,8 +102,8 @@ func newController(ctx context.Context, cmw configmap.Watcher, brds *brokerdeliv
 
 	brokerInformer.Informer().AddEventHandlerWithResyncPeriod(
 		cache.FilteringResourceEventHandler{
-			// Only reconcile brokers with the proper class annotation
-			FilterFunc: pkgreconciler.AnnotationFilterFunc(eventingv1beta1.BrokerClassAnnotationKey, brokerv1beta1.BrokerClass, false /*allowUnset*/),
+			// Only reconcile brokers with proper brokerclass
+			FilterFunc: reconcilerutils.BrokerClassFilter,
 			Handler:    controller.HandleAll(impl.Enqueue),
 		},
 		reconciler.DefaultResyncPeriod,
```
danyinggu, line 109: Should we add the brokerclass filter in the brokercell handler to only filter the broker with googleclass at line 125?

zhongduo: I think so. I remember that's part of our discussion in the chat too.

danyinggu: Although I found that https://github.com/google/knative-gcp/blob/121999757c34d7fc08b03b624f09ca293e863937/pkg/client/injection/reconciler/broker/v1beta1/broker/reconciler.go#L223 here, seems like it filter out all non googleclass broker from reconcilation, but I guess we should keep this filter and the brokerInformer filter consistent?

zhongduo: I think the original problem was solved by the change in brokercell. But it is always good to check if the problem persists.

cathyzhyi: I think the plan is to eventually use `r.brokerLister.List(labels.SelectorFromSet(map[string]string{"brokercell":bc.Name, "brokercellns":bc.Namespace}` to include only brokers related to a single brokercell as commented in https://github.com/google/knative-gcp/blob/ebc58cdbe0cb1c803b059a849ee031eebd5640eb/pkg/reconciler/broker/controller.go#L118. Before this happens, another class filter in reconciler would filter the class anyways.

pkg/reconciler/trigger/controller.go
zhongduo: Changing the broker class is not allowed and will be rejected in the webhook, so the original question does not make sense any more.

PR #2104, opened 2021-01-28
Split data plane GSA to broker & sources GSAs in e2e scripts

MohamedElhawaty:
Fixes #

## Proposed Changes

- Add the logic to split the data plane GSA to two separate broker & sources GSAs in the e2e scripts to match our documentation.
- Rename variables related to the control plane GSA to be consistent with the other GSAs variables.
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/test/e2e-common.sh
+++ b/test/e2e-common.sh
@@ -22,8 +22,11 @@ which gcloud &> /dev/null || gcloud() { echo "[ignore-gcloud $*]" 1>&2; }
 # Constants used for creating ServiceAccount for the Control Plane if it's not running on Prow.
 readonly CONTROL_PLANE_SERVICE_ACCOUNT_NON_PROW="events-controller-gsa"
 
-# Constants used for creating ServiceAccount for Data Plane(Pub/Sub Admin) if it's not running on Prow.
-readonly PUBSUB_SERVICE_ACCOUNT_NON_PROW="cre-pubsub"
+# Constants used for creating ServiceAccount for the Sources if it's not running on Prow.
+readonly SOURCES_GSA_NON_PROW="events-sources-gsa"
+
+# Constants used for creating ServiceAccount for the Broker if it's not running on Prow.
+readonly BROKER_GSA_NON_PROW="events-broker-gsa"
 
 # Vendored eventing test images.
 readonly VENDOR_EVENTING_TEST_IMAGES="vendor/knative.dev/eventing/test/test_images/"
@@ -44,7 +47,8 @@ function knative_setup() {
 # Tear down tmp files which store the private key.
 function test_teardown() {
   if (( ! IS_PROW )); then
-    rm "${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}"
+    rm "${SOURCES_GSA_KEY_TEMP}"
+    rm "${BROKER_GSA_KEY_TEMP}"
   fi
 }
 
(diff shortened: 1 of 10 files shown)


test/e2e-secret-tests.sh
```diff
@@ -24,27 +24,32 @@ source $(dirname "${BASH_SOURCE[0]}")/e2e-common.sh
...
-readonly CONTROL_PLANE_SERVICE_ACCOUNT_NON_PROW_KEY_TEMP="$(mktemp)"
+readonly CONTROL_PLANE_GSA_NON_PROW_KEY_TEMP="$(mktemp)"
 
-# Constants used for creating ServiceAccount for Data Plane(Pub/Sub Admin) if it's not running on Prow.
-readonly PUBSUB_SERVICE_ACCOUNT_NON_PROW_KEY_TEMP="$(mktemp)"
+# Constants used for creating ServiceAccount for the Sources if it's not running on Prow.
+readonly SOURCES_GSA_NON_PROW_KEY_TEMP="$(mktemp)"
+
+# Constants used for creating ServiceAccount for the Broker if it's not running on Prow.
+readonly BROKER_GSA_NON_PROW_KEY_TEMP="$(mktemp)"
```
zhongduo, line 33: Do we need this at all? Can we just do sth like `readonly SOURCES_GSA_KEY_TEMP="$(mktemp)"`?

test/e2e-secret-tests.sh
```diff
@@ -24,27 +24,32 @@ source $(dirname "${BASH_SOURCE[0]}")/e2e-common.sh
...
+    readonly CONTROL_PLANE_GSA_KEY_TEMP="${GOOGLE_APPLICATION_CREDENTIALS}"
+    readonly SOURCES_GSA_KEY_TEMP="${GOOGLE_APPLICATION_CREDENTIALS}"
+    readonly BROKER_GSA_KEY_TEMP="${GOOGLE_APPLICATION_CREDENTIALS}"
   fi
 }
 
 # Setup resources common to all eventing tests.
 function test_setup() {
-  pubsub_setup "secret" || return 1
+  sources_auth_setup "secret" || return 1
```
zhongduo, line 52: sth to think about, is "source" a better name than "pubsub"? What if we have a new source that does not use pubsub?

MohamedElhawaty: > What if we have a new source that does not use pubsub?

I feel this is one more reason not to call the GSA variables or the service itself pubsub. This GSA doesn't only have the pubsub permission but also `cloudtrace.agent` & `monitoring.metricWriter`.  Also, `pubsub` doesn't tell anything about its user.

My guess if we have a new source that doesn't use pubsub is to still add the permission the sources GSA or create a new one.

test/e2e-secret-tests.sh
MohamedElhawaty: No you are right. I had this as part of my initial PR but I thought we can do this in a separate PR. Anyways, I will include this.

test/e2e-secret-tests.sh
zhongduo: make sense.

zhongduo:
/lgtm
/approve
/hold
just to give others a chance to review, since this is mostly renaming, it should be functionally safe.

test/e2e-common.sh
```diff
@@ -106,24 +110,32 @@ function pubsub_setup() {
...
 # Create resources required for GCP Broker authentication setup.
-function gcp_broker_setup() {
+function broker_auth_setup() {
   echo "Authentication setup for GCP Broker"
   local auth_mode=${1}
 
   if [ "${auth_mode}" == "secret" ]; then
-    kubectl -n "${CONTROL_PLANE_NAMESPACE}" create secret generic "${GCP_BROKER_SECRET_NAME}" --from-file=key.json="${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}"
+    if (( ! IS_PROW )); then
+      init_pubsub_service_account "${E2E_PROJECT_ID}" "${BROKER_GSA_NON_PROW}"
```
AlexandraRoatis, line 119: Perhaps this function should be renamed as well?

test/e2e-common.sh
```diff
@@ -106,24 +110,32 @@ function pubsub_setup() {
...
 
 # Create resources required for GCP Broker authentication setup.
-function gcp_broker_setup() {
+function broker_auth_setup() {
   echo "Authentication setup for GCP Broker"
   local auth_mode=${1}
 
   if [ "${auth_mode}" == "secret" ]; then
-    kubectl -n "${CONTROL_PLANE_NAMESPACE}" create secret generic "${GCP_BROKER_SECRET_NAME}" --from-file=key.json="${PUBSUB_SERVICE_ACCOUNT_KEY_TEMP}"
+    if (( ! IS_PROW )); then
```
AlexandraRoatis, line 118: Can you add a comment here explaining why this `if` statement was necessary?

test/e2e-common.sh
MohamedElhawaty: Now that we use a different GSA for the broker we have to create it and grant the right permissions, just similar to the `sources_auth_setup`. I can add a comment similar to the one there:

```# When not running on Prow we need to set up a service account for broker.```

test/e2e-common.sh
MohamedElhawaty: Now that we use a different GSA for the broker we have to create it and grant the right permissions, just similar to the `sources_auth_setup`. I can add a comment similar to the one there:

```# When not running on Prow we need to set up a service account for broker.```

test/e2e-common.sh
MohamedElhawaty: I was think of renaming to `init_gsa_with_pubsub_editor` but @zhongduo had a fair point (https://github.com/google/knative-gcp/pull/2089#discussion_r564623456) to keep the old name as the new one will be a bit confusing if later we decided not to use editor role. Do you think of a specific name? This is used both the setup the sources & broker GSA.

zhongduo: Just the clarify, my previous comment was not about "pubsub" or not since the new name was using pubsub too. I was saying that too much implementation details in the name is not a good idea, in this case "editor" role.

MohamedElhawaty: Right, I guess that's what I understood too. I am happy to change this if there is a better name.

PR #2111, opened 2021-01-29
Update error message verification in Auth Check

grac3gao:
Fixes #

## Proposed Changes

- Auth check test has flakiness due to secret cache sync failure. In auth check, I only filtered k8s events with key word "Message:MountVolume.SetUp failed for volume", so if there is a cache sync up k8s event for the Pod, it will be filtered in as an authentication check failure. Update to filter more specific messages.
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/utils/authcheck/list.go
+++ b/pkg/utils/authcheck/list.go
@@ -97,7 +97,7 @@ func isAuthMessage(message string) bool {
 
 // isSecretFailureMessage checks if the message is for a specific secret's failure.
 func isSecretFailureMessage(message, namespace string, secret *corev1.SecretKeySelector) bool {
-	return strings.Contains(message, fmt.Sprintf(`MountVolume.SetUp failed for volume "%s"`, secret.Name)) ||
+	return strings.Contains(message, fmt.Sprintf(`secret "%s" not found`, secret.Name)) ||
 		strings.Contains(message, fmt.Sprintf(`couldn't find key %s in Secret %s/%s`, secret.Key, namespace, secret.Name))
 }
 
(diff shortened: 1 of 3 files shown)


Harwayne (commented):
/lgtm

zhongduo:
I am not sure I fully understand the whole message. 
1. What's the original message that make both the old and new code work? 
2. What's the message that works with only new code but not old code?

grac3gao:
The full error message we want for authentication check: `MountVolume.SetUp failed for volume \"google-cloud-key\"  :  secret  \"google-cloud-key\"  not found`
The full error message we don't want: `MountVolume.SetUp failed for volume \"google-cloud-key\"  : failed to sync secret cache: timed out waiting for the condition`

old key word: `MountVolume.SetUp failed for volume \"google-cloud-key\"`
new key word: `secret  \"google-cloud-key\"  not found`

why not use the whole sentence to check `MountVolume.SetUp failed for volume \"google-cloud-key\"  :  secret  \"google-cloud-key\"  not found` ?
I found there is a white space right after `MountVolume.SetUp failed for volume \"google-cloud-key\"`, before `:` in the event message. In other event message, this white space doesn't appear. I guess it is a kind of small typo in the original k8s codebase. In case they might delete this white space, and break our check, I only include half of the sentence to be the key words. Also, before verifying the event message, we filter out k8s events only for specific Pod object and for warning type, so it is unlikely that there might be other event message has the same keyword for this Pod object.

zhongduo:
Thanks for clarification, this makes sense to me. And I will vote for partial matching too.
/lgtm

PR #2112, opened 2021-01-29
Remove empty spec

zhongduo:
Empty spec will now cause error. The error is about webhook, but I believe this is just because it is not parsed properly.


grantr (approved):
Seems fine as long as the webhook accepts it

/lgtm

PR #2115, opened 2021-02-01
Fix failure during serving installation in e2e test

zhongduo:
Install serving in case of failure to avoid race condition of webhook and config map.

### DIFF
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -40,7 +40,8 @@ function start_knative_gcp_monitoring() {
 
 # Install all required components for running knative-gcp.
 function start_knative_gcp() {
-  start_latest_knative_serving || return 1
+  # Try reapply serving yaml again in case of race condition between webhook and config map
+  start_latest_knative_serving || start_latest_knative_serving || return 1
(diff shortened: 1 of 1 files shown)


test/lib.sh
```diff
@@ -40,7 +40,8 @@ function start_knative_gcp_monitoring() {
 
 # Install all required components for running knative-gcp.
 function start_knative_gcp() {
-  start_latest_knative_serving || return 1
+  # Try reapply serving yaml again in case of race condition between webhook and config map
+  start_latest_knative_serving || start_latest_knative_serving || return 1
```
Harwayne, line 44: Wow, we are really desperate.

Harwayne (commented):
/lgtm

test/lib.sh
zhongduo: We are doing that for eventing too before the webhook is ignored.

zhongduo:
@Harwayne some jobs hanged in the previous one, I just pushed it again without change.

Harwayne (commented):
/lgtm

PR #2116, opened 2021-02-01
Update PS & Topic reconcilers to use Base instead of PubSubBase

MohamedElhawaty:
## Proposed Changes

- Both PullSubscription and Topic reconcilers are using PubSubBase which is a bit confusing as they don't need the functionalities it offers, changing this to use our [reconciler base](https://github.com/google/knative-gcp/blob/master/pkg/reconciler/reconciler.go) instead.

**Release Note**

```release-note
NONE
```


Harwayne (commented):
/lgtm

PR #2118, opened 2021-02-02
WIP: drop dependency of net-istio

zhongduo:
Fixes #

## Proposed Changes

-
-
-

**Release Note**

```release-note

```

**Docs**


zhongduo:
This apparently is not the problem.

PR #2120, opened 2021-02-02
Change the receiver to pointer type

cathyzhyi:
## Proposed Changes

- Change the receiver to a pointer:
    - to avoid copying when calling those methods
    - to update object fields correctly

**Release Note**

```release-note
NONE
```


Harwayne (commented):
/lgtm

zhongduo:
For the second point, is there any test that might capture it?

cathyzhyi:
> For the second point, is there any test that might capture it?

opened an issue for adding unit tests https://github.com/google/knative-gcp/issues/2122

PR #2124, opened 2021-02-03
Dump more system namespace controller pods in case of failure

zhongduo:
Add more namespace dump to capture possible problem from serving and istio.


nlopezgi (commented):
/lgtm

PR #2127, opened 2021-02-03
Mark pullSubscription as empty when it is deleted

danyinggu:
Fixes #

## Proposed Changes

- The topic ID and projectID for topic is set to empty when topic is deleted, so does the sinkURI. Mark pullSubscription to be empty to keep them consistent.
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/reconciler/intevents/reconciler.go
+++ b/pkg/reconciler/intevents/reconciler.go
@@ -288,5 +288,6 @@ func (psb *PubSubBase) DeletePubSub(ctx context.Context, pubsubable duck.PubSuba
 		return fmt.Errorf("failed to delete PullSubscription: %w", err)
 	}
 	status.SinkURI = nil
+	status.SubscriptionID = ""
 	return nil
 }


pkg/reconciler/intevents/reconciler.go
```diff
@@ -288,5 +288,6 @@ func (psb *PubSubBase) DeletePubSub(ctx context.Context, pubsubable duck.PubSuba
 		return fmt.Errorf("failed to delete PullSubscription: %w", err)
 	}
 	status.SinkURI = nil
+	status.SubscriptionID = ""
```
zhongduo, line 291: Any reason not to move it to after line 281? These are all pubsub related fields IIUC.

danyinggu: the subscriptionID was originally set up by the pullSubscription (only after pullSubscription turns ready), https://github.com/google/knative-gcp/blob/master/pkg/reconciler/intevents/reconciler.go#L198,
so it should be unset if pullsubscription is fully deleted.

Line 238 is for topic.

PR #2128, opened 2021-02-03
Get deployment of receive adapter by deployment's name

danyinggu:
Fixes #

## Proposed Changes

- Previously we get the deployment of receive adapter by labels (controllerAgentName and pullSubscription's name). We could directly get the deployment by its name, because the deployment name of a pullSubscription is generated from UID, which is unique. 
(https://github.com/google/knative-gcp/blob/350ed608feb43474488be97275723599389d76e2/pkg/reconciler/intevents/pullsubscription/resources/names.go#L44)
- 
-
-

**Release Note**

```release-note

```

**Docs**

### DIFF
--- a/pkg/reconciler/intevents/pullsubscription/reconciler.go
+++ b/pkg/reconciler/intevents/pullsubscription/reconciler.go
@@ -338,18 +338,19 @@ func (r *Base) reconcileDataPlaneResources(ctx context.Context, ps *v1.PullSubsc
 
 func (r *Base) GetOrCreateReceiveAdapter(ctx context.Context, desired *appsv1.Deployment, ps *v1.PullSubscription) (*appsv1.Deployment, error) {
 	existing, err := r.KubeClientSet.AppsV1().Deployments(ps.Namespace).Get(ctx, desired.Name, metav1.GetOptions{})
+	if err == nil {
+		return existing, nil
+	}
+	if !apierrors.IsNotFound(err) {
+		logging.FromContext(ctx).Desugar().Error("Unable to get an existing Receive Adapter", zap.Error(err))
+		ps.Status.MarkDeployedUnknown("ReceiveAdapterGetFailed", "Error getting the Receive Adapter: %s", err.Error())
+		return nil, err
+	}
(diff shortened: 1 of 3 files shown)


zhongduo:
Ingress not reconciled error in CI: https://prow.knative.dev/view/gs/knative-prow/pr-logs/pull/google_knative-gcp/2128/pull-google-knative-gcp-integration-tests/1357110720625381376

/retest

pkg/reconciler/intevents/pullsubscription/reconciler.go
```diff
@@ -338,13 +337,13 @@ func (r *Base) reconcileDataPlaneResources(ctx context.Context, ps *v1.PullSubsc
...
 func (r *Base) GetOrCreateReceiveAdapter(ctx context.Context, desired *appsv1.Deployment, ps *v1.PullSubscription) (*appsv1.Deployment, error) {
-	existing, err := r.getReceiveAdapter(ctx, ps)
-	if err != nil && !apierrors.IsNotFound(err) {
-		logging.FromContext(ctx).Desugar().Error("Unable to get an existing Receive Adapter", zap.Error(err))
-		ps.Status.MarkDeployedUnknown("ReceiveAdapterGetFailed", "Error getting the Receive Adapter: %s", err.Error())
-		return nil, err
-	}
-	if existing == nil {
+	existing, err := r.KubeClientSet.AppsV1().Deployments(ps.Namespace).Get(ctx, desired.Name, metav1.GetOptions{})
+	if err != nil {
```
zhongduo, line 341: Changed to 
```
if err == nil {
   return existing, nil
}
// logic without indent
```

danyinggu: Thanks for the comment! Does this change cause any differences for the logic?
It seems like if changing to the way you suggested, it will `return existing, nil`  again at the end after creating the new deployment, I'd rather keep it as this.

zhongduo: It won't change the logic, this is a common way to remove unnecessary nested logic. Eg. the if body might later become big with more nested logic, and this is one way of keeping the code cleaner.

danyinggu: Makes sense! Thanks for the explanation :)

PR #2203, opened 2021-05-04
Use nightly serving and eventing in e2e tests

### LINKED ISSUE #2201: Use serving & eventing nightly in e2e tests
#2199 pins the serving version to 0.22.0 and eventing version to 0.22.1 as installed in the e2e tests. We should revert back to using serving and eventing nightly ASAP once the changes goes in.

tommyreddad:
Fixes #2201

We temporarily pinned the serving and eventing versions in e2e tests in order to get confidence about the upgrade tests passing due to eventing nightly being incompatible with knative-gcp at the time. Now we should go back to testing against serving and eventing nightly.

## Proposed Changes

- 🧽 Use nightly serving and eventing in e2e tests.

### DIFF
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -41,15 +41,11 @@ function start_knative_gcp_monitoring() {
 # Install all required components for running knative-gcp.
 function start_knative_gcp() {
   # Try reapply serving yaml again in case of race condition between webhook and config map
-  start_release_knative_serving 0.22.0 || start_release_knative_serving 0.22.0 || return 1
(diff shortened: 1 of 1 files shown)


tommyreddad:
The upgrade test is still running with cloud-run-events 0.21 which has the incompatible Broker v1beta1 type, so that's why it is failing.

tommyreddad:
Unpinning the versions in #2205 instead since the tests here still use v1beta1 Broker.

PR #2214, opened 2021-05-13
Check custom metrics only if running E2E tests

tommyreddad:
This unblocks #2212.

The "build" GitHub action has been changed to actually perform the build test. It ends up running our e2e tests a little bit and failing because the metrics client can't authenticate. This change makes it so the metrics check only runs when the e2e tests are running, not when the build test is running.

Tested with:

```shell
tags="$(grep -I  -r '// +build' . | \
      grep -v '^./vendor/' | \
      grep -v '^./hack/' | \
      grep -v '^./third_party' | \
      cut -f3 -d' ' | \
      sort | uniq | \
      grep -v '^!' | \
      tr '\n' ' ')"
go test -vet=off -tags "${tags}" -run=^$ ./...
```

### DIFF
--- a/test/conformance/main_test.go
+++ b/test/conformance/main_test.go
@@ -51,7 +51,8 @@ func TestMain(m *testing.M) {
 	os.Exit(func() int {
 		test.InitializeFlags()
 		eventingtest.InitializeEventingFlags()
-		projectID, e2e_test := os.LookupEnv(lib.ProwProjectKey)
+		// If there's no project ID environment variable, it's probably a build test
+		projectID, projectProvided := os.LookupEnv(lib.ProwProjectKey)
 		channelTestRunner = eventingtestlib.ComponentsTestRunner{
 			// ChannelFeatureMap saves the channel-features mapping.
 			// Each pair means the channel support the given list of features.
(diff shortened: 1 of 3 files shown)


coryrc:
Revert the build test to the original too?

https://github.com/knative/serving/blob/main/.github/workflows/knative-go-build.yaml

test/conformance/main_test.go
```diff
@@ -51,7 +51,7 @@ func TestMain(m *testing.M) {
 	os.Exit(func() int {
 		test.InitializeFlags()
 		eventingtest.InitializeEventingFlags()
-		projectID = os.Getenv(lib.ProwProjectKey)
+		projectID, e2e_test := os.LookupEnv(lib.ProwProjectKey)
```
coryrc, line 54: This is too clever IMO...

also not supposed to have underscores in Go variable names

would prefer

```
projectID, projectProvided :=
```

or 

```
// If there's no project ID environment variable, it's probably a build test
projectID, e2eTest :=
```

tommyreddad: Done.

coryrc:
> Revert the build test to the original too?
> 
> https://github.com/knative/serving/blob/main/.github/workflows/knative-go-build.yaml

Oh, I see that's the other change pending merge.

PR #2224, opened 2021-05-21
Manual update dependency and use new upgrade test infra

zhongduo:
The upgrade test infra has been updated, this PR changes to the latest infra.


zhongduo:
it turns out that the upgrade test now use `.` in the name of the trigger, which will be copied as label in the pubsub topic as labels, but pubsub topic does not accept `.` in the label value.....

PR #2225, opened 2021-05-24
run update to pin to 0.23 deps for release

nlopezgi:
Prepare for 0.23 release. Changes created running
```
RELEASE=0.23
./hack/update-deps.sh --upgrade --release ${RELEASE}
./hack/update-codegen.sh
```


go.mod
```diff
@@ -33,10 +34,11 @@ require (
 	k8s.io/api v0.19.7
 	k8s.io/apimachinery v0.19.7
 	k8s.io/client-go v0.19.7
-	knative.dev/eventing v0.23.1-0.20210520143017-869220a8c3c3
+	knative.dev/eventing v0.23.0
 	knative.dev/hack v0.0.0-20210428122153-93ad9129c268
-	knative.dev/pkg v0.0.0-20210520062216-e749d6a2ad0e
-	knative.dev/serving v0.23.1-0.20210520141217-a59495bbd17d
+	knative.dev/networking v0.0.0-20210519080815-4a780203e35b // indirect
```
tommyreddad, line 39: We have a dependency on knative networking now? I wonder how this appeared.

nlopezgi: since its saying its `indirect` we could maybe assume that one of our deps added it?

PR #2227, opened 2021-05-26
feat/broker: convert "." to "-" in topic and subscription labels

### LINKED ISSUE #2226: Trigger with `.` will causes error due to pubsub label constraints
**Describe the bug**
When creating a trigger with name or namespace including `.` or any other unsupported character, controller will fail to create the topic and subscription as it will try to add labels `name: TRIGGER_NAME, namespace: NAMESPACE` to the pubsub resources, and since label value can only be alphabet or `-` `_`, this will cause failure.

 **Expected behavior**
Controller is smart enough to skip labels or convert it to a supported form.

 **To Reproduce**
create a trigger with name `test.trigger`.

zhongduo:
Fixes #2226 

## Proposed Changes

- Convert all `.` with `-` in topic and subscription labels

- 🐛 Fix topic and subscription label issue with '.'.

### DIFF
--- a/pkg/reconciler/celltenant/target.go
+++ b/pkg/reconciler/celltenant/target.go
@@ -19,6 +19,7 @@ package celltenant
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	metadataClient "github.com/google/knative-gcp/pkg/gclient/metadata"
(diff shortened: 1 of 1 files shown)


tommyreddad:
Can you write a unit test for this?

